PR #6, opened 2015-07-29
Make the Github repository the authoritative source for GTFS Realtime

egorich239:
Make the Github repository the authoritative source for gtfs-realtime.proto.

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -20,7 +20,7 @@
 // arrival times.
 //
 // This protocol is published at:
-// https://developers.google.com/transit/gtfs-realtime/
+// https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto
 
 syntax = "proto2";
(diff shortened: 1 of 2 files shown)


gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -20,7 +20,7 @@
 // arrival times.
 //
 // This protocol is published at:
-// https://developers.google.com/transit/gtfs-realtime/
+// https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto
```
magdalar: Just link to the gtfs-realtime directory of the project, rather than the proto file directly?

We should probably add a README.md directly at the gtfs-realtime directory level to quickly point to the most-relevant files/entry-points too (which will be especially helpful once there's more than an 'en' directory under /spec/).

egorich239: @magdalar Addressed, added gtfs-realtime/README.md

magdalar:
LGTM

PR #10, opened 2015-07-29
Fix Element Index and links

barbeau:
This patch adds a few missing elements to the Element Index (I believe all under `VehiclePosition`), and fixes the links within the document so they reference the correct heading for each element in markdown.

I signed the ICLA referenced [here](https://groups.google.com/forum/#!topic/gtfs-realtime/m_VhGYfYeDY) so this should be good to go for review.


barbeau:
Just to be clear - this PR isn't proposing any changes to the spec, just fixing the existing docs.

barbeau:
@egorich239 Ok!  :) Will push another commit into PR shortly with CONTRIBUTORS addition.

barbeau:
@egorich239 Ok, all set to merge.  Added myself to CONTRIBUTORS.

PR #15, opened 2015-11-13
Adding clarifications about trip matching

msbreadloaf:
I propose adding information on how to match trips in systems for which trip_ids are not unique or trip_ids are not stable/available.

### DIFF
--- a/gtfs-realtime/spec/en/trip-updates.md
+++ b/gtfs-realtime/spec/en/trip-updates.md
@@ -50,6 +50,7 @@ The information provided by the trip descriptor depends on the schedule relation
 In most cases, you should provide the trip_id of the scheduled trip in GTFS that this update relates to. 
 
 #### Systems with repeated trip_ids
+
 For systems using repeated trip_ids, for example trips modeled using frequencies.txt, that is frequency-based trips, the trip_id is not in itself a unique identifier of a single journey, as it lacks a
 specific time component. In order to uniquely identify such trips within a
 TripDescriptor, a triple of identifiers must be provided:
@@ -72,7 +73,8 @@ as (T, 2015-05-25, 10:10:00) but provide a StopTimeUpdate with departure from
 first stop at 10:13:00.
 
 #### Alternative trip matching
-Trips which are not frequency based can also be uniquely identified by a
+
+Trips which are not frequency based may also be uniquely identified by a
 TripDescriptor including the combination of:
 
 *    __route_id__
@@ -80,17 +82,8 @@ TripDescriptor including the combination of:
 *    __start_time__
 *    __start_date__
 
+where start_time is the scheduled start time as defined in the static schedule, as long as the combination of ids provided resolves to a unique trip.
 
-start_time should be the scheduled start time.
-Once appeared, start_time must stay the same in all TripDescriptor representing
-the same trip instance across all feed versions. StopTimeUpdates should be used
-to indicate adjustments to start_time.
-For example, let’s say we decide at 10:00, May, 25th 2015, that a trip with
-route_id=R and direction_id=0 will start at start_time=10:10:00, and provide
-this information via realtime feed at 10:01. By 10:05 we suddenly know that the
-trip will start not at 10:10 but at 10:13. In our new realtime feed we can still
-identify this trip as (R, 0, 2015-05-25, 10:10:00) but provide a StopTimeUpdate
-with departure from first stop at 10:13:00.
 
 ## Uncertainty
 
--- a/gtfs-realtime/spec/en/trip-updates.md
+++ b/gtfs-realtime/spec/en/trip-updates.md
@@ -49,6 +49,7 @@ The information provided by the trip descriptor depends on the schedule relation
 
 In most cases, you should provide the trip_id of the scheduled trip in GTFS that this update relates to. 
 
+#### Frequency based systems
 For trips modeled using frequencies.txt, that is frequency-based trips, the
 trip_id is not in itself a unique identifier of a single journey, as it lacks a
 specific time component. In order to uniquely identify such trips within a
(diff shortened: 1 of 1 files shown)


skinkie:
Could you describe the impact of this change? What are the risks for current implementations? Is there a producer/consumer pair want to implement it?

magdalar:
This describes the rules that Google is currently applying.

We are thusly such a consumer, and you (openov) are one of the producers :-D

As for current implementations, I don't think there's a risk, as this is
only adding clarity where there was none.

gtfs-realtime/spec/en/trip-updates.md
```diff
@@ -47,7 +47,43 @@ The information provided by the trip descriptor depends on the schedule relation
...
+should be used to indicate adjustments; start_time does not have to be precisely
+the departure time from the first station, although it should be pretty close to
+that time.
+
+For example, let’s say we decide at 10:00, May, 25th 2015, that a trip with
+trip_id=T will start at start_time=10:10:00, and provide this information via
+realtime feed at 10:01. By 10:05 we suddenly know that the trip will start not
+at 10:10 but at 10:13. In our new realtime feed we can still identify this trip
+as (T, 2015-05-25, 10:10:00) but provide a StopTimeUpdate with departure from
+first stop at 10:13:00.
```
barbeau: There is a corner case not covered here, previous raised by @kurtraschke in another thread.  A transit agency may run two buses back-to-back in rush hour to double capacity.  This is effectively the same `trip_id`, `start_time`, and `start_date` for both vehicles.  In that case, you would also need `vehicle_id` to disambiguate.

@kurtraschke what agency was it that you said you observed this at?

barbeau:
An implementation of GTFS-rt support for frequency-based trips (exact_times=0) in OpenTripPlanner that one of our students worked on made similar assumptions, in terms of requiring the combination `trip_id` and `start_time` to be unique in the GTFS-rt feed - see https://github.com/opentripplanner/OpenTripPlanner/pull/1647.  The plain text output of the GTFS-rt feed we implemented for frequency-based trips (exact_times=0) for our campus bus system is [here](http://mobullity.forest.usf.edu:8088/trip-updates?debug).  We didn't need `start_date` in our case because service doesn't span midnight, but I'm for including `start_date` as a requirement in the spec - others may not realize when exactly its needed and when its not.

I raised a corner case above in which `vehicle_id` is also needed, but generally speaking +1 for this.

It should be noted that some of this is also already outlined in the [current definition of `start_time`](https://developers.google.com/transit/gtfs-realtime/reference?hl=en#TripDescriptor).

gtfs-realtime/spec/en/trip-updates.md
kurtraschke: The conversation @barbeau describes is in the comments on this document (along with some other points which may be relevant to this issue): https://docs.google.com/document/d/13pkZPVxWcphCLzSO-CpetUZ4onqZ0BN11oIpf78gNp4/edit#heading=h.fy8gzwwva0d4

I believe this is what WMATA does with some trips on the airport buses during the holidays.

egorich239: @barbeau This corner-case does not really exist. As it is written in the paragraph (start_date + start_time) does not have to correspond to the departure time from the first stop. For such back-to-back buses one can specify start_time differing by, say, 1 second, and still have them depart at the same moment from the first stop per stop_time_update.

egorich239:
@barbeau Re: start_date + GTFS.

Fun fact: if your transit system has a bus starting at 00:30:00 local wall time, then the only proper way I know to encode such bus in GTFS (static) is by defining it as 24:30:00 of the previous day. It is otherwise impossible to encode this bus on a day when time is moved back by 1 hour (that is, the calendar day, when 13 hours pass between 00:00 wall time and 12:00 wall time) -- GTFS definition of time is 12 hours before local noon, which corresponds to 1am wall time for that day, and you'd need negative time to encode such a trip at such a day; GTFS 00:30:00 means wall time 01:30:00 on such day (assuming usual switch-over time between 2am and 4am). In the same way on the short day (moving time forward), GTFS 00:30:00 will correspond to 23:30:00 wall time of the previous day. If I were to give a recommendation, I'd recommend to always provide times in GTFS (static) feeds in the time window of 03:00:00-27:00:00 (or 02:00:00-26:00:00, depending on when the typical DST-switchover happens in the system), because these times correspond to wall times on every day of the year.

Now, back to GTFS RT. Resolving start_time's of 24:00:00+ without start_date is not very fruitful (as in: possible, but may trigger ambiguity). Therefore I'd also encourage start_date to be specified properly (and by properly I mean -- consistent with GTFS static).

gtfs-realtime/spec/en/trip-updates.md
magdalar: @barbeau @egorich239 

I'm not particularly a fan of adopting `vehicle_id` as a part of the unique identifier here; first of all, it's not an attribute of `TripDescriptor`, and so would need to be added (or moved) there.

Secondly,  I think that I agree with Ivan that offsetting the `start_time` by a second or so might be the better approach here: it's not actually feasible [in the real world] for the vehicles to have departed at the same second anyway. In essence however, the `start_time` here is less the actual `start_time`, but more the unique identifier of the trip within the given `start_date`+`trip_id` combination anyway.

barbeau: >In essence however, the start_time here is less the actual start_time, but more the unique identifier of the trip within the given start_date+trip_id combination anyway.

That's really my main point - if we require `start_time` to be unique we're overloading the `start_time` field with something that isn't really a time.

>offsetting the start_time by a second or so might be the better approach here

If everyone else agrees, I'll go with the consensus.

magdalar:
Vote +1

barbeau:
Vote +1

bboissin:
Vote +1

skinkie:
Vote +1

barbeau:
Looks like the HTML docs for this PR haven't been updated either:

bboissin:
FYI I'm trying to get the issue with the devsite@google resolved.

barbeau:
Thanks @bboissin!

PR #16, opened 2016-01-07
Clarify behavior for dropping StopTimeUpdates for vehicles running ahead of schedule

barbeau:
In the "Stop Time Updates" section, the [GTFS-realtime spec](https://developers.google.com/transit/gtfs-realtime/trip-updates#stop-time-updates) says:

>A trip update consists of one or more updates to vehicle stop times, which are referred to as StopTimeUpdates. These can be supplied for past and future stop times. You are allowed, but not required, to drop past stop times. When doing this, be aware that you shouldn't drop a past update if it refers to a trip that isn't yet scheduled to have finished (i.e. it finished ahead of schedule) as otherwise it will be concluded that there is no update on this trip.”

My interpretation of the highlighted portion is that when multiple StopTimeUpdates exist in a trip (i.e., per-stop predictions), individual StopTimeUpdates shouldn't be dropped from the GTFS-rt feed if the vehicle is running ahead of schedule until *after* the scheduled arrival time for that stop (from GTFS stop_times.txt).

For example, if the following data appears in the GTFS-rt feed:

* Stop 4 –Predicted at 10:18am (scheduled at 10:20am – 2 min early)
* Stop 5 –Predicted at 10:30am (scheduled at 10:30am – on time)

...the prediction for Stop 4 cannot be dropped from the feed until 10:21am, even if the bus actually passes the stop at 10:18am.  If the StopTimeUpdates for Stop 4 was dropped from the feed at 10:18am or 10:19am, and the scheduled arrival time is 10:20am, then the consumer should assume that no real-time information exists for Stop 4 at that time.

A vendor is arguing that this text applies only to the TripUpdate, not to the StopTimeUpdates, and they are allowed to drop updates for vehicles running early as soon as the vehicle passes the stop (no matter what the scheduled arrival time is).  If the vendor's interpretation is correct, it results in a very poor end user experience for consumers - for example, if a vehicle was running 5 minutes early, and the user checks their app 4 minutes before the scheduled arrival time, the app would only show scheduled information, and would show that the vehicle was expected to arrive in 4 minutes (even though at a system level we know that the vehicle already passed the stop).

I'm proposing this change to clarify the expected behavior in the context of dropping StopTimeUpdates with a vehicle running ahead of schedule.

See also https://groups.google.com/forum/#!topic/gtfs-realtime/3rAf6UIhAsQ.

### DIFF
--- a/gtfs-realtime/spec/en/trip-updates.md
+++ b/gtfs-realtime/spec/en/trip-updates.md
@@ -6,7 +6,14 @@ There should be **at most** one trip update for each scheduled trip. In case the
 
 ## Stop Time Updates
 
-A trip update consists of one or more updates to vehicle stop times, which are referred to as [StopTimeUpdates](reference.md#StopTimeUpdate). These can be supplied for past and future stop times. You are allowed, but not required, to drop past stop times. When doing this, be aware that you shouldn't drop a past update if it refers to a trip that isn't yet scheduled to have finished (i.e. it finished ahead of schedule) as otherwise it will be concluded that there is no update on this trip.
+A trip update consists of one or more updates to vehicle stop times, which are referred to as [StopTimeUpdates](reference.md#StopTimeUpdate). These can be supplied for past and future stop times. You are allowed, but not required, to drop past stop times.  Producers should not drop a past `StopTimeUpdate` if it refers to a stop with a scheduled arrival time in the future for the given trip (i.e. the vehicle has passed the stop ahead of schedule), as otherwise it will be concluded that there is no update for this stop.  
+
+For example, if the following data appears in the GTFS-rt feed:
+
+* Stop 4 – Predicted at 10:18am (scheduled at 10:20am – 2 min early)
+* Stop 5 – Predicted at 10:30am (scheduled at 10:30am – on time)
+
+...the prediction for Stop 4 cannot be dropped from the feed until 10:21am, even if the bus actually passes the stop at 10:18am. If the `StopTimeUpdate` for Stop 4 was dropped from the feed at 10:18am or 10:19am, and the scheduled arrival time is 10:20am, then the consumer should assume that no real-time information exists for Stop 4 at that time, and schedule data from GTFS should be used.
 
 Each [StopTimeUpdate](reference.md#StopTimeUpdate) is linked to a stop. Ordinarily this can be done using either a GTFS stop_sequence or a GTFS stop_id. However, in the case you are providing an update for a trip without a GTFS trip_id, you must specify stop_id as stop_sequence has no value. The stop_id must still reference a stop_id in GTFS.
 


egorich239:
Sean, I am very much in favor of your interpretation. 
The reason is that it should be possible to describe a system at any given moment using its GTFS + non-incremental GTFS-RT feeds, without any additional memory.

Therefore in your example omitting Stop 4 between 10:18 and 10:20 the state of the system (if no Stop 4 is listed): "Stop 4 is according to GTFS schedule".

Now, from the producer's point of view it may be not very convenient (I don't know). In such a case I'd think about providing a hint at Stop 5 that it is the next stop that this vehicle approaches. Such a hint allows a producer to also avoid having memory. Problem is, StopTimeUpdate currently does not have such a hint.

gtfs-realtime/spec/en/trip-updates.md
```diff
@@ -6,7 +6,7 @@ There should be **at most** one trip update for each scheduled trip. In case the
 
 ## Stop Time Updates
 
-A trip update consists of one or more updates to vehicle stop times, which are referred to as [StopTimeUpdates](reference.md#StopTimeUpdate). These can be supplied for past and future stop times. You are allowed, but not required, to drop past stop times. When doing this, be aware that you shouldn't drop a past update if it refers to a trip that isn't yet scheduled to have finished (i.e. it finished ahead of schedule) as otherwise it will be concluded that there is no update on this trip.
+A trip update consists of one or more updates to vehicle stop times, which are referred to as [StopTimeUpdates](reference.md#StopTimeUpdate). These can be supplied for past and future stop times. You are allowed, but not required, to drop past stop times. When doing this, be aware that you shouldn't drop a past StopTimeUpdate if it refers to stop that a vehicle isn't yet scheduled to have passed (i.e. running ahead of schedule) as otherwise it will be concluded that there is no update for this stop.
```
egorich239: "a vehicle isn't yet scheduled to have passed" -- this blows up my mind, can we somehow re-formulate it?

barbeau:
>Sean, I am very much in favor of your interpretation. 
The reason is that it should be possible to describe a system at any given moment using its GTFS + non-incremental GTFS-RT feeds, without any additional memory.

Agreed, this is definitely the most efficient design from a consumer's perspective and keeps complexity low.

>Therefore in your example omitting Stop 4 between 10:18 and 10:20 the state of the system (if no Stop 4 is listed): "Stop 4 is according to GTFS schedule".

I agree - I also believe this expected behavior for consumers is fairly clear given the current spec as a whole, although it might help to add another bullet with "stop_sequences 1, 2 have an unknown delay" to Example 2 [here](https://developers.google.com/transit/gtfs-realtime/trip-updates#stop-time-updates) - I can add that in another PR/proposal if you agree.  

>Now, from the producer's point of view it may be not very convenient (I don't know).

Yes, the main question I'm attempting to address in this PR is the requirement for producers.

>In such a case I'd think about providing a hint at Stop 5 that it is the next stop that this vehicle approaches.

Interesting thought - so I guess this would require the producer to include something like `VehicleStopStatus` within `StopTimeUpdate`, IFF the vehicle is running early?  So if Stop 5 had `VehicleStopStatus` = `IN_TRANSIT_TO`, and that prediction is ahead of schedule, then consumers would propagate the delay from Stop 5 upstream to Stop 4?  I assume the upstream propagation would only be 1 stop, so Stops 1, 2, and 3 would still have an unknown delay (i.e., would be "according to GTFS schedule").

I could see two challenges, assuming I'm interpreting this correctly:

* The delay value at Stop 5 for update `n` is not necessarily the same delay value that existed in the previous update at Stop 4 in update `n-1` (where `n` is the number of GTFS-rt  `FULL_DATASETs` generated by the producer).  The vehicle could have been running ahead of schedule by 2 minute at Stop 4 and then fall behind schedule by 3 minutes on its way to Stop 5.  In update `n`, this could result in us showing the 3 min delay upstream at Stop 4, which could indicate that the bus hadn't left yet.  As a result, I believe the producer would be required to only include `VehicleStopStatus` = `IN_TRANSIT_TO` IFF both update `n` for Stop 5 and update `n-1` for Stop 4 are early AND the Stop 5 delay for update `n` isn't earlier than the update `n-1` delay for Stop 4 (otherwise we could propagate an illogical delay upstream).  So, I believe we're back to the producer needing memory to retain the state of update `n-1`.
* If the producer doesn't provide `arrival_time` and `departure_times` for non-timepoint stops in GTFS `stop_times.txt`, and it provides `arrival` or `departure` fields instead of `delays` in the GTFS-rt feed, the consumer doesn't actually know if the vehicle is running late or early - it can only guess based on its interpolation of scheduled times from the timepoint schedules.  So, again we'd need a requirement that `VehicleStopStatus` = `IN_TRANSIT_TO` (or whatever we want to call the hint) is only published by the producer when the vehicle is actually running early, based on the producer's internal schedule.  This would also require the producer to have delay information in its implementation even if its not publishing `delay` values.  I think this gets very confusing from both producers and consumers perspectives.

It seems to me that the far simpler implementation and specification requirement would be to require that producers keep `StopTimeUpdates` for early arrivals until the scheduled arrival time for that stop has passed.  The vendor we're working with just agreed to implement this approach within the next two weeks.  From a spec perspective, this also matches the existing requirement for TripUpdates (there must be at least one prediction retained in the feed until the entire trip is scheduled to have completed).  

cc @cagryInside

barbeau:
BTW, I haven't mentioned this previously, but this proposal is in the context of our implementation of multiple `StopTimeUpdate` support for OneBusAway (https://github.com/OneBusAway/onebusaway-application-modules/pull/142).

gtfs-realtime/spec/en/trip-updates.md
barbeau: Yeah, that's not the clearest - language is hard :).

What if we include an example to make sure this is clear?  How about the following:

>Producers should not drop a past `StopTimeUpdate` if it refers to a stop with a scheduled arrival time in the future for the given trip (i.e. the vehicle has passed the stop ahead of schedule), as otherwise it will be concluded that there is no update for this stop.  

>For example, if the following data appears in the GTFS-rt feed:

>Stop 4 – Predicted at 10:18am (scheduled at 10:20am – 2 min early)
Stop 5 – Predicted at 10:30am (scheduled at 10:30am – on time)

>...the prediction for Stop 4 cannot be dropped from the feed until 10:21am, even if the bus actually passes the stop at 10:18am. If the `StopTimeUpdate` for Stop 4 was dropped from the feed at 10:18am or 10:19am, and the scheduled arrival time is 10:20am, then the consumer should assume that no real-time information exists for Stop 4 at that time, and schedule data from GTFS should be used.

barbeau:
@egorich239 I've rephrased the spec language for this proposal to be a little clearer, and included the same example I originally posted to the list.  I forced-pushed this new commit into this same PR to replace the previous contents, so it's ready for your review.

bboissin:
+1

nevi-me:
+1, this is progress; though I'd prefer keeping the update for `x` time longer so that developers/users have the flexibility of seeing how early a service was. I've expressed same in the mailing lists.

barbeau:
@egorich239 Can we get a +1 from you on this proposal?  That should be enough for the 3 required votes to get this clarification adopted.

magdalar:
+1

barbeau:
Alright, we're past the end of the voting window, and we have at least 3 votes with unanimous consent.  I believe we should be set for this to be officially adopted.  Could we please get this merged?

magdalar:
Merging... thanks for your contribution!

barbeau:
Thanks!

barbeau:
Looks like the HTML docs haven't been updated for this PR either:

PR #18, opened 2016-01-20
Clarify that stops upstream of predictions have unknown delay

barbeau:
In the process of attempting to clarify behavior for producers as to when they are allowed to drop per-stop predictions (https://github.com/google/transit/pull/16), it became apparent that at least one vendor was assuming that when using per-stop predictions, consumers could either propagate predictions upstream or hold over predictions from a previous feed update and show these to end users.

This patch clarifies that in the absence of any predictions upstream of a stop-specific prediction, it should be assumed that these upstream stops have an unknown delay.

### DIFF
--- a/gtfs-realtime/spec/en/trip-updates.md
+++ b/gtfs-realtime/spec/en/trip-updates.md
@@ -32,6 +32,7 @@ For the same trip instance, three [StopTimeUpdates](reference.md#StopTimeUpdate)
 
 This will be interpreted as:
 
+*   stop_sequences 1,2 have unknown delay.
 *   stop_sequences 3,4,5,6,7 have delay of 300 seconds.
 *   stop_sequences 8,9 have delay of 60 seconds.
 *   stop_sequences 10,..,20 have unknown delay.


bboissin:
Is there a reason not to expand the example to also have unknown delay in the middle?

barbeau:
>Is there a reason not to expand the example to also have unknown delay in the middle?

I'm not against that - for this proposal I was targeting the smallest possible change to the existing example.  Perhaps it would make sense to add a more complex 3rd example that would illustrate this.

c3pohara:
+1

magdalar:
+1

bboissin:
+1

magdalar:
Looks like we're past the voting period on this one too, and it looks like all in favor. Merging.

Thanks for your contributions!

barbeau:
Thanks!

barbeau:
Just a heads up - it looks like this change hasn't been pushed over to the main HTML docs.  If you look at Example 2 at https://developers.google.com/transit/gtfs-realtime/guides/trip-updates, it doesn't have the  `stop_sequences 1,2 have unknown delay` statement added via this PR.

barbeau:
Another friendly request to please update the main [GTFS HTML docs](https://developers.google.com/transit/gtfs-realtime/guides/trip-updates) with this change.  Many users consider that the canonical source for the spec, and there is already at least [one case](https://github.com/opentripplanner/OpenTripPlanner/issues/2295#issuecomment-237779707) of someone who could of benefited from this information not finding it in the HTML docs.

PR #20, opened 2016-02-11
Conditionally require stop_sequence in StopTimeUpdate

barbeau:
If the same stop is visited more than once in a trip, producers must provide the stop_sequence for all StopTimeUpdates that apply to that stop.  Otherwise, if a producer only provides a stop_id, it is unclear to consumers to which stop_sequence that prediction should be applied.

Proposal announced on GTFS-realtime group here - https://groups.google.com/forum/#!topic/gtfs-realtime/E9jcHxBjJ0k.


riteshwarade:
+1

gtfs-realtime/spec/en/reference.md
```diff
@@ -140,7 +140,7 @@ Uncertainty applies equally to both time and delay. The uncertainty roughly spec
 Realtime update for arrival and/or departure events for a given stop on a trip. Please also refer to the general discussion of stop time updates in the [TripDescriptor](#message-tripdescriptor) and [trip updates entities](trip-updates.md) documentation.
 
 Updates can be supplied for both past and future events. The producer is allowed, although not required, to drop past events.
-The update is linked to a specific stop either through stop_sequence or stop_id, so one of these fields must necessarily be set.
+The update is linked to a specific stop either through stop_sequence or stop_id, so one of these fields must necessarily be set.  If the same stop_id is visited more than once in a trip, then stop_sequence must be provided in all StopTimeUpdates for that stop_id on that trip.
```
magdalar: Should there be a similar update to trip-updates.md line 18?

I might degrade 'must' to 'should' though. We've seen a number of producers for which it's just not feasible to keep stop_sequences synchronized with GTFS in the AVL system without some tricks.

Note also that stop_sequence is one of the most-prone-to-breaking identifiers across static feed versions (since they're often not pre-gapped for future-proofing), and GTFS-RT still lacks a means of indicating the target static feed version.

barbeau: >Should there be a similar update to trip-updates.md line 18?

Good catch!  Just forced pushed a new commit that includes this addition to trip-updates.md as well.

>I might degrade 'must' to 'should' though.

In the commit I just force-pushed I also changed "must" to "should".

magdalar:
Should we open this for a vote?

magdalar:
Ah -- I see that the vote was called via gtfs-realtime@ 10 days ago, but it seems I was not the only one to miss that.

With that, I recommend we re-start the voting from today.

+1 from me.

barbeau:
+1

barbeau:
@riteshwarade The proposal text changed slightly from your original vote - would you mind reviewing again, and if you still agree chime in with another "+1"?

ixtano:
+1

riteshwarade:
+1

magdalar:
Voting concluded; looks like we have 4 yeses, and no objections. Mering.

barbeau:
:+1:

barbeau:
Looks like the HTML docs for this PR haven't been updated either:

* https://developers.google.com/transit/gtfs-realtime/reference/StopTimeUpdate

PR #21, opened 2016-03-16
Initial conversion

bboissin:


egorich239:
Tables require some love though

PR #26, opened 2016-05-03
Added Metro-North and realCity extension ids

TewWe:
As mentioned on the mailing list, I've created the extension ids for these projects (based on advice from Google engineers). Please merge it.


TewWe:
I signed it!

PR #27, opened 2016-08-04
Added Missing Info Under fare_attributes.txt

eburkman:
Added missing agency_id requirement (for feeds with multiple agency fare
information) under fare_attributes.txt.

See also: https://groups.google.com/forum/#!topic/gtfs-changes/VRb-Et-wWow

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -1,6 +1,6 @@
 ## General Transit Feed Specification Reference
 
-**Revised March 16, 2016. See [Revision History](../../CHANGES.md) for more details.**
+**Revised August 4, 2016. See [Revision History](../../CHANGES.md) for more details.**
 
 This document explains the types of files that comprise a GTFS transit feed and defines the fields used in all of those files.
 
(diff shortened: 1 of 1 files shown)


egorich239:
The usual requirements for the amendment process apply: https://github.com/google/transit/blob/master/gtfs/CHANGES.md

eburkman:
A few weeks ago I updated a multi-agency feed that includes fare information. The feed passed the validator offered in transit tools, but then failed the Google Partner Dash validator. A screen shot of the error thrown by the Partner Dash validator is attached.

I looked to the specification documentation to find that it does not mention requiring agency_id in fare_attributes.txt (despite the Google validator looking for that field in multi-agency feeds). Once I updated the feed to include the field in that file, it successfully passed the Partner Dash validator and was launched.

I posted to GTFS feed spec changes Google Group, and at least one other person has experienced the same issue with the incomplete documentation.

This pull request is to modify the specification reference documentation to include the agency_id field under the fare_attributes.txt section. The field is marked as optional with the following description: "Required for feeds with multiple agencies defined in the agency.txt file. Each fare attribute must specify an agency_id value to indicate which agency the fare applies to."

capture

eburkman:
Is there any other discussion on this?

Since this is essentially regarding updating the documentation to meet what is apparently already required, if there is no more discussion by Monday, August 22, we will continue into the voting process. The minimum required 7 days has already elapsed.

eburkman:
OK, we will move into the voting period for this proposal.
The vote starts now, and will end at 23:59:59 UTC on Tuesday, August 30, 2016.

antrim:
I support this change. Note that all Trillium feeds include fare_attributes.agency_id. e.g. http://transitfeeds.com/p/arcata-mad-river-transit-system/148/latest/file/fare_attributes.txt

skinkie:
+1

RideSchedules:
+1

eburkman:
The vote has ended, and we have unanimous consensus with at least 3 votes, so the proposal passes.

Thanks to all who participated!

PR #28, opened 2016-08-24
WIP - Host GTFS/GTFS-realtime HTML docs in repo via Github pages

barbeau:
This is a proof-of-concept PR showing the ability to host the GTFS HTML docs in the same Github repository as the canonical markdown.

~This PR contains the HTML source for the page:~
~https://developers.google.com/transit/gtfs-realtime/~

~The contents of this PR can be seen rendered by visiting:~
~http://barbeau.github.io/transit/gtfs-realtime/~

~If merged, the site would be visible at:~
~http://google.github.io/transit/gtfs-realtime/~

The purpose of this PR is to explore possible ways that contributors outside of Google could help streamline the process to update the HTML docs.  Currently there are four PRs dating back to Nov. 2015 that have been accepted and merged in this repo, but the HTML docs were never updated:
- https://github.com/google/transit/pull/15
- https://github.com/google/transit/pull/16
- https://github.com/google/transit/pull/18
- https://github.com/google/transit/pull/20

This is a problem because many users consider the GTFS HTML docs the canonical source for the spec.

GTFS-realtime group post at https://groups.google.com/forum/#!topic/gtfs-realtime/Gl933YAr4js.

~**Process change options**~
~1. **Contribute proposal changes to markdown and HTML** - Using a github.io site, contributors could propose changes in the HTML docs along with the changes in the markdown in the same PR.  Then, when the PR is merged it would automatically update the HTML (**EDIT** - because the HTML pages need to be in a separate `gh-pages` branch, and Github PRs only support one branch per PR, markdown and HTML edits would actually need to be split into two PRs).~
~2. **Change to using HTML docs as canonical document, get rid of markdown** - This would avoid duplication of effort in editing both markdown and HTML, but at the cost of readability of the source documents - the HTML source is much harder to read and edit by hand than markdown (IMHO).  IDEs can help with this, but it raises the learning curve for contributing.~

**EDIT Dec. 12, 2016**

Github will render markdown as HTML, so we shouldn't need to keep a separate Markdown and HTML version of the spec.

You can see an example of this here:
http://barbeau.github.io/transit/

It renders the main markdown README as:

~~~
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title></title>
    <link href="http://barbeau.github.io/transit/assets/style.css" rel="stylesheet">
  </head>
  <body>
    <div class="container">
[...]

PR #30, opened 2016-10-26
add station entrance (stops.location_type = 2)

antrim:
following
https://developers.google.com/transit/gtfs/reference/gtfs-extensions#sta
tion-entrances


antrim:
I signed it!

mmorang:
Here are a few comments:

1. In the section describing location_type=2, change stop id -> stop ID to match the rest of the page.
2. Perhaps we should add some further descriptive text explaining the meaning of a station entrance, like "a location where passengers can enter a station from the street"
3. Do we need to amend the text about wheelchair_boarding being inherited from the parent station?  The same rule for regular stops should apply to station entrances, but the text is a bit awkward.  Not sure what to suggest.

mmorang:
(To provide some history on this issue and support for it)

Station entrances with location_type=2 is already on unofficial GTFS extension from the pre-GitHub GTFS datys: https://developers.google.com/transit/gtfs/reference/gtfs-extensions?visit_id=1-636117987477708017-2049630037&rd=1

It is already being used in several existing feeds and applications:
- LA Metro, Madrid, and NY/New Jersey PATH feeds include station entrances with location_type=2.
- Trillium's GTFS Manager software has the capability to publish station entrances.
- Esri's [Add GTFS to a Network Dataset tool] (http://www.arcgis.com/home/item.html?id=0fa52a75d9ba4abcad6b88bb6285fae1) supports location_type=2.

skinkie:
+1

abyrd:
+1 on this feature, but I agree with @mmorang that a short explanation of the meaning of "entrance" means and how it's used should be added.

barbeau:
+1, agree that meaning of entrance and example uses should be included.

antrim:
It looks as though voting has already started. Let's continue voting through Wednesday of next week to see if there are any objections or counter-proposals.

barbeau:
Per my comment on the GTFS changes post for this proposal (https://groups.google.com/forum/#!topic/gtfs-changes/H_nfzbvbD1E) - the discussion period of the proposal needs to last 7 days from Oct. 26th, when this was announced on the GTFS Changes list.  After that window, then we can call for official vote and start tallying them.

antrim:
@barbeau: Got it! Thanks. I had misunderstood the process.

barbeau:
@antrim No worries!  This is new to everyone - I believe you get the honor of having the first GTFS proposal for a new feature in the Github workflow :).

mmorang:
I think it's been 7 days now, so voting can start?  @antrim, can you address the comments above about adding a description about the meaning of a station entrance before calling the vote?

antrim:
I just made some modifications to this proposal:
* Changes to respond to comments from @mmorang and @abyrd .
* Added note that our understanding of entrances is that they are also used for exiting.

mmorang:
Looks good.  Just one typo in the wheelchair_boarding section:
Should be:  "there exists no accessible path _from_ the entrance to station platforms" (you forgot the word "from").
If you fix that, then it looks good to me.

antrim:
Fixed. Let's give through Monday for any other comments and then start the vote after that.

eburkman:
The purist in me feels that stops.txt should focus on _where the service is provided_ as opposed to _how to access the service_, especially when there could be additional helpful information related to entrances and exits.

A more durable solution could be to include an additional file with points of access (entrance only, exit only, entrance/exit). These could also include other attributes, such as elevator service (possible tie-in for GTFS-realtime for communicating elevator outages), fare purchasing available (some entrances/exits have no fare machines), or possibly even hours (as some entrances are daytime only, etc).

Stops.txt could then reference an access_point or similarly named field from the new file (or this could be structured some other way, possibly like fare_zones). This would allow the additional information related to specific entrances and exits without further cluttering stops.txt, but maybe that level of detail isn't desired or necessary.

I recognize that a number of agencies already use the proposed extension, but in a way the extensions focus on using the preexisting GTFS specification to fit a previously unexpected condition. This modification process allows for deeper changes that make more sense if the current spec's constraint is removed.

antrim:
@eburkman: Thoughtful point. My own position is that because data following these specifications is already in use for a few applications, we should incorporate into the Spec, **_unless_** those application developers and all of the feed publishers that use location_type=2 (station entrance) are ready to commit to a change.

There is a proposal for **pathways.txt** that is related to stops and station entrances, here: https://groups.google.com/forum/#!msg/gtfs-changes/Gk88QYQYgtw/UNdVCH0v-RIJ. I don't believe that it covers elevators specifically.

Amenities: There is an existing proposal for a **stop_features.txt** file: https://sites.google.com/site/gtfschanges/proposals/stop-amenity. I believe that TriMet publishes a feed with this file.

Anyone else: Thoughts?

mmorang:
Although I share the same concern @antrim does regarding overcomplicating matters, I do kind of like the idea of a more comprehensive street entrance dataset.  The more information about how to get passengers from the street to the platform the better.  I would update my tools to support whatever we decide here.

In particular, here are some things that would be helpful to routing besides what's already been mentioned:
- Travel time from the entrance to the platform.  Super important in really large stations.  Geometry could be included as well and might be helpful for guidance, but travel time is more important to me.
- Some way to link a stop or station entrance to a specific street.  Often when I have to guess which street to "snap" a stop to, particularly bus stops along major roads, the stop will fall closer to the centerline of a side street and will snap to the side street instead of the main street the bus actually runs on.  There's no way for me to associate the stop with a particular street.  However, in order to make this distinction, one would have to tag each stop with some kind of street ID, which would necessitate picking some "official" street dataset, which I don't think is a good thing.  Even if we chose something like OSM, it would be hard to keep the GTFS dataset and the OSM street IDs in sync.  So...I don't know what to do about this problem.

Regarding amenities: I'm under the impression that GTFS is meant to be primarily for routing passengers through the transit system and not for, say, asset management.  We wouldn't want to clutter up the spec with list of amenities that aren't relevant to routing passengers from one place to another.  That said, I don't think anything that's been mentioned so far (elevators, fare boxes, entrance vs exit, hours of operation) are inconsistent with the routing aspects.

Also, if we broke the station entrances out into a separate file, it's a little weird then that parent stations are included in stops.txt already.

skinkie:
I think from entrances we should move to tourniquets for check-in. But in my understanding transfers.txt can now be filled with all those distances/walking times.

barbeau:
I think we should bring the other existing producers into this discussion, as their willingness to adopt to a new format for something they have been previously publishing is important.

Does anyone have good contacts at the existing producers?  I don't have direct contacts at LA Metro, Madrid, and NY/New Jersey PATH, but I believe some of my colleagues do.

@antrim How many agencies are you currently publishing station entrances for in the existing format?

mmorang:
I have contacts at LA Metro and Madrid and will ask them to view this thread and comment.

antrim:
NY/New Jersey PATH is the only Trillium client that publishes station entrances. https://transitfeeds.com/p/port-authority-of-new-york-and-new-jersey/384
The Trillium GTFS Manager software has this capability but only PATH takes advantage of it (our client portfolio is mainly surface bus, not rail or underground services).

zhsh:
Russian agency "EasyWay" (www.eway24.ru) has been publishing data with stops.location_type=2 too.

antrim:
@mmorang : Have you got in touch with LA Metro and Madrid? I'm not sure what the rules are on open pull requests for change proposals, but I worry we are letting the discussion period become protracted.

barbeau:
Given that there seems to be a good amount of inertia already behind the existing proposal (i.e., producers/consumers), it would make sense to me to vote on this as it stands.  I don't believe there is anything precluding us from a new proposal in the future that would flesh out additional files to capture more precise paths, such as the pathways.txt proposal.

antrim:
@barbeau : Agreed. Let's call the vote (7 days).

barbeau:
+1

andreyz:
+1

mmorang:
@antrim, I reached out to both.  My contact with LAMetro said he would weigh in, but he hasn't.  My contact at Madrid said they had previously discussed location_type=2 with Google and gotten approval to use the unofficial extension and didn't seem interested in further discussion.

I vote yes for the current proposal.

antrim:
The 7 days of voting has ended. There are votes of support from people affiliated with Conveyal (@abyrd), ESRI (@mmorang), Trillium (me), USF CUTR (@barbeau), and support from @andreyz and @skinkie. I don't count any dissenting votes. I'm going to go ahead and close the pull request.

mmorang:
So...this proposal was adopted, and the commits were merged into master.  This means they're officially part of the spec now, right?  That's how this is supposed to work?

When does this page get updated?  It appears that there is now a discrepancy between the GitHub GTFS spec and the Google developers page spec.  Google, how/when do these discrepancies get resolved?  I'm confused.
https://developers.google.com/transit/gtfs/reference/stops-file

barbeau:
@mmorang You're right, the HTML GTFS spec should be in sync with the markdown in this repo, but it's been lagging behind.  It's a known issue with some of the past PRs as well - see https://github.com/google/transit/pull/28 for details.  https://github.com/google/transit/pull/28 also proposes generating the HTML directly from the markdown to address this issue.  We've been discussing this with Google and they are looking at solutions internally too.  I'd suggest concentrating any discussion regarding the lag in comments on https://github.com/google/transit/pull/28.

mmorang:
Thanks for the info @barbeau.  Another thought: Was the official adoption of this proposal ever announced on gtfs-changes, as a follow-up to the introduction of that discussion?  I don't see anything there.  Should that be part of the official procedure, too?

barbeau:
@mmorang Here's the Google Groups thread for this proposal - https://groups.google.com/forum/#!topic/gtfs-changes/H_nfzbvbD1E.  Doesn't look like we closed the loop and announced the adoption at the end of this thread.

I agree that we should make two items part of the [change process](https://github.com/google/transit/blob/master/gtfs/CHANGES.md):

1. Cross-link the PR and Google Group proposals - Google Groups post should have a link to the Github PR (this is implied but not explicitly stated), and the Github proposal should have a link to the Google Groups post.
1. After PR is merged, announce this to the same Google Groups thread.

I'll open a PR and do Google Groups post proposing this :).

barbeau:
PR for proposal https://github.com/google/transit/pull/35.

PR #32, opened 2016-11-26
transfer rules for routes and trips, and in-seat transfers

antrim:
* route to route and trip to trip transfers:
https://developers.google.com/transit/gtfs/reference/gtfs-extensions#TripToTripTransfers
* allow overrides to in-seat block transfers -
https://groups.google.com/forum/#!msg/gtfs-changes/PkNAWbN9wcs/LFEqHGemBAAJ
* new field: agency.disallow_block_transfer

Updates [see below discussion]:

* agency.disallow_block_transfer field proposal has been removed.
* change Spec reference so that block transfers are not assumed by default


antrim:
Notes and justification for proposal:
1. Some feed publishers with to specify trip-to-trip and route-to-route transfers. Google Maps has defined this in a GTFS extension, which is incorporated here: https://developers.google.com/transit/gtfs/reference/gtfs-extensions#TripToTripTransfers
2. Current interpretations of block_id are inconsistent. Real-time prediction systems such as [Transitime](https://github.com/Transitime/core) require block_id, but the current GTFS assumes that in-seat transfers are always allowed. The new field **agency.disallow_block_transfer** and new values for **transfer.transfer_type**, 4 and 5 would allow overriding this default. This feature would work with the trip-to-trip and route-to-route transfer capabilities.
3. For backwards-compatibility, block transfers (in-seat transfers) are assumed to be allowed unless overriding parameters are set.

antrim:
Does anyone have comments on these proposed changes? Since there are changes in a few files addressing a few issues, there is more here to think about than with some other recent proposed changes.

barbeau:
@antrim Please remind me - do we have existing consumers and producers for this?

antrim:
We do for route-to-route and trip-to-trip transfers (Google Maps is the consumer). But I don’t believe we do for other components of this proposal. Perhaps we need to put voting aside and recruit some consumers & producers.

mgilligan:
+1

abyrd:
OpenTripPlanner uses this GTFS extension, as does Google Maps. @antrim reports that Trillium is generating these transfers and I thought that OpenOV did for the Netherlands, but I don't see them in a recent feed. @skinkie are you still generating transfers with trip/stop/route specificity?

About "disallow block transfers" and the corresponding transfer types, I talked to someone at TriMet about this and they were in fact fine with exhaustively listing out all the allowed block transfers in their feed. It is going to be confusing to have some feeds that assume block transfers are allowed (where you have to individually disallow them) and other feeds that assume they are not allowed (where you have to individually enable them). I'm leaning toward just requiring all block transfers to be allowed individually, and if no such transfers are present in the feed, but block IDs are in use, then the consumer can take a guess at what that means.

antrim:
Trillium is actually not yet producing these transfer types. We could very quickly add support for transfer_type = {4,5} but it would take more time to add route-to-route and trip-to-trip transfers because we will probably wait a month or two while transitioning to an overhauled version of the GTFS Manager software.

quoting @abyrd:
> It is going to be confusing to have some feeds that assume block transfers are allowed (where you have to individually disallow them) and other feeds that assume they are not allowed (where you have to individually enable them).

I proposed a default value of 0 (NO) for the new **agency.disallow_block_transfer** field. This was designed to support backward compatibility with the spec today -- block transfers are assumed to be allowed unless otherwise specified.

abyrd:
I should have been more clear. This proposal is really two different things. First, route-to-route and trip-to-trip transfers. Trillium, OTP, Google, and OpenOV (I believe) are already using this. Second, better handling of in-seat transfers, for which nothing has been implemented yet. 

After all the mailing list and personal discussion, I would propose doing this a different way than previously suggested. In my view, in-seat transfers should never be assumed, and all in-seat transfers should be enumerated in transfers.txt using the trip-to-trip or route-to-route mechanism. Block_ids would only imply that a vehicle was passing from one trip to the next, and would only be relevant for consumers propagating real-time data. This means we only need one new transfer type (4) for adding an in-seat transfer.

In the event that a feed has no in-seat transfers listed but has block_ids, the GTFS consumer might want to infer that some in-seat transfers exist but this would be a heuristic based on knowledge about a particular agency, not a standardized behavior.

I'm aware that this could be considered breaking backward compatibility, but I don't get the sense that there was ever strong agreement on how this feature was supposed to work.

DaveBarker:
+1 to the overall proposal and to abyrd's recommendation of approach that in-seat transfers should never be assumed.

skinkie:
>  are you still generating transfers with trip/stop/route specificity?

Yes, and also for FlixBus.

antrim:
I changed the proposal so that in-seat transfers are not assumed -- disallowed by default. This can be set with the new **agency.block_transfer_default** field. 0e149ca

abyrd:
@antrim do we ever want to enable blanket block transfers using agency.block_transfer_default ? It doesn't seem clear that this should be enabled at the feed, agency, or route level, and I don't personally think this deserves cluttering the spec with a new field. Any agency technically sophisticated enough to consider using block IDs to communicate in-seat transfer availability can probably also pull off the transfers.txt approach.

antrim:
@abyrd : We could remove the **disallow_block_transfer**.

However, my reasoning for **block_transfer_default** was that it allows any feed publisher that had assumed block transfers are allowed to publish a new feed that is consistent with the amended spec by adding only one field.

In public Trillium feeds that are provided to trip planners, we presently assume that block transfers are always allowed. This is consistent with how the Spec is written. For arrival prediction software, we actually have a function to provide different feeds (technically not Spec compliant) with block_id values for all trips (including where in-seat transfer is not allowed).

Trillium can quickly change the way we are publishing feeds. I worry that some other feed publishers assumed block transfers and it will take time for them to catch up with the Spec. I propose that we provide opportunity for a broader scan of practice.

For convenient reference, this was the proposed added field in **agency.txt**.

> |  Field Name | Required | Details |
> | --- | --- | --- |
> |  disallow_block_transfer | Optional | Specifies whether in-seat transfers are allowed by default for the agency (see **block_id** field in [trips.txt](tripstxt)). Overriding parameters are specified in [transfers.txt](transferstxt).|
> |   |  | * **0** or **(empty)** - In-seat transfers are disallowed unless an override is set in transfers.txt. This is the default behavior. |
> |   |  | * **1** - In-seat transfers are allowed unless an override is set in transfers.txt.  |

abyrd:
I am mainly concerned about the proliferation of rarely-used fields that exist only for historical reasons, or to preserve backward compatibility when all producers are prepared to accept a small change in semantics to yield a new, clearer specification. Among the main benefits of GTFS are its simplicity and pragmatic approach. There should not be many different ways to represent the same thing, and special fields only used by one operator somewhere.

antrim:
@abyrd @mgilligan : Point taken. I removed the proposed agency.block_transfer_default field. This aimed to provide a convenient bridge to forwards-compatibility, but would create Spec baggage, confusion, and burden over the long term.

gtfs/spec/en/reference.md
```diff
@@ -381,14 +381,35 @@ Trip planners normally calculate transfer points based on the relative proximity
...
-|   |  | * **1** - This is a timed transfer point between two routes. The departing vehicle is expected to wait for the arriving one, with sufficient time for a passenger to transfer between routes. |
+|   |  | * **1** - This is a timed transfer point between two vehicles. The departing vehicle is expected to wait for the arriving one, with sufficient time for a passenger to transfer between routes. |
 |   |  | * **2** - This transfer requires a minimum amount of time between arrival and departure to ensure a connection. The time required to transfer is specified by **min_transfer_time**. |
-|   |  | * **3** - Transfers are not possible between routes at this location. |
+|   |  | * **3** - Transfers are not possible. |
+|   |  | * **4** - Passengers can transfer from one route to another by staying onboard the same vehicle (an "in-seat transfer") in cases where sequential trips are made by the same vehicle (same block_id).  In-seat transfer between these two stops / routes / trips is allowed. |
+|   |  | * **5** - In-seat transfers are not allowed between sequential trips with the same block ID. The passenger must alight from the vehicle and re-board. |
```
antrim: I suppose we should remove transfer_type=5 because in-seat transfers are disallowed by default with this current version.

antrim:
I signed it!

antrim:
Summary of the latest proposal - 4b77a2d.

#### Original motivation

It is not presently possible to specify blocks for all trips (useful for arrival prediction software) AND to precisely specify in-seat transfer availability. The original (and currently published) Spec defines that "A block consists of two or more sequential trips made using the same vehicle, where a passenger can transfer from one trip to the next just by staying in the vehicle." ([at developers.google.com](https://developers.google.com/transit/gtfs/reference/trips-file))

#### What the proposal accomplishes
* Maintains backwards compatibility with the current Spec (in-seat transfers are allowed by default)
* Allows changing the default of whether in-seat transfers are allowed for an agency
* Provide for granular control of in-seat transfers (by stop, trip, and route)

#### Why the 360° turn in approach back to the original proposal?
Off-list discussion has brought up the need for backwards compatibility, given the number of feeds that presently follow the spec and are integrated in present systems. This is an attempt to minimize disruption to existing feed producers and consumers.

#### Suggested path forward
Let's continue to discuss on GitHub for as long as it takes to arrive at a solution. Given that this has lingered for some time (and the original motivating problem has not been solved), I suggest we attempt to limit discussion to 2 weeks from today before bringing this proposal, or any revised proposal, to a vote.

gtfs-realtime/CHANGES.md
```diff
@@ -19,7 +19,6 @@ The official specification, reference and documentation are written in English.
     If a voter changes her vote, it is recommended to do it by updating the original vote comment by striking through the vote and writing the new vote.
   - Votes before the start of the voting period are not considered.
 1. The proposal is accepted if there is a unanimous consensus yes with at least 3 votes.
-  - The proposer's vote does not count towards the 3 vote total. For example, if Proposer X creates a pull request and votes yes, and User Y and Z vote yes, this is counted as 2 total yes votes.
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/CHANGES.md
```diff
@@ -5,22 +5,20 @@ The official specification, reference and documentation are written in English.
 
 1. Create a git branch with update of all relevant parts of protocol definition, specification and documentation files (except for translations).
 1. Create pull request on https://github.com/google/transit. Pull request must contain an extended description of the patch. The creator of the pull request becomes the _advocate_.
-1. Once pull request is registered, it must be announced by its advocate in the [GTFS Changes mailing list](https://groups.google.com/forum/#!forum/gtfs-changes), including a link to the pull request. Once announced, the pull request is considered a proposal.  The pull request should also be edited to contain a link to the Google Groups announcement so they can easily be cross-referenced.
+1. Once pull request is registered, it must be announced by its advocate in the [GTFS Changes mailing list](https://groups.google.com/forum/#!forum/gtfs-changes). Once announced, the pull request is considered a proposal.
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/CHANGES.md
```diff
@@ -5,22 +5,20 @@ The official specification, reference and documentation are written in English.
...
   - Since the advocate is a contributor, they must sign the [Contributor License Agreement](../CONTRIBUTING.md) before pull request can be accepted.
 1. The discussion of the proposal follows. Pull request comments should be used as the sole discussion forum.
   - The discussion lasts for as long as the advocate feels necessary, but must be at least 7 calendar days.
   - The advocate is responsible for timely update of the proposal (i.e. pull request) based on the comments for which they agree to.
   - At any point in time the advocate can claim proposal abandoned.
 1. The advocate can call for a vote on a version of the proposal at any point in time following the initial 7-day interval required for discussion.
-  - Before calling for a vote, at least one GTFS producer and one GTFS consumer should implement the proposed change.  It is expected that the GTFS producer(s) include the change in a public-facing GTFS feed and provide a link to that data within the pull request comments, and that the GTFS consumer(s) provides a link in the pull request comments to an application that is utilizing the change in a non-trivial manner (i.e, it is supporting new or improved functionality).
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/CHANGES.md
```diff
@@ -5,22 +5,20 @@ The official specification, reference and documentation are written in English.
...
 1. Vote lasts the minimum period sufficient to cover 7 full calendar days and 5 full Swiss business days. Vote ends at 23:59:59 UTC.
   - The advocate should announce the specific end time at the start of the vote.
   - During voting period only editorial changes to the proposal are allowed (typos, wording may change as long as it does not change the meaning).
   - Anyone is allowed to vote yes/no in a form of comment to the pull request, and votes can be changed until the end of the voting period.
     If a voter changes her vote, it is recommended to do it by updating the original vote comment by striking through the vote and writing the new vote.
   - Votes before the start of the voting period are not considered.
 1. The proposal is accepted if there is a unanimous consensus yes with at least 3 votes.
-  - The proposer's vote does not count towards the 3 vote total. For example, if Proposer X creates a pull request and votes yes, and User Y and Z vote yes, this is counted as 2 total yes votes.
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/CHANGES.md
```diff
@@ -31,7 +29,6 @@ The official specification, reference and documentation are written in English.
   - Google is committed to merging the voted upon version of the pull request (provided that the contributors have signed the [CLA](../CONTRIBUTING.md)), and performing the pull request within 5 business days.
   - Translations must not be included into the original pull request.
     Google is responsible for eventually updating relevant translations into supported languages, but pure translation pull requests from the community are welcome and will be accepted as soon as all editorial comments are addressed.
-1. The final result of the pull request (accepted or abandoned) should be announced on the same Google Groups thread where the pull request was originally announced.
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/CHANGES.md
```diff
@@ -196,4 +193,4 @@ Added trips.txt 'wheelchair_accessible' proposal to spec: [discussion](https://g
 
 #### September 25, 2006
 
-* Initial version.
+* Initial version.
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/examples/README.md
```diff
@@ -5,7 +5,7 @@ This example GTFS feed shows comma-delimited data samples for each file in a tra
 ### agency.txt
 
 ~~~
-agency_id,agency_name,agency_url,agency_timezone,agency_phone,agency_lang
+agency_id, agency_name,agency_url,agency_timezone,agency_phone,agency_lang
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/examples/README.md
```diff
@@ -127,4 +127,4 @@ from_stop_id,to_stop_id,transfer_type,min_transfer_time
 S6,S7,2,300
 S7,S6,3,
 S23,S7,1,
-~~~
+~~~
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/examples/sample-feed-1/stop_times.txt
```diff
@@ -1,4 +1,4 @@
-trip_id,arrival_time,departure_time,stop_id,stop_sequence,stop_headsign,pickup_type,drop_off_type,shape_dist_traveled
+trip_id,arrival_time,departure_time,stop_id,stop_sequence,stop_headsign,pickup_type,drop_off_time,shape_dist_traveled
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/examples/sample-feed-1/stop_times.txt
```diff
@@ -26,4 +26,4 @@ AAMV2,11:00:00,11:00:00,BEATTY_AIRPORT,2
 AAMV3,13:00:00,13:00:00,BEATTY_AIRPORT,1
 AAMV3,14:00:00,14:00:00,AMV,2
 AAMV4,15:00:00,15:00:00,AMV,1
-AAMV4,16:00:00,16:00:00,BEATTY_AIRPORT,2
+AAMV4,16:00:00,16:00:00,BEATTY_AIRPORT,2
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -85,44 +85,41 @@ File: **Required**
...
-|  stop_code | Optional | The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. Stop codes are often used in phone-based transit information systems or printed on stop signage to make it easier for riders to get a stop schedule or real-time arrival information for a particular stop.  The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. The **stop_code** can be the same as **stop_id** if it is passenger-facing. This field should be left blank for stops without a code presented to passengers. |  |  |
-|  stop_name | **Required** | The **stop_name** field contains the name of a stop, station, or station entrance. Please use a name that people will understand in the local and tourist vernacular. |  |  |
+|  stop_id | **Required** | The **stop_id** field contains an ID that uniquely identifies a stop or station. Multiple routes may use the same stop. The **stop_id** is dataset unique. |  |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -85,44 +85,41 @@ File: **Required**
...
+|  stop_code | Optional | The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. Stop codes are often used in phone-based transit information systems or printed on stop signage to make it easier for riders to get a stop schedule or real-time arrival information for a particular stop.  The stop_code field contains short text or a number that uniquely identifies the stop for passengers.  For internal codes, use **stop_id**. This field should be left blank for stops without a code. |  |  |
+|  stop_name | **Required** | The **stop_name** field contains the name of a stop or station. Please use a name that people will understand in the local and tourist vernacular. |  |  |
 |  stop_desc | Optional | The **stop_desc** field contains a description of a stop. Please provide useful, quality information. Do not simply duplicate the name of the stop. |  |  |
-|  stop_lat | **Required** | The **stop_lat** field contains the latitude of a stop, station, or station entrance. The field value must be a valid WGS 84 latitude. |  |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -85,44 +85,41 @@ File: **Required**
...
-|  location_type | Optional | The **location_type** field identifies whether this stop ID represents a stop, station, or station entrance. If no location type is specified, or the location_type is blank, stop IDs are treated as stops. Stations may have different properties from stops when they are represented on a map or used in trip planning.  The location type field can have the following values: |  |  |
+|  location_type | Optional | The **location_type** field identifies whether this stop ID represents a stop or station. If no location type is specified, or the location_type is blank, stop IDs are treated as stops. Stations may have different properties from stops when they are represented on a map or used in trip planning.  The location type field can have the following values: |  |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -85,44 +85,41 @@ File: **Required**
...
 |   |  | * **0** or blank - Stop. A location where passengers board or disembark from a transit vehicle. |  |  |
 |   |  | * **1** - Station. A physical structure or area that contains one or more stop. |  |  |
-|   |  | * **2** - Station Entrance/Exit. A location where passengers can enter or exit a station from the street. The stop entry must also specify a parent_station value referencing the stop ID of the parent station for the entrance. |  |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -85,44 +85,41 @@ File: **Required**
...
...d contains the timezone in which this stop, station, or station entrance is located. Please refer to [Wikipedia List of Timezones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) for a list of valid values. If omitted, the stop should be assumed to be located in the timezone specified by **agency_timezone** in [agency.txt](#agencytxt).   When a stop has a parent station, the stop is considered to be in the timezone specified by the parent station's **stop_timezone** value. If the parent has no stop_timezone value, the stops that belong to that station are assumed to be in the timezone specified by **agency_timezone**, even if the stops have their own **stop_timezone** values. In other words, if a given stop has a **parent_station** value, any **stop_timezone** value specified for that stop must be ignored.  Even if **stop_timezone** values are provided in stops.txt, the times in [stop_times.txt](#stop_timestxt) should continue to be specified as time since midnight in the timezone specified by **agency_timezone** in agency.txt. This ensures that the time values in a trip always increase over the course of a trip, regardless of which timezones the trip crosses. |  |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -85,44 +85,41 @@ File: **Required**
...
 |   |  | * **0** (or empty) - indicates that there is no accessibility information for the stop |  |  |
 |   |  | * **1** - indicates that at least some vehicles at this stop can be boarded by a rider in a wheelchair |  |  |
 |   |  | * **2** - wheelchair boarding is not possible at this stop |  |  |
 |   |  | When a stop is part of a larger station complex, as indicated by a stop with a **parent_station** value, the stop's **wheelchair_boarding** field has the following additional semantics: |  |  |
 |   |  | * **0** (or empty) - the stop will inherit its **wheelchair_boarding** value from the parent station, if specified in the parent |  |  |
 |   |  | * **1** - there exists some accessible path from outside the station to the specific stop / platform |  |  |
 |   |  | * **2** - there exists no accessible path from outside the station to the specific stop / platform |  |  |
-|   |  | For station entrances, the **wheelchair_boarding** field has the following additional semantics: |  |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -175,42 +172,26 @@ File: **Required**
 |   |  | * **1** - indicates that the vehicle being used on this particular trip can accommodate at least one bicycle |
 |   |  | * **2** - indicates that no bicycles are allowed on this trip |
 
-#### Example showing blocks and service day
```
barbeau: Is this related to this proposal?

gtfs/spec/en/reference.md
```diff
@@ -175,42 +172,26 @@ File: **Required**
...
...day on which the trip schedule begins. If you don't have separate times for arrival and departure at a stop, enter the same value for **arrival_time** and **departure_time**.<br/><br/>Scheduled stops where the vehicle strictly adheres to the specified arrival and departure times are timepoints. For example, if a transit vehicle arrives at a stop before the scheduled departure time, it will hold until the departure time. If this stop is not a timepoint, use either an empty string value for the **arrival_time** field or provide an interpolated time. Further, indicate that interpolated times are provided via the **timepoint** field with a value of zero. If interpolated times are indicated with **timepoint**=0, then time points must be indicated with a value of 1 for the **timepoint** field. Provide arrival times for all stops that are time points.<br/><br/>An arrival time must be specified for the first and the last stop in a trip. Times must be eight digits in HH:MM:SS format (H:MM:SS is also accepted, if the hour begins with 0). Do not pad times with spaces. The following columns list stop times for a trip and the proper way to express those times in the **arrival_time** field: |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -175,42 +172,26 @@ File: **Required**
...
...s for arrival and departure at a stop, enter the same value for **arrival_time** and **departure_time**.<br/><br/>Scheduled stops where the vehicle strictly adheres to the specified arrival and departure times are timepoints. For example, if a transit vehicle arrives at a stop before the scheduled departure time, it will hold until the departure time. If this stop is not a timepoint, use either an empty string value for the **departure_time** field or provide an interpolated time (further, indicate that interpolated times are provided via the **timepoint** field with a value of zero). If interpolated times are indicated with **timepoint**=0, then time points must be indicated with a value of 1 for the **timepoint** field. Provide departure times for all stops that are time points.<br/><br/>A departure time must be specified for the first and the last stop in a trip even if the vehicle does not allow boarding at the last stop.  Times must be eight digits in HH:MM:SS format (H:MM:SS is also accepted, if the hour begins with 0). Do not pad times with spaces. The following columns list stop times for a trip and the proper way to express those times in the **departure_time** field: |  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -348,8 +329,6 @@ For examples that demonstrate how to specify a fare structure with fare_rules.tx
 
 File: **Optional**
 
-Shapes describe the physical path that a vehicle takes, and are defined in the file shapes.txt. Shapes belong to Trips, and consist of a sequence of points. Tracing the points in order provides the path of the vehicle. The points do not need to match stop locations.
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -384,8 +363,8 @@ This table is intended to represent schedules that don't have a fixed list of st
 |  Field Name | Required | Details |
 |  ------ | ------ | ------ |
 |  trip_id | **Required** | The **trip_id** contains an ID that identifies a trip on which the specified frequency of service applies. Trip IDs are referenced from the [trips.txt](#tripstxt) file. |
-|  start_time | **Required** | The **start_time** field specifies the time at which the first vehicle departs from the first stop of the trip with the specified frequency. The time is measured from "noon minus 12h" (effectively midnight, except for days on which daylight savings time changes occur) at the beginning of the service day. For times occurring after midnight, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins. E.g. 25:35:00. |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -405,14 +384,35 @@ Trip planners normally calculate transfer points based on the relative proximity
 |  ------ | ------ | ------ |
 |  from_stop_id | **Required** | The **from_stop_id** field contains a stop ID that identifies a stop or station where a connection between routes begins. Stop IDs are referenced from the [stops.txt](#stopstxt) file. If the stop ID refers to a station that contains multiple stops, this transfer rule applies to all stops in that station. |
 |  to_stop_id | **Required** | The **to_stop_id** field contains a stop ID that identifies a stop or station where a connection between routes ends. Stop IDs are referenced from the [stops.txt](#stopstxt) file. If the stop ID refers to a station that contains multiple stops, this transfer rule applies to all stops in that station. |
+|  from_route_id | **Optional** | The from_route_id field can contain a route_id (as specified by [routes.txt](#routestxt)), reducing the scope to which the given transfer applies. If from_route_id is specified, the transfer will only apply to the arriving trip with the given route id, at the given from_stop_id.  |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -405,14 +384,35 @@ Trip planners normally calculate transfer points based on the relative proximity
...
+|  from_trip_id | **Optional** | The from_trip_id and to_trip_id fields can contain a trip_id, as specified by [trips.txt](#tripstxt). If from_trip_id is given, from_route_id is ignored. If from_trip_id is specified, the transfer will only apply to the arriving trip with the given trip id, at the given from_stop_id. |
+|  to_trip_id | **Optional** | The to_trip_id field can contain a trip_id, as specified by [trips.txt](#tripstxt). If to_trip_id is given, to_route_id is ignored. If to_trip_id is specified, the transfer will only apply to the departing trip with the given trip id, at the given to_stop_id.  |
 |  transfer_type | **Required** | The **transfer_type** field specifies the type of connection for the specified (from_stop_id, to_stop_id) pair. Valid values for this field are: |
 |   |  | * **0** or **(empty)** - This is a recommended transfer point between two routes. |
-|   |  | * **1** - This is a timed transfer point between two routes. The departing vehicle is expected to wait for the arriving one, with sufficient time for a passenger to transfer between routes. |
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
```diff
@@ -405,14 +384,35 @@ Trip planners normally calculate transfer points based on the relative proximity
...
-|   |  | * **3** - Transfers are not possible between routes at this location. |
+|   |  | * **3** - Transfers are not possible. |
+|   |  | * **4** - Passengers can transfer from one route to another by staying onboard the same vehicle (an "in-seat transfer") in cases where sequential trips are made by the same vehicle (same block_id).  In-seat transfer between these two stops / routes / trips is allowed. |
+|   |  | * **5** - In-seat transfers are not allowed between sequential trips with the same block ID. The passenger must alight from the vehicle and re-board. |
 |  min_transfer_time | Optional | When a connection between routes requires an amount of time between arrival and departure (transfer_type=2), the **min_transfer_time** field defines the amount of time that must be available in an itinerary to permit a transfer between routes at these stops. The min_transfer_time must be sufficient to permit a typical rider to move between the two stops, including buffer time to allow for schedule variance on each route. |
 |   |  | The min_transfer_time value must be entered in seconds, and must be a non-negative integer. |
 
+**Specificity of a transfer**
```
barbeau: Unrelated to this proposal - needs to be removed from PR.

gtfs/spec/en/reference.md
antrim: Hi @barbeau : The additional transfer types are intended to be part of this proposal/PR.  If block transfers are turned off by default for an agency (as many agencies will likely choose), then they will also need a way to override and allow block transfers in specific cases.

antrim:
@barbeau Your instructions worked perfectly. Thank you. The proposed changes are now all in 03a4592.
.

gtfs/spec/en/reference.md
antrim: Intentional to provide additional capabilities to manage transfers broadly, and block transfers as a subset.

barbeau: Ah, right - gotcha.

gtfs/spec/en/reference.md
barbeau: Ok, thanks - makes sense.

antrim:
No new objections have been raised. Let's go ahead and call a vote on this. To cover the full 7 required calendar days, this vote will end on Tuesday, 6/13 at 23:59:59 UTC.

barbeau:
+1

abyrd:
Thanks for all your work on this proposal. I am hesitant to prioritize backward compatibility when it increases complexity, and when past behavior was never well-defined in the first place. If blocks imply in-seat transfers by default, every consumer will need to include heuristic logic for generating these in-seat transfers. Do we have any examples of feeds that use block_id to create in-seat transfers, where the feed publisher is not planning to export explicit in-seat transfer records in transfers.txt?

DaveBarker:
I vote yes.

joovel:
:+1:

Do we have any feed producers ready to provide this, or ones that are planning to in the near-term?

skinkie:
I'll see what I can do (and more importantly: if Google Transit don't choke on it)

RachM:
I vote yes.

RachM:
@abyrd: I'm a strong advocate on maintaining backwards compatibility on this issue for a few reasons:
1) I'm not convinced that this proposal adds significantly more complexity than the equivalent non-backwards compatible change. Furthermore, confusions always ensues when compatibility is broken, so I'd argue this proposal is clearer for consumers.
2) Getting producers to change is hard; this change will give producers an opportunity to clarify their intentions, without disadvantaging those who aren't able to adapt as easily.
3) I know of many producers that use block ids to indicate in-seat transfers. Indeed, almost all of our (Google's) high priority feeds currently use block ids to indicate in-seat transfers.

Overall, these changes allow flexibility, backwards compatibility, and importantly, aren't overly complex for producers or consumers to use. Seems like a win win to me.

antrim:
@abyrd : Trillium did a very limited survey of GTFS feeds w.r.t. blocks and in-seat transfers: https://docs.google.com/spreadsheets/d/1v8X7DPq2pc6sKGK4OgxiheNWpFIYOsfpvE4ZLtjXW4U/edit#gid=279055228. Again, I acknowledge this was very limited survey. It revealed varied practices and expectations. This proposal is also partially informed by the internal surveys of other GTFS feed consumers.

My interest as a feed producer is to see this clarified in a way that major feed consumers accept. In some cases currently, we need to publish different feeds for different applications (as MBTA and several other agencies also report).

antrim:
@joovel Trillium (feed producer) will implement this in its final approved form, shortly after it is approved.

barbeau:
According to the official process we need to have a producer and consumer
prior to this being accepted into the spec. Technically this should happen
before the vote:
https://github.com/google/transit/blob/master/gtfs/CHANGES.md

On Jun 7, 2017 9:32 AM, "Aaron Antrim" <notifications@github.com> wrote:

@joovel <https://github.com/joovel> Trillium (feed producer) will implement
this in its final approved form, shortly after it is approved.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/google/transit/pull/32#issuecomment-306795758>, or mute
the thread

drewda:
Mapzen can volunteer to support consuming this change within [Transitland](https://transit.land) (aggregation and querying of GTFS data) and [Valhalla](https://github.com/valhalla/valhalla) (routing engine powered by Transitland data).

abyrd:
Hi @RachM, thanks for your response. The fact that most of your high-priority feeds use the existing block-transfer feature is important, I didn't realize that. In that case I agree, we don't want to break backward compatibility.

mgilligan:
TriMet will also work on producing this data next week.

mike-swiftly:
I vote yes on this change on the behalf of Swiftly.

Note with respect to real-time prediction generation systems: while a motivation for making this change is to make sure that block_ids can be used in generating real-time predictions and other such information, I don't believe that such systems need to actually be modified to accommodate this change. Swiftly Transitime system for example has already long assumed that blocks are for tying together trips and do not relate to in-seat transfers. Once this change is implemented by GTFS producers then this assumption will always hold true. 

Instead, this change directly affects trip planners, which are quite different from real-time information systems. Therefore changes to trip planners will need to be made in order to generate proper trip plans.

If anybody has a difference in opinion with respect to this change and real-time information generation, please let us know!

barbeau:
Hi all - just wanted to check in and see if this had been implemented by any consumers or producers yet.

antrim:
Since this has been open for some time, it seems prudent to revisit.

Trillium has not yet implemented as a producer yet, though it is planned in our GTFS Manager software roadmap. Has any other producer implemented? @mgilligan, has TriMet made any progress on this (or would there be need for TriMet to suppor this)? Looks like the transfers.txt file follows current spec: https://transitfeeds.com/p/trimet/43/latest/file/transfers.txt

Finally, Google has documented its support for Route-to-route and trip-to-trip transfers (https://developers.google.com/transit/gtfs/reference/gtfs-extensions#TripToTripTransfers) so I assume that some feeds producers are providing this data. Could a feed publisher or someone at Google or Mapzen point us to those feeds?

I suspect the new field for which there will not be immediate support is agency.block_transfer_default. We could remove that to bring this to vote, but I think it will be better if we are able to vote on this entire proposal as a package.

mgilligan:
@antrim, unfortunately I haven't had any time to implement the in-seat transfers in TriMet's GTFS and I don't think I will any time to.

gtfs/spec/en/reference.md
```diff
@@ -405,13 +408,34 @@ Trip planners normally calculate transfer points based on the relative proximity
...
 |   |  | * **0** or **(empty)** - This is a recommended transfer point between two routes. |
-|   |  | * **1** - This is a timed transfer point between two routes. The departing vehicle is expected to wait for the arriving one, with sufficient time for a passenger to transfer between routes. |
+|   |  | * **1** - This is a timed transfer point between two vehicles. The departing vehicle is expected to wait for the arriving one, with sufficient time for a passenger to transfer between routes. |
 |   |  | * **2** - This transfer requires a minimum amount of time between arrival and departure to ensure a connection. The time required to transfer is specified by **min_transfer_time**. |
-|   |  | * **3** - Transfers are not possible between routes at this location. |
+|   |  | * **3** - Transfers are not possible. |
+|   |  | * **4** - Passengers can transfer from one route to another by staying onboard the same vehicle (an "in-seat transfer") in cases where sequential trips are made by the same vehicle (same **block_id**).  In-seat transfer between these two stops / routes / trips is allowed. |
```
DaveBarker: Does the reference to block_id mean that block_id must be included to use this transfer_type? 

Which of the following are valid uses of transfer_type 4?
A. transfers.txt contains route_id's and stop_id's. The trips on the routes do have block_id's. 
B. transfers.txt contains trip_id's. The trips on the routes do have block_id's. 
C. transfers.txt contains trip_id's. The trips on the routes DO NOT have block_id's. 
D. transfers.txt contains route_id's and stop_id's. The trips on the routes DO NOT have block_id's.

abyrd:
@DaveBarker if these transfers were provided the block ID would not even be necessary to convey the information. Just the transfer entry is sufficient for all routing and passenger information applications. 

So I would say that all of your four cases should be valid. But this needs to be made clear in the spec.

I would not be in favor of the spec requiring block_ids in the case that type-4 transfers are specified; if block IDs are included, they are basically redundant information that allows checking the coherency of the whole data set. They are also useful for propagating real-time delays.

paulswartz:
Quick update from @mbta:

- we are using trip-to-trip transfers in the feed we provide to Google, and will be providing them in the public feed in the short/medium term
- we could provide the in-seat transfer data as well, but it seems like the specification would need to be clarified

antrim:
Sounds like there are some implementation possibilities. Let's see if we can chart forward a path for implementations and spec amendment.

@paulswartz : Would @mbta support the `block_transfer_default` field as it is specified?

skinkie:
@antrim do you have any reference how blocks should be interpreted in GTFS? Hence, how would a consumer derive a block pattern? I am figuring out if the schedule should be duplicated to implement this at trip level, while for the rider the scheduled times remain the same.

gtfs/spec/en/reference.md
antrim: I agree with @aybrd that all these uses should be valid.

@abyrd: I think you're right that block_id should not be required in the case that transfer_type=4 is specified, but I still anticipate that feed producers may want to supply block_id to propagate vehicle delays as you suggest, to configure real-time passenger information systems (e.g. @TheTransitClock), and for other purposes.

antrim:
@skinkie: Sorry, I'm not following completely. Can you give an example of how or when the schedule might be duplicated? Then I can delve into that.

skinkie:
@antrim 
given that my weekday schedule on friday has a deviation to facilitate night life, typically the block_id would change, since the series of trips following each other change.
Given that in my weekly pattern have on all weekdays a bus departing at 23pm with trip_id=trip23pm.
Now because my block_id becomes an attribute that is a references outwards [blockref], instead of inwards [servicejourneyref] I must end up with:

trip_id=trip2300pm, service_id=mo-no-th block_id=bus1weekly
trip_id=trip2300pm-friday, service_id=friday block_id=bus1weekly-friday
trip_id=trip2330pm-friday, service_id=friday block_id=bus1weekly-friday

(and now envision the duplication in stop_times.txt for exactly the same timetable, for all trips the vehicle makes over the entire day)

Where what I would like to have is:
trip_id=trip2300pm, service_id=weekdays
trip_id=trip2330pm, service_id=friday

block_id=bus1weekly, service_id=block1weekly, trip_id=trip2300pm
block_id=bus1weekly-friday, service_id=block1weekly-friday, trip_id=trip2300pm
block_id=bus1weekly-friday, service_id=block1weekly-friday, trip_id=trip2330pm

(and I can even argue that people would still get angry when they would have to change a single other attribute in trip and therefore create a new blocking pattern because the reference changes or tricks with validity have to be applied)

Doing blocks this naive will never work for any serious operation, unless data duplication becomes the new norm.

antrim:
@skinkie 
Fortunately, that is defined! And service days can be "layered" - where various trips have different service_id but share same block_id. Definition:
https://github.com/google/transit/blob/fef4def9f684f4f6c8d04dad2d5f0c1658fd4fd3/gtfs/spec/en/reference.md#example-blocks-and-service-day

2017 discussion related to this change: https://github.com/google/transit/pull/44

skinkie:
@antrim the way it is currently defined results in the duplication.

antrim:
@skinkie Yes, my recollection was completely wrong. How embarrassing, as I even authored the clarification to make the spec consistent with consumer expectations. I see the value in what you describe -- though I expect we're stuck with vehicle block being defined as it is.

paulswartz:
> @paulswartz : Would @mbta support the `block_transfer_default` field as it is specified?

We'd want to be sure that Google Maps is going to use the new data. Our public feed has block information, but we strip most of it for Google out so that Google Maps doesn't show transfers we don't allow. We also have the issue @skinkie raises, where a block_id can be active on different service IDs.

devinbraun:
@antrim We can make the change block_transfer_default immediately as our block_ids are vehicle tasks (for real-time information) and aren't specific to in-seat transfers which we don't allow except if it's another trip on the same route (for example, a looping route).  We can add in the route-to-route in-seat transfers to transfers.txt for those few routes where it is allowed.

davidlattimore:
Something I had been thinking about, that's related to this, is when two trips share a block_id, but the later trip starts from a different stop than the earlier trip finishes. Currently Google still allows an in-seat transfer. I'm wondering however if perhaps it shouldn't. I've observed at least some data providers that have given us such trips who didn't intend for them to be in-seat transfers.

[GTFS best practices](https://gtfs.org/best-practices/#transferstxt) says:

> If in-seat (block) transfers are allowed between trips, then the last stop of the arriving trip must be the same as the first stop of the departing trip.

I wonder what proportion of unwanted in-seat transfers would go away if we only allowed in-seat transfers when the last/first stops matched up.

Regardless of the more explicit rules in this proposal as to whether in-seat transfers are allowed, it'd be good to clarify the expected behavior when the stops don't match. Is it an error? Is it OK, but in-seat transfers are disallowed, even if the default is to allow them? The best practices doc seems to suggest the latter.

Also, there's still a Non-Backwards-Compatible label on this pull request. My impression is that it's no longer required, right?

antrim:
tl;dr: Should we even allow blocked trips where the last stop of the earlier trip is different than the first stop of the subsequent trip? This strikes me as problematic and unnecessary. Longer discussion below.

Quoting @davidlattimore :

> Something I had been thinking about, that's related to this, is when two trips share a block_id, but the later trip starts from a different stop than the earlier trip finishes. Currently Google still allows an in-seat transfer. I'm wondering however if perhaps it shouldn't. I've observed at least some data providers that have given us such trips who didn't intend for them to be in-seat transfers.

Question to all -- particularly feed producers: What would be the purpose of specifying subsequent trips where the later trip starts from a different stop than the earlier trip finishes at? @davidlattimore: Are there particular feeds you can point to that do this?

How should a feed consumer interpret a case where the last stop of the earlier (blocked) trip and first stop of the subsequent trip are different? This strikes me as problematic, especially in cases where a real-time system is consuming the GTFS. Opinions / insights?: @scrudden @mike-swiftly @barbeau @nselikoff

If the vehicle goes into non-revenue service, that can be represented in GTFS (see Best Practices RE deadhead trips: https://gtfs.org/best-practices/#stop_timestxt).

skinkie:
>  Should we even allow blocked trips where the last stop of the earlier trip is different than the first stop of the subsequent trip? This strikes me as problematic and unnecessary. Longer discussion below.

If Trip 1 ends at a drop-off only stop, and starts at the pick-up that is in The Netherlands at least, a very common situation to get disembarking passengers as close to the train entrance as possible, while picking up new passengers at a dedicated platform. The block will not start and end in the same stop.

davidlattimore:
Sure. Here's a couple of feeds:
https://bkk.hu/gtfs/budapest_gtfs.zip
http://opendata.tec-wl.be/Current%20GTFS/TEC-GTFS.zip
I have no idea if they're representative. Both have a mix of block transfers where the first and last stop match and ones that don't.

skinkie:
@antrim a compromise might be to state that a block should end and begin at the same parent station. Given the entrances proposal, every station will get one anyway.

mike-swiftly:
Sorry for the late reply, but with interlining, where a vehicle might cover
multiple routes with a single block could also cause differences between
last stop of a trip and first stop of next trip. Interlining is pretty
common for larger agencies.

*Michael Smith*
*CIO, **Swiftly, Inc*
(415) 260-4700

gtfs/spec/en/reference.md
abyrd: The text does need to be updated to remove the reference to block_id, which only applies to trips. This PR includes elements of the transfer specificity proposal, which would allow transfer_type 4 transfers to be created between routes or stops, not just trips. It definitely makes sense for a producer using transfer_type 4 to still include block_ids for realtime propagation purposes, but it doesn't seem relevant to the present section of the spec.

abyrd:
Thanks @mike-swiftly. It also seems completely normal to me to allow in-seat transfers between sequential trips in blocks, independent of whether the last stop of trip 1 is the first stop of trip 2. I would expect this to be very common, and maybe even the most common case (any time route 2 begins at a stop a bit farther down the road from where route 1 ends).

I am not sure what we would accomplish by creating more rules about when in-seat transfers are allowed or not. It seems like any such set of more complicated heuristics would give better results for certain agencies and worse results for others, while requiring every consumer to implement more complex code, so I'd prefer the simplest possible behavior. When the broad backward-compatible default is used to enable in-seat transfers, I would expect in-seat transfers to then exist after every trip within each block. Each consumer could perhaps implement a few sanity-check heuristics of their own design for good measure, but such things should not be part of the spec.

If a producer needs to communicate more nuanced information they can disable in-seat transfers by default and employ transfer_type 4. Based on input in the comments above about existing feeds, we do need a field creating backward-compatible behavior, and allowing producers to switch that behavior off and use transfer_type 4.

abyrd:
I have just re-read most of the comments and the current content of the pull request. I'm still quite supportive of having a way to individually enable in-seat transfers, using transfers.txt. Some comments on the current state of the proposal:

## transfer_type 4 and block_id

The text describing transfer_type 4 should be updated to clarify the reference to block_id. To me, block_ids do not seem relevant when a producer is listing individual transfers instead of using the backward-compatible block-based default in-seat transfer behavior. Also, consider that the block_id field only exists for trips. This PR allows transfer_type 4 transfers to be created between routes or stops, not just trips, in which case that transfer should (?) still be available to the rider even in the absence of block_ids. Alternatively, if you really do want to enforce creation of in-seat transfers only between trips with the same block_id even when the enabling transfers.txt entry is between stops or routes, then this behavior needs to be very clearly defined in the spec, including a description of what happens when trips do not have defined block_ids but are nonetheless tied together by a transfer_type 4 transfer. Some additional information should also be given clarifying whether the transfer_type 4 only applies when it connects the last stop of the first trip to the first stop of the next trip (i.e. what if the transfer is between stops that are not at the end of a trip).

My proposed system (which of course needs scrutiny from other people) is this: When default in-seat transfers are enabled, in-seat transfers are available after every trip in every block, with no restrictions. Whether or not this default is enabled, transfers with the new transfer type have the same function: they allow in-seat transfers from the final stop of all trips that meet its from_stop_id, from_route_id, and from_trip_id criteria, toward the initial stop of all trips that meet its to_stop_id, to_route_id, and to_trip_id criteria, with no regard to block_ids at all. In this way, the new transfer_type can be used to exhaustively specify all in-seat transfers, or to augment default in-seat transfers inferred from block_ids.

It definitely makes sense for a producer using transfer_type 4 to still include block_ids for realtime propagation purposes, but how block_id meaning and effect changes in response to the default in-seat transfer flag must be clearly described.

## transfer_type 5

I am thinking transfer_type 5 should be eliminated. This leaves only two modes of operation: default in-seat transfers enabled, augmented by any transfer_type 4 transfers; or default in-seat transfers disabled, with all in-seat transfers exhaustively listed with transfer_type 4. Since the mode with default in-seat transfers enabled exists primarily for backward compatibility, I am hesitant about additional complexity allowing producers to subtractively pare down the full default set of in-seat transfers.

gcamp:
I tried to read as much as possible from this propositions and comments, but I didn't read everything. I'm sorry if I'm repeating discussion that happened before. 

My biggest problem with `block_id` is how they are hard to interpret since they can change depending on the current service day, which is a pain (see this [example](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#example-blocks-and-service-day)). The trip to trip transfer solves this. 

My proposition would be to actually **deprecate completely block_id and replace it with trip to trip with transfers type 4 and 5**. If block_id are deprecated, that also means we can remove the `block_transfer_default` value. `from_route_id` and `to_route_id` are also not required in this PR and can be moved to an other proposition. 

If that makes sense for everybody we can implement this fairly quickly in Transit. We can also potentially open source a convertor that transforms GTFSs with `block_ids` with GTFSs with trip to trip transfers. I know this is a big change, but in my opinion it would really clarify the spec and reduce the amount of potential bugs in both producers and consumers.  

For those interested, we came to this conclusion because we are actually fixing bugs in our `block_id` support (actually bug which are basically the [example](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#example-blocks-and-service-day) from the spec). Our final solution was basically have a first step of transformation of block_id to trip to trip transfers inspired by this proposition to simplify processing later on. Instead of duplicating work, we could just share our transformation step and hopefully become part of the spec.

skinkie:
@gcamp I consider this certainly a very good direction to go to. Considering how blocks in GTFS never represent what they are in more complex systems downstream unless all data is denormalised. But I do want to suggest something that I hope you could take in a account. Would your consider adding an optional service_id to the trip_to_trip transfers?

gcamp:
@skinkie could you have the same trip to trip transfer with multiple service_id? You want that to avoid duplication of trips just to create different continuations? 

In our case we would want to duplicate the trip anyway to simplify storage of trip to trip transfers, but that can be done on our side. If this reduce size and complexity a lot for producers, that's something we can support.

skinkie:
@gcamp I indeed wat to prevent that new trips are created (while the internal trip identifiers are equal) that on certain dates will not share the same block id therefore would not have that trip-to-trip transfer.

gcamp:
Bump on proposition above ☝️

paulswartz:
Would this deprecate `block_id` only for in-seat transfers, or entirely? I imagine it is still useful for consumers to know that two different trips are scheduled to be provided by the same vehicle.

gcamp:
@paulswartz I would deprecate entirely. If the producer wants to provide the information about two different trips are provided by the same vehicle with a trip to trip transfer type = 5.

paulswartz:
@gcamp 
> @paulswartz I would deprecate entirely. If the producer wants to provide the information about two different trips are provided by the same vehicle could do so with a trip to trip transfer type = 5.

Hmm: our current GTFS has ~85k trips on ~5k blocks; that's a lot of trip-to-trip transfers to add to the file. But if that's what the consensus is, we can make it work.

antrim:
Closing this PR. This function is covered in #278.

PR #33, opened 2016-11-26
correction to interpolated times language

antrim:
make arrival_/departure_time field definition consistent with
definitions for stop_times.timepoint.


antrim:
Before the introduction of the **stop_times.timepoint** field, the Spec instructed "Do not interpolate stops." Today, this statement remains in the Spec.

Stop time interpolation is now widespread and recommended for many applications. This is a proposal to make GTFS internally-consistent (with timepoint field definition) and consistent with widespread and recommended practice.

abyrd:
I agree with the proposed change. Operators can often provide a much better sense of expected arrival times at non-timepoint stops than the heuristics in a GTFS consumer application. The timepoint field is generally a better solution than leaving the arrival and departure times blank. One part that could use some additional clarification is "Stops without arrival times will be scheduled based on the nearest preceding timed stop", but I suppose the precise behavior here (how to provide useful and accurate instructions to end users outside of timepoints) is part of what makes each GTFS consumer application unique.

antrim:
> One part that could use some additional clarification is "Stops without arrival times will be scheduled based on the nearest preceding timed stop"

@abyrd : Do you think we should just remove this and let consuming applications decide what's best? I believe the Google Maps behavior for a stop without times (for awhile anyway -- I haven't checked recently) has been to give the arrival time of the next timed stop as the arrival (drop-off) time, and the departure time of the preceding stop as the departure (pick up) time.

gtfs/spec/en/reference.md
```diff
@@ -176,7 +176,7 @@ File: **Required**
...
...oon minus 12h" (effectively midnight, except for days on which daylight savings time changes occur) at the beginning of the service date. For times occurring after midnight on the service date, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins. If you don't have separate times for arrival and departure at a stop, enter the same value for **arrival_time** and **departure_time**.  If this stop isn't a time point, use an empty string value for the **arrival_time** and **departure_time** fields or provide an interpolated time, indicated by providing a value of 0 for the **timepoint** field. Stops without arrival times will be scheduled based on the nearest preceding timed stop. To ensure accurate routing, please provide arrival and departure times for all stops that are time points. You must specify arrival and departure times for the first and last stops in a trip.  Times must be eight digits in HH:MM:SS format (H:MM:SS is also accepted, if the hour begins with 0). Do not pad times with spaces. The following columns list stop times for a trip and the proper way to express those times in the **arrival_time** field: |  |
```
barbeau: >Stops without arrival times will be scheduled based on the nearest preceding timed stop

I vote for removing this.  This is the behavior of one specific GTFS-consuming application (Google Maps), and other applications may have different (and potentially improved) ways to interpolate/extrapolate times.  I believe it's also possible for the first stop in the trip to have `timepoint=0` (I recall someone having this case), in which case the above method isn't possible.

barbeau:
I agree with this proposed change as well - we missed clarifying this language in the original timepoint proposal (https://groups.google.com/forum/#!msg/gtfs-changes/Ah-J9JP2rJY/PxP0_g_NBkIJ).

antrim:
I just added commits that address comments from @barbeau and @abyrd .

antrim:
Does anyone have other comments before we vote?

gtfs/spec/en/reference.md
```diff
@@ -176,14 +176,14 @@ File: **Required**
...
... changes occur) at the beginning of the service date. For times occurring after midnight on the service date, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins. If you don't have separate times for arrival and departure at a stop, enter the same value for **arrival_time** and **departure_time**.  If this stop isn't a time point, use an empty string value for the **arrival_time** field or provide an interpolated time, indicated by providing a value of 0 for the **timepoint** field. If interpolated times are indicated with timepoint=0, then times that are strictly adhered to by the transit vehicle (i.e., timepoints) must be indicated with a value of 1 for the timepoint field. To ensure accurate routing, please provide arrival and departure times for all stops that are time points. You must specify arrival and departure times for the first and last stops in a trip.  Times must be eight digits in HH:MM:SS format (H:MM:SS is also accepted, if the hour begins with 0). Do not pad times with spaces. The following columns list stop times for a trip and the proper way to express those times in the **arrival_time** field: |  |
```
fpurcell: If this stop is not a timepoint, use either an empty string value for the **arrival_time** field or provide an interpolated time (further, indicate that interpolated times are provided via the **timepoint** field with a value of zero). If interpolated times are indicated with timepoint=0, then times that are strictly adhered to by the transit vehicle (i.e., timepoints) must be indicated with a value of 1 for the **timepoint** field. To ensure accurate routing, please provide arrival and departure times for all stops that are time points.

devinbraun:
Should it be specifically noted that having timepoint=1 means the bus will not depart until the  departure_time of the record?  

I also regarding the sentence "You must specify arrival and departure times for the first and last stops in a trip."  I believe this should be worded as, "You must specify a departure time for the first stop in a trip, and an arrival time for the last stop in a trip."  Does having an arrival_time for the first stop in a trip make sense?  I'm guessing feed consumers might use it to signify when a vehicle arrives and begins loading before the trip begins.  I'm also not sure about a departure_time at the last stop of a trip.  I can't think of a reason that would be needed as the next trip's departure time would be signified as the first departure_time of a new trip.

barbeau:
>"You must specify arrival and departure times for the first and last stops
in a trip."

This has always been a confusing part of the spec.  However, I think there
are a lot of consumers that assume this data exists.  IMHO changing this
should be part of a GTFS v2.

abyrd:
@devinbraun I would agree that the meaning of "timepoint" should be made clear. Unfortunately for the departure and arrival times at first and last stops, as @barbeau said this line appears to be protecting consumers who assume any non-empty stoptimes will have both an arrival and departure time, and for interpolation purposes the first and last stop must be non-empty. If this is the case, then the sentence should still be clarified, because it can be interpreted the way you saw it. Instead, "You must specify both an arrival time and a departure time for the first and the last stop in a trip."

barbeau:
>Should it be specifically noted that having timepoint=1 means the bus will not depart until the departure_time of the record?

@devinbraun Yes, good catch! That's actually very important if we're going to use timepoint to more intelligently propagate real-time info (i.e., not propagate early predictions past a timepoint).

gtfs/spec/en/reference.md
antrim: @fpurcell , I am going to to incorporate your language with a change, below. I removed the reference to "accurate routing" because GTFS is now used for applications besides trip planning.

@devinbraun , @abyrd : What do you think of this definition of the a time point?

> If this stop is not a timepoint, use either an empty string value for the **arrival_time** field or provide an interpolated time (further, indicate that interpolated times are provided via the **timepoint** field with a value of zero). If interpolated times are indicated with **timepoint**=0, then stops at which the vehicle will hold until the scheduled departure time (i.e., timepoints) must be indicated with a value of 1 for the **timepoint** field. Provide arrival and departure times for all stops that are time points.

Interesting note: GTFS is referenced in the [Wikipedia article "Timing Point".](https://en.wikipedia.org/wiki/Timing_point)

antrim:
Spec:
> "You must specify arrival and departure times for the first and last stops in a trip."

@barbeau @devinbraun : This has bothered me for years, because it means that we sometimes have to supply meaningless values for first arrival or last departure time. Should we open up an issue to discuss this? I think we'd need to find out what applications require this and work with them if we were to propose a change which is not backwards-compatible.

antrim:
I'd like to officially call for a vote now given this proposal was amended Dec 4 in response to comments. As per this spec amendment process, the vote lasts 7 days.
https://github.com/google/transit/blob/master/gtfs/CHANGES.md#specification-amendment-process

barbeau:
@antrim Thanks!  I support this proposal, although I think something went wrong with the formatting in these changes - check out https://github.com/trilliumtransit/transit/blob/9683278f015ecb32a78f8a493fcc9ab43f0ea105/gtfs/spec/en/reference.md.  It looks like all fields after `arrival_time` aren't in the table format.

Also, it seems like there is a duplicate section of "If this time isn't a timepoint" under `arrival_time`.  I'd recommend cutting the middle paragraph in the changes and making sure all those points are in the following paragraph?

Also, there is a sentence `Provide arrival times for all stops that are time points.`  I think this should be changed to `Provide arrival and departure times for all stops that are time points.`?

antrim:
I remedied the formatting issue: https://github.com/trilliumtransit/transit/blob/e353215d377013ea9024307ec6261dd46ca32d12/gtfs/spec/en/reference.md#stop_timestxt

A departure time must be specified for the first and the last stop in a trip even if the vehicle does not allow boarding at the last stop.

>Also, there is a sentence Provide arrival times for all stops that are time points. I think this should be changed to Provide arrival and departure times for all stops that are time points.?

There was duplicated information between the definitions for arrival_time and departure_time before. So I made statements refer to each field separately in their definitions. There is an equivalent statement for **departure_time**:
> Provide departure times for all stops that are time points.

barbeau:
@antrim Thanks!  Looks much better.

I'd propose one final change here to better align with the `stop_times.timepoint` field definition:

>Scheduled stops where the vehicle will hold until the departure time are time points.

should be changed to:

>Scheduled stops where the vehicle strictly adheres to the specified arrival and departure times are timepoints.  For example, if a transit vehicle arrives at a stop before the scheduled departure time, it will hold until the departure time.

The current text addresses adherence to the departure time but doesn't say anything about the arrival time.  The modified text addresses arrival times as well with the same language as `stop_times.timepoint`.  This is important if we want to use `arrival_time` for calculating on-time performance and determining if a vehicle arrived early/late to a stop.

antrim:
Good changes, @barbeau. I checked them in. Since we are still discussing & amending does that cancel the voting period?

barbeau:
@antrim Yes, I think we should reset the voting period as there were changes that aren't just formatting.  So if you want to call for a vote again now here and on gtfs-changes I believe we can just restart it - luckily no one has voted yet anyway.

antrim:
Ok. Let's start the voting period again, then, unless anyone has other comments. So that there are 7 full calendar days, then the vote would end on Friday, 1/20 @ 23:59:59 UTC.

barbeau:
+1!

andreyz:
+1

LeoFR:
+1

fpurcell:
+1

abyrd:
A belated commentary on @barbeau's new version of the text: The text was updated to say that at timepoints, "the vehicle strictly adheres to the specified arrival and departure times." It seems to me that the text was correct before when it only mentioned strict adherence to departure times. Generally it is unimportant and harmless to passengers when vehicle arrives early. But in many places timepoints exist specifically to transform early or late arrivals into on-time departures. It is fully expected that the arrival time will not be respected (when the departure time on the previous non-timepoint stop was not respected) but the departure time must nonetheless be respected.  

I would prefer the text: "Scheduled stops where the vehicle strictly adheres to the specified departure time are timepoints. If a transit vehicle arrives at a timepoint before the scheduled departure time, it will hold until that departure time."

abyrd:
Another detail: "except for days on which daylight savings time changes occur". Although colloquially referred to as "daylight savings time" it's always written "daylight saving time" (with no S on "saving").

barbeau:
@abyrd if I'm calculating schedule deviation, though, especially looking at historical vehicle data, I need to know if a vehicle arrived late at a timepoint stop. If there is a 10 minute layover and the vehicle arrives 8 min late, as you say it can and should hold to depart, but the late arrival should still be captured based on strict adherence to the arrival time.

abyrd:
@barbeau these are two different meanings for "strictly adhere" though. "Strictly adhere to departure time" is an action: in practice vehicles will adopt a specific behavior, which must be accounted for in real-time delay propagation. "Strictly adhere to arrival time", on the other hand, is a normative statement: that it's considered "against the rules" to arrive late and potentially sanctioned with fines etc. A vehicle will not take some action to ensure that it does not arrive at the timepoint before the scheduled arrival time. A late arrival would not in itself have an effect on routing, delay propagation, or passenger information display, and is therefore not so relevant to a typical GTFS consumer.

barbeau:
>these are two different meanings for "strictly adhere" though.

True

>and is therefore not so relevant to a typical GTFS consumer

So I think this is the key question :) - the scope of consumers is quickly changing, and we're seeing feedback loops now as well, where GTFS is becoming the foundation of real-time systems and analytics.  For example, a GTFS dataset and raw vehicle location info is used with the TransiTime open-source project to generate arrival time predictions.  TransiTime also captures actual arrival and departures, and uses that to generate schedule deviation values and on-time performance metrics.  That information is then used to adjust the schedules/GTFS data and/or used to generate better real-time predictions, which does have an effect on customer information systems (although definitely more latent than the departure time example).  I'd argue that there is no harm (to my knowledge - but let me know if there is) in including language about strictly adhering to the arrival time and it does fall within the scope of existing GTFS consumers.

Also note that the language proposed here simply mirrors the existing "strictly adhere to arrival/departure times" language in stops_times.txt for timepoint field there.  So if we want to remove the reference to arrival times here, we'd need to remove it there too.

abyrd:
@barbeau thanks for the clarifications, and of course I am happy to see GTFS being used in new ways. I'd say the key question/issue is not the relevance to the "typical" consumer (who we don't know) but just clarity on what a timepoint is and means (the first point you quoted above). "Strictly adhere" is poorly defined, yet the same expression is applied to both arrivals and departures. This does not give much information to the GTFS consumer about how to interpret and make use of a timepoint field.

I think it is fairly common that the operator is in fact not bound to adhere to the arrival time at a timepoint, and "late" arrival is not sanctioned. Consider the case you mentioned, where there's a 10 minute layover and the vehicle arrives 8 minutes late. It's entirely possible that this is considered normal (not late) operation, as the operator is given 10 minutes of slack at this timepoint to compensate for traffic conditions. @skinkie isn't this how buses function in the Netherlands?

So again, we do need language explaining that buses will hold for departure, but I'm not sure there's agreement across all networks about how vehicles are constrained by timepoint arrival times.

skinkie:
Indeed. The only late arrival that is sanctioned is the final stop.

abyrd:
> On 19 Jan 2017, at 04:47, Sean Barbeau <notifications@github.com> wrote:
> Also note that the language proposed here simply mirrors the existing "strictly adhere to arrival/departure times" language in stops_times.txt for timepoint field there. So if we want to remove the reference to arrival times here, we'd need to remove it there too.
> 
Agreed, the language would need to be changed in both places. Since there is general approval for this pull request as it is, and it simply mirrors language present elsewhere, I’d be happy to push this one through and start another PR to redefine “timepoint”.

Andrew

barbeau:
>I’d be happy to push this one through and start another PR to redefine “timepoint”.

Sounds good, let's get this approved for consistency and then we can hash out a more exact definition - see also https://groups.google.com/forum/#!topic/transit-developers/dwd96EwJqIc for confusion related to the current timepoint text.

antrim:
We just hit the end of the voting period it looks like there is broad support for the Spec change. Let's close this and amend the spec. @zhsh Can you please merge this?

Also, there is work to do to define what the stop_times.timepoint field means. Quoting @abyrd,
> Agreed, the language would need to be changed in both places. Since there is general approval for this pull request as it is, and it simply mirrors language present elsewhere, I’d be happy to push this one through and start another PR to redefine “timepoint”.

PR #34, opened 2016-12-12
Add missing agency_email field

barbeau:
`agency_email` was missed in the original HTML->Markdown conversion - it exists in the current HTML version at https://developers.google.com/transit/gtfs/reference/agency-file, and in the Revision History (https://developers.google.com/transit/gtfs/guides/revision-history) added Feb 3, 2016.


barbeau:
Thanks!

PR #36, opened 2016-12-25
correct example arrival_/departure_times

antrim:
There appear to be errors in the example arrival_times and departure_times. (See https://developers.google.com/transit/gtfs/reference/stop_times-file for comparison.) This change corrects those errors.


barbeau:
Whoah, not sure how that happened :).  Thanks for catching it @antrim!  You're right, this should be merged, it fixes a typo in the example data.

antrim:
@zhsh Can you please merge this?

PR #37, opened 2017-01-03
propose location_type=4 (Internal timing point)

antrim:
A proposed field, stops.hidden, indicates if a stop, station, or other facility is relevant for display in passenger-facing applications.
* **0** or blank - passenger-relevant facility should be shown in maps and other applications.
* **1** - an internal-only location or facility such as timing point that does not allow passengers to board or alight.

Currently, internal timing points are sometimes published in GTFS datasets. stop_times.drop_off_type and stop_times.pickup_type are not adequate to indicate there is not service to those location because a stop that is temporarily closed might still be useful to show in a map.

[Update, 2017-January-4: The proposal for a stops.hidden field has been withdrawn and replaced by a new location_type=4.]

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -100,10 +100,12 @@ File: **Required**
 |  stop_lon | **Required** | The **stop_lon** field contains the longitude of a stop, station, or station entrance. The field value must be a valid WGS 84 longitude value from -180 to 180. |  |  |
 |  zone_id | Optional | The **zone_id** field defines the fare zone for a stop ID. Zone IDs are required if you want to provide fare information using [fare_rules.txt](#fare_rulestxt). If this stop ID represents a station, the zone ID is ignored. |  |  |
 |  stop_url | Optional | The **stop_url** field contains the URL of a web page about a particular stop. This should be different from the agency_url and the route_url fields.  The value must be a fully qualified URL that includes **http**:// or **https**://, and any special characters in the URL must be correctly escaped. See http://www.w3.org/Addressing/URL/4_URI_Recommentations.html for a description of how to create fully qualified URL values. |  |  |
-|  location_type | Optional | The **location_type** field identifies whether this stop ID represents a stop, station, or station entrance. If no location type is specified, or the location_type is blank, stop IDs are treated as stops. Stations may have different properties from stops when they are represented on a map or used in trip planning.  The location type field can have the following values: |  |  |
+|  location_type | Optional | The **location_type** field identifies whether this stop ID represents a stop, station, station entrance, or internal timing point. If no location type is specified, or the location_type is blank, stop IDs are treated as stops. Stations may have different properties from stops when they are represented on a map or used in trip planning.  The location type field can have the following values: |  |  |
 |   |  | * **0** or blank - Stop. A location where passengers board or disembark from a transit vehicle. |  |  |
 |   |  | * **1** - Station. A physical structure or area that contains one or more stop. |  |  |
 |   |  | * **2** - Station Entrance/Exit. A location where passengers can enter or exit a station from the street. The stop entry must also specify a parent_station value referencing the stop ID of the parent station for the entrance. |  |  |
+|   |  | * **4** - Non-passenger-relevant location. An internal-only location or facility such as timing point that does not allow passengers to board or alight. |  |  |
+
 |  parent_station | Optional | For stops that are physically located inside stations, the **parent_station** field identifies the station associated with the stop. To use this field, stops.txt must also contain a row where this stop ID is assigned location type=1. |  |  |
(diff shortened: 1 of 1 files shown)


antrim:
Important note: There are no consuming or publishing applications for this field right now, so before this can be accepted we will need to application developers and data suppliers to support this field.

barbeau:
It should also be noted that, as proposed, this could be considered a breaking change for existing consumers.  If producers start adding hidden stops to stops.txt and existing consumers aren't updated to consume this field, those consumers will show these hidden stops to users as if they were normal stops.

barbeau:
I prefer adding a new `location_type` solution to adding the `hidden` field.  IMHO this is a better distinction that these items are logically different, and also is a better step towards backwards compatibility.  Consumers at least have some indication that the data is different if a new field value for `location_type` is detected, vs. ignoring a new `hidden` field and the data otherwise appearing identical.

antrim:
I removed stops.hidden and added location_type=4
https://github.com/trilliumtransit/transit/commit/ff5ad2ee2627ed1736ed3c8115525cf83488468f

mmorang:
Shouldn't all stops be "passenger-relevant locations"?  I mean, if they're not relevant to passengers, shouldn't they not be included in GTFS, for which the primary purpose is routing applications?  I am concerned that agencies might be tempted to clog up their GTFS feeds with asset management type stuff.

antrim:
Hi @mmorang : The thinking here is that feed producers are _already_ including non-passenger-relevant locations in their feeds (usually AVL timing points) and this information is useful for arrival prediction systems that consume GTFS. Non-passenger-relevant locations can and should be marked with pickup_type and drop_off_type values of 1 (no passenger boarding alighting) but it's also useful to designate that the location should not appear on a map (stops.location_type=4). It's not perfect, but it's the best compromise to adapt to existing practice and need without completely breaking backwards compatibility.

barbeau:
@mmorang I have some of the same concerns.  However, things like timepoints that may not be co-located with stops are definitely needed in GTFS - without them you can't even calculate on-time performance from GTFS data (see page 87 and 100 in [this report](https://www.dropbox.com/s/v05z1w7td1kpqd9/Transit_Service_Reliability_Final_Report.pdf?dl=0) for details).

After thinking about this more, I think we should add a `location_type=4` that is specifically for schedule timepoints, vs. a more general catch-all for "non-passenger relevant location".  This is the primary use case for this field anyway (as far as I'm aware - someone correct me if there are others), and it would constrain usage so that producers and consumers have better guidance and consistency for what this element represents.  If we do this, there should also be some language added surrounding `stop_times.timepoint` - if a stop is `location_type=4`, then all occurences of that stop_id for records in `stop_times.txt` the `stop_times.timepoint` field, if provided, should be set to `1`.

LeoFR:
+1 for the `location_type=4`

LeoFR:
+1 for the `location_type=4` meaning "Internal timing point. A timing point location where passengers cannot board or alight. When referenced in stop_times.txt, pickup_type and drop_off_type values must be 1 for an internal timing point."

mgilligan:
+1, location_type=4 with Leo's language. No `hidden` field.

abyrd:
I agree, no `hidden` field. location_type=4 seems like the right approach. I prefer the narrow definition stated above by @LeoFR, which can be expanded in the future if some other use case comes up.

antrim:
This is operational data that is relevant to the discussion of runs in GTFS. See #195.

antrim:
Another use for a "not a stop" location type: designating the start or end of "continuous stops" sections ("flag stops" or "hail-and-ride") along a trip when those points are not also published stops (see GTFS-ContinuousStops, #208). Like internal timing points, this can currently be indicated by setting `drop_off_type` and `pickup_type` to `1`, but the meaning could be made less ambiguous for consuming applications.

Below are examples of flag stop and "no stop" zones:
* MVRTA (http://www.mvrta.com/routes/route-54/)
* GATRA (https://www.gatra.org/routes/route-6/)

abyrd:
It looks like there was a fair amount of agreement on this proposal. Should we revise and vote on it? Is anyone in a position to start including location_type=4 in their feeds so we have a producer?

davidlewis-ito:
I think location_type=4 is already taken as  a Boarding Area.

timMillet:
Indeed, `location_type=4` is a boarding area.

Also, just a heads-up: according to [MobilityData's GTFS-Vehicles proposal](https://docs.google.com/document/d/156RiBjI6FnWJvO8_XWX11Q9nLdOiBdS_rilA-oamlv8/edit#heading=h.yibgg5lmoq9j), `location_type=5` would be a head boarding area, i.e a boarding area that allows sequencing the other boarding areas of the platform (goal: matching vehicle doors with boarding areas, useful on train and subway routes to tell the user where to wait on the platform). This proposal may evolve, though.

PR #39, opened 2017-01-09
clarify relationship of shape points and stops

antrim:
I wonder if we should propose a change to the Spec, as many feed validators do produce warnings when stops are far from route alignments. It seems as though shapes do need to match stop locations within some parameter.

My guess is that the original intent of this statement was to say that stop locations do not need to intersect the route alignment exactly (which would result in jaggies from centerline to curb). Of course, I'm only making a guess.


skinkie:
I my perception, and in standards such as NeTEx, a stoppoint is projected on a route which follows the road.

sabre23t:
Shapes defining bus routes along roads, only need to have points at intersections/curves where the roads change directions. Very rarely we have bus stops exactly at intersection/curves. We generally have bus stops along straight segments of the route shapes, where points are not needed to define the shape. I guess that "points do not need to match stop locations" was intended to describe that.

Anyway, I agree with your merge to remove "The points do not need to match stop locations."

abyrd:
@antrim the original language makes sense to me, I think it just needs more clarification. As @sabre23t points out, when roads are not curved geometry points may be only at intersections, or even farther apart. @skinkie's interpretation is the same I've seen in most GTFS consumers, that the stops are projected onto the shape geometry using the shortest possible line. 

In most cases stop locations cannot touch the shape exactly, because the shape is re-used for both directions of the route, following the centerline of the road with the stops more accurately located off to the sides. As @antrim points out, forcing the shape to touch all the stops would make the shape zig-zag across the road, sometimes even having sawtooth shapes where stops are across from one another. This is clearly bad, so some form of this statement needs to remain.

The confusion here is between the points and the geometry. I would say: "The points do not need to match stop locations, but all stops on a trip should lie within a small distance of straight line segments between the shape points for that trip."

antrim:
@sabre23t @abyrd Do you want to propose new language? Or do you want me to take a stab at new language?

abyrd:
"Shapes describe the physical path that a vehicle takes, and are defined in the file shapes.txt. Shapes are associated with Trips, and consist of a sequence of points through which the vehicle passes in order. The vehicle's path between successive points is generally assumed to be approximately a straight line. There is no need for points defining Shapes to match the location of Stops, but all Stops on a trip should lie within a small distance of the shape for that trip, i.e. close to straight line segments connecting the shape points in order."

antrim:
Ok. I adapted that a bit.

"Shapes describe the path that a vehicle travels along a route alignment, and are defined in the file shapes.txt. Shapes are associated with Trips, and consist of a sequence of points through which the vehicle passes in order. Shapes do not need to intercept the location of Stops exactly, but all Stops on a trip should lie within a small distance of the shape for that trip, i.e. close to straight line segments connecting the shape points."

barbeau:
Looks good to me.

acollier88:
I agree with this change.

antrim:
It looks as though there is support for this change. Let's start the 7 day voting period for this change, unless anyone has other comments. So that there are 7 full calendar days, then the vote would end on Friday, February 3 @ 23:59:59 UTC.

barbeau:
+1

barbeau:
Vote should also be announced on gtfs-changes and cross linked.

sabre23t:
+1

LeoFR:
+1

RachM:
+1

abyrd:
+1

barbeau:
@antrim It looks like this vote passed.  Do you want to officially wrap it up and request merge?

antrim:
@barbeau Good call. I will request a merge.

antrim:
I signed it!

gtfs/spec/en/reference.md
```diff
@@ -332,7 +332,7 @@ For examples that demonstrate how to specify a fare structure with fare_rules.tx
 
 File: **Optional**
 
-Shapes describe the physical path that a vehicle takes, and are defined in the file shapes.txt. Shapes belong to Trips, and consist of a sequence of points. Tracing the points in order provides the path of the vehicle. The points do not need to match stop locations.
+Shapes describe the path that a vehicle travels along a route alignment, and are defined in the file shapes.txt. Shapes are associated with Trips, and consist of a sequence of points through which the vehicle passes in order. Shapes do not need to intercept the location of Stops exactly, but all Stops on a trip should lie within a small distance of the shape for that trip, i.e. close to straight line segments connecting the shape points.
```
antrim: Need to change capitalization to match the reference right now.

"stops" -> "Stops"

PR #40, opened 2017-01-12
Clarify stop_id (internal only) and stop_code (passenger-facing)

antrim:
Earlier the Spec was not entirely clear that:
* stop_id is internal-only (primary key for the "table")
* stop_code is passenger-facing
* stop_code and stop_id can be the same

This proposal attempts to bring greater clarity to the definitions.


gtfs/spec/en/reference.md
```diff
@@ -93,8 +93,8 @@ File: **Required**
...
+|  stop_id | **Required** | The **stop_id** field contains an ID that uniquely identifies a stop, station, or station entrance. Multiple routes may use the same stop. The **stop_id** is used by systems as an internal identifier of this record (e.g., primary key in database), and therefore the **stop_id** must be dataset unique.  |  |  |
+|  stop_code | Optional | The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. Stop codes are often used in phone-based transit information systems or printed on stop signage to make it easier for riders to get a stop schedule or real-time arrival information for a particular stop.  The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. The **stop_code** can be the same as **stop_id** if it is passenger-facing. This field should be left blank for stops without a code. |  |  |
```
skinkie: I would explicitly end with "without a clearly visible code."

barbeau:
I like this proposal, I think it provides additional clarity on what should go in each and encourage producers to supply stop_code when it exists.

gtfs/spec/en/reference.md
abyrd: @skinkie I guess you are referring to the second line about stop_code rather than the one Aaron edited about stop_id, i.e.: The **stop_code** can be the same as **stop_id** if it is passenger-facing. This field should be left blank for stops without a code clearly visible to passengers.

abyrd:
I am in favor of this change. It is good to clearly state which information is for passengers and which is for internal use (by database systems).

antrim:
I made this change in 24fdf37.
> …This field should be left blank for stops without a code presented to passengers.

There has been limited discussion, but it seems as though there is support for this change. 

Let's start the 7 day voting period for this change, unless anyone has other comments. So that there are 7 full calendar days, then the vote would end on Wednesday, 2/1 @ 23:59:59 UTC.

abyrd:
+1

skinkie:
+1

barbeau:
+1

laidig:
My agency has passenger-facing stop_ids and no stop_codes. so I ask the question:

If stop_code is not populated, can it be assumed that stop_id is passenger facing?

skinkie:
@laidig no. stop_code should be populated if the stop_id is meaningful to something else than a system. Like clearly written on them. Not like a very small serial number somewhere.

laidig:
@skinkie Are you implying that in our case we should explicitly include stop_code == stop_id ? That duplication doesn't seem right to me (although I have seen it in other feeds). 

I would add to the proposed stop_id definition something along the lines of "if stop_code is not populated, stop_id may be (is?) considered as a passenger-facing ID"

barbeau:
>If stop_code is not populated, can it be assumed that stop_id is passenger facing?

We're trying to get away from that assumption with these changes.

>Are you implying that in our case we should explicitly include stop_code == stop_id

Yes - this makes it clearer to consumers that the ID (`stop_code`) is definitely intended to be user-facing

skinkie:
@laidig plus if two strings on the same line have the name sequence, it compresses very nicely. So there is virtually no overhead with the duplication.

laidig:
I've said my piece. It's incredibly unlikely that we will update our GTFS to conform to this.

skinkie:
@laidig what is your technical objection in doing so?

antrim:
This proposal intended to codify existing practice and make this more clear for feed producers. It is not intended to change the predominant existing practice. To my knowledge, no major GTFS-consuming application exposes the stop_id in the user interface.

laidig:
@skinkie , @antrim : bustime.mta.info does ;)

abyrd:
@laidig in your case I'd say yes, there should be a stop_code field with all values equal to those in stop_id. The purpose of GTFS is data interchange, and all other GTFS tools must be able to understand a feed. I don't see any better way to communicate that a system (or especially part of a system) uses its stop_codes equal to its stop_ids. If a particular GTFS consumer site exposes stop_ids in its interface, it is understandable that this was done for convenience, but it does not follow the reading I've always had of the spec. 

Stop IDs are already conceptually distinct from rider-facing stop codes, and this proposal clarifies that distinction. Some systems have public facing stop codes, others do not, and GTFS must fit the general case. The presence of a stop_code column indicates that a system has stop codes, and provides those stop codes. The absence of a stop_code column indicates that a system does not use rider-facing stop codes. Likewise, if the feed has a stop_code column, the absence of a value on certain rows indicates that a particular stop does not have a code, or as advocated by @skinkie that a stop's code is not prominently displayed to riders.

In short I would say that a GTFS consumer should never assume that riders will understand stop_ids. Lack of a stop_code column or lack of an entry in a stop_code field would just mean that there is no user-visible code for that stop.

I imagine that copying stop_id values into stop_code fields is quite trivial in most systems that would export GTFS. @laidig can you tell us why are you opposed to doing this, what feeds you are managing, and whether the software you use somehow makes this difficult?

laidig:
I personally don't manage creation of the feeds, I manage importing feeds
into our bus AVL platform, which is one of the largest in the world (if not
the largest). We have 12,000 stops out on the street with the stop_id on
them, and millions of SMS are being sent to us each month where users
request info for the stop_id.

Yes, stop_code could be added rather easily (csvkit to the rescue!), but
IMHO it's just artless if it's exactly the same. I'll bite my tongue if
this goes through.

sheldonabrown:
Late to the party here, but I completely agree with @laidig.  BusTime, and in fact OneBusAway, has always made the assumption that stop_id can be used as the external identifier if no stop_code is present.  If the intent it to clarify the spec, why not simply state that stop_code is optional and defaults to stop_id?  There seems like enough established precedent/momentum (from my perspective at least) for that to be the case.

skinkie:
@sheldonabrown because you can't negate it.

sheldonabrown:
I don't follow.  If you change your mind, then supply the stop_code.

skinkie:
In The Netherlands we don't have **visible** stop codes, thus we don't want applications to show the internal unique codes per stop. In the case of a **defaults to** an empty (non-visible) stop code cannot be represented.

abyrd:
@sheldonabrown I think what @skinkie means is that in the system you describe there is no way to indicate that a stop does not have a code, or that no stops anywhere in the system have codes. What you're describing is a convenient shortcut if you know for a fact that the network described by the GTFS data has stop codes. But when a network does not have stop codes, the same GTFS consumer app would display internal, machine-readable codes that are meaningless to a rider. For the convenience of not having to include an additional column in the GTFS, we would cause transit riders around the world to see confusing codes that do not correspond to anything visible to them at stations.

abyrd:
@laidig it's noteworthy that several GTFS consumers already assume that in the absence of stop_code, the stop_code is equal to the stop_id. But the existing spec is already fairly clear on this point: "The stop_code field should only be used for stop codes that are displayed to passengers. For internal codes, use stop_id. This field should be left blank for stops without a code." 

As a convenience some software is deviating from the specification, but making this behavior official means assuming that all stations in all transit networks everywhere have stop codes, which is not the case. Whatever approach is adopted needs to have a way of indicating that a stop or the whole feed does not have stop_codes. Leaving the specification unchanged allows this by leaving out the stop_code column or leaving the field empty for certain stops.

sdjacobs:
It sounds like this is a GTFS practice question that differs across both GTFS-producing agencies and GTFS-consuming software applications. For example as OneBusAway assumes that stop IDs are stop codes, many agencies' GTFS make that assumption. Perhaps it's worth getting a broader sense of how widespread both these practices are before making a decision that will have impacts on a number of agencies.

LeoFR:
As a GTFS-consumer, we are only displaying the`stop_code` if it is defined in the GTFS. We are not falling back on the `stop_id`.

antrim:
Quoting @sdjacobs :
> Perhaps it's worth getting a broader sense of how widespread both these practices are before making a decision that will have impacts on a number of agencies.

Agreed. My understanding is that the most common practice is that stop_id is assumed to be internal (not for customers). I believe this is the case with Google Maps and OpenTripPlanner. Some trip planners do not seem to show any sort of stop number -- I looked at Bing Maps and HERE Maps.

Other application developers: What are your practices and expectations?

sheldonabrown:
On Mon, Jan 30, 2017 at 11:46 AM, Aaron Antrim <notifications@github.com>
wrote:

> Quoting @sdjacobs <https://github.com/sdjacobs> :
>
> Perhaps it's worth getting a broader sense of how widespread both these
> practices are before making a decision that will have impacts on a number
> of agencies.
>
> Agreed. My understanding is that the most common practice is that stop_id
> is assumed to be internal (not for customers). I believe this is the case
> with Google Maps and OpenTripPlanner. Some trip planners do not seem to
> show any sort of stop number -- I looked at Bing Maps and HERE Maps.
>
> Other application developers: What are your practices and expectations?
>
>
>
As Simon pointed out, OneBusAway fairly prominently attaches an id to a
stop in all cases.  If stop_code is not present, it uses the stop_id.  This
has influenced the following GTFS feeds (that I know of, there are likely
more):

MTA (NYC)
King County Metro (Seattle)
Sound Transit (Seattle)

MTA is obviously the big one, but KCM is a decent size as well.

Sheldon

RachM:
stop_id should be internal only, and is similar to route_id and trip_id. Its purpose is like a primary key; to allow the stop_id to be referenced from other data files (e.g. in stop_times.txt). At Google, we show stop_code to users, and never surface the stop_id.

Some stops don't have a code (e.g. trains in Australia generally don't), and so a blank stop_code is a good mechanism to tell consumers that there is no short-hand code, and that stop_name should be used. If stop_id were used by consumers, then you wouldn't be able to tell which stops have a 'code', and which stops should use stop_name.

RachM:
Also, +1 in favor of the proposal.

antrim:
I put a [call out on gtfs-changes](https://groups.google.com/d/msg/gtfs-changes/tefWsdezj0I/N6tPqp1MAgAJ) to ask if there are other applications that treat stop_id as customer-facing. If there are no other applications besides OneBusAway that do this, then I would ask that we support this change proposal to establish greater clarity going forward.

pailakka:
I agree strongly with @abyrd in this case. In Finland we have a national database consisting of over 90 000 bus stops, only approximately 17 000 of those stops have stop_code defined. In reality, a lot more stops have some kind of passenger facing code but the national database is missing that information. Every stop however has a technical national identifier which is used as a stop_id in our national GTFS dataset. Showing this technical identifier instead of empty stop_code can be considered very confusing and harmful for the passengers. 

So in this case Finnish Transport Agency disagrees strongly with assuming `stop_code == stop_id` if the stop_code is missing or empty, and we are supporting the original suggestion of clarifying the definitions of stop_code and stop_id in the specification.

obeid:
+1 in favor of the proposal.

I also think defaulting to `stop_id` in the absense of `stop_code` is a bad idea.

Some of the feeds we manage carry slightly adjusted locations for stops from one service period to another. These stops share the same `stop_code` if it's present for an agency, but obviously each has its own unique `stop_id`. This is quite common for agencies with consistent seasonal changes (e.g., universities and colleges).

antrim:
The voting period has ended. All votes (+1) are in favor of this change. It appears as though (at least some) implementations of OneBusAway do not follow this interpretation, but we did not find any other applications with a similar interpretation of stop_id as customer-facing. @zhsh: Can you please merge and close?

Issue #43, opened 2017-01-21
"Service date" vs. "service day"


antrim:
Current Spec refers to "service date". Trips occur on a particular "service date". The service date begins on a particular calendar date, and may continue to the next calendar date. (i.e. if there is transit service past midnight, trips may begin at 25:00:00, etc.).

It appears that "service day" is a more accepted term in transit, and this feels more intuitive to me.

Examples of the use of the term "service day":

* Sacramento Regional Transit ([glossary of terms](https://www.sacrt.com/transitglossary.stm)):

> Service Span — The span of hours over which service is operated, e.g., 6 a.m. to 10 p.m. or 24 hr (owl). Service span often varies by weekday, Saturday, or Sunday.
>
> Synonyms: Span of Service, Service Day

* King County Metro ([Regional Day Pass](http://metro.kingcounty.gov/fares/pdf/regional-day-pass-faq.pdf)):

> The pass covers unlimited rides for a "service day" starting at first use and expiring at 3 a.m. the next day. Each Regional Day Pass covers rides with a one-way adult fare up to $3.50 ($1.75 senior/disabled).

barbeau:
Agreed, "service day" makes more sense to me in this context.

abyrd:
I confirm, "service day" is widely used in the FOSS GTFS community, source code comments, etc.

LeoFR:
+1

antrim:
Closing. Spec changed in #45.

PR #44, opened 2017-01-22
Blocks and service days - Clarify relationship of block_id & service_id

antrim:
It was previously unclear how a block can be defined across service days.

New/added text: "A block consists of a single trip or or many sequential trips made using the same vehicle, defined by shared service day and block_id."

Removed: "The block_id must be referenced by two or more trips in trips.txt."

Note: This proposed change introduces the term of "service day" (see https://github.com/google/transit/issues/43 for discussion).

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -166,7 +166,7 @@ File: **Required**
 |   |  | * `trip_id,...,trip_headsign,direction_id` |
 |   |  | * `1234,...,to Airport,0` |
 |   |  | * `1505,...,to Downtown,1` |
-|  block_id | Optional | The **block_id** field identifies the block to which the trip belongs. A block consists of two or more sequential trips made using the same vehicle, defined by shared service day and block_id. The block_id must be referenced by two or more trips in trips.txt. A block_id can have trips with different service days, making distinct blocks. (See [example below](#example-showing-blocks-and-service-day)) |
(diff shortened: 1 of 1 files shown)


mgilligan:
+1

barbeau:
I think this definition works.  Small typo - `of a single trip or or many` should be `of a single trip or many` (there is an extra `or`).

antrim:
Thanks @barbeau! Fixed in f64f770

antrim:
It looks as though there is support for this change. Let's start the 7 day voting period for this change, unless anyone has other comments. So that there are 7 full calendar days, then the vote would end on Friday, February 3 @ 23:59:59 UTC.

I’m not in love with the formatted example, but I don’t know how better to do this in markdown and in the current document structure. Therefore, I propose that we get block_id definition written into the Spec reference, and leave the door open for reformat/restructure later.

barbeau:
+1

LeoFR:
+1

abyrd:
+1

antrim:
The voting period has ended. It looks as though there is support for this change. @zhsh: Can you please merge and close? We may want to do a little formatting rearrangement (for the example table) down the line.

barbeau:
@dbabramov Thanks for merging!  Friendly request - in the future, could you please squash commits before merging?  I think it's useful to keep a multiple-commit history in a pull request so we can track changes based on comments, but personally I'd prefer to keep the history of the `master` branch clean with one commit per proposal.  Makes it easier to track proposals and changes over time.

PR #46, opened 2017-02-06
Specify that prior to vote GTFS producer/consumer should implement change

barbeau:
Since the beginning of GTFS, one of the defining characteristics of the spec has been that at least one consumer and one producer has been expected to implement any new change/addition to the spec before the change was officially adopted.  This was enshrined in the [following guiding principle](https://github.com/google/transit/blob/master/gtfs/CHANGES.md#speculative-features-are-discouraged):

>**Speculative features are discouraged**
Every new feature adds complexity to the creation and reading of feeds. Therefore, we want to take care to only add features that we know to be useful. Ideally, any proposal will have been tested by generating data for a real transit system that uses the new feature and writing software to read and display it.

Now that we have a formal process for changes, I'd like to include this principle within the steps to ensure it's reviewed by all proposing parties.  I've proposed this as a pre-requisite step before calling for a vote, but I'm open to better ideas as well.

Proposal announced on gtfs-changes list here:
https://groups.google.com/forum/#!topic/gtfs-changes/L_02vz71Ib4

### DIFF
--- a/gtfs-realtime/CHANGES.md
+++ b/gtfs-realtime/CHANGES.md
@@ -12,6 +12,7 @@ The official specification, reference and documentation are written in English. 
   - The advocate is responsible for timely update of the proposal (i.e. pull request) based on the comments for which they agree to.
   - At any point in time the advocate can claim proposal abandoned.
 1. The advocate can call for a vote on a version of the proposal at any point in time following the initial 7-day interval required for discussion.
+  - Before calling for a vote, at least one GTFS producer and one GTFS consumer should implement and test the proposed change.
 1. Vote lasts the minimum period sufficient to cover 7 full calendar days and 5 full Swiss business days. Vote ends at 23:59:59 UTC.
   - The advocate should announce the specific end time at the start of the vote.
   - During voting period only editorial changes to the proposal are allowed (typos, wording may change as long as it does not change the meaning).
--- a/gtfs-realtime/CHANGES.md
+++ b/gtfs-realtime/CHANGES.md
@@ -12,7 +12,7 @@ The official specification, reference and documentation are written in English. 
   - The advocate is responsible for timely update of the proposal (i.e. pull request) based on the comments for which they agree to.
   - At any point in time the advocate can claim proposal abandoned.
 1. The advocate can call for a vote on a version of the proposal at any point in time following the initial 7-day interval required for discussion.
-  - Before calling for a vote, at least one GTFS producer and one GTFS consumer should implement and test the proposed change.
+  - Before calling for a vote, at least one GTFS-realtime producer and one GTFS-realtime consumer should implement the proposed change.  It is expected that the GTFS-realtime producer(s) include the change in a public-facing GTFS-realtime feed and provide a link to that data within the pull request comments, and that the GTFS-realtime consumer(s) provides a link in the pull request comments to an application that is utilizing the change in a non-trivial manner (i.e, it is supporting new or improved functionality).  Producers and consumers should use the [gtfs-realtime.proto](proto/gtfs-realtime.proto) file modified in the pull request that proposes the change to ensure that the final merged change exactly matches the example producers/consumers.  
 1. Vote lasts the minimum period sufficient to cover 7 full calendar days and 5 full Swiss business days. Vote ends at 23:59:59 UTC.
   - The advocate should announce the specific end time at the start of the vote.
   - During voting period only editorial changes to the proposal are allowed (typos, wording may change as long as it does not change the meaning).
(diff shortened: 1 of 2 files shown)


mmorang:
I think this is a good idea, but how do you propose this requirement be met?  Just by a comment in the pull request thread saying "I'm using this", or by some more thorough/extensive vetting process?

barbeau:
@mmorang Good question.  I'd suggest we follow the same approach taken historically on the gtfs-changes list, which is basically trusting the word of whoever is saying "I'm using this" as a consumer or producer.

abyrd:
We could go one step beyond just trusting them by requiring a link to specific lines of open source software implementing the feature, and a specific URL for a feed containing the new feature.

abyrd:
I am definitely in favor of this rule, but I think we should avoid referring to this as "testing" the feature (as in the title of this PR). It must be a full, working implementation. The idea is that only features useful enough to merit the investment of someone's time and effort will get into the spec, and problems or inconsistencies may be spotted during the implementation.

eburkman:
Originally there had to be at least one producer and one consumer for any change that was proposed, but I'm not sure if it was ever "enforced." See #3 under the general outline, [here](https://developers.google.com/transit/gtfs/guides/changes-overview).

Having changes implemented by (at least) one producer and one consumer is important for a number of reasons, including:
1. Avoiding speculative changes ("Let's add a service type for flying buses, which could exist in the future!");
2. Avoiding changes to the spec that are so specific to one agency or situation as to be unnecessary for inclusion in the broader spec; and,
3. Avoiding adding changes that unintentionally break or confuse other functionality, which might not be realized until implementation of the change.

It would be unlikely for a change to the spec to violate one of the above ideas while also being produced and consumed.

barbeau:
Good points everyone - I just pushed commits to this PR that removes the word "test", and elaborates on what we expect from consumers/producers.  

Also - tl;dr - for this proposal, to simplify things I'm changing it to focus just on GTFS - I'll focus on GTFS-rt in a new proposal following the completion of this one.

Explanation:

Note that this proposal was originally for both GTFS and GTFS-rt.  However, the GTFS-rt process has been a little different than GTFS for requiring producers/consumers - the current practice is that following a vote, fields are added to the documentation but marked them as "experimental", and then at a later date to change them to "official" following implementation/deployment  (see the [original proposal here](https://groups.google.com/forum/#!searchin/gtfs-realtime/extensions/gtfs-realtime/bWQAgGdj3f0/mdcKwR3J8xAJ) for this practice).  This was originally done due to the hassle with implementing protobuf extensions, and then having to go back and publish a new version of the feed after official acceptance switching from the extension to the canonical .proto file (which can incur further consultant/AVL vendor costs for agencies, etc).  This change happened prior to the move to using Github as a tool for managing proposals, and this practice was brought up for discussion again when transitioning to using Github (see [this post](https://groups.google.com/forum/#!topic/gtfs-realtime/bwFaI6c_K1c)), although the consensus was to stick to the process of adding "experimental" fields to GTFS-rt.  The downside to the "experimental" field process is that it seems to potentially float-on endlessly - to my knowledge I don't think any "experimental" fields have been finalized in the spec or removed to date.  And, speculative features are more likely to creep into the spec because no implementation is expected up front.

mmorang:
@abyrd We can't assume that all valid implementations will be open source, so requiring a link to the line of code isn't going to work for everyone.

barbeau:
I'd like to call for a vote on this proposal - voting will end Feb 23rd at 23:59:59 UTC.

mmorang:
+1

DaveBarker:
+1

barbeau:
So right now we have 2 votes not counting myself (3 if you count me).  We need 3 total yes votes for this to pass.

So - should we count my own vote?  If so, this votes passes and we can merge.  If not, I need to call for another 7 day voting period.

eburkman:
It's after the deadline, but I vote yes.

barbeau:
Thanks @eburkman!  I'd like to take the more conservative stance here that my vote doesn't count, so to follow the rules (given that your vote was outside the voting window) I'm calling another 7 day voting period starting now which will end on March 4th at 23:59:59 UTC.  So if you previously voted please vote again.

On a related note, I opened a separate proposal at https://github.com/google/transit/pull/50 to clarify that proposer's votes shouldn't count towards the total.

mmorang:
+1

eburkman:
I vote yes on this, and within the proper voting time frame!

antrim:
+1

sabre23t:
+1

abyrd:
+1

mgilligan:
+1

RachM:
+1

barbeau:
Alright, voting period is closed, and it looks like we got the required number of votes this time and it's approved!  Thanks everyone!

@dbabramov Could you please squash and merge this?

PR #47, opened 2017-02-07
Clarify definition of "frequency-based" trips

barbeau:
For some time, the term "frequency-based trips" (trip in frequencies.txt with `exact_times`=0, or empty `exact_times`) has been conflated with the representation of schedule-based trips in frequencies.txt (trips with `exact_times`=1).  This conflation makes it very difficult and confusing to discuss any concept related to frequencies.txt.

This proposal clearly specifies the difference between `exact_times`=0 vs 1 trips using the following definitions:
* `exact_times`=0 trips are the only trips referred to as "frequency-based"
* `exact_times`=1 trips are referred to as "headway-defined schedules"
* When `exact_times` 0 and 1 trips are both referenced, this proposal refers to them as "trips defined in frequencies.txt" (instead of "frequency-based trips")

Announced on the gtfs-changes list at https://groups.google.com/forum/#!topic/gtfs-changes/tgRrog7wM78.


gtfs/spec/en/reference.md
```diff
@@ -379,20 +379,25 @@ Shapes describe the physical path that a vehicle takes, and are defined in the f
...
 |   |  | `A, 05:00:00, 07:00:00, 600` |
 |   |  | `B, 07:00:00, 12:00:00, 1200` |
-|  exact_times | Optional | The **exact_times** field determines if frequency-based trips should be exactly scheduled based on the specified headway information. Valid values for this field are: |
-|   |  |* **0** or **(empty)** - Frequency-based trips are not exactly scheduled. This is the default behavior. |
-|   |  |* **1** - Frequency-based trips are exactly scheduled. For a frequencies.txt row, trips are scheduled starting with trip_start_time = start_time + x * headway_secs for all x in (0, 1, 2, ...) where trip_start_time < end_time. |
+|  exact_times | Optional | The **exact_times** field determines if trips represent frequency-based service or scheduled-based service using on the specified headway information. Valid values for this field are: |
+|   |  |* **0** or **(empty)** - Trips are frequency-based. Operators make their best effort to serve stops periodically based on a headway, but there is no fixed schedule.  This is the default behavior. |
```
trilliumtransit: @barbeau What do you think about adding an extra note in here to specify expectations around a UI? Or is this superfluous? Proposed additions are in **_bold italic_**.

  |  ------ | ------ | ------ |
|  exact_times | Optional | The **exact_times** field determines if trips represent frequency-based service or scheduled-based service using on the specified headway information. Valid values for this field are: |
 |   |  |* **0** or **(empty)** - Trips are frequency-based. Operators make their best effort to serve stops periodically based on a headway, but there is no fixed schedule. **_A user interface should therefore not show scheduled pickup and drop-off times._** This is the default behavior. |
|   |  |* **1** - Trips are scheduled. For a frequencies.txt row, trips are scheduled starting with trip_start_time = start_time + x * headway_secs for all x in (0, 1, 2, ...) where trip_start_time < end_time. **_A user interface may show scheduled pickup and drop-off times._** |

gtfs/spec/en/reference.md
```diff
@@ -379,20 +379,25 @@ Shapes describe the physical path that a vehicle takes, and are defined in the f
...
+
+* frequency-based trips - Service that does not have a schedule of stop times (**exact_times** is **0** or **empty**).  Instead, operators make their best effort to serve stops periodically based on a headway.  For example, a rider can expect a bus to depart approximately every 15 minutes, but the vehicle will not attempt to adhere to a preset scheduled departure time such as 10:00 am.  Therefore, in the absence of real-time information, every subsequent departure is expected with a interval of **headway_secs** from the most recent departure. 
+* headway-defined schedules - Service that adheres to a schedule (**exact_times** is **1**).  This table can be used to represent simple scheduled service based on a fixed interval without fully listing out all scheduled trips via the [trips.txt](#tripstxt) and [stop_times.txt](#stop_timestxt) files.  Instead, the scheduled start time is defined in this table along with the set time between each departure (i.e., headway).  Consumers will then expand this to represent scheduled service as if each of these trips were defined in [trips.txt](#tripstxt) and [stop_times.txt](#stop_timestxt). 
+
```
antrim: I am having trouble with the term "headway-defined schedules" for service that adheres to a schedule (**exact_times** is **1**). I keep thinking about "headway" as particularly meaningful because of its context with customers. i.e.. _"Service every 10 minutes"_ shows how headway is presents a "schedule" in customer-facing context. (exact_times = 0).

In transit operations, the similar term "headway-based" seems to often indicate the absence of an exact schedule. Or, at least appeared to, based on a Google search:

* [Seattle Transit blog](https://www.seattletransitblog.com/2012/11/15/transit-information-systems-schedules/): "Headway based schedules should primarily be used when time savings for riders from improved reliability due to active headway maintenance exceeds the added uncertainty or wait time due to lack of a set schedule."
* TRB Paper, [_Reliability Improvement in Short Headway Transit Services Schedule- and Headway-Based Holding Strategies_](http://trrjournalonline.trb.org/doi/abs/10.3141/2143-09): "Holding vehicles can improve reliability, resulting in shorter travel times and less crowding. In this paper both schedule-based and headway-based holding strategies in short headway services are analyzed."

antrim:
We definitely need to introduce some strong clear terms here so this is not so confusing to talk about. I'm having a hard time getting comfortable with some of these new terms (see inline comments above). However, I must also confess my feedback is not tremendously constructive at this point. I'll try to think of some alternative terms.

In the meantime, I think it could be really useful if people could brainstorm further and contribute their knowledge of terminology usage in the industry around these concepts.

gtfs/spec/en/reference.md
barbeau: Yeah, that's not a bad idea.  It helps shape expectations of how the information is used by consumers.  I'll work on tweaking this.

gtfs/spec/en/reference.md
barbeau: Yeah, I'm not in love with "headway-defined schedule" either - I'm open to proposals for how to name exact_times=1 service. "Headway" can definitely be associated with frequency-based service, but unfortunately we're stuck with the field name `headway_secs` for both types of service.  The main rule is that we can't use the word "frequency" for exact_times=1, and we can't use "schedule" for exact_times=0.

abyrd:
For what it's worth, it's become common practice in the projects and organizations I work with to refer to these as "pure frequencies" and "exact-time frequencies". I'm not saying these are great terms either, but I think the second usage grew out of a clear connection to the field value exact_times=1.

antrim:
To brainstorm, here are some more term ideas.

**frequency-defined service** (any service defined by frequencies.txt):
- **headway-specified operations** (exact_times = 0)
- **scheduled operations** (exact_times = 1)

These two later invoke the perspective of the operator. Vehicles will hold differently with service managed for headways. Speaking of, should we look at definitions for the timepoint field w.r.t. exact_times = 0, or is that getting too fussy?

barbeau:
I'd really like to stick with "frequency-based service" (or something very similar) for `exact_times`=0, and avoid using the term "frequency" as an umbrella for all data in frequencies.txt.  "Frequency-based service" is (in my experience) the industry-standard term for the `exact_times`=0 type of service, and IMHO calling it anything else, or implying that exact_times=1 is somehow frequency-related, leads to confusion.

Here are some other ideas for `exact_times`=1 (instead of "headway-defined schedule"):

* schedule-based
* scheduled service
* compressed schedule
* rolled-up schedule
* abbreviated schedule

The industry usually categorizes the two types of service as "frequency-based" and "schedule-based", so maybe it's best to just keep it simple and use these definitions as shorthand, and provide a more detailed explanation in the docs of how schedule-based service in frequencies.txt differs from the standard expression of schedules in GTFS.

>should we look at definitions for the timepoint field w.r.t. exact_times = 0, or is that getting too fussy?

Good point - I hadn't thought about this, but I believe that `timepoint` shouldn't be populated for exact_times=0 trips.  Let's hold off on discussing this for now though - we can tackle it when we expand the `timepoint` field definition.

sabre23t:
Some comments from my part of transit world, Kuala Lumpur, Malaysia.

Our usage of "frequency", "schedule", "headway" operations and services are most like [Antrim's comments on 8 Feb 2017](https://github.com/google/transit/pull/47#issuecomment-278240955).

Our LRT/MRT operations are **internally headway based**, and published [to users as frequencies](https://www.myrapid.com.my/traveling-with-us/how-to-travel-with-us/rapid-kl/monorail-and-lrts/train-frequency).

Our Bus operations are **internally schedule based**, but generally not published to users (with limited exceptions), instead [frequencies range are published](https://www.myrapid.com.my/traveling-with-us/how-to-travel-with-us).

antrim:
**frequencies.txt span** (represented by any record in frequencies.txt. The _span_ term focuses on the start_time and end_time.

with sub-terms for **exact_times = 0** and **exact_times = 1**

barbeau:
There is some other work on simplifying spec language that this may end up getting folded into, so I'm closing this PR for now.

PR #50, opened 2017-02-24
Clarify that a proposer's vote does not count towards total

barbeau:
As far as I can tell, it's not currently clear whether or not the proposer's vote is counted towards the total 3 yes votes needed for a proposal to be accepted for both GTFS and GTFS-rt.  For example, in https://github.com/google/transit/pull/46 there were two yes votes by users other than the proposer (me) during the voting period.

This pull request clarifies that a proper's vote does not count towards the total needed 3 yes votes.

Hopefully in the vote on this we can get it to pass by more than 2 other votes :).

Posted to gtfs-changes at https://groups.google.com/forum/#!topic/gtfs-changes/8jONl-MbyY0.

### DIFF
--- a/gtfs-realtime/CHANGES.md
+++ b/gtfs-realtime/CHANGES.md
@@ -19,7 +19,7 @@ The official specification, reference and documentation are written in English. 
     If a voter changes her vote, it is recommended to do it by updating the original vote comment by striking through the vote and writing the new vote.
   - Votes before the start of the voting period are not considered.
 1. The proposal is accepted if there is a unanimous consensus yes with at least 3 votes.
-  - The proposer's vote does not count towards the 3 vote total.  For example, if Proposer X creates a pull request and votes yes, and User Y and Z vote yes, this is counted as 2 total yes votes.
+  - The proposer's vote does not count towards the 3 vote total. For example, if Proposer X creates a pull request and votes yes, and User Y and Z vote yes, this is counted as 2 total yes votes.
   - Votes against shall be motivated, and ideally provide actionable feedback.
   - If the vote has failed, then the advocate may choose to continue work on the proposal, or to abandon the proposal.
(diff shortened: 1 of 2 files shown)


antrim:
This makes sense to me. Essentially, a minimum of 4 parties must agree (including the proposer). That bar seems just sufficiently high to make sure that enough parties are interested and in agreement without being too high and forcing uninterested parties to weigh in.

mmorang:
Quick question while we're thinking about this.  I don't think we've run into this so far, but what happens if there are explicit "no" votes?  Does each no vote cancel out a yes vote?  Or does it still pass if four parties agree, even if one additional person explicitly disagrees?

barbeau:
A single No vote kills the proposal:
https://github.com/google/transit/blob/master/gtfs/CHANGES.md

LeoFR:
@barbeau If I may make a tiny tiny remark, the specification currently contains almost only single-space after a full stop (37 single space vs 1 double-space for `gtfs/CHANGES.md`).

Your PR contains a double-space after a full stop.

😄

barbeau:
Thanks @LeoFR!  Old habits die hard :).  Fixed in https://github.com/google/transit/pull/50/commits/8192cabab7c32630c3e1a8d5e3a4ad86dc75cd75.

barbeau:
I'd like to call for a vote on this proposal - voting will end March 13th at 23:59:59 UTC.

devinbraun:
Yes

antrim:
+1

RachM:
+1

abyrd:
+1

sabre23t:
+1

barbeau:
Alright, voting is now wrapped on this proposal, and it passes with 5 yes votes and 0 no votes!

@dbabramov Could you please squash and merge?

PR #55, opened 2017-04-03
Clarify that header timestamp should never be less than entity timestamps

barbeau:
This patch clarifies that the GTFS-rt header timestamp should be the "newest" timestamp in the feed - it should never be less than the timestamp in any of the entities (i.e., it must always be greater than or equal to the feed entity timestamps).  

If the header timestamp is less than the entity timestamps, this implies time skew within the feed - either the clock for the server producing the GTFS-rt is off (assuming VehiclePosition/TripUpdate timestamps are generated from GPS, this is probably the more likely scenario), or the timestamps for the VehiclePosition/TripUpdate is incorrect.

Announced on the GTFS-realtime Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/ie_ZmTQ7hJo.

**EDIT** Adding additional context:

My interpretation of FeedHeader.timestamp has always been that it should reflect the time at which the contents of the feed was updated - essentially, a shortcut to the timestamp of the most recently updated entity if one of the entities was changed, or if no entity has changed it reflects the age of the information in the feed. Under this interpretation, if one of the entity timestamps is ahead of the header timestamp, then either the header timestamp is wrong or the entity timestamp is wrong.

For additional context, we're adding this as either a warning or error in our GTFS-rt validation tool:
https://github.com/CUTR-at-USF/gtfs-realtime-validator


skinkie:
-1

No the header in the GTFS-RT timestamp should be the time that the message was started to be generated. For serial systems this value would be the time data was started to be acquired from origin system. For parallel systems this value will indicate the time when the GTFS-RT message itself was being serialised.

barbeau:
>No the header in the GTFS-RT timestamp should be the time that the message was started to be generated. 

https://developers.google.com/transit/gtfs-realtime/reference/FeedHeader says:

>This time identifier shows when the feed content was created on the server, in POSIX time (seconds since January 1st 1970 00:00:00 UTC).

I interpret "feed content" being the protobuf itself.  No matter how long it takes to generate and aggregate the underlying data, IMHO the timestamp should indicate when the binary that's returned to the client is assembled (i.e., the content is created) - as you say, when the underlying data is serialized to protobuf.  Otherwise it's somewhat of a meaningless value if it's arbitrary to some time during which the data is gathered.  We're also talking precision on the order of seconds here, as the timestamp is POSIX time.

But, that's a good question - do others consider "feed content" to be the protobuf representation of the content, or a looser description of some time during which data is collected from the origin system?  Do any producers view an implementation of setting the header timestamp just before serialization being problematic?

skinkie:
It is not a meaningless value if it is the timestamp of data acquisition.  Hence I am telling you that some time ago I started to acquire the contents of this feed. As producer it wouldn't be problematic, as consumer it would! Hence it would look like the feed represents newer data the state as of the timestamp, while the information can only be guaranteed to be accurate at the first acquisition time.

barbeau:
So in your scenario how big of an offset are you talking about between the header timestamp and the entity timestamps, where the header timestamp is less than the entities?

skinkie:
Minutes.

barbeau:
And what is their interval between feed updates? For example, the difference between two sequential header timestamps?

skinkie:
That would be something like fetchtime + backoff interval. So in case of STIF fetchtime + 5 minutes.

barbeau:
Wow...that seems like a long time in between GTFS-rt updates.  Is this a production feed?  If so you can share the GTFS and GTFS-rt URLs?

skinkie:
Yes, don't ask. It is a pre-production limitation.

barbeau:
Ok - so feed hasn't launched to production yet?

skinkie:
This on is not. The origin data can be found here, including the request limitation per day. https://opendata.stif.info/api/v1/documentation

barbeau:
Does anyone else think that the header timestamp should always been the most recent timestamp in the feed (i.e., all TripUpdate and VehiclePosition entity timestamps should always be equal to or less than the header timestamp)?

To recap, the rationale for this proposal is to attempt to address timestamp quality problems.

For example, consider the following - I'm a consumer and I want to show a "Last updated" time in my app, like:

image

I see the following data in a feed:

~~~
FeedHeader.timestamp = 09:02:00
|
|-->VehiclePosition.timestamp = 09:04:00
~~~

Which timestamp do I trust?  

The `FeedHeader.timestamp` definition says:

>This time identifier shows when the feed content was created on the server

...while `VehiclePosition.timestamp` definition says:

>Time when the vehicle position was measured

I don't see how the vehicle position can be measured and included in a feed after the feed content was created on the server, which means IMHO I should throw out the `VehiclePosition.timestamp` as an error.  But, this behavior isn't specified anywhere, which means other consumers might just pass the `VehiclePosition.timestamp` through to the end user.

So, this proposal attempts to clarify what is expected in terms of relationship between timestamps in entities, to guide producers when QA'ing feeds, and to guide consumers to have consistent behavior for end users.

If I'm the only one who has this opinion I'll drop this proposal, but I'd like to hear what others think.

skinkie:
Server starts polling per vehicleinformation information on moment t1, while polling the server gets vehicle information on t2, your fetch is t3, which may just have updated. Thus trust the vehicle information, discard any information that is greater than t3 as error.

barbeau:
There doesn't seem to be sufficient interest in making this part of the spec, so I'm closing this proposal.  If others are interested, please let me know and I'll re-open.

PR #64, opened 2017-06-28
Define semantic cardinality for GTFS-realtime fields

barbeau:
* As originally discussed in https://github.com/google/transit/pull/19 and https://groups.google.com/forum/#!msg/gtfs-realtime/wm3W7QIEZ9Y/DLyWKkknJyoJ, the current GTFS-realtime spec documentation describes field Protocol Buffer cardinality, not semantic cardinality.  This has created confusion for consumers and producers where fields have been omitted based on them being labeled as "optional" in the GTFS-realtime spec, even if they were required under certain logical transit conditions (e.g., stop_sequence must be provided if a trip contains a loop).
* ~~This patch changes the "Cardinality" documentation to define the semantic cardinality for each data element as "required", "conditionally required", and "optional", with an explanation for when "conditionally required" fields are required in the "Description" section of that field.~~ **EDIT** - Field requirements are now being defined in a new field "Required" - see https://github.com/google/transit/pull/64#discussion_r124941448 and discussion that follows. It also bumps the gtfs_realtime_version in the .proto file to 2.0 so validators can strictly enforce semantic cardinality based on the gtfs_realtime_version.

Announced on GTFS-realtime Google Group at:
https://groups.google.com/forum/#!topic/gtfs-realtime/J-csZSxeWjs


barbeau:
I should also mention that with the exception of making the header timestamp required, I believe all of the other field cardinalities reflect existing GTFS-rt documentation. So, this isn't so much proposing something new as it is trying to capture existing logic spread around the docs into a concise description in the Cardinality and Description fields.

gtfs-realtime/spec/en/reference.md
```diff
@@ -114,7 +129,7 @@ Note that the update can describe a trip that has already completed.To this end,
...
 | **vehicle** | [VehicleDescriptor](#message-vehicledescriptor) | optional | Additional information on the vehicle that is serving this trip. |
-| **stop_time_update** | [StopTimeUpdate](#message-stoptimeupdate) | repeated | Updates to StopTimes for the trip (both future, i.e., predictions, and in some cases, past ones, i.e., those that already happened). The updates must be sorted by stop_sequence, and apply for all the following stops of the trip up to the next specified one. |
+| **stop_time_update** | [StopTimeUpdate](#message-stoptimeupdate) | required | Updates to StopTimes for the trip (both future, i.e., predictions, and in some cases, past ones, i.e., those that already happened). The updates must be sorted by stop_sequence, and apply for all the following stops of the trip up to the next specified one.  At least one stop_time_update must be provided for the trip. |
```
jxeeno: When a trip is canceled and its `ScheduleRelationship` is reported as `CANCELED`, some (most?) operators don't list any stops in `StopTimeUpdate` as its assumed that the trip will not be making any stops.

Do we want this to be the case?  Or, would it preferred to list all stops with the `StopTimeUpdate`'s `ScheduleRelationship` as `SKIPPED` for all stops?

gtfs-realtime/spec/en/reference.md
```diff
@@ -146,10 +161,10 @@ The update is linked to a specific stop either through stop_sequence or stop_id,
...
-| **arrival** | [StopTimeEvent](#message-stoptimeevent) | optional |
-| **departure** | [StopTimeEvent](#message-stoptimeevent) | optional |
+| **stop_sequence** | [uint32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | conditionally required | Must be the same as in stop_times.txt in the corresponding GTFS feed.  Either stop_sequence or stop_id must be provided within a StopTimeUpdate - both fields cannot be empty.  stop_sequence is required for trips that visit the same stop_id more than once (e.g., a loop) to disambiguate which stop the prediction is for. |
+| **stop_id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | conditionally required | Must be the same as in stops.txt in the corresponding GTFS feed. Either stop_sequence or stop_id must be provided within a StopTimeUpdate - both fields cannot be empty. |
+| **arrival** | [StopTimeEvent](#message-stoptimeevent) | conditionally required | Either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. |
```
jxeeno: If a `StopTimeUpdate` has a `ScheduleRelationship` of `SKIPPED`, I don't think it would be necessary to provide an `arrival` or a `departure` `StopTimeEvent`.

gtfs-realtime/spec/en/reference.md
RachM: I'm concerned that `repeated` has been removed; I think it should be included so that people know they can supply 0 or more.

gtfs-realtime/spec/en/reference.md
```diff
@@ -1,14 +1,29 @@
 A GTFS Realtime feed lets transit agencies provide consumers with realtime information about disruptions to their service (stations closed, lines not operating, important delays, etc.) location of their vehicles, and expected arrival times.
 
-Version 1.0 of the feed specification is discussed and documented on this site.
+Version 2.0 of the feed specification is discussed and documented on this site.
 
 ### Term Definitions
 
-*   **required**: Exactly one
-*   **repeated**: Zero or more
```
RachM: I'm concerned that `repeated` has been removed; I think it should be included so that people know they can supply 0 or more.

barbeau: @RachM I had concerns about removing `repeated` too, but wasn't sure how to represent this in the table.  Currently I added text to the "Description" field, but I do like the ability to scan the table and easily see which fields are 0+ elements.

Questions for this:
* Should we add a new column for one/many relationships?  If so, what should this be called?  
* Do we leave this as the "Cardinality" column and change the name of the column for the new semantic fields to "Requirement", or something similar?

Also, I've never liked the term `repeated` for this concept because to my knowledge it's very protocol buffer (implementation)-specific.  I'd prefer to find another term - suggestions?

gtfs-realtime/spec/en/reference.md
barbeau: @jxeeno Good catch!  I missed this when converting some of our gtfs-rt-validator rules to the text.  I agree, if `ScheduleRelationship` is `CANCELED` then no `stop_time_updates` need to be provided.  For my reference this is [E041](https://github.com/CUTR-at-USF/gtfs-realtime-validator/blob/master/RULES.md#e041---trip-doesnt-have-any-stop_time_updates) for us.  I'll push an update to fix this.

gtfs-realtime/spec/en/reference.md
barbeau: @jxeeno Good catch again!  Yes, for `SKIPPED` neither `arrival` nor `departures` need to be provided ([E043](https://github.com/CUTR-at-USF/gtfs-realtime-validator/blob/master/RULES.md#e043---stop_time_update-doesnt-have-arrival-or-departure) for our validator).  Also, it's actually an error to provide `arrival` and `departure` for `ScheduleRelationship` of `NO_DATA` ([E042](https://github.com/CUTR-at-USF/gtfs-realtime-validator/blob/master/RULES.md#e042---arrival-or-departure-provided-for-no_data-stop_time_update)).  I'll fix both these items.

gtfs-realtime/spec/en/reference.md
barbeau: Fixed in https://github.com/google/transit/pull/64/commits/bf3d2a5a8c3e67f259fbf7a867b64632bb41db27.

gtfs-realtime/spec/en/reference.md
barbeau: Fixed via https://github.com/google/transit/pull/64/commits/56066ee33f0606ec89f7245d3a68ea95b55b45e4

barbeau:
@jxeeno Good catch on the `schedule_relationship` conditional states - I've pushed commits to fix both of these.

gtfs-realtime/spec/en/reference.md
RachM: > Do we leave this as the "Cardinality" column and change the name of the column for the new semantic fields to "Requirement", or something similar?

I think that's the best approach. My reasoning:
- Cardinality best describes the number of entities allowed in that field, so let's keep it purely to numbers. Of course the savvy user may determine the "requirement" from the number, but I think it's useful to explicitly state the "requirement" in a separate column.
- A new column, `Required` or something similar makes it very clear that the user should check that column to see if the field is needed. 

I think it's good to separate the two concepts; a) is it needed (check `Required` column) and 2) how many can I provide (check `Cardinality` column).

I agree about `repeated`; it's not user friendly. If you have the `Cardinality` column, then the `Required` column entries become: Yes (for `required`) and No (for `optional` and `repeated`).

barbeau: Sounds good on adding the new column.

So here's what I'm working on now to update this proposal, in summary:

* `Required` column will address semantic requirements
* `Cardinality` column will include information about how many elements should be provided for a particular field

The goal of this proposal is to represent the semantic requirements of fields, based on transit use cases and domain logic.  The main challenge here is that there isn't a direct mapping between semantic requirements in `Required` and protobuf cardinality.  As a result I think it will be confusing to show semantic requirements in `Required` and protobuf cardinality in `Cardinality`.

I'd prefer to keep the spec documentation independent of the implementation - in other words, the docs should be purely semantic, with the `.proto` file being the reference for the protobuf cardinality.

With this in mind, I propose that we define semantic cardinality something like the following:

> *Cardinality* represents the number of elements that may be provided for a particular field:
>
>* **One** - A single one element may be provided for this field.  This maps to the [Protocol Buffer *required* and *optional* cardinalities](https://developers.google.com/protocol-buffers/docs/proto#simple).
>* **Many** - Many elements (0, 1, or more) may be provided for this field.  This maps to the [Protocol Buffer *repeated* cardinality](https://developers.google.com/protocol-buffers/docs/proto#simple).
>
>Always reference the *Required* and *Description* fields to see when a field is required, conditionally required, or optional.  Please reference [`gtfs-realtime.proto`](https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto) for Protocol Buffer cardinality.

Using these definitions, `FeedMessage` and `TripUpdate` would look like the following:

#### FeedMessage

| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
|------------------|------------|----------------|-------------------|-------------------|
|**header** | [FeedHeader](#message-feedheader) | Required | One | Metadata about this feed and feed message. |
|**entity** | [FeedEntity](#message-feedentity) | Required | Many | Contents of the feed. |

#### TripUpdate

| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
|------------------|------------|----------------|-------------------|-------------------|
| **trip** | [TripDescriptor](#message-tripdescriptor) | Required | One | XXXX |
| **vehicle** | [VehicleDescriptor](#message-vehicledescriptor) | Optional | One | XXXX |
| **stop_time_update** | [StopTimeUpdate](#message-stoptimeupdate) | Conditionally required | Many | XXXX |
| **timestamp** | [uint64](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | XXXX |
| **delay** | [int32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | XXXX |

barbeau:
I've pushed a new commit https://github.com/google/transit/pull/64/commits/040f49cd9f2088a2a55b8513ca175d8bdf743bb7 into this PR that creates a new column for "Required", and uses semantic cardinality (`One` and `Many`) in "Cardinality" instead of protobuf cardinality (as proposed in https://github.com/google/transit/pull/64#discussion_r125075490).

Feedback welcome!  I'm definitely open to comments on how to improve this.

barbeau:
Any additional comments on this proposal?  I'd like to make any further changes based on feedback prior to calling a vote.

RachM:
LGTM - I'm happy for voting to commence.

barbeau:
Great!  I'd like to call for a vote on this proposal then - voting will end Thursday July 20th at 23:59:59 UTC.

RachM:
I vote yes.

jxeeno:
+1

barbeau:
Sorry for the delay updating this proposal - I was out on vacation last week.

Unfortunately we currently don't have the 3 needed votes for this to pass (we're at 2), and the voting period ended July 20th.

This ends up being a good thing, as I realized there was one issue in this proposal (FeedMessage.entity should be "Conditionally required" - I just fixed this in https://github.com/google/transit/pull/64/commits/3c24ff7fdb4095398a970f22999a6d1a74a73b3d, see commit description for more info).

So, I'd like to call for another vote on this PR - voting will end Tuesday Aug 8th at 23:59:59 UTC.

RachM:
I vote yes.

jxeeno:
+1 I vote yes.

qwasar:
I vote yes.

barbeau:
Alright, looks like this passed the vote - 3 yes votes and 0 nos!  Congrats everyone, we have GTFS-realtime v2.0! :confetti_ball: :confetti_ball: :confetti_ball: 

This includes better documentation for semantic cardinality and requirements - which fields are required and under what conditions.

@RachM Would you be able to squash and merge this, so one commit ends up in the master branch?

benkaday3218 (commented):
Help

Cristhian-HA:
@benkaday3218  would you be able to provide details about the areas where you need help?

gtfs-realtime/spec/en/reference.md
ChTalhaazhar: I have identity theft issue that's why I can't share next any thing without my secret code (CHAUDHARY) copy that 
Screenshot_20241102-181346

PR #65, opened 2017-07-06
Remove requirement for stop_time_update arrival AND departure for SCHEDULED trips

barbeau:
The following wording for `stop_time_updates` that have `schedule_relationship` of `SCHEDULED` (or no `schedule_relationship` - `SCHEDULED` is default):

> If the schedule for this stop contains both arrival and departure times then so must this update

...is problematic from two perspectives:

1. **GTFS requires both `arrival_time` and `departure_time` in `stop_times.txt` to be populated** - Looking at https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt, it says:

    >*If you don't have separate times for arrival and departure at a stop, enter the same value for arrival_time and departure_time*.... If this stop is not a timepoint, use either an empty string value for the arrival_time field or provide an interpolated time...Provide arrival times for all stops that are time points...

    `departure_time` has similar text.

    So, `arrival_time` and `departure_time` should always be populated, which means that the above wording *always* applies to `stop_time_update` `arrival` and `departure` for `SCHEDULED` (normal) trips.

2. **There are valid cases where an `arrival` prediction may not exist, but a `departure` does, and vice versa** - For example, for the first stop in a trip, there may not be a rider-facing prediction for when the bus will arrive at that stop, but there is for when it will depart.  And, vice versa for the last stop in the trip.

To fix this, I'm proposing removing the following text that specifies that both arrival and departure must be populated, which will clarify that providers are allowed to provide only one of the two:

> If the schedule for this stop contains both arrival and departure times then so must this update

This proposal was announced on GTFS-rt Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/hHWOJBMyDws.

**EDIT October 30th 2017**

Reviewing the spec, elsewhere it's clear that arrival *OR* departure can be provided.

[Stop Time Updates](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/trip-updates.md#stop-time-updates) narrative section says:

>The update can provide a exact timing for arrival and/or departure at a stop in StopTimeUpdates using StopTimeEvent.

From [StopTimeUpdate](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-stoptimeupdate), for `arrival` and `departure`:

>If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty

The text I'm proposing to eliminate clearly conflicts with both of these, and is likely left over from an early version of the spec and Google's initial implementation.

### DIFF
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -160,7 +160,7 @@ The relation between this StopTime and the static schedule.
 
 | _**Value**_ | _**Comment**_ |
 |-------------|---------------|
-| **SCHEDULED** | The vehicle is proceeding in accordance with its static schedule of stops, although not necessarily according to the times of the schedule. This is the **default** behavior. At least one of arrival and departure must be provided. If the schedule for this stop contains both arrival and departure times then so must this update. |
+| **SCHEDULED** | The vehicle is proceeding in accordance with its static schedule of stops, although not necessarily according to the times of the schedule. This is the **default** behavior. At least one of arrival and departure must be provided. |
 | **SKIPPED** | The stop is skipped, i.e., the vehicle will not stop at this stop. Arrival and departure are optional. |
 | **NO_DATA** | No data is given for this stop. It indicates that there is no realtime information available. When set NO_DATA is propagated through subsequent stops so this is the recommended way of specifying from which stop you do not have realtime information. When NO_DATA is set neither arrival nor departure should be supplied. |
 


skinkie:
If the departure is before the scheduled arrival, and you don't have a prediction for arrivals. Do you want to force the consumer to infer the value?

barbeau:
>If the departure is before the scheduled arrival, and you don't have a prediction for arrivals. Do you want to force the consumer to infer the value?

So I think the following is the scenario you're presenting:

* In GTFS, you have both a scheduled arrival and departure time, and you don't have a GTFS-rt prediction for the arrival, but you do have a real-time prediction for the departure, and the predicted departure is before the scheduled arrival.  

Are you saying that in this case, you would want the GTFS-rt spec to force producers to populate the arrival prediction anyway?  If so, what should this value be?

skinkie:
We go again into the discussion where the ideal for the consumer is to have an as much expanded feed as possible (for the purpose of filling journeyplanners and applications) and GTFS-RT should not be used to exchange 'raw' data (you would use SIRI/VDV like standards for that). I would like us to move away from exchanging partial information where work is to be done for the consumer.

barbeau:
@skinkie Producers should always include data in the feed if they have it - I definitely won't argue with that.  But I don't think it makes sense to have producers fill in data for values they don't know.  There is definitely room for clarifying how consumers should present such unknown information.  For example, in OneBusAway is a real-time arrival/departure prediction exists, we show that information to the user, and if it does not, then we show the scheduled time to the user (alternately, an app could just show "UNKNOWN").  I'd certainly be supportive of clarifying expected behavior for consumers to create less fragmented end user experience.

Back to the above case, though - in the scenario you present, how would you suggest a producer populate the `arrival` prediction field in GTFS-rt that is unknown?  IMHO it's better to leave the field empty than create a value that isn't a legitimate value a transit rider should see.  

For example, if we require providers fill both fields, I think most providers would (or do) copy arrival to departure, and vice versa.  But this creates false expectations for transit riders by telling them a vehicle will arrive or depart when it may not actually do that.

skinkie:
I would suggest the producer would fill the field as it wants it to be interpreted. If it doesn't know the value it is clear that the scheduled value must be available. If that doesn't make sense? Copy the the departure field. But by all means: do not allow consumers to make choices what the data _could_ be.

barbeau:
>But by all means: do not allow consumers to make choices what the data could be.

I certainly agree with this, and I think the spec is currently pretty clear about how a lack of real-time data should *NOT* be interpreted by consumers, although, as with any spec, I'm sure it could be improved further.

https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/trip-updates.md says:

~~~
In case there is no trip update for a scheduled trip, it will be concluded that no realtime data is available for the trip. The data consumer should not assume that the trip is running on time.
~~~

>I would suggest the producer would fill the field as it wants it to be interpreted. If it doesn't know the value it is clear that the scheduled value must be available. 

@skinkie So, in summary, I don't think we have differing opinions here as it relates to this specific proposal, in terms of relaxing the requirements on a producer where it doesn't make sense to provide a data element if it doesn't exist.  Let me know if you think differently.

skinkie:
In my perspective a producer will always have a value. It either would be a prediction, or it would be the scheduled information. Hence, the producer will be able to fill both fields?

barbeau:
Yes - GTFS-rt `stop_time_updates` should contain only real-time predictions - schedule times come from GTFS.  If a producer doesn't have a prediction for a given arrival or departure, they should leave this field empty in the GTFS-rt feed.  In that case, the consumer follows the rules for GTFS-rt downstream propagation if there are other predictions for that trip in GTFS-rt feed, and if not the consumer should not show any real-time information for that trip - it's up to them to show a scheduled time from GTFS or show no information for that trip.  But the spec is clear that the consumer should not make up a real-time value for that arrival/departure.

skinkie:
You are contradicting yourself. If we follow your proposal and allow for empty arrival xor empty departure times, the consumer **must** infer values. By my given example (schedule arrival exceeds the predicted departure) this already suggest the consumer must copy the departure to the arrival time to have positively increasing times. In any other case where no prediction (at this stop) is available but the next stop has a prediction which is before the previous stop scheduled time, the consumer also must infer times in order to maintain positively increasing stop times. Here business logic is being pushed to all clients, where the business logic itself (incl. corner cases) is not being documented.

I want us to move away from producers that don't want to provide complete - directly consumable information. Your suggestion is exchange what is known (at the source). In my personal opinion: that is not good enough for travel information services, the producer fails to supply a high quality realtime feed, and must improve its GTFS-RT producing software, not avl system.

barbeau:
>If we follow your proposal and allow for empty arrival xor empty departure times, the consumer must infer values.

No - the same GTFS-rt rules apply.  In your example, if no real-time arrival prediction is in the GTFS-rt feed for Stop A, then the consumer follows this logic:

1. If there is an upstream delay of Stop A, it gets propagated downstream to be the arrival prediction for Stop A, and this is the prediction shown to the rider
1. If there is no upstream delay of Stop A, the consumer can show a scheduled arrival time for Stop A to the rider, or no arrival time for Stop A

So ultimately it's up to the producer to provide data that makes logical sense, either based on upstream predictions or by specifying values for a specific stop.  Garbage in, garbage out - consumers aren't responsible for trying to reconcile an upstream delay that's propagated downstream and results in a predicted arrival time after a predicted departure time - it's the producers responsibility to make sure they provide the `NO_DATA` field to stop downstream propagation or provide an accurate prediction. 
 Better GTFS-rt validation tools like the we've been working on should help flag these problems going forward.  But IMHO it is worse for a producer to make up a predicted arrival time for the first stop in the trip, and have consumers show this to transit riders when it doesn't actually represent what the transit vehicle is doing.  It's better to leave this empty as an unknown, to tell consumers not to show any predicted arrival information to riders.

Also - I know of at least one production GTFS-rt feed from a major US AVL vendor that is only providing departure for first stop on the trip, and is only providing arrival for last stop on the trip, so some producers are already ignoring the text I'm proposing to remove.  So, in our [GTFS-rt validation tool](https://github.com/CUTR-at-USF/gtfs-realtime-validator), we'd either need to flag this feed as having errors, or remove the text requiring producers to always populated both `arrival` and `departure`.

barbeau:
Reviewing the spec, elsewhere it's clear that arrival OR departure can be provided.

[Stop Time Updates](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/trip-updates.md#stop-time-updates) narrative section says:

>The update can provide a exact timing for arrival and/or departure at a stop in StopTimeUpdates using StopTimeEvent.

From [StopTimeUpdate](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-stoptimeupdate), for `arrival` and `departure`:

>If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty

The text I'm proposing to eliminate clearly conflicts with both of these, and is likely left over from an early version of the spec and Google's initial implementation.

barbeau:
I've updated my proposal above to include references elsewhere in the spec that a stop_time_update only needs to contain an arrival or a departure, but not both.  The text I'm proposing to remove conflicts with these other definitions and is likely left over from an early draft version of the spec.

I'd like to call for a vote on this proposal.  Voting will end on November 6th at 23:59:59.

barbeau:
Well, we officially got crickets for the vote over the last week (i.e., no votes).

I'd like to call for another vote - hopefully everyone's calendars are less packed this week.

To summarize, currently the GTFS-realtime spec has conflicting information over whether an arrival AND a departure prediction must be provided for each stop_time_update.  This proposal addresses the conflict by removing the text that says you must provide both an arrival AND departure, which appears to be left over from an earlier version of the spec.  The remaining language in the GTFS-realtime spec then clearly states that producers can provide an arrival OR a departure prediction.

Voting will end on Tuesday November 14th at 23:59:59 UTC.

Please vote via comment on this pull request.

DaveBarker:
+1

ritesh-warade-ibigroup:
+1

mgilligan:
+1

mike-swiftly:
+1

It is indeed clear that not all stops should have both an arrival and departure time, such as first and last stops of trips. Plus because of the old requirement most agencies have just been using the same value even though the arrival time should usually differ from the departure time if both are going to be set.

gaetano7:
+1

harringtonp:
+1  Don't see a problem with it. And in my own implementation cater for it

              // If only one of arrival or departure is specified in StopTimeUpdate, let the other assume its value
              arrival = departure = null;
              if (stoptimeUpdate.hasArrival()) arrival = stoptimeUpdate.getArrival();
              if (stoptimeUpdate.hasDeparture()) departure = stoptimeUpdate.getDeparture();
              if (arrival == null  || (!arrival.hasTime() && !arrival.hasDelay())) arrival = departure;
              if (departure == null  || (!departure.hasTime() && !departure.hasDelay())) departure = arrival;

skinkie:
@harringtonp the point with people not seeing the problem is the actual problem. These 5 lines of code in your programming language will be needed to be implemented for **all** consumers, and ran for each stoptimeUpdate processed, opposed to one single time at a producer, with no complexity of interpretati or ambiguity.

abyrd:
@skinkie this seems to be intended for the case where the data producer actually does not have a prediction for either arrival or departure. If I am not mistaken, it is the equivalent of a NULL value in a database (as opposed to a default value or empty string). In that case, forcing the data provider to duplicate these fields is a sort of artificial precision. Can a supporter of this proposal confirm my interpretation?

skinkie:
@abyrd GTFS-RT is always about a prediction from the source, it is not a punctuality (raw value) like we have in The Netherlands. Hence the producer is always using _artificial_ precision. My reading of the current standard with this line is: you must update both arrival/departure if both were specified in GTFS. My interpretation of that is preventing arrival RT can become larger than departure Scheduled, of which the prediction system at the producer may not be aware.

abyrd:
@skinkie with your comment in mind I've re-read this whole comment chain. To summarize: 

@barbeau said of a feed missing arrival predictions: "we either need to flag this feed as having errors, or remove the text requiring producers to always populate both arrival and departure (i.e., this proposal)." This is the core issue: is this feed wrong, or is the spec wrong?

In his comment https://github.com/google/transit/pull/65#issuecomment-337032447 he points out that requiring both arrival and departure times contradicts existing text in the specification. Since specification coherency is obviously a requirement, rejecting the present proposal implies another change where the rest of the spec is changed to always require both arrivals and departures.

So @skinkie your rejection of this proposal seems to be to be a small element of a larger, more general objective that has perhaps not been completely articulated: eliminate missing information from all RT feeds. Your response to the question above would be that the feed is wrong, should be considered invalid, and all GTFS-RT producers should be updated to follow the revised spec.

The logic behind that broad change is that all stop time updates are predictions rather than observations, and are coming out of a system that was designed to perform predictions. Such systems should fully assume their role and provide complete predictions using principled methods, rather than forcing countless consumers to perform their own ad-hoc / amateur propagation.

Broadly I'm in agreement with this architectural position -- it's one that we debated in the Netherlands where we all came to the conclusion that @skinkie is advocating. But I don't know how realistic it is to require such modifications to RT producing systems at this stage.

TeXitoi:
-1

I completely agree with @abyrd comment just bellow.

barbeau:
> In that case, forcing the data provider to duplicate these fields is a sort of artificial precision. Can a supporter of this proposal confirm my interpretation?

@abyrd Correct, and well put.

>Since specification coherency is obviously a requirement, rejecting the present proposal implies another change where the rest of the spec is changed to always require both arrivals and departures.

Correct.  And given the number of +1's on this proposal (and the GTFS-rt change process requiring unanimous consent to adopt a proposal) an counter proposal for removing the arrival OR departure text and requiring that arrival AND departure predictions always be provided will also fail.  So if this proposal fails we'll likely be stuck in a deadlock with a spec that contradicts itself.

>But I don't know how realistic it is to require such modifications to RT producing systems at this stage.

I think very unlikely (hence this proposal), although I'd welcome other producer's comments on this.

skinkie:
> I think very unlikely (hence this proposal), although I'd welcome **other** producer's comments on this.

this. But I also hope that @RachM could give some accurate number of implementations that are unambiguous in arrival/departure times and which would fail to pass this test.

gaetano7:
TriMet could easily provide both arrival and departure StopTimeEvents all the time. For example, include an arrival at the start of trip that doesn't have a real arrival time. This would just be a copy of the departure time as it is in the GTFS schedule. And this would be silly just like its generally agreed that including both in the GTFS is silly when one doesn't really exist. And the result might  be riders are shown silly data by consumers of GTFS-RT.

As a tentative suggestion, this could be resolved if the specification included a way to indicate the arrival time or departure is not useful and should be shown to riders.

If StopTimeEvent could have a ScheduleRelationship, and ScheduleRelationship had a new value called NOT_RELEVANT, NOT_RELEVANT could be set on the arrival or departure StopTimeEvents that are always present to indicate to consumers its not a good idea to display it to riders.

mike-swiftly:
Requiring both arrival and departure is problematic in that it causes
agencies to almost universally specify the same time for the arrival and
departure fields. But we don't know if it is an arrival time or a departure
time. Each application has to make assumptions. By being able to specify
just an arrival time or just a departure time then agencies are also
specifying whether the time is an arrival or departure. No assumptions need
to be made.

*Michael Smith*
*CIO, **Swiftly, Inc*
(415) 260-4700

barbeau:
FWIW, I did a quick analysis on some GTFS-realtime TripUpdate feeds cataloged in [TransitFeeds.com](http://transitfeeds.com/search?q=gtfsrt) (using [this tool](https://github.com/CUTR-at-USF/transit-feed-quality-calculator) and [a branch](https://github.com/CUTR-at-USF/gtfs-realtime-validator/tree/arrivals-and-or-departures) of the gtfs-realitime-validator).

Of the 29 TripUpdates feeds analyzed:
* 25 provided an arrival OR departure (not both) for at least one trip
* 15 provided an arrival AND departure that were the SAME for at least one trip
* 12 provided an arrival AND departure that were DIFFERENT for at least one trip

I know at least one of the 15 providers with identical arrivals and departures (Tampa) doesn't actually have independent departure predictions - in the early days of GTFS-realtime Google requested that we fill in both fields (i.e., copy `arrival` to `departure`) after we submitted it to Google Transit, even though in the underlying data we had only arrival times.  This wasn't originally intended to be a long-term change but it's stuck so far and is still in production. I'm guessing that others went through a similar process.

JFWade:
It really doesn't matter much to us, RTD Denver, since 99% of our arrival and departure times are the same (95% for timing points). So the impact to is is negligible. Still, to us it seems the flexibility of the OR is better.  So this would be a vote 'for' this proposal.

barbeau:
Voting closed last night at midnight UTC time, and here are the results:
* Yes - 7
* No - 2

The [current spec amendment process](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md) requires unanimous consensus, so this proposal failed to pass.

IMHO this is a significant problem, as the GTFS-realtime spec currently contradicts itself on this topic (see [this post](https://github.com/google/transit/pull/65#issue-241056495)).

@skinkie @TeXitoi Being the two no votes, where do you suggest we go from here?

As @gaetano7 outlined in https://github.com/google/transit/pull/65#issuecomment-342918760, we could require both arrival and departure in feeds but add another field to signify if each arrival and departure should be shown to riders.  Pros are that maybe we could get to a unanimous consensus with this approach, but IMHO this adds more complexity and potential confusion for both producers and consumers.  It seems much simpler to just omit values from the feed that shouldn't be used by consumers (this proposal).

skinkie:
@barbeau I would go for consensus if we would introduce an explicit level of quality in the feed. Hence something like the Tim Berners Lee star system, but then for the quality of realtime transit data. As discussed (years ago) on the mailinglist some data is more purified then others. That means that GTFS-RT is used as interexchange between systems, opposed to directly consumable by a journey planner or departure time system. I would grade a feed which does not include both departure times lower than a feed which has "artificial predictions", and can be directly consumed. Tomorrow and Friday I am probably close to @abyrd to discuss a proposal like this.

TeXitoi:
@skinkie seems promising.

abyrd:
We have already needed to confront the fact that in practice some prediction systems provide more complete data than others. Although it seems impossible for a prediction system to function without knowing the planned departure times of each vehicle, and impossible for a prediction system to guess whether and how much that planned schedule will be delayed without predicting the corresponding arrival times, the fact is that some existing systems do not export this information. These are rigid systems put in place by rigid processes that are unlikely to be revised anytime soon.

We don't want to dogmatically prevent sharing of this data, which requires error-prone consumer-side heuristics, interpolation, and propagation. 

On the other hand, we want to allow complete high quality predictions to be transmitted where they are available, and for consumers to strictly validate and trust that data. The practical solution we came up with in NL was the strict / non-strict GTFS-RT distinction. So I agree with @skinkie that we should make a proposal to standardize this concept. I'm in Delft until Saturday, we should be able to find some time to formalize that proposal.

abyrd:
I should clarify though that I don't really consider this a "grade" or star system. Just an on/off distincton: should the consumer need to guess and interpolate, or can it trust the data and expect it to be coherent?

abyrd:
@barbeau I agree that this proposal cannot just be set aside: refusing it requires a counter-proposal, because unless action is taken the spec remains contradictory. It's seems quixotic to force all existing producers to "behave" by voting on details of a specification, but at the same time we don't want to pull all systems downward to a lowest common denominator of ambiguity. We should probably allow systems to declare their target level of completeness and reliability - we'll submit a proposal that attempts to resolve this.

abyrd:
To add another layer to this debate: I can't imagine how or why a prediction system would have arrival times but not departure times. Has anyone ever encountered this in practice? Is the arrival time even relevant or significant to passengers? Aren't these systems leaving it out specifically because it's not considered relevant?

Is it then accurate to say that all prediction systems always provide a *departure* prediction, and only some provide a corresponding arrival prediction? If that's the case, then an alternative is to always require a departure prediction (since that's the one the passenger cares about) and make the arrival prediction optional.

TeXitoi:
@abyrd except the arrival time is needed for journey planning.

abyrd:
Please neglect my previous (deleted) comment about relevancy of arrival times. Of course the arrival time is equally relevant at the final stop of any trip, where the passenger alights.

barbeau:
@abyrd @skinkie Alternate proposal would be great!

>Just an on/off distincton: should the consumer need to guess and interpolate, or can it trust the data and expect it to be coherent?

Our biggest collective problem in the industry at this point is feed data quality.  My sense (although I could be wrong) is that the American AVL vendor ecosystem is more diverse and less regulated than the European one, which might be part of what's driving different perspectives on this problem.

One item I'd like to see addressed in an alternate proposal is how a consumer can objectively evaluate the provided data (e.g., an agency during acceptance testing of a vendor's product) to determine if it can indeed be trusted and coherent. My main concern is that strongly encouraging or requiring the population of certain fields will result in artificial information that makes it even more difficult for consumers to provide quality information to end users.  In the U.S., feed consumers get the vast majority of negative feedback from end users when there are data quality issues, because the consumer and producer software are typically owned by different parties.  Delivering end-user feedback to producers in an actionable format is not simple (we actually just finished a project trying to address this by leveraging Open311 with OneBusAway - details [here](https://www.slideshare.net/sjbarbeau/2017-seeclickfix-workshop-closing-the-loop-improving-transit-through-crowdsourced-information), with a forthcoming TRB 2018 paper).

For example - `vehicle.timestamp`. I want this value ONLY if it's the ACTUAL vehicle GNSS timestamp, not another timestamp in the data pipeline.  However, determine whether a timestamp actually DOES come from the vehicle is not a simple task.  I'd prefer that vendors omit this value if they don't have it, rather than populate it with sub-par artificial data.  In the former case we just don't show this missing information to end users.  In the later case, end users complain to consumers who then carry the burden of figuring out why the user is complaining and providing actionable information to the producer if it is indeed an error in the feed.

abyrd:
Another nuance I'd like to commit to the record: Many predictions supplied by an API are expected values somewhere in a probability distribution. When a value is missing and the consumer copies an arrival prediction into a departure prediction field, that departure transforms into something like a *lower bound* on departure time, conditional on the accuracy of the arrival time prediction; when the consumer copies a departure into an arrival field, it becomes something like an *upper bound* on arrival time, but again conditional on the accuracy of the departure prediction. While these bounds are reasonably conservative / safe for finding transfers or other trip planning operations (*if* the copied prediction is accurate) they are of a different nature than the original prediction (unless it was itself inteded as a conservative bound rather than expected value). There are a lot of assumptions and shifts in meaning when applying generic consumer-side heuristics. We do need a way for a producer to signal that it has made a serious effort to properly account for the nature of its predictions and keep everything coherent.

abyrd:
Thanks for the additional details @barbeau. Clearly we don't want data pipelines to "lie" about the existence and quality of data when they simply don't have it. We want to allow innovative consumers to make better, more relevant guesses or simply tell users when information is absent. 

But on the other hand, in cases where more coherent data pipelines have been achieved, we want to enable data consumer systems to trust their data sources. I don't think there's a one-size-fits-all rule here, we'll probably have to let producers declare their intent to comply with stricter definitions.

mike-swiftly:
Ever since the beginning of real-time info for buses, arrival predictions,
instead of departure predictions, have been provided by many systems
because 1) they are more conservative in general since they are of course
before the departure time and therefore passengers are then less likely to
miss their bus; 2) if a bus doesn't need to stop because no passengers are
getting off then the departure prediction will be slightly off, resulting
in a passenger at a stops to possibly miss their bus; and 3) you don't want
passengers arriving when the doors have shut (the departure time). Yes, the
difference is only the dwell time, which might only be around 20 seconds or
so. But if the goal is to provide the most accurate and useful predictions
possible then the difference is important.

Vehicles that always stop for a while, such as rail vehicles, or course
behave differently, though arrival predictions are still useful for them as
well.

And yes, people really do use real-time info systems to time exactly when
to get to a bus stop. I don't recommend it, but people definitely do so.

*Michael Smith*
*CIO, **Swiftly, Inc*
(415) 260-4700

abyrd:
I had the chance to discuss this in person with @skinkie and came to agreement that:

1. Clearly the specification must be coherent, the existing spec apparently intended "arrival and/or departure" (rather than both) to be required in each message (see original PR comment edit on October 30th 2017), and we cannot break backward compatibility;
2. Though in some systems we want to use GTFS-RT as an interface with no ambiguity or consumer-side interpretation, it is extremely unlikely that many existing systems will be retrofitted to do so, and they will continue to produce data.

Therefore we can both vote +1 or +0 to establish a consensus on this ticket, with the understanding that an additional proposal will be made allowing RT producers to declare their data to have certain stricter semantics and characteristics. That proposal will focus on what I referred to above as the more general objective that has not been articulated: eliminate missing information from RT feeds.

I will make that proposal on the RT mailing list and @barbeau we should be able to proceed with a final vote on this PR as soon as that proposal is posted.

barbeau:
@abyrd and @skinkie It seems the effort for the "strict flag" proposal is dragging out (https://groups.google.com/forum/#!topic/gtfs-realtime/JGITCQLw8Ww) - it's been over 2 months since we closed this pull request.  In the mean time, we've got a GTFS-realtime spec that contradicts itself.

Would you be open to re-voting to adopt the language in this PR to remove the contradiction, and we can pursue the "strict flag" discussion/proposal in parallel?

skinkie:
I would like to solve this at the same time, not in parallel.

abyrd:
Personally I'm not in a big hurry to see the "strict" flag adopted. I am prepared to vote +1 and finish up the change in this PR just knowing that the "strict" idea has been proposed and is being considered by the community. I am just following through on this idea we formulated in the past. 

In the proposal discussion on the mailing list Sean does make a good point: if you don't want the consumer to apply the propagation rules, you can just provide stop times for every stop and the consumer should not infer anything. The main thing is validation (knowing whether to issue warnings when such completeness is not attained).

skinkie:
> The main thing is validation

And the problem with validation is that you would need to do it for every message. Which is bad.

abyrd:
@skinkie do you just mean that applying "strict" validation to every message in a feed that had (allowed) missing values would yield huge amounts of spurious warnings? That's what I was thinking  was bad.

skinkie:
If a feed announces strict, it should go in to a fast-path of the consumer code, ignoring all validation. If strict was not provided, validation must be applied, because it can't be known that the producing application took effort to produce a complete prediction.

barbeau:
Ok - I'm re-opening this proposal and calling for a vote so we can hopefully eliminate the contradiction from the spec.

**Summary**

Of 29 TripUpdates feeds [I analyzed](https://github.com/google/transit/pull/65#issuecomment-343218895) (both for time and delay):

* 25 provide an arrival OR departure (not both) for at least one trip

The GTFS-realtime spec currently [has conflicting information](https://github.com/google/transit/pull/65#issue-241056495) over whether an arrival AND a departure prediction must be provided for each stop_time_update.  As a result, all 25 of these feeds could be considered non-compliant given the current spec language.

**Proposal**

This proposal addresses the conflict by removing the text that says you must provide both an arrival AND departure. The remaining language in the GTFS-realtime spec then clearly states that producers can provide an arrival OR a departure prediction.  As a result, all 25 of these feeds are considered compliant with the GTFS-realtime spec.

To be clear, if this proposal is approved, it does **not** prevent the "strict flag" proposal at https://groups.google.com/forum/#!topic/gtfs-realtime/JGITCQLw8Ww from being adopted.  That proposal can continue.

Voting will close Monday Feb 12th at 23:59:59 UTC.

skinkie:
Would you be willing to add something in proper English like: "If an arrival prediction is provided that is after the scheduled departure time and no departure prediction is provided, the receiving system must resolve the prediction, or ignore the delay. If a departure prediction is provided that is before the scheduled arrival time and no arrival prediction is provided, the reiving system must resolve the prediction with respect to potential provided timepoints, or ignore the speed up. If the prediction contains a conflict, the user must ignore the message."

barbeau:
@skinkie I'd really prefer to resolve this conflict in the spec by removing the conflicting sentence without any other additions, and then if there are holes in the current propagation logic patch those via other individual proposals.

skinkie:
@barbeau I am not going to let this standard to conform to existing poor implementations, without its users being warned over these implementations. Hence I would rather be in favor to make them both mandatory _"Both arrival and departure must be provided."_ and make all existing implementations incompliant than make existing producing implementations compliant without telling new existing and users of the implications of such.

mike-swiftly:
+1 on accepting this pull request

abyrd:
> On Feb 2, 2018, at 21:18, Stefan de Konink <notifications@github.com> wrote:
> If a feed announces strict, it should go in to a fast-path of the consumer code, ignoring all validation. If strict was not provided, validation must be applied, because it can't be known that the producing application took effort to produce a complete prediction.
> 
Although I’m sympathetic to the idea of more complete data, I don’t really understand this line of argument. A consumer component must always validate received updates before they are applied to a production system to avoid nonsense outcomes or crashes in the event that corrupted or non-conforming data are received. 

Besides being necessary, I would expect this validation to have very low computational cost, possibly low enough to be entirely negligible in  server environments. Even if there was some risk it would have an impact, that impact has not been measured or even demonstrated theoretically, so adjusting the specification on this basis amounts to premature optimization, upstream of even writing code.

It does seem necessary to remove the contradictions from the spec independent of whether people want a “strict” flag or not. The important thing is that serious consideration be given to the strict mode idea, which seems more likely if it’s a separate discussion item.

abyrd:
> On Feb 5, 2018, at 16:22, Stefan de Konink <notifications@github.com> wrote:
> @barbeau <https://github.com/barbeau> I am not going to let this standard to conform to existing poor implementations, without its users being warned over these implementations. Hence I would rather be in favor to make them both mandatory "Both arrival and departure must be provided." and make all existing implementations incompliant than make existing producing implementations compliant without telling new existing and users of the implications of such.
> 
The GTFS family of specifications has a long history of tolerating spotty data and remaining backward compatible. These are among the reasons it is preferred in many places. This change would go against both characteristics.

I understand the idea of pushing people toward clearer, less ambiguous data sources, but it seems to me that would be better achieved by the backward-compatible strict mode, or simply creating real-time producer components that produce complete unambiguous output.

Again though, that discussion should probably happen in a separate proposal. As long as people are considering such a separate proposal, then it seems like a good idea for this one eliminating contradictions to proceed.

harringtonp:
As before +1 on accepting the request

abyrd:
I am also +1 on this PR since discussion of strict mode etc. are happening in other places.

DaveBarker:
+1 on accepting the request.

ritesh-warade-ibigroup:
+1 for Ritesh (IBI)

gaetano7:
+1

barbeau:
So the current tally is:

Yes - 6

@skinkie Could you please clarify your vote on this proposal?  Is it +1, -1, or abstain (e.g., +/-0)?

skinkie:
@barbeau I have asked Google to give their reply on this subject. I am happy if _they_ would overrule my -1.

barbeau:
@skinkie Voting is closed (it ended last night), so there are no more votes.  Are you saying your vote is -1?  If so then this proposal fails.

abyrd:
I don't understand the idea of Google "overruling a -1". Are there rules that allow special intervention by representatives of Google?

skinkie:
I have asked Google to give their engineering point of view on this topic. I am _personally_ happy to follow their +1.

barbeau:
There are no rules that allow special intervention by Google (see the official spec amendment process [here](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md)).

@skinkie Follow the official spec amendment process, as announced in my call for votes [earlier in this thread](https://github.com/google/transit/pull/65#issuecomment-363125418) voting ended Monday Feb 12th at 23:59:59 UTC.  I'm giving you the courtesy of clarifying your vote, as it appeared that you were voting -1 in [your comment](https://github.com/google/transit/pull/65#issuecomment-363136302) but you never explicitly stated this.

So, last chance - do you want to clarify your earlier comment as a vote, or not?

skinkie:
Then I would have to vote -1.

barbeau:
Ok, here are the results:
* Yes - 6
* No - 1

The [current spec amendment process](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md) requires unanimous consensus, so this proposal failed to pass.

Again, [as I stated earlier](https://github.com/google/transit/pull/65#issuecomment-344642111), IMHO this is a significant problem, as the GTFS-realtime spec currently contradicts itself on this topic (see [this post](https://github.com/google/transit/pull/65#issue-241056495)).

@skinkie Thoughts for moving forward from here?  You're the only -1.

skinkie:
@barbeau I'll send you an e-mail. I am in no way blocking out of principle. Hence I think we may solve this in a few days.

skinkie:
@ca9 thanks for participating too :)

My vote++, that makes it 0, we can continue :)

barbeau:
Thanks @ca9!

Ok, hopefully third time is a charm. I'm calling for a vote again so we can hopefully eliminate the contradiction from the spec.

**Proposal recap**

This proposal addresses the conflict by removing the text that says you must provide both an arrival AND departure. The remaining language in the GTFS-realtime spec then clearly states that producers can provide an arrival OR a departure prediction.

To be clear, if this proposal is approved, it does **not** prevent the "strict flag" proposal at https://groups.google.com/forum/#!topic/gtfs-realtime/JGITCQLw8Ww from being adopted.  That proposal can continue.

Voting will close Tuesday April 17th at 23:59:59 UTC.

harringtonp:
+1

DaveBarker:
+1

skinkie:
+0

mike-swiftly:
+1

gaetano7:
+1

JymDyer:
+1

mgilligan:
+1

JFWade:
+1

barbeau:
Ok, here are the results:

Yes - 9
No - 0
Abstain - 1

So it passes! Thanks to everyone for helping to clarify this issue in the spec. 

@RachM Could you please merge?

PR #66, opened 2017-07-11
Add repeated CarriageDescriptor to VehicleDescriptor to encapsulate carriage specific information

RachM:
This PR adds `CarriageDescriptor` as a repeated field to `VehicleDescriptor`. The goal is to allow carriage-specific information to be specified in `TripUpdates` and `VehiclePositions`.

The repeated field was chosen as it allows producers to specify information about every carriage. In the instance where there is a single carriage (e.g. a standard bus), then producers can specify one.

The type of information that can vary by carriage that I'm interested in is:
- Wheelchair accessibility
- Occupancy status
- Air conditioning
- Toilet facilities
- WiFi
- Bicycle storage

Announced on GTFS-realtime Google Group at: https://groups.google.com/forum/#!topic/gtfs-realtime/VFkZYh0pnWM


abyrd:
Hi @RachM, it does seem like a good idea to describe the composition of multi-carriage trains. Many rail travelers would appreciate this kind of information. 

In an existing system in the Netherlands two additional fields are present: train-equipment-type and train-equipment-code. In many cases these imply the other amenity fields so are somewhat redundant, but it's could imply additional detail (I'm thinking of double-deck trains or large seatless areas near doors). In a particularly good implementation, it may actually allow you to visually recognize the train in the station. Check out https://drgl.nl/stop/NL:S:ut and click on any train to see its status with images and names for the carriage types, derived from such fields.

@skinkie or @somoht might be able to give you more information about this data source.

gtfs-realtime/spec/en/reference.md
```diff
@@ -340,9 +354,89 @@ Identification information for the vehicle performing the trip.
...
+| **label** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | optional | User visible label, i.e., something that can be shown to the passenger to help identify the correct carriage. |
+| _**occupancy_status**_ | _[OccupancyStatus](#enum-occupancystatus)_ | _optional_ |The degree of passenger occupancy of the carriage.<br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future.|
+| **wheelchair_accessible** | [WheelchairAccessible](#enum-wheelchairaccessible) | optional | Whether the carriage is wheelchair accessible. |
+| **toilet_facilities** | [ToiletFacilities](#enum-toiletfacilities) | optional | Whether the carriage has toilet facilities onboard. |
+| **wifi_availability** | [WifiAvailability](#enum-wifiavailability) | optional | Whether the carriage has WiFi onboard. |
+| **air_conditioning** | [AirConditioning](#enum-airconditioning) | optional | Whether the carriage is air conditioned. |
+| **bicycles_allowed** | [BicyclesAllowed](#enum-bicyclesallowed) | optional | Whether bicycles are allowed in the carriage. |
+
+## _enum_ WheelchairAccessible
```
barbeau: Wheelchair accessibility always seems to be a tricky one to tackle.  I'd love to see this included, although I'd be curious to hear from producers whether or not carriage wheelchair accessibility can be appropriately modeled as a boolean (effectively, given the below enumerations).  For example, does `WHEELCHAIR_ACCESSIBLE` mean that a wheelchair can board the carriage (i.e., travel through the doorway)?  Or does it mean that there is accessible seating/tiedowns for a wheelchair?  Are toilet facilities accessible to a wheelchair?  All or none of the above?

gtfs-realtime/spec/es/reference.md
```diff
@@ -347,7 +422,7 @@ Un mensaje internacionalizado que contiene versiones por idioma de un fragmento
 
 | _**Nombre del campo**_ | _**Tipo**_ | _**Cardinalidad**_ | _**Descripción**_ |
 |------------------------|------------|--------------------|-------------------|
-| **translation** | [Translation](#Translation) | repetido | Se debe proporcionar al menos una traducción. |
+| **translation** | [Translation](#mensaje-translation) | repetido | Se debe proporcionar al menos una traducción. |
```
barbeau: I don't believe this change should be in the PR?

gtfs-realtime/spec/es/reference.md
```diff
@@ -336,7 +411,7 @@ Un selector para una entidad en un feed GTFS. Los valores de los campos deben co
 | **agency_id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | opcional |
 | **route_id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | opcional |
 | **route_type** | [int32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | opcional |
-| **trip** | [TripDescriptor](#TripDescriptor) | opcional |
```
barbeau: I don't believe this change should be in the PR?

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -574,6 +574,115 @@ message VehicleDescriptor {
...
+      // The carriage has air conditioning.
+      AIR_CONDITIONED = 1;
+
+      // The carriage does not have air conditioning.
+      NOT_AIR_CONDITIONED = 2;
+    }
+
+    optional AirConditioning air_conditioning = 7 [default = UNKNOWN];
+
+    // Whether bicycles are allowed in the carriage.
```
barbeau: I think `BicyclesAllowed` would be useful outside of just the carriage use case - agencies have told me they'd like to see this for buses, including real-time capacity of racks, although I'm not aware of any real-time system that currently exposes this via an API.

gtfs-realtime/spec/en/reference.md
```diff
@@ -20,13 +20,27 @@ Version 1.0 of the feed specification is discussed and documented on this site.
...
+                    *   [BicyclesAllowed](#enum-bicyclesallowed)
             *   [StopTimeUpdate](#message-stoptimeupdate)
                 *   [StopTimeEvent](#message-stoptimeevent)
                 *   [ScheduleRelationship](#enum-schedulerelationship)
         *   [VehiclePosition](#message-vehicleposition)
             *   [TripDescriptor](#message-tripdescriptor)
                 *   [ScheduleRelationship](#enum-schedulerelationship-1)
             *   [VehicleDescriptor](#message-vehicledescriptor)
+                *   [CarriageDescriptor](#message-carriagedescriptor)
+                    *   [OccupancyStatus](#enum-occupancystatus)
```
barbeau: There was a design decision made when originally adding `OccupancyStatus` that `VehicleDescriptor` should only contain immutable information, while `VehiclePosition` should contain mutable statuses (discussion starting at https://groups.google.com/d/msg/gtfs-realtime/_HtNTGp5LxM/86U3GGxDbngJ).  

This proposal breaks from that design decision by putting `OccupancyStatus` as a child element of `VehicleDescriptor->CarriageDescriptor`.  You could also argue that some of the other elements are mutable as well (e.g., bathroom closed for cleaning/maintenance, bikes allowed only at certain times of the day).  

At the time of the `OccupancyStatus` proposal, I actually argued for including `OccupancyStatus` in `VehicleDescriptor` instead of `VehiclePosition`, but was overruled :).  I bring this up for discussion - please discuss :).

gtfs-realtime/spec/en/reference.md
astein6: Hi everyone, Sean asked me to comment on this thread. I lead a research center on accessible transportation which includes both IT and vehicle accessibility research. Being able to specify a vehicle is wheelchair accessible is important but there are some subtle nuances you all might want to consider. 

As Sean mentioned, there's a difference between accessible boarding and there being accessibility features inside the vehicle. There are also cases where the accessible boarding is only at certain stations that have the right height platforms or portable lifts. Furthermore, wheelchair access is sometimes limited to specific cars on a train. Since boarding is often tied to stations or even train platforms, you may want to split boarding out from vehicle interior. Perhaps something like AccessibleBoarding (yes, no, accessible platforms only, unknown) and WheelchairParkingArea (integer from 0). These are just brainstorms since I don't know the rest of the vehicle spec discussions.

I recommend adding something to the toilet and dining car elements (if any) to indicate whether those facilities are accessible, rather than trying to fold it into a single accessibility element.

As an aside, you may also want to think about in-vehicle announcements as an element too. Visual message signs, audio announcements, etc. Those have universal value beyond just deaf and blind riders.

gtfs-realtime/proto/gtfs-realtime.proto
RachM: The idea is that a bus amenities/info can be modelled by having a single CarriageDescriptor. Hence this works for buses.

Capacity for the rack is interesting; it could certainly be added at a later stage.

gtfs-realtime/spec/es/reference.md
RachM: No worries - I'll split it out into a separate PR.

gtfs-realtime/spec/es/reference.md
RachM: No worries - I'll split it out into a separate PR.

gtfs-realtime/spec/en/reference.md
RachM: Thanks so much for the insightful comments! It's fantastic to have an accessibility expert weigh in; I'm keen to get this right.

In regards to having the accessible boarding tied to the vehicle; I've experienced both a stop/station needing to be accessible as well as the vehicle. We already capture stop accessibility (wheelchair_boarding in stops.txt), but I thought it would be useful to capture vehicle as well (for the scenario where only _some_ vehicles servicing the stop are wheelchair-boarding accessible). The use case I have in mind is [buses in Sydney](http://www.cityofsydney.nsw.gov.au/explore/getting-around/accessibility/accessible-public-transport); not all are wheelchair accessible, and I think it would be great to show this information to users so they can better plan their trips.

I think it's clear that my original accessibility field is too simplistic. Maybe we need a whole message to encapsulate all of the relevant accessibility information? Maybe it doesn't all belong in a message? I'll think more about this....

Summarizing the potentially important accessibility features:
- Accessible boarding
- Accessible boarding instructions?
- Wheelchair parking area
- Accessible seating
- Accessible toilet facilities
- Braille signage
- Audio announcements
- Visual announcements

Anything else that I'm missing?

jxeeno: Potentially hearing "T" loop for the audio announcements as well?

A common complaint from wheelchair users is that there's no way to determine if the designated wheelchair spaces on-board are already occupied.  It's one thing to have a wheelchair parking space, but they're only useful if they're available.  It's a similar scenario for accessible seating.

Perhaps availability of wheelchair parking can be expressed as:
* There is a parking space, but they are all occupied
* There is a parking space, and there is at least one available
* There is a parking space, and its availability is unknown (if the operator cannot determine its availability in real-time)
* There is no parking space

gtfs-realtime/proto/gtfs-realtime.proto
RachM: Also, maybe 'Carriage' is the wrong word to use? I want it to be used for all vehicle types, and if that's not clear, we should change the name.

Let me know if you have other name suggestions. Car? Compartment? Section?

gtfs-realtime/spec/en/reference.md
RachM: Great point! One concern; are producers able to capture this type of information?

jxeeno:
With in-seat power being made available in some services now, that could be a feature that's of interest to commuters.  You could identify types of ports available e.g. USB, GPO, both, none.

gtfs-realtime/spec/en/reference.md
astein6: Note that audio announcements can vary between computer announcements and operator announcements. The latter are notoriously unintelligible in most systems since operators frequently mumble or fail to speak loudly.

We've looked at wheelchair space occupancy in our crowdsourcing transit information app (Tiramisu) and intentionally opted for overall vehicle loading instead. In most cases there is a direct relationship between loading and wheelchair space availability. Our concern about about using wheelchair space in a crowdsource system was twofold. First, most non-experts might not properly classify an opening if people without disabilities are seated in those areas. Second, we were concerned about malicious reporting that spots were full so people with wheelchairs would opt for another bus, thereby discouraging long dwell times for ingress/egress.

gtfs-realtime/spec/en/reference.md
RachM: Thanks for the context Sean!

My reasoning for putting it in VehicleDescriptor:
- It seems like a property of the vehicle, and I (perhaps naively) thought VehicleDescriptor encapsulated properties of vehicles. I don't see why we should strive to maintain immutability, especially when this isn't mentioned (or enforced?) anywhere.
- By putting it in VehicleDescriptor, it gets to be in either/both TripUpdates and VehiclePositions.

I didn't realise VehicleDescriptor was intended to be immutable. If people care deeply about maintaining this, then perhaps we should at least document this in the spec?

Keen to hear others thoughts :-)

RachM:
@jxeeno and @abyrd: great idea, although I have no idea if producers a) have this info and b) can incorporate it in gtfs-rt.

**Producers:** are you able to capture this info for individual carriages?

gtfs-realtime/spec/en/reference.md
barbeau: @RachM That was my exact same reasoning for originally proposing `OccupancyStatus` in `VehicleDescriptor`.  So, I'm also interested to hear others thoughts :).

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: >Let me know if you have other name suggestions. Car? Compartment? Section?

I couldn't come up with anything better and more general than `Carriage`.  I think as long as we make it clear in the spec that this can apply to any transit vehicle this would work for buses too.

gtfs-realtime/spec/en/reference.md
```diff
@@ -340,9 +354,89 @@ Identification information for the vehicle performing the trip.
...
-| **id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | optional | Internal system identification of the vehicle. Should be **unique** per vehicle, and is used for tracking the vehicle as it proceeds through the system. This id should not be made visible to the end-user; for that purpose use the **label** field |
+| **id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | optional | Internal system identification of the vehicle. Should be **unique** per vehicle, and is used for tracking the vehicle as it proceeds through the system. This id should not be made visible to the end-user; for that purpose use the **label** field. |
 | **label** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | optional | User visible label, i.e., something that must be shown to the passenger to help identify the correct vehicle. |
 | **license_plate** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | optional | The license plate of the vehicle. |
+| **carriage_descriptor** | [CarriageDescriptor](#message-carriagedescriptor) | repeated | An ordered list of carriage information. |
```
jxeeno: Maybe we should define how the carriages should be ordered. The original proposal from NSW Trains was to use a well-defined central point. A more logical one for me is to start with the first carriage in the direction of travel and move back from there.

barbeau: >Maybe we should define how the carriages should be ordered. 

That's a really good point, especially on the accessibility attributes.  The whole point of providing this data would be to allow a rider in a wheelchair to position themselves in the proper place on the platform before the vehicle arrives, and if the order isn't well defined, they will end up in the wrong place.

>A more logical one for me is to start with the first carriage in the direction of travel and move back from there.

This would be good.  Maybe there should also be a required `carriage_sequence` int, similar to `stop_sequence`, that would explicitly define ordering?  Otherwise it will be difficult from a consumers perspective to know if implicit ordering is correct.

gtfs-realtime/proto/gtfs-realtime.proto
RachM: Cool - I've clarified the single-vehicle case in the reference as well as the proto.

gtfs-realtime/spec/en/reference.md
RachM: I've added carriage_sequence to CarriageDescriptor. Can you take a quick look and see if it looks OK?

qwasar: I definitely support adding such field, but I'm a bit unsure how it would handle a situation when a train changes direction of its travel during its journey e.g. in a dead-end station.

barbeau: @qwasar Good point. It would be good to know how this scenario is typically modeled with trip_ids in GTFS and GTFS-realtime. If the trip_id changes, which I would imagine it would, then I would expect carriage_sequence to change with the trip_id, which should have a different direction_I'd. One option would be to make GTFS trips.txt direction_I'd conditionally required if CarriageDescriptor is provided to make the behavior and expectations clearer. Consumer could change display for outgoing trip a certain amount of time become outgoing departure.

barbeau: Actually, I suppose direction_id could just be provided in TripDescriptor in GTFS-rt, but I think having it in GTFS would help too.

RachM: Yes, I think we need to add `direction_id` in `TripDescriptor` in order to help disambiguate the vehicle. Here's two ways that producers could model a vehicle with a dead-end stop in its journey:
1. Have separate `trip_id` in GTFS-static for the legs before and after the dead-end station. Carriage sequence will change (reverse) when the `trip_id` (in `TripDescriptor`) changes. Based on my own analysis of a small set of random feeds, I'm expecting this to be the most-used way.
2. If one `trip_id` is used for the journey, have separate `direction_id` (in trips.txt) in GTFS-static for the legs before and after the dead-end station. Carriage sequence will change (reverse) when the `direction_id` (in `TripDescriptor`) changes.

Does that make sense? What do you think?

qwasar: @barbeau Well, here in Europe it is not that rare for a train to change its direction (even multiple times) and still hold its designation, i.e. being modeled as a single trip. At least this is how I handle it. It'd be great if @RachM could check how it is modeled for trains running through e.g. Leipzig Hbf (DB on route Berlin-Erfurt), Hamburg Hbf (direct ICE trains on Berlin-Copenhagen route), Munich Hbf (OEBB on route Salzburg-Ulm) or Kristianstad (Öresundståg on Copenhagen-Karlskrona route). All major stations in Budapest are dead-ends as well, so long distance trains have to change direction there too.

I'm afraid employing direction_id for that purpose doesn't feel right to me semantically. If we would take this way I'd suggest a separate field for it, something like `trip/consist_orientation=FORWARD|BACKWARD`.

But I'm afraid there is another drawback with this approach as there could be only a single `TripDescriptor` (thus single `CarriageDescriptor[]`) at a time. So up till the time of arrival to the dead-end station the `CarriageDescriptor[]` would be reversed for the rest of the journey after such station. I think it'd be better to use approach similar to headsigns - i.e. `trip_headsign` can be overridden with `stop_headsing` later on. Such approach would enable us to describe even adding and removing carriages along the route of the train as well (again a technique still used for long distance trains in Europe). I know it complicates things, but I'm afraid that's reality :). Thoughts?

qwasar:
How about adding an information on dining facilities as well? With enum values like:
- restaurant (tables with seating)
- bistro (standing or bar stools only)
- in-seat service/minibar
- vending machine

Counting on having a breakfast in a restaurant carriage which is unexpectedly not present is not a good start of a day...

gtfs-realtime/spec/en/reference.md
RachM: To clarify, when I say "Carriage sequence will change (reverse)", I mean producers will send `TripDescriptors` with the elements in `carriage_descriptor` reversed. For example, using this scenario
trips
 as a reference, the `VehicleDescriptors` corresponding to a trip with `direction_id = 1` would be:
```
carriage_descriptor {
  label = 'A'
  carriage_sequence: 1
},
carriage_descriptor {
  label = 'B'
  carriage_sequence: 2
},
carriage_descriptor {
  label = 'C'
  carriage_sequence: 3
}
```
While `direction_id = 0` would be:
```
carriage_descriptor {
  label = 'C'
  carriage_sequence: 1
},
carriage_descriptor {
  label = 'B'
  carriage_sequence: 2
},
carriage_descriptor {
  label = 'A'
  carriage_sequence: 3
}
```
Note also producers can change the carriage_descriptor at any time in the trip. If carriages are added/removed at a stop, then producers simply add/remove carriage_descriptor elements.

qwasar: I think I understand your idea. My objection was that it is probably not a feasible solution for cases where such train is modeled as a single trip (which can have only a single value of `direction_id`). It would require to slice the train journey to several trips arbitrarily just for the sake of export to GTFS. Even so, using `direction_id` for this purpose still feels wrong to me, as the train (even with the order of carriages reversed) still follows the same direction in the sense of trips.txt's `direction_id` semantics.

Take the [Copenhagen-Karlskrona](https://goo.gl/maps/X3AZg8mKuj82) route as an example:
`Copenhagen-Kristianstad-Karlskrona, direction_id=0`
`Karlskrona-Kristianstad-Copenhagen, direction_id=1`
and the order of carriages has got always reversed in Kristianstad.

And if `CarriageDescriptor[]` would be tight to a `TripDescriptor` only, then while the train would be in the Copenhagen-Kristianstad section, the feed would advertise misleading order of carriages for the rest of its journey after Kristianstad. That might be quite inconvenient for passengers boarding at stations in that section and checking the status of the train in advance.

RachM: Thanks for the example; it's a really good one! I agree my use of direction_id is not appropriate, so let's scratch that idea.

I have a new solution that I think solves our problem; have **TWO** fields for the sequence:
- **next_station_arrival_sequence**: the order that the carriage will arrive at the next station
- **next_station_departure_sequence**: the order that the carriage will depart from the next station

For non-dead end stations, next_station_arrival_sequence = next_station_departure_sequence. For dead-end stations, they will be the reverse of each other. So for my dead-end picture above:

TUs/VPs before arriving at the station:
carriage_descriptor {
  label = 'A'
  next_station_arrival_sequence: 1
  next_station_departure_sequence: 3
},
carriage_descriptor {
  label = 'B'
  next_station_arrival_sequence: 2
  next_station_departure_sequence: 2
},
carriage_descriptor {
  label = 'C'
  next_station_arrival_sequence: 3
  next_station_departure_sequence = 1
}

Consumers can know from this that it's a dead-end station, and can decide on an appropriate UI/info to show to the user.

Thoughts?

qwasar: @RachM Sorry for being late with reaction, I've been off-grid.

I'm still afraid your latest proposal wouldn't allow us to handle all the cases I'd like to see addressed.

Let me expand on the example of the Copenhagen-Karlskrona train. Before boring the Citytunneln in Malmö in 2010, the Malmö Central used to be a dead-end station as well. So that train had changed its direction twice during its journey (in Malmö and Kristianstad). I can't help myself, but I think to describe such situation we really have to add `CarriageDescriptor[]` as an optional field to `StopTimeUpdate` as well. Something like this:
`TripUpdate
    TripDescriptor
        CarriageDescriptor[]=[(A,B,C)+(D,E,F)] // two three-carriages units due to high demand
    StopTimeUpdate
        stop_id=Malmö, 
        CarriageDescriptor[]=[(F,E,D)+(C,B,A)] // first reversal
    StopTimeUpdate
        stop_id=Kristianstad, 
        CarriageDescriptor[]=[(A,B,C)] // second reversal, only single unit continues to Karlskrona
`
I found it very analogous to how `Trip.trip_headsign` can be overridden by `StopTime.stop_headsign`s in GTFS.

And on top of that it would be really great to have such info in GTFS too as it is quite common that the very same trip (train number) is served by quite different carriages on weekdays and on weekend.

qwasar: More examples:
railjets at [Wien-Zürich HB](https://goo.gl/maps/DYRjNGGNvtH2) route changes direction twice at Buchs and Sargans stations.

LeoExpress at [Košice-Prague](https://goo.gl/maps/cwgEP6uTi4G2) route changes direction twice at Prešov and Přerov stations.

RachM: You're right - I was thinking mainly in terms of VehiclePositions (and hence only considered the **next** station). However, TripUpdates can specify **many** stations, so we need info about the carriage directions for each of them.

I still think it's important to know what both the arriving and departing sequence is at a particular station.

Revising:
- `CarriageDescriptor` still sits in `VehicleDescriptor`, but only contains info about the carriage facilities (no sequencing info)
- TUs: `StopTimeUpdates` contain the carriage sequences
- VPs: the carriage sequence is at the top level, and is for the stop specified in `stop_id`

More concretely:
```
message VehicleDescriptor {
  ....
  repeated CarriageDescriptor carriage_descriptor;
  message CarriageDescriptor {
    optional string id = 1;
    ... wifi, occupancy, bicycles etc
  }
}

message CarriageSequence {
  optional string carriage_descriptor_id = 1;
  optional int sequence = 2;
}

message TripUpdate {
  ...
  message StopTimeUpdate {
    ...
    optional string stop_id = 4;
    ...
    repeated CarriageSequence arrival_carriage_sequence = X;
    repeated CarriageSequence departure_carriage_sequence = X+1;
  }
}

message VehiclePosition {
  ...
  optional string stop_id = 7;
  ...
  repeated CarriageSequence arrival_carriage_sequence = X;
  repeated CarriageSequence departure_carriage_sequence = X+1;
}
```

PR #69, opened 2017-07-14
Update README.md

laidig:
Good to know that the spec is available in Spanish. Pero todo es mejor con un enlace!

### DIFF
--- a/gtfs-realtime/README.md
+++ b/gtfs-realtime/README.md
@@ -2,7 +2,8 @@ This directory contains GTFS Realtime Specification and documentation.
 
 ### Quick links
 - [GTFS Realtime protocol buffer definition](proto/gtfs-realtime.proto)
-- [Documentation](spec/en)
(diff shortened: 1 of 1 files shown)


laidig:
I signed it!

gtfs-realtime/README.md
```diff
@@ -3,5 +3,6 @@ This directory contains GTFS Realtime Specification and documentation.
 ### Quick links
 - [GTFS Realtime protocol buffer definition](proto/gtfs-realtime.proto)
 - [Documentation](spec/en)
+- [Documentación en Español](spec/es)
```
barbeau: It would probably make sense to split languages out as sub-bullets, like:

- Documentation
    * [English](spec/en)
    * [Español](spec/es)

laidig: Good point, I'm sure there are other languages coming some time soon.

PR #72, opened 2017-08-14
Sanitary additions to the specification

ca9:
This change includes sanitary suggestions to the specification based on observation of several GTFS feeds, specifically proper use of stop/route name fields, and gotchas in trip specification. 

It also adds recommendations (Ex: feed_start_date, feed_end_date, stop_timezone) for better
GTFS that are not captured in "required" vs "optional" categories, whilst respecting backward compatibility.


gtfs/spec/en/reference.md
```diff
@@ -95,10 +95,11 @@ File: **Required**
...
 |  stop_code | Optional | The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. Stop codes are often used in phone-based transit information systems or printed on stop signage to make it easier for riders to get a stop schedule or real-time arrival information for a particular stop.  The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. The **stop_code** can be the same as **stop_id** if it is passenger-facing. This field should be left blank for stops without a code presented to passengers. |  |  |
-|  stop_name | **Required** | The **stop_name** field contains the name of a stop, station, or station entrance. Please use a name that people will understand in the local and tourist vernacular. |  |  |
+|  stop_name | **Required** | The **stop_name** field contains the name of a stop, station, or station entrance. Please use a name that people will understand in the local and tourist vernacular, and avoid abbreviations. The stop name should not contain symbols, or the stop_code or stop_id. |  |  |
```
skinkie: Actually remove the 'or stop_id' there is no reason why the stop_id can't be the full stop name (as long as it is unique).

gtfs/spec/en/reference.md
```diff
@@ -95,10 +95,11 @@ File: **Required**
...
 |  stop_desc | Optional | The **stop_desc** field contains a description of a stop. Please provide useful, quality information. Do not simply duplicate the name of the stop. |  |  |
 |  stop_lat | **Required** | The **stop_lat** field contains the latitude of a stop, station, or station entrance. The field value must be a valid WGS 84 latitude. |  |  |
 |  stop_lon | **Required** | The **stop_lon** field contains the longitude of a stop, station, or station entrance. The field value must be a valid WGS 84 longitude value from -180 to 180. |  |  |
+| | | Note: Stops less than 2m away should be matched by a parent/child relationship. Similarly, parent-child stops should not be more than 1 km away. |
```
skinkie: The 1 km is very reasonable. But still some stop areas are bigger than that. Since topology is not possible in GTFS I would prefer another solution than an arbitrary maximum distance.

gtfs/spec/en/reference.md
```diff
@@ -132,8 +133,8 @@ File: **Required**
...
-|  route_long_name | **Required** | The **route_long_name** contains the full name of a route. This name is generally more descriptive than the *route_short_name* and will often include the route's destination or stop. At least one of *route_short_name* or *route_long_name* must be specified, or potentially both if appropriate. If the route does not have a long name, please specify a *route_short_nam*e and use an empty string as the value for this field. |
+|  route_short_name | **Required** | The **route_short_name** contains the short name of a route. This will often be a short (6 characters or less), abstract identifier like "32", "100X", or "Green" that riders use to identify a route, but which doesn't give any indication of what places the route serves. At least one of *route_short_name* or *route_long_name* must be specified. If the route does not have a short name, please specify a *route_long_name* and use an empty string as the value for this field. |
```
skinkie: On Google Maps we have seen that the route_short_name might also indicate the product formula (or modality brand name) of the specific transit mode especially true for trains. This should be fixed up here as well.

gtfs/spec/en/reference.md
```diff
@@ -401,10 +402,13 @@ File: **Optional**
...
 
+**Note**: Transfers between stops more than 2 km apart are discouraged.
+
 |  Field Name | Required | Details |
 |  ------ | ------ | ------ |
 |  from_stop_id | **Required** | The **from_stop_id** field contains a stop ID that identifies a stop or station where a connection between routes begins. Stop IDs are referenced from the [stops.txt](#stopstxt) file. If the stop ID refers to a station that contains multiple stops, this transfer rule applies to all stops in that station. |
 |  to_stop_id | **Required** | The **to_stop_id** field contains a stop ID that identifies a stop or station where a connection between routes ends. Stop IDs are referenced from the [stops.txt](#stopstxt) file. If the stop ID refers to a station that contains multiple stops, this transfer rule applies to all stops in that station. |
+|| | Note: **from_stop_id** and **to_stop_id** should not refer to station-entrance type stops. |
```
skinkie: So theoretically, are they aloud between between parent stations?

gtfs/spec/en/reference.md
```diff
@@ -198,23 +199,23 @@ File: **Required**
...
...day on which the trip schedule begins. If you don't have separate times for arrival and departure at a stop, enter the same value for **arrival_time** and **departure_time**.<br/><br/>Scheduled stops where the vehicle strictly adheres to the specified arrival and departure times are timepoints. For example, if a transit vehicle arrives at a stop before the scheduled departure time, it will hold until the departure time. If this stop is not a timepoint, use either an empty string value for the **arrival_time** field or provide an interpolated time. Further, indicate that interpolated times are provided via the **timepoint** field with a value of zero. If interpolated times are indicated with **timepoint**=0, then time points must be indicated with a value of 1 for the **timepoint** field. Provide arrival times for all stops that are time points.<br/><br/>An arrival time must be specified for the first and the last stop in a trip. Times must be eight digits in HH:MM:SS format (H:MM:SS is also accepted, if the hour begins with 0). Do not pad times with spaces. The following columns list stop times for a trip and the proper way to express those times in the **arrival_time** field: |  |
```
linusnorton: `An arrival time must be specified for the first and the last stop in a trip.`

A) This is redundant as it's a mandatory field
B) How the first stop of a trip have an arrival_time in most cases it will only have a departure_time.

gtfs/spec/en/reference.md
```diff
@@ -198,23 +199,23 @@ File: **Required**
...
 |   |  | Time | departure_time value |
 |   |  | 08:10:00 A.M. | 08:10:00 or 8:10:00 |
 |   |  | 01:05:00 P.M. | 13:05:00 |
 |   |  | 07:40:00 P.M. | 19:40:00 |
 |   |  | 01:55:00 A.M. | 25:55:00 |
-|   |  | **Note:** Trips that span multiple dates will have stop times greater than 24:00:00. For example, if a trip begins at 10:30:00 p.m. and ends at 2:15:00 a.m. on the following day, the stop times would be 22:30:00 and 26:15:00. Entering those stop times as 22:30:00 and 02:15:00 would not produce the desired results. |  |
+|   |  | **Notes:** <br /> 1. Trips that span multiple dates will have stop times greater than 24:00:00. For example, if a trip begins at 10:30:00 p.m. and ends at 2:15:00 a.m. on the following day, the stop times would be 22:30:00 and 26:15:00. Entering those stop times as 22:30:00 and 02:15:00 would not produce the desired results. <br /><br /> 2. Every **departure_time** should monotonically increase for a given trip, and specify times in the associated agency timezone value for consistency across the trip. Thus the reference timezone for times specified in a trip does not change.  |  |
```
linusnorton: `monotonically`

The word is redundant and needlessly complicated.

gtfs/spec/en/reference.md
ca9: Thats fair.

gtfs/spec/en/reference.md
ca9: Happy to reword it to "ideally not be more than 1km away". Does that seem reasonable?

gtfs/spec/en/reference.md
ca9: I'm not sure what you mean by "product formula (or modality brand name)". Could you elaborate please?

gtfs/spec/en/reference.md
ca9: The spec/changes do not explicitly comment on that. Is that a blocker?

gtfs/spec/en/reference.md
ca9: Reasonable points, but I would say this is unrelated to this *specific change*. Yes?

gtfs/spec/en/reference.md
ca9: Thanks for sharing. Happy to update.

gtfs/spec/en/reference.md
linusnorton: True

ca9:
Thank you @linusnorton and @skinkie for valuable feedback 👍 

Do you feel better about current changes?

gtfs/spec/en/reference.md
skinkie: Consider the French TGV, the German ICE, or the more international IC trains. The route_short_name in these cases is TGV, ICE, IC, Sprinter, etc. I think the above text still doesn't make that clear because it is still on a line numbering opposed to group identification.

gtfs/spec/en/reference.md
skinkie: If you want to clear things up and explictly mark that it should not refer to one location type, I would also like to know more about the other location types possible ;)

gtfs/spec/en/reference.md
skinkie: I don't know if my comment yesterday didn't make it, I don't see it here. But for GTFS only the first departure and last arrival (or: duration of the trip) should be known. Hence arrival_time and departure_time are only mandatory for the first and last stop.

ca9: While I do see sense in that but I hope to have that discussion on a separate thread, 
if that's okay? :-) 

The change here about explicitly specifying a reference timezone for all the stop times
across a trip.

gtfs/spec/en/reference.md
ca9: "TGV" is fair use of this field if local riders identify the same route (without 
confusing another) as "TGV". 

Identification of route by number: "32" is certainly common practice in UK, Singapore, etc.
"Green" and "Red" are also often to identify tube lines in a city.

As such, "TGV" is acceptable use given context but not sure if ideal sample use case.

barbeau:
@ca9 I haven't had a chance to review this yet, but have you taken a look at the GTFS Best Practices (http://gtfs.org/best-practices/)?  Is there any overlap between this proposal and the Best Practices?

gtfs/spec/en/reference.md
antrim: What do you think of how the GTFS Best Practices defines short and long route names? http://gtfs.org/best-practices/#routes_2

gtfs/spec/en/reference.md
antrim: I've interpreted that the spec allows transfer rules involving stations:

|  from_stop_id | **Required** | The **from_stop_id** field contains a stop ID that identifies a stop ***or station** [emphasis mine] where a connection between routes begins. Stop IDs are referenced from the [stops.txt](#stopstxt) file. If the stop ID refers to a station that contains multiple stops, this transfer rule applies to all stops in that station. |

gtfs/spec/en/reference.md
antrim: Quoting @linusnorton 

> ```An arrival time must be specified for the first and the last stop in a trip.```
> 
> A) This is redundant as it's a mandatory field
> B) How the first stop of a trip have an arrival_time in most cases it will only have a departure_time.

The Spec has long required the first and last stops to have arrival and departure times. Several consuming apps and validation tools require this. This has always struck me as odd, and it precipitates other issues, like block overlap. See: https://groups.google.com/d/msg/gtfs-changes/u1UuL1VNVSk/W0CdAPN0DgAJ

I'd support a change here but represent just one organization. I think that there are significant issues to consider and hurdles to overcome w.r.t. backwards compatibility.

gtfs/spec/en/reference.md
```diff
@@ -424,6 +428,6 @@ The file contains information about the feed itself, rather than the services th
...
+|  **feed_start_date** | Optional | The feed provides complete and reliable schedule information for service in the period from the beginning of the **feed_start_date** day to the end of the **feed_end_date** day. While *optional* for backwards compatibility, this is a highly recommended field. Both days are given as dates in YYYYMMDD format as for [calendar.txt](#calendartxt), or left empty if unavailable. The **feed_end_date** date must not precede the **feed_start_date** date if both are given. Feed providers are encouraged to give schedule data outside this period to advise of likely future service, but feed consumers should treat it mindful of its non-authoritative status. If **feed_start_date** or **feed_end_date** extend beyond the active calendar dates defined in [calendar.txt](#calendartxt) and [calendar_dates.txt](#calendar_datestxt), the feed is making an explicit assertion that there is no service for dates within the **feed_start_date** or **feed_end_date** range but not included in the active calendar dates. |
+|  **feed_end_date** | Optional | (see above) <br /><br /> While *optional* for backwards compatibility, this is a highly recommended field. |
```
antrim: Reference in GTFS Best Practices: http://gtfs.org/best-practices/#feed_info_4
Long term: We need to set a practice for whether "highly recommended" belongs in the Spec itself. I think there is a strong case for yes.

gtfs/spec/en/reference.md
```diff
@@ -95,10 +95,11 @@ File: **Required**
...
 |  stop_code | Optional | The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. Stop codes are often used in phone-based transit information systems or printed on stop signage to make it easier for riders to get a stop schedule or real-time arrival information for a particular stop.  The **stop_code** field contains short text or a number that uniquely identifies the stop for passengers. The **stop_code** can be the same as **stop_id** if it is passenger-facing. This field should be left blank for stops without a code presented to passengers. |  |  |
-|  stop_name | **Required** | The **stop_name** field contains the name of a stop, station, or station entrance. Please use a name that people will understand in the local and tourist vernacular. |  |  |
+|  stop_name | **Required** | The **stop_name** field contains the name of a stop, station, or station entrance. Please use a name that people will understand in the local and tourist vernacular, and avoid abbreviations. The stop name should not contain symbols, or the stop_id. |  |  |
```
antrim: This is one of the recommendations in GTFS Best Practices: https://gtfs.org/best-practices/#stopstxt

antrim:
@ca9: In the interest of curating the open pull requests, could we close this? Are there particular changes that we should re-open in their own subject-specific pull requests? (Listed below)

* "It is required that the agency_lang field in a single feed is the same for all its agencies."
* **stop_name:** "…avoid abbreviations. The stop name should not contain symbols, or the stop_id." (We addressed this in the GTFS Best Practices.)
* "Stops less than 2m away should be matched by a parent/child relationship. Similarly, parent-child stops should ideally not be more than 1 km away."
* "**Note**: Depending on implementations, this value [stop_timezone] may be completely ignored as the stop_timezone is evident from respective stop coordinates."
* "**[route_long_name]** should not be the same as, nor contain the short name."
* "Transfers between stops more than 2 km apart are discouraged."

skinkie:
> * "It is required that the agency_lang field in a single feed is the same for all its agencies."

I know this is the case for Google Maps, but it is very inconvenient for anything bilingual country. Please do not impose this.

> * "Stops less than 2m away should be matched by a parent/child relationship. Similarly, parent-child stops should ideally not be more than 1 km away."

Care to elaborate a bit more? Do you mean that nearby stops should be grouped with a parent station? Shouldn't *any* stop currently have a parent station to facilitate an entrance?

> * "**Note**: Depending on implementations, this value [stop_timezone] may be completely ignored as the stop_timezone is evident from respective stop coordinates."

If the value is provided, it should be used, not inferred.

antrim:
@skinkie: Note that I am not expressing a position on any of @ca9's proposed changes I listed.

If we discuss any of these changes, let's do so in new topic-specific pull requests.

[Closing for now.]

antrim:
Suggestion: This issue/discussion thread is a candidate to be closed. This is part of a longer discussion about the relationship between GTFS reference and Best Practices.

PR #76, opened 2017-09-29
Add *_icon_url to facilitate icons on maps and itineraries

skinkie:


RachM:
Custom icons would be great for users, but one concern I have is that different consumers will likely have different guidelines that the icons must meet. For example, at Google we'd need **multiple** versions of an icon in varying sizes, formats and potentially foreground colors. To see what I mean, have a look at the 'Maps' icons at https://material.io/icons/.

We'd probably need the url to point to an icon set/package (maybe have them all zipped?), where the icons are named such that consumers know which ones they want. This would have to be documented in the spec.

Also, what happens when a consumer needs a new icon that isn't in the set? How would they go about getting it added?

skinkie:
It might be a good practise to define the requirement for a vector resource, in a specific aspect ratio. But I agree, this is a discussion we must have. Then again, favicons gracefully solve this problem, so we could make it complex or just define the intent by an URI, where the consumer knows that the URI is the intent and local resources may replace it.

barbeau:
I agree that multiple icons would be needed.  I think we'd also need to give very specific examples of how the icons are used.  

@RachM It looks like Google Maps shows some agency-specific icons on the map tiles - see [San Francisco BART example](https://www.google.com/maps/place/San+Francisco+International+Airport/@37.6155256,-122.3921975,17z/data=!4m6!3m5!1s0x808f778c55555555:0xa4f25c571acded3f!4b1!8m2!3d37.6213129!4d-122.3789554) below:

image

Does Google have a standard template icon set that they ask agencies for?  Or is this done more on an ad-hoc basis?

I was thinking if Google can share this template it would be a good starting point, especially if multiple agencies are already sharing icons in this format.

If not, maybe the best way to approach this is to look at the specific use cases in apps (e.g., stop icons on map tiles) and back into a generic format from there, and then see if agencies are able to provide this.

RachM:
Good question Sean - I will find out and report back.

LeoFrachet:
Maybe pointing to vectorial icon, like an `.svg`, could be a solution. On WikiMedia you can either access the `.svg` or the `.png` (when the `.svg` is available).

Example for the bullet of the Paris Subway Line 1:

As `.svg`:
https://upload.wikimedia.org/wikipedia/commons/f/f4/Paris_m_1_jms.svg

As `.png`, width 200px:
https://upload.wikimedia.org/wikipedia/commons/thumb/f/f4/Paris_m_1_jms.svg/200px-Paris_m_1_jms.svg.png

RachM:
Reporting back:
- Customisation of icons is done on a case-by-case basis, so unfortunately I don't have a set of guidelines for you
- We require a few different sizes, so svg would be a good solution
- Transparent background is ideal

stevenmwhite:
Posting the same comment I did over on #269 because we closed that one.

I think generally there's an agency "logo" that represents the agency, but it may not be the icon/symbol that represents their services. One agency may have multiple services with distinct brands (circulators vs commuter routes, or rapid vs local) so perhaps the service icons should be defined at the route level? Both are important, and the route-level icons could be treated as an override to the agency-level one. For many agencies, just the single agency-level logo/icon will suffice.

skinkie:
@stevenmwhite we have a regular discussion if agency.txt is actually used as "brands.txt" since many map service providers only have a single way to distinguish between "the operator" and "the product formula". Further complexity is observed when "the authority" that hires the "the operator" wants to be the brand visible. The solution is not about route level, because routes on their own can have different branding styles which are currently being facilitated by line color et al. In addition even a _direction_ (headsign) may have a distinct icon which GTFS does not support as as normalised concept, rather only as denormalised trip_headsign.

skinkie:
Keep open.

skinkie:
Keep open.

skinkie:
Keep open.

scmcca:
@skinkie Is there something actionable to work on for this proposal (i.e., working through solutions, listing producers/consumers, calling a vote)? I think that would be a more effective way to advocate.

gcamp:
This is generally not something we could use directly because we have specific requirement that don't make sense to be in the specification (mono-color, some specific code to use the route color inside the image, height requirements, etc). 

However, we could use those fields are "good default" that we base ourselves on for the final value that will be used in the app. Would that be enough to be considered as a "consumer"? We use route_color the same way because we modify it most of the time for our context. 

Other comments : 
- Heavy +1 to enforce SVG
- Some guidelines/best practice on the level of details that is required on the image would be good. 
- I think we need some additional field to define if the image is enough to represent the route by itself. Images on routes are often stylized text with the same content as route_short_name, we need a way to know if we should show the route_short_name in addition to the image or not. Same to know if the color is embedded inside the image or it should be shown in addition.

skinkie:
Keep open.

skinkie:
Keep open.

skinkie:
Keep open.

skinkie:
Keep open.

gtfs/spec/en/reference.md
```diff
@@ -188,11 +190,14 @@ Primary key (`route_id`)
...
 |  `continuous_drop_off` | Enum | Optional | Indicates that the rider can alight from the transit vehicle at any point along the vehicle’s travel path as described by `shapes.txt`, on every trip of the route. Valid options are: <br><br>`0` - Continuous stopping drop off. <br>`1` or empty - No continuous stopping drop off. <br>`2` - Must phone agency to arrange continuous stopping drop off. <br>`3` - Must coordinate with driver to arrange continuous stopping drop off. <br><br>Values for `routes.continuous_drop_off` may be overridden by defining values in `stop_times.continuous_drop_off` for specific `stop_time`s along the route. |
+|  `route_icon_url` | Optional | The **route_icon_url** specifies the URL of the icon that allows a rider to identify a route visually, sometimes observed as alternative to the *route_short_name*. The value must be a fully qualified URL that includes **http**:// or **https**://, and any special characters in the URL must be correctly escaped. See http://www.w3.org/Addressing/URL/4_URI_Recommentations.html for a description of how to create fully qualified URL values. |
 
 ### trips.txt
 
 File: **Required**
 
+#### Example showing blocks and service day
```
skinkie, line 199:
```suggestion
```

gtfs/spec/en/reference.md
```diff
@@ -188,11 +190,14 @@ Primary key (`route_id`)
...
+=======
```
skinkie, line 200:
```suggestion
```

Bertware:
The allowed image formats should be defined, and in case that formats other than SVG are allowed, a recommended image size could help as well to prevent big differences in image size/quality between feeds. Agree with @gcamp that SVG would be ideal.

skinkie:
While I generally agree that SVG is the ultimate format for these things. It only works (well) from a minimum dpi. Let's say the 16x16 icon you see on a map is very difficult to achieve with SVG, because it requires pixel alignment prior to rasterization. I am sure we can work out the best way to do this.

felixguendling:
GBFS has this in `system_information.json`. Since this approach seems to work well for GBFS, it could work for GTFS as well?

Field Name | REQUIRED | Type | Defines
---|---|---|---
`license_id` <br/>*(added in v3.0)* | Conditionally REQUIRED | String | REQUIRED if the dataset is provided under a standard license. An identifier for a standard license from the [SPDX License List](https://spdx.org/licenses/). Provide `license_id` rather than `license_url` if the license is included in the SPDX License List. See the GBFS wiki for a [comparison of a subset of standard licenses](https://github.com/MobilityData/gbfs/blob/master/data-licenses.md). If the `license_id` and `license_url` fields are blank or omitted, this indicates that the feed is provided under the [Creative Commons Universal Public Domain Dedication](https://creativecommons.org/publicdomain/zero/1.0/legalcode).
`license_url` | Conditionally REQUIRED <br/>*(as of v3.0)* | URL | REQUIRED if the dataset is provided under a customized license. A fully qualified URL of a page that defines the license terms for the GBFS data for this system. Do not specify a `license_url` if `license_id` is specified. If the `license_id` and `license_url` fields are blank or omitted, this indicates that the feed is provided under the [Creative Commons Universal Public Domain Dedication](https://creativecommons.org/publicdomain/zero/1.0/legalcode). *(as of v3.0)*
`attribution_organization_name` <br/>*(added in v3.0)* | OPTIONAL | Array&lt;Localized String&gt; | If the feed license requires attribution, name of the organization to which attribution should be provided.
`attribution_url` <br/>*(added in v3.0)* | OPTIONAL | URL | URL of the organization to which attribution should be provided.
`brand_assets` <br/>*(added in v2.3)*  | OPTIONAL | Object | Object containing branding information about the system.
`brand_assets.brand_last_modified` <br/>*(added in v2.3)*  | REQUIRED | Date | Date that indicates the last time any included brand assets were updated or modified.
`brand_assets.brand_terms_url` <br/>*(added in v2.3)*   | OPTIONAL |  URL |  A fully qualified URL pointing to the location of a page that defines the license terms of brand icons, colors, or other trademark information.  This field MUST NOT take the place of `license_url` or `license_id`.
`brand_assets.brand_image_url` <br/>*(added in v2.3)*  |  REQUIRED |  URL | A fully qualified URL pointing to the location of a graphic file representing the brand for the service. File MUST be in SVG V1.1 format and MUST be either square or round.


> While I generally agree that SVG is the ultimate format for these things. It only works (well) from a minimum dpi. Let's say the 16x16 icon you see on a map is very difficult to achieve with SVG, because it requires pixel alignment prior to rasterization. I am sure we can work out the best way to do this.

Maybe this could be solved by adding a 16x16 URL field? This could be another PR to keep things simple for now.


Edit: I am re-starting this discussion because we would like to bring icons into Transitous and MOTIS and of course it would be nice to not do a custom solution that works only for us if there could also be a solution that would help the whole ecosystem.

PR #82, opened 2017-10-25
fix gtfs_realtime_version in FeedHeader

JNaftali:
in gtfs-realtime/proto/gtfs-realtime.proto the FeedHeader message was misconfigured. gtfs_realtime_version was set = to 2 (the version number) instead of 1 (the field number)


JNaftali:
submitted cla

barbeau:
Whoops! Thanks for catching that @JNaftali. LGTM.

PR #83, opened 2017-11-02
Proposal for new field: route.route_sort_order

LeoFrachet:
### Issue
Some agencies have specific ways of ordering their routes when displayed to customers. Riders expect this specific ordering of routes when they encounter information about an agency. We cannot infer this order from the ordering of the routes in the dataset, therefore, we need another field in the GTFS to define it.

Currently, at least three GTFS producers already add such values in their GTFS: TriMet, MBTA and Trillium (for the PATH dataset). Also, at least one GTFS consumer already uses it: Transit.

### Proposal
_(Inspired from the MBTA specification)_

`route.route_sort_order` (Optional): The **route_sort_order** field can be used to order the routes in a way which is ideal for presentation to customers. It must be a non-negative integer.

#### NYC MTA
Routes are grouped by so-called "line"
Good order: ACE BDFM ...
Wrong order: ABCDEF...
Source: http://www.mta.info/

#### Berlin U-Bahn
U55 is a disconnected branch of the U5
Good order: U1, U2, U3, U4, U5, U55, U6, U7, U8, U9
Wrong order: U1, U2, U3, U4, U5, U6, U7, U8, U9, U55
Source: http://www.bvg.de/en/Travel-information?#tab-list-0-u

#### Boston MBTA
Mattapan is a distinct route branded as part of the Red Line.
Good order: Green, Orange, Blue, Red, Mattapan, Silver
Wrong order (alphabetical): Blue, Green, Mattapan, Orange, Red, Silver
Source: see MBTA GTFS URL below

(Also, MBTA uses it to force the following ordering: Subway, then BRT, then Bus, then Ferry)

#### Trois-Rivières STTR (Québec)
Every eighty-something route is the night service of the route something.
Good order: 1, 81, 2, 82, 3, 83, 4, 5, 6, 86, ...
Wrong order: 1, 2, 3, 4, 5, 6, 7, 8, ..., 81, 82, 83
Source: http://www.sttr.qc.ca/carte.html

#### Boston MBTA
Specification: http://realtime.mbta.com/Portal/Content/Documents/MBTA_GTFS_Documentation.html
GTFS: https://www.mbta.com/uploadedfiles/MBTA_GTFS.zip

#### Portland TriMet
Specification: https://developer.trimet.org/gtfs_ext.shtml
GTFS: https://developer.trimet.org/schedule/gtfs.zip

#### Trillium in the PATH's GTFS
Specification: not found (@aantrim are you going to provide one?)
GTFS: http://data.trilliumtransit.com/gtfs/path-nj-us/path-nj-us.zip


antrim:
+1

> Trillium in the PATH's GTFS
> Specification: not found (@antrim are you going to provide one?)
> GTFS: http://data.trilliumtransit.com/gtfs/path-nj-us/path-nj-us.zip

@LeoFrachet : We provide this field in all GTFS that we publish, but most of the time values are null. We do use it in a few cases.

Thank you for the prompt to begin providing documentation for our feeds. We are going to begin providing documentation along the lines of [MBTA's](http://realtime.mbta.com/Portal/Content/Documents/MBTA_GTFS_Documentation.html), and will share that when it is ready.

Another application uses the route_sort_order field: Trillium's [Interactive Map](https://trilliumtransit.com/gtfs/interactive-map/).

mike-swiftly:
+1. Swiftly implemented this long ago. Would be great to have more agencies use this proposed field.

DaveBarker:
+1

mgilligan:
+1 (TriMet)

LeoFrachet:
@barbeau: I agree that we have to specify the ordering. I added that portion to my proposal. My bad.

About the right-to-left vs left-to-right, I would tend to think it is overkill to speak about it in the spec. We are defining that such route should be "before" such other route, which is enough IMHO at the spec level.

But maybe it should be in the Best Practices?

abyrd:
+1 in response to @barbeau, I do agree that it should be made clear that the sort is in ascending order (smaller numbers first). I also agree with @LeoFrachet that it is unnecessary to specifically address RTL languages. In an RTL language "before" would clearly mean "to the right of" when elements were inlined into text.

barbeau:
Thinking more, I agree re: specifying RTL being overkill (my suggested text was awful too). At the time I was thinking of the issues that Android has had with RTL, but they are more layout-related and not related to ordering text. As long as we don't use the words "left" or "right" I think it's fine.

LeoFrachet:
@googlebot CLA signed and GitHub Account Name updated.

LeoFrachet:
Let's officially call for the vote.

The vote is open until Thursday, November 23rd 2017, midnight UTC.

antrim:
+1

barbeau:
+1 (for @CUTR-at-USF)

DaveBarker:
+1 (for MBTA, Boston)

invisiblefunnel:
+1 (for @remix)

abyrd:
+1

LeoFrachet:
5 votes in favor.
0 vote against.

@RachM: Could you merge it? 🎉

LeoFrachet:
@RachM :wave: Is there anything else I should do?

ShailShar:
@LeoFrachet Rach is on vacation until 11 Dec. She'll be on it once back.

RachM:
Sorry about the delay - should be merged now.

mmorang:
Let me re-ask my question over here in this old thread where the people who created the route_sort_order field may see it.

Is route_sort_order meant to be used to control map draw order?
Detailed question: https://github.com/google/transit/issues/353

PR #86, opened 2018-01-18
Pathways (and levels) proposal

LeoFrachet:
The goal is to describe the inside of a station, so that a trip planner could process the impact of an elevator being down on the accessibility (for example).

This proposal contains two new files:

*Pathways* describes "ways" within the stations (which can be seen as edges in a graph where the stop/station/entrance/node are the vertices), with a description of their physical characteristics (stairs vs elevator, slope, length) and on their descriptive characteristics (name, code, indication).

*Levels* describes the vertical distribution of a station. In most complex stations, the platforms are layered, without any indication in the current GTFS specification of this repartition.

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -494,7 +494,7 @@ This file allows a better modelling of the pedestrian walkways within a station 
 |  Field Name | Required | Details |
 |  ------ | ------ | ------ |
 |  level_id | **Required** | The **level_id** field contains an ID that identifies a level. The **level_id** is dataset unique. |
-|  level_index | **Required** | The **from_stop_id** field contains a numeric index of the level that indicates relative position of this level in relation to other levels (levels with higher indices are assumed to be located above levels with lower indices). Ground level should have index 0, with levels above ground indicated by positive indices and levels below ground by negative indices. |
+|  level_index | **Required** | The **level_index** field contains a numeric index of the level that indicates relative position of this level in relation to other levels (levels with higher indices are assumed to be located above levels with lower indices). Ground level should have index 0, with levels above ground indicated by positive indices and levels below ground by negative indices. |
 |  level_name | Optional | The **level_name** field contains the name of the level (that matches level lettering/numbering used inside the building or the station). Example: "1", "Subway Level", "Mezzanine", "RDC"... |
 |  level_elevation | Optional | The **level_elevation** field contains the elevation above ground, in meters (with negative values indicating underground locations). |
 
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -22,6 +22,8 @@ This document explains the types of files that comprise a GTFS transit feed and 
     -   [shapes.txt](#shapestxt)
     -   [frequencies.txt](#frequenciestxt)
     -   [transfers.txt](#transferstxt)
+    -   [pathways.txt](#pathwaystxt)
+    -   [levels.txt](#levelstxt)
     -   [feed\_info.txt](#feed_infotxt)
 
 ## Term Definitions
@@ -50,6 +52,8 @@ This specification defines the following files along with their associated conte
 |  [shapes.txt](#shapestxt)  | Optional | Rules for drawing lines on a map to represent a transit organization's routes. |
 |  [frequencies.txt](#frequenciestxt)  | Optional | Headway (time between trips) for routes with variable frequency of service. |
 |  [transfers.txt](#transferstxt)  | Optional | Rules for making connections at transfer points between routes. |
+|  [pathways.txt](#pathwaystxt) | Optional | Ways inside the station, and their description. |
+|  [levels.txt](#levelstxt) | Optional | Levels used in the different stations of the dataset, and their description. |
 |  [feed_info.txt](#feed_infotxt)  | Optional | Additional information about the feed itself, including publisher, version, and expiration information. |
 
 ## File Requirements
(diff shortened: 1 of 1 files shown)


gtfs/spec/en/reference.md
```diff
@@ -413,6 +419,85 @@ Trip planners normally calculate transfer points based on the relative proximity
...
+### levels.txt
+
+File: **Optional**
+
+This file allows a better modelling of the pedestrian walkways within a station by placing its locations on different levels.
+
+|  Field Name | Required | Details |
+|  ------ | ------ | ------ |
+|  level_id | **Required** | The **level_id** field contains an ID that identifies a level. The **level_id** is dataset unique. |
+|  level_index | **Required** | The **from_stop_id** field contains a numeric index of the level that indicates relative position of this level in relation to other levels (levels with higher indices are assumed to be located above levels with lower indices). Ground level should have index 0, with levels above ground indicated by positive indices and levels below ground by negative indices. |
```
jxeeno: I think this is meant to be `level_index` here?

gtfs/spec/en/reference.md
```diff
@@ -413,6 +419,85 @@ Trip planners normally calculate transfer points based on the relative proximity
...
+|   |  | * **(empty)** - Unknown number of stairs. |
+|   |  | * **(non-negative integer)** - Total number of stairs of the pathway. |
+|   |  | Note: This field is allowed only for pathway_mode 2 (stairs) and 4 (escalator). |
+|   |  | Note: The approximation of 1 floor = 15 stairs (or 12 for an escalator) can be used to generate approximative values. |
+|  pathway_name | Optional | The **pathway_name** field contains the name of the pathway, if any. Please use a name that people will understand in the local and tourist vernacular. |
+|  pathway_code | Optional | The **pathway_code** field contains a short text or a number that uniquely identifies the pathway for passengers. The **pathway_code** can be the same as pathway_id if it is passenger-facing. This field should be left blank for pathway without a code presented to passengers. Example: Elevator “A”. |
+|  signposted_as | Optional | The **signposted_as** field helps to generate schematic directions like 'follow signs to <signposted as>' where <signposted as> part represents specific text on the signs that should remain unaffected by translation into other languages. |
```
jxeeno: Did you mean "semantic directions" rather than "schematic directions" here?

gtfs/spec/en/reference.md
LeoFrachet: Fixed. My bad.

gtfs/spec/en/reference.md
LeoFrachet: @dbabramov: This question is for you.

dbabramov: Apologies, I am not a native speaker, so writing well-formed English sentences can be quite a challenge for me.
A native speaker I asked to take a look at this sentence suggested ditching the whole thing in favour of:
"The **signposted_as** field is an exact string of text from physical signage visible to transit riders. The string can be used to provide textual directions to users, such as 'follow signs to <signposted as>'. The language as printed on the signs should not be translated in this field."

LeoFrachet: Updated.

LeoFrachet:
About the beta implementations:

@dbabramov: Can we say that Google has already implemented most of this proposal, and that we have a beta consumer?

@DaveBarker for MBTA, @kurtraschke for NYMTA, @antrim for Trillium or @sdjacobs for CS: Is one of you planning to release a test GTFS with those fields in the close future?

dbabramov:
We have not implemented the full spec yet.
At the moment we only have a very basic implementation that does not yet allow for levels or generic station nodes, with several data providers supporting this basic version.

LeoFrachet:
Quick update, we are currently working with the STM (Montréal, CA-QC) to have a beta implementation of this proposal in theirs GTFS. They already have the station entrances in their GTFS.

gtfs/spec/en/reference.md
```diff
@@ -413,6 +419,85 @@ Trip planners normally calculate transfer points based on the relative proximity
...
+|   |  | * **2** - Connects stops/stations for transfers off the streets or within stations. |
+|   |  | * **3** - Connects stops/stations/entrances for transfers on the streets. |
+|  traversal_time | Optional | The **traversal_time** field specifies the time needed to go through the pathway on foot. Valid values for this field are: |
+|   |  | * **(empty)** - Unknown travel time. |
+|   |  | * **(non-negative integer)** - Number of seconds needed to travers this pathway on foot. |
+|  wheelchair_traversal_time | Optional | The **wheelchair_traversal_time** field specifies the time needed to go through the pathway in a wheelchair. Valid values for this field are: |
+|   |  | * **(empty)** - Unknown travel time. |
+|   |  | * **-1** - Pathway not wheelchair accessible. |
+|   |  | * **(non-negative integer)** - Number of seconds needed to travers this pathway in a wheelchair. |
+|  ramp_slope | Optional | The **ramp_slope** field specifies the slope ratio of the pathway. Valid values for this field are: |
```
barbeau: How exactly is `ramp_slope` calculated?  Is is the steepest slope between stops?

After some discussion with @astein6 who lead a research center on accessible transportation it sounds like this may need to be split into two values - steepest incline and decline, or perhaps positive and negative elevation change.

Here's how Google shows elevation change in walking directions:

image

Do we have any producers providing the slope data yet?  I'm interested to see how this translates to real data.

LeoFrachet: True.

My assumption was that Pathways would only describe paths inside a human-made construction, and that therefore they wouldn't be any complex slope.

But now I'm thinking about those \_/ slopes you find when tunnel go under tracks... Either we modelize them with two distinct ramps, or we need indeed two values.

Do you have an example where this could be needed, knowing that this is just for a pathway and not for a full 40min journey?

barbeau: @LeoFrachet I (unfortunately) don't have an exact real-world example off the top of my head, which is why I was curious to see real producer data for slope values.

Issue #95, opened 2018-08-15
Simplification of the language fields (agency_lang vs feed_lang)


LeoFrachet:
#### Current Issue
Current specification gives two distinct definitions for the two language fields:
- `agency_lang` allows only two-letter language codes, using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) (e.g. "en" for English,  "es" for Spanish or "de" for German);
- `feed_lang` allows the two letter-language codes, but also more complex [language tags](https://www.w3.org/International/articles/language-tags/), using [BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag) (e.g. "en" for English, but also "en-US" for American English or "es-419" for Spanish used in Latin America).

This implies that the high-level definition of the language (with `feed_lang`) can be specific more specific than the low-level one (with `agency_lang`), which is the opposite of what could be expected.

This distinction:
- forces producers to be less specific in the `agency_lang` than in the `feed_lang`, or discourage them to be specific in the `feed_lang`,
- forces consumers to handle two different field formats for validation, without knowing how they should handle a `feed_lang` more specific than an `agency_lang` one (e.g. "en-US" as `feed_lang` but "en" as `agency_lang`). 

##### Option 1: Unify the two definitions by allowing both fields to use language tags.
- This will reduce the complexity of the specification without reducing it's precision.
- This keeps all old datasets valid, since the two-letter language codes is are included in BCP 47.
- This will only require producers to simplify their code, by allowing `agency_lang` to be like `feed_lang`.

##### Option 2: Deprecated the `agency_lang` field once-for-all.
- This was the original intention when the field `feed_lang` has been added (see [Joe Hughes 2010 message on the subject](https://groups.google.com/d/msg/gtfs-changes/muxms00iBos/TrCb9YYteCMJ))
- This would reduce even more the complexity of the specification.
- Both fields are optional so we wouldn't be loosing required information.
- BUT, this wouldn't allow anymore to specify distinct languages for distinct agencies in the dataset. It it used? 
  - Official dataset for Switzerland (which has four official languages) arbitrarily set the `feed_lang` as "DE" and then lists all agencies `agency_lang` at "DE" aka speaking German (see [GTFS](https://opentransportdata.swiss/de/dataset/timetable-2018-gtfs/resource/432eaef2-8e65-4eed-b8c9-8ffe55c60581)).
[...]

PR #98, opened 2018-08-17
Unify `agency_lang` and `feed_lang` definitions

LeoFrachet:
Hi GTFS community,

This pull request is the result of the short discussion we had on [Issue 95](https://github.com/google/transit/issues/95). No need to read the whole issue discussion, here is the sum up:

#### Current Issue
Current specification gives two distinct definitions for the two language fields:
- `agency_lang` allows only two-letter language codes, using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) (e.g. "en" for English,  "es" for Spanish or "de" for German);
- `feed_lang` allows the two letter-language codes, but also more complex [language tags](https://www.w3.org/International/articles/language-tags/), using [BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag) (e.g. "en" for English, but also "en-US" for American English or "es-419" for Spanish used in Latin America).

This implies that the high-level definition of the language (with `feed_lang`) can be specific more specific than the low-level one (with `agency_lang`), which is the opposite of what could be expected.

This distinction:
- forces producers to be less specific in the `agency_lang` than in the `feed_lang`, or discourage them to be specific in the `feed_lang`,
- forces consumers to handle two different field formats for validation, without knowing how they should handle a `feed_lang` more specific than an `agency_lang` one (e.g. "en-US" as `feed_lang` but "en" as `agency_lang`). 

#### Proposal: Unify the two definitions by allowing both fields to use language tags.
- This will reduce the complexity of the specification without reducing it's precision.
- This keeps all old datasets valid, since the two-letter language codes are included in BCP 47.
- This will only require producers to simplify their code, by allowing `agency_lang` to be like `feed_lang`.
- The only non-backward compatibility part of this proposal is that `agency_lang` was loose on the capitalization (language code should always be lower case according to ISO 639-1, aka "en", but `agency_lang` was oddly also allowing upper case, aka "EN"). I suggest we do not officialy handle this case in the spec anymore. GTFS consumers will still have to allow "EN" in short term anyway, and it's a good think IMHO to suggest to agency to follow the ISO/BCP standard in the future.

Thoughts?


barbeau:
+1

skinkie:
+1 of the proposal
+2 for the effort

gcamp:
+1

abyrd:
+1

LeoFrachet:
This pull request has been open one week ago. As per the Official Process, we can now open the vote on it.

@barbeau, @skinkie, @gcamp, @abyrd & @mgilligan: Please re-vote (since all +1 given before the official voting period are not counted).

Vote will be closed on Friday August 31st at 23:59:59 UTC.

ritesh-warade-ibigroup:
+1

gcamp:
+1

mgilligan:
+1

barbeau:
+1

LeoFrachet:
Vote is closed.

In favor:
- IBI (by @ritesh-warade-ibigroup)
- Transit (by @gcamp)
- TriMet (by @mgilligan)
- CUTR at USF (by @barbeau)
- Conveyal (by @abyrd)

Against:
[nobody]

So we have:
5 votes in favor
No vote against
=> The proposal is therefore adopted.

skinkie:
With the recent changes, is gtfs.org also updated automatically?

LeoFrachet:
@antrim: I remember we discussed it but I don't remember the conclusion. Is it pulling from GitHub automatically? Or should we update it by hand?

PR #103, opened 2018-09-13
Field types

LeoFrachet:
Hi GTFS community!

We are starting to have some heavy duplication in the specification. The definition of an URL is given five times (`stop_url`, `route_url`, `feed_publisher_url`, `agency_url`, `agency_fare_url`), the definition of a timezone is given twice, just like the definition of a language code or a service day or a color.

This pull request offers to factorize those definitions of types on the top of the specification, to avoid duplication and make it more concise. It also makes it easier for consumers who wants to parse a GTFS and who wants to see which data is in which field.

Last but not least, it highlights the specific definition of "date" and "time" in a GTFS dataset, which isn't the usual definition of date and time.

You can read the pull request using the "Files changed" tab, or you can read the result formatted markdown, which might be easier:

https://github.com/MobilityData/transit/blob/field_types_2/gtfs/spec/en/reference.md#field-types

Thoughts?

Side note: I would have loved to tighten some definitions, e.g. to force the phone numbers to be in the E123 format, but I didn't do it because I wanted to keep this proposal as being just a factorization of definitions and not a change in the specification per se.


LeoFrachet:
Since I didn't kept Evan work down the road, there is no need for him to sign a CLA. I'm moving that to another branch. Sorry.

Issue #109, opened 2018-10-18
GTFS-rt change process - define experimental consensus, limited time window


barbeau:
In https://github.com/google/transit/pull/105 we just updated the GTFS-rt change process docs to match the currently followed process as outlined by Brian Ferris in https://groups.google.com/forum/#!msg/gtfs-realtime/bWQAgGdj3f0/mdcKwR3J8xAJ.

@gcamp suggested two changes to the process, which we agreed to table until after the above PR was adopted - so now we can discuss in this issue.

First comment thread is at https://github.com/google/transit/pull/105#discussion_r222000795, although for some reason @gcamps comment doesn't appear there anymore (?).  It was a suggestion to put a limited time window on experimental fields, and after than window expires they would be adopted (or potentially discarded).  For example, `OccupancyStatus` has been experimental [since 2015](https://groups.google.com/forum/#!msg/gtfs-realtime/_HtNTGp5LxM/23vN1VIHKfkJ).

Second comment at https://github.com/google/transit/pull/105#pullrequestreview-160776233:

>Should there be a definition of consensus? The amendment process is way more formal but it seems to me that if an experimental fields is approved, it becomes hard to change after the fact due to protobuf backward compatibility issues. My proposition would be same process as the spec amendment process but with a "strong majority vote (80%)" instead of unanimous.

@gcamp any further thoughts or proposals on the above?

gcamp:
@barbeau I removed it because I wanted to clean up the timeline of your PR but that didn't work and couldn't revert. That's why it's not showing up. 

I would make the time window for experimental field of 1 year. Hopefully most field should take less time than this. Nothing else significant to add. Strong majority vote still make sense to me.

gcamp:
@barbeau I would agree in principle but in practice we've had big problems integrating producers into the discussion on Github. I've had that with 2 producers who wanted changes on the spec but didn't follow on Github nor put energy on it. (
https://github.com/google/transit/pull/87 and 
https://github.com/google/transit/pull/111)

Maybe that's a symptom of a bigger problem with the producers we are working with, so I would agree.

barbeau:
@gcamp Ok, thanks.  One other question here - after the expiration windows passes, and assuming an experimental field doesn't get adopted, how exactly does it get deprecated?

Looks like our options are, in increasing order of severity:
1. `EXPERIMENTAL_FIELD_THAT_NOBODY_WANTED = 4 [deprecated=true];`
[...]

PR #111, opened 2018-10-19
Add prediction certainty

gcamp:
We have seen that `uncertainty` values in TripUpdates are sometime used to convey the information if a prediction is precise enough to be shown as real time in consumer applications. Different producers have different values where `uncertainty` has different meanings. 

This causes problems because consumers has to have different interpretation for different producers. It would be better to have a clear value for each prediction that defines if a prediction is precise enough to be shown as real time or not. 

This is a proposition to add `prediction_type` TripUpdates as an _experimental field_. Predictions can either be 

- `REALTIME` (current default)
- `IMPRECISE_REALTIME` where the vehicle is followed in real time but doesn't have enough precision to be shown as real time to the user. The prediction threshold for the switch between `REALTIME` and and `IMPRECISE_REALTIME` should be the same threshold that the producer uses to hide the real time icon in other interfaces like bus displays. 
- `UPDATED_SCHEDULE` where a vehicle is not yet followed in real time but some changes to the schedule was done after the latest GTFS export, very likely due to changes from the control center. 

I would also be curious to know if anybody uses the `uncertainty` value in GTFS-rt without having some magic threshold value. If that's not the case and usage of `prediction_type` is accepted and successful, I would propose to deprecate `uncertainty`

Note : I might be unresponsive in the next few weeks. @juanborre has my full confidence and can respond in my stead and represent Transit.


mike-swiftly:
I certainly think it is a good idea to add a prediction_type attribute to the GTFS-realtime spec. This will allow systems, like Swiftly's Transitime system, to clearly indicate whether predictions in the user interfaces should be displayed with a real-time icon or not. 

I think there are two issues that should first be dealt with, clarifying the possible values and determining if any other related additions are needed.

With respect respect to clarifying the possible values, I'm not clear on UPDATED_SCHEDULE. My current opinion is that there should simply be a value SCHEDULED to indicate that there is currently no location information for the vehicle for that trip and that the predictions are just based on the schedule. Could be that the standard schedule is used or that an improved scheduled based on historical travel times is used. Either way, I think it would be good to simply indicate that the predictions are schedule based. For such predictions a real-time icon should of course not be shown in the user interface.

With respect to other possible additions, I think another key one is to indicate if a prediction is based on when the driver is supposed to leave a terminal or a wait stop. Predictions for after such a stop are directly affected by driver behavior, when the driver actually leaves the terminal. Such predictions are inherently not as accurate. Some user interfaces could indicate such. A possible way to deal with this is to have an additional possible value of REALTIME_DRIVER_AFFECTED.

One last thing, I suggest that "IMPRECISE_REALTIME" be changed to "REALTIME_IMPRECISE" so that it is obvious that the main goal of this attribute is to indicate whether or not the predictions are real time or schedule based.

gcamp:
@mike-swiftly 

The reason I put UPDATED_SCHEDULE instead of simply SCHEDULE was because the role of simply showing scheduled is usually reserved for the static GTFS. UPDATED_SCHEDULE was to make clear it wasn't a regular scheduled case. 

Re REALTIME_DRIVER_AFFECTED, I'm not strongly against it but it for a consumer application it's not something that's useful. What's useful is to make sure that we display the same information in all interfaces. For example, if Swifly shows departure dependant on driver behavior as real time in the SMS interface, the departure should be tagged as REALTIME. If Swifly doesn't tag them as real time then it should be REALTIME_IMPRECISE. 

Agreed on the REALTIME_IMPRECISE change, will do the modifications.

gcamp:
@abyrd 

Sometimes real time prediction can have low accuracy so users shouldn't rely on it the same way they do for other real time predictions, but they are still more precise than the actual GTFS schedule data, so it's worth sending. 

Re-How to use `UPDATED_SCHEDULE` for consumer application. At Transit we plan to show them initially as regular schedule information, but we do want to eventually change that and show that's planned information but changed from the original schedule. 

In a few words, we want producers to send the latest information that they have, even if it's not up to regular real time standards.

gcamp:
@mike-swiftly @abyrd Any responses on the follow ups earlier? I would like to call for a vote soon. If you think the answer were satisfactory let me know :)

gtfs-realtime/spec/en/reference.md
```diff
@@ -177,6 +177,21 @@ The update is linked to a specific stop either through stop_sequence or stop_id,
...
 | **departure** | [StopTimeEvent](#message-stoptimeevent) | Conditionally required | One | If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED.  If schedule_relationship is NO_DATA, arrival and departure must be empty. |
 | **schedule_relationship** | [ScheduleRelationship](#enum-schedulerelationship) | Optional | One | The default relationship is SCHEDULED. |
+| **prediction_type** | [PredictionType](#enum-predictiontype) | Optional | One | The default relationship is REALTIME. |
+
+## _enum_ PredictionType
+
+Experimental field, subject to change.
+PredictionType represent the source of data of the prediction. This lets consumer adjust their display of the information depending on the source (for example, with a real time symbol).
```
barbeau: `PredictionType` doesn't really seem like a source of data based on the enumerated values.  It's more like a qualifier of the type of data.

barbeau:
>I'm not sure though why we would need a flag for imprecise data. Considering this is primarily a consumer-facing format, if the real-time information is not precise enough to be used by passengers, shouldn't the producer just not send it?

I tend to agree here. @gcamp Can you give an example of how a consumer would use data of type `REALTIME_IMPRECISE`?

skinkie:
You want to mention that the vehicle is driving (hence it is not cancelled) but are not aware of the whereabouts. Opposed to "there is no realtime status".

barbeau:
@gcamp I agree that `uncertainty` doesn't seem very useful as it exists today, and I agree that if producers can differentiate between the two levels of predictions we should give them the ability to communicate that info.

However, to me it seems like there are two very different uses of `prediction_type` here:
1. Communicate whether a stop_time_update should be shown in the consumer's UI to riders (`REALTIME` and `REALTIME_IMPRECISE`)
1. Communicate an updated schedule from GTFS (`UPDATED_SCHEDULE`)

It seems unnatural to me to bundle these two very different concepts into the same field.

For the first two values (`REALTIME` and `REALTIME_IMPRECISE`), I'd suggest directly addressing the end goal in the name of the field and values.  I think a relatively vague name like `REALTIME_IMPRECISE` could suffer from the same varying interpretations as `uncertainty`.  IMHO having a boolean flag that clearly indicates whether or not a prediction should be shown to a rider would be preferred, such as `show_to_user`, if that's the primary goal of the first two field values.

I'd be interested to hear from producers to know whether they can actually differentiate between the two quality levels of predictions internally.

Given that `UPDATED_SCHEDULE` has a completely different purpose from the first two values, what are your thoughts on removing `UPDATED_SCHEDULE` from this proposal and instead using the GTFS-ServiceChanges spec as mentioned by @LeoFrachet in https://github.com/google/transit/issues/113?

One motivation for using a separate channel for schedule updates is that `UPDATED_SCHEDULE` as proposed here isn't backwards-compatible from a consumer's perspective.  If producers start publishing updated schedules as stop_time_updates, and a consumer isn't aware of the new `UPDATED_SCHEDULE` enum, they will by default show this info as real-time to riders.

(FWIW, `REALTIME_IMPRECISE` really isn't backwards compatible either in the same way, although the effects may be hidden from consumers if agencies just label the existing predictions they are already pushing out in `stop_time_updates` as  `REALTIME_IMPRECISE`, as opposed to publishing new `stop_time_updates` they aren't currently producing).

mike-swiftly:
I think that there is lack of agreement on how to move forward because we are trying to put too much information into a single field. I think a key goal should be to be able to clearly specify a prediction is due to GPS from a vehicle or whether it is something else, like schedule based. 

A secondary goal could be to specify more information about the source and the quality of the prediction. 

Therefore I think should have a boolean parameter called something like `gps_based`. Default would be true. Easy to understand and to implement. We can also have a second parameter called `prediction_type` that could specify any additional information.

I think splitting things into these two parameters would allow us to move ahead pretty quickly with `gps_based` and achieve a lot of what we are looking to accomplish.

skinkie:
gps_based is something that would not work for me, especially since most vehicles use multiple sources like odometers etc.

mike-swiftly:
Stefan make a good point about the name "gps_based". Other examples of real-time position systems include train control systems for when vehicles are underground. So a different name would be an improvement. Perhaps "real_time_information" or "real_time_location" or ???

barbeau:
@mike-swiftly What are your thoughts in terms of naming the field after the action producers want the consumer to take (e.g., `show_to_user` or `hide_from_user`, or something better...) instead of qualifying the type of data?

gcamp:
@barbeau you would want to show departure that are marked as `REALTIME_IMPRECISE`, even if they aren't as precise as other departures. It's imprecise, but still better than schedule from the GTFS. `show_to_user` or similar doesn't achieve the same goal here. 

The reason I would push for `UPDATED_SCHEDULE` is that updating schedule information is often done by the same system that pushes real time trip updates. It feels a bit wasteful not to use the spec we already have to update the schedule, since it's very very close to what it already achieves. 

That said, I'm not against having a single flag that denote if the departure is precise enough to be shown a real time symbol if that would fit everyone and move quicker.

abyrd:
> On 31 Oct 2018, at 02:54, Stefan de Konink <notifications@github.com> wrote:
> You want to mention that the vehicle is driving (hence it is not cancelled) but are not aware of the whereabouts. Opposed to "there is no realtime status”.
> 
Isn’t the default that scheduled service will be carried out as planned, unless the real-time message cancels or modifies it? Meaning, if you are receiving a GTFS-RT feed for an operator, and there is no message saying the trip is canceled or supplying times, doesn’t this mean the operator will make the trip but we don’t have location and delay information? This situation doesn’t seem like something that could be described as “precise” or “imprecise”.

abyrd:
Overall I agree with @barbeau's analysis, though I'd encourage people to also consider ServiceChanges #113 from the opposite point of view. As I mentioned in a comment on that ticket #113, moving schedule updates to a completely new data spec (ServiceChanges) requires all GTFS consuming software to be extended to handle that new information, otherwise it will silently produce an incorrect view of the system by applying GTFS-RT directly on top of the static GTFS without first applying the ServiceChanges. The alternative of legacy/simpler systems interpreting GTFS-RT modified schedules as real-time predictions is comparatively harmless because it still results in the end user seeing the service the operator plans to provide them.

abyrd:
> On 23 Oct 2018, at 01:47, Guillaume Campagna <notifications@github.com> wrote:
> @abyrd <https://github.com/abyrd> Sometimes real time prediction can have low accuracy so users shouldn't rely on it the same way they do for other real time predictions, but they are still more precise than the actual GTFS schedule data, so it's worth sending.
> 

I still don’t understand this. Can you give an example of how you would use data that are declared by the producer to be too poor for riders to see? How would you know if the data were useable at all? Can you give an example of a source that produces such poor data that still has some purpose? 

What is the specific threshold between those two levels of precision, and what is the threshold at which the data become useless? How does the operator know where these thresholds lie and when they have crossed them?

barbeau:
>Can you give an example of how you would use data that are declared by the producer to be too poor for riders to see?

Same question here - if the end goal isn't to show or hide certain info from the user (which is the proposed threshold the producer uses to pick the enum), then the expected consumer behavior for these values isn't clear to me at all.

gcamp:
> I can imagine the situation that an operator wants to communicate that the bus left the terminus and is serving the scheduled route, but doesn’t have any realtime position/delay information. It seems increasingly unlikely that an operator would have a digital representation of that one bit of information yet not have a position or any predictions, but if there’s a field that can communicate this situation why not.

From the uses-cases that we have heard : 
- Weak cellular data spots, where buses lose connections but are known to be actively running. (Apparently this happening a lot in Gatineau)
- Broken hardware where a bus doesn't have GPS but still reports in real time 
- Prediction that have are dependant on driver behaviour. i.e. driver has a 5 minutes pause but might take a 8 minutes pause. 

However, that would be a better question for producers. 

> Can you give an example of how you would use data that are declared by the producer to be too poor for riders to see? @abyrd @barbeau 

The point is to hide the real time symbol for some predictions, not hide the predictions themselves.

In most consumer application there's a real time symbol that indicates that a prediction is based on real time information. When that's shown, users have different behaviour depending if the information shown is real time or scheduled. For example, a user might go a the stop 1 minute or less if the prediction is real time, but will allow 3-5 minutes if the information is has schedule-based. 

In the case of a low-accuracy prediction, we still want to show the prediction because it's more precise than the schedule information. However, we don't want user to expect the same level of precision, so we want to hide the real time symbol. `prediction_type` achieves that. 

> What is the specific threshold between those two levels of precision, and what is the threshold at which the data become useless? How does the operator know where these thresholds lie and when they have crossed them? @abyrd 

This PR specifically doesn't specify the specific threshold because it should be the same the producer users in the own display of information. Producers will need to make a similar decision of a threshold in their implementation, the point here is that the threshold is the same on all interfaces. 

From this PR : 
> The prediction threshold for the switch between REALTIME and and IMPRECISE_REALTIME should be the same threshold that the producer uses to hide the real time icon in other interfaces like bus displays.

Specific example : MTA Baltimore uses an SMS interface powered by Swifly. 
textimage

Here the (s) marks scheduled departure. However, the departures returned by the SMS and marked as scheduled are *not* always the same as found in the original static GTFS. In consumer applications, departures should be marked the same way. The goal is to have consistency in the interfaces, even if we don't know how imprecise the predictions actually are. 

I still think `UPDATED_SCHEDULE` is useful, but there seems to be a consensus that Service-changes should be the way forward. I'll update this PR to reflect that.

barbeau:
Generally I'm on-board with a way to better qualify prediction accuracy to riders - we've had the same discussions in context of OneBusAway.

And, narrowing down to two buckets for producers to label prediction accuracy (vs. the `uncertainty` integer that likely isn't very precise) may help with this.  Consumers could then pass on this qualifier to riders.

However, I strongly disagree that these `IMPRECISE` predictions should be marked as scheduled arrivals/departure information to the end riders.  This is bound to create confusion for anyone referencing the actual agency schedule, whether printed or surfaced in another app based on the GTFS data.

Could we maybe agree on some additional generic language in the PR for expected consumer behavior in the area of differentiating predictions to end users that isn't prescriptive, but also doesn't allow them to be displayed as schedule data?

juanborre:
There seems to be consensus so far that we need to decouple the real time precision concept from the updated schedule.
I updated the PR to reflect that change in the approach and simplify the discussion.
There is now just a single flag RealTimePrecision taking values `PRECISE` / `IMPRECISE`.
Please, share your comments and thoughts.

@barbeau The _Imprecise_ real time information can be shown to the user with a different real-time symbol or with no symbol. It can depend on each consumer.

gcamp:
@googlebot please check for CLA again

gcamp:
@barbeau I'm ok with @juanborre 's commit

abyrd:
> On 1 Nov 2018, at 11:50, Guillaume Campagna <notifications@github.com> wrote:
> I still think UPDATED_SCHEDULE is useful, but there seems to be a consensus that Service-changes should be the way forward. I'll update this PR to reflect that.
> 
I am not sure there is a true consensus. Among the group commenting on these tickets there is indeed general interest in the idea of service changes. But if the audience were extended to include the large number of people maintaining GTFS consumer software, who will need to revise and increase the complexity of their systems at their own effort and cost just to maintain their accurate view of the transit network, I suspect there would be significant skepticism. I would also like to know what proportion of those people would simply not implement the new layers of patching, breaking the unity of GTFS spec interpretation.

gcamp:
@abyrd Agree, but I'd like to make this PR move relatively fast, and not waiting for that decision is one way to do it. @juanborre 's changes are compatible with the eventual UPDATED_SCHEDULE, if we decide to go that way.

abyrd:
@juanborre further to @barbeau's comments about how consumers are allowed to visualize `IMPRECISE` data:
After the last commits, the text still says in several places that `IMPRECISE` means "not precise enough to show the real time indicator". This is a particular (in some people's opinion, unacceptable) strategy for visually rendering imprecise information. I would recommend removing this sentence entirely. I would also remove "(for example with a real-time symbol)".

About the overall proposal:

The proposal is now much clearer after discussion. What was confusing before were the statements that "imprecise" data would not be shown to end users. In reality it seems we're talking about exactly the opposite - making it possible to show these poor estimates to the end user with a warning that their quality is degraded.

The core idea is that the existing quantitative variable (`uncertainty` field) should be replaced with a qualitative one with a limited number of discrete values. It is more realistic for producers to choose between only two values, stating that the real-time data is either dependable and can be taken at face value, or is known to be degraded and should be taken with a grain of salt.

Reduced to this statement, this proposal seems completely sane to me. 

However I still take issue with implying in the spec that the imprecise data flag should be used to "hide the realtime symbol". This seems like an attempt to engineer rider behavior by telling them a low-quality real-time update is a scheduled arrival. This strikes me as a kind of behavioralist perspective where the value of information is reduced to the influence it has in training people to take certain actions. I hope consumers will instead decide to give the riders more agency in interpreting and acting on this new information. I don't think this should be presented as the normal way to use this new field.

I think there is also some ambiguity in the posted examples about what "scheduled" means. I don't think the riders would agree that something is a "scheduled departure" if the time was changed on the same day by the operator, yet some of the examples seem to show this situation. We seem to be mixing up the idea of poor real-time data (due to broken GPS, tunnel, etc.) with dynamically planned departures, and confusing both of these with pre-scheduled trips.

My preference would be to accompany the new field with guidance strongly suggesting that consumers have a distinct visual representation for each of three situations (schedule with no realtime update / precise realtime update / imprecise realtime update).

juanborre:
Thanks for your comments, @abyrd 
I updated the description accordingly to allow the consumer to display the real time information differently depending on the precision and existence of RT.

abyrd:
Thanks @juanborre, the proposal looks very reasonable to me in its current state.

mike-swiftly:
I'm having a bit of trouble understanding the how the current proposal meets ours goals. I agree that PRECISE vs IMPRECISE is a clear and useful concept and I fully support it. But I don't see how those values sufficiently relate to whether a real-time icon should be displayed in the client, which was the original goal of this PR. I see a need for two separate attributes so that we can clearly deal with all situations. We could have real_time_precision  to specify precision and real_time_location (or other name) to specify whether real time location information is available. 

Perhaps the best way to explain my thinking is to look at the possible 4 states if we used two separate variables to indicate the state.

real_time_location = REAL_TIME (show real-time icon)
real_time_precision  = PRECISE 
Normal situation. Have a vehicle assigned to the trip, getting GPS location information for vehicle, and vehicle likely to behave as expected.

real_time_location = REAL_TIME (show real-time icon)
real_time_precision  = IMPRECISE 
Have a vehicle assigned to the trip, getting GPS location information for vehicle, at the layover at the beginning of the trip, but not completely certain about when vehicle will start trip. Perhaps there is no precisely defined start time or the drivers are know to not start as scheduled.

real_time_location = INFERRED (do not show real-time icon)
real_time_precision  = PRECISE 
There are some situations where GPS info is not available until a vehicle actually starts on the trip. This means that the location info is only inferred. But the trips are usually fulfilled and start on time, and therefore could be considered to be precise.

real_time_location = INFERRED (do not show real-time icon)
real_time_precision  = IMPRECISE 
No GPS info available and not completely certain that vehicle will actually run the service or start on time.

juanborre:
Hi @mike-swiftly, do you have any thoughts about the last comment? We would like to call the vote for this unless there are remarks.

abyrd:
@mike-swiftly my take on your question is that any data that are not from a pre-published schedule are realtime, and should be presented as such. No additional information is needed in the RT feed for this, all data displayed from an RT feed are real-time.

Of course some people will disagree with this and no one is going to actively prevent end-user apps from using their own approach for rendering or presenting the information available. But I don't think we should add more fields or values to the spec for such special use cases.

The point about wanting to warn users when the data are of lower quality than usual is certainly valid, but it seems like all we need is a single on/off field for that.

I'm getting the impression from some of the discussion that people are displaying the real-time predictions without displaying the underlying schedules. Some confusion could probably be eliminated by showing schedules and real-time separately. For a train that is 10 minutes late, I am accustomed to real-time information displayed like:

`Intercity to Brussels 12:30 (+10 minutes)`

or alternatively in a list of stations

|station|Scheduled | estimated|
|-------|-----------|-----------|
|a|12:30 | 12:40|
|b|12:35 | 12:44|
|c|12:40 | 12:45|

anatolip-rt:
> The bad news is that it appears that one or more commits were authored or co-authored by someone other than the pull request submitter. We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that here in the pull request.

@juanborre can you confirm and I will update the cla label.

juanborre:
Hi @anatolip-rt , I'm OK with my commits being contributed to this project.

abyrd:
@mike-swiftly thanks a lot for the clarification. The confusion is mostly about terminology then: when you said something is "realtime" or not, you meant information collected continuously from vehicles as they drive. When I said "realtime" I just meant any kind of updates sent while the vehicles drive, independent of their source.

Indeed predictions based on past observations are different than reporting information coming from vehicles as they drive. I now understand your suggestion with two variables (precision and source). I just didn't understand the meaning of "inferred" in that proposal, because all RT messages are inferred using some model - what changes is the nature and quality of inputs to that model. I would suggest changing the term "inferred" and precisely defining the new term.

Are the different kinds of inputs as follows?
{Schedule data, historical data, intermittent realtime, continuous realtime}

And then the output of the model can be subjectively categorized by the producer as precise or imprecise in another variable.

The question is just whether consumers or end users care or should even know about the inputs to your realtime prediction model, or should just know whether you consider it accurate or not.

gcamp:
This pull request didn't have discussion for more than a week and everybody seems to agree on the type of the field and the usefulness. I'd suggest we consider we merge this pull request since I feel the consensus was reached as required by the current Official Process. 

@barbeau if you or anybody else feels more secure to do a vote similar to what I suggested for experimental fields (https://github.com/google/transit/issues/109) I'm ok with this too.

barbeau:
@gcamp I was waiting for you to call a vote on this, when I then realize that we don't have a process yet :).

I just opened https://github.com/google/transit/pull/126 to define the experimental consensus process (I think we can vote on that independently of the limited time window of experimental fields) - are you ok if we vote/merge that, and then use that process for this issue?

gcamp:
@barbeau maybe call a vote here and work on the other PR at the same time? Waiting for #126 means this PR won't be approved for at least 3 weeks :/

mike-swiftly:
We have had discussions but I don't see the issue as fully resolved yet. I
think it would be good to first do a pre-vote summary to see if in advance
whether people agree to the proposal.

Key thing is that I still question 1) the naming of the element(s); and 2)
whether we need a second element or at least a third possible value.

With respect to the naming of the element, I don't see the issue as
*precision*. The issue is to be able to signify whether or not the
real-time icon is to be displayed. I still think it would better to have a
element that clearly indicates whether or not the info is real-time based,
instead of whether it is precise.

I'll provide a specific example of the problem with the currently proposed
terminology. The predictive system might indicate that a vehicle should
depart at precisely 10:00am based on the schedule even though no GPS data
has been yet received for the corresponding vehicle. The prediction is
precise, most likely accurate, but it might also be wrong if there is no
vehicle to fulfill that trip. I don't think specifying that the prediction
is precise is adequate for this situation because the real-time icon should
not be displayed.

This all brings me back to suggest 3 possible states (I originally
suggested 4, but Juan showed that 3 should be sufficient):

   - Real time that I can trust
   - Real time that I cannot trust that much
   - No real time

The name of the element could be something besides RealTimePrecision since
precision isn't the key issue.

But to move things forward  perhaps we could make just minimal changes to
the proposal before voting on it. We could keep the element name as
RealTimePrecision  but simply change the possible values to
REAL_TIME_PRECISE, REAL_TIME_IMPRECISE, and NOT_REAL_TIME.

What do others think with respect to modifying the proposal with this
suggestion?

mike-swiftly:
@Barbeau, Sean <barbeau@cutr.usf.edu> , I wasn't actually assuming the
departure was based on historical data. It might be based on historical
data or a time in the GTFS. Either way the prediction might be *precise*,
but not based on real-time GPS data, and therefore possibly wrong and
should be indicated as such in the user interface.

abyrd:
> @barbeau, Sean <barbeau@cutr.usf.edu> , I wasn't actually assuming the departure was based on historical data. It might be based on historical data or a time in the GTFS. Either way the prediction might be *precise*, but not based on real-time GPS data, and therefore possibly wrong and should be indicated as such in the user interface.

Can you clarify what you mean here? If you're referencing the strict definition of the word "precise" then yes, all predictions have the same precision (1 second) so the term is not quite right. Perhaps accuracy, reliability, or (as @barbeau suggested) (un)certainty is more appropriate.

But even if we avoid the term "precision" that still leaves us with two outcomes: reliable/unreliable or certain/uncertain.

abyrd:
@barbeau said:
> I agree, but I may rephrase this to:
> > * Predictions that I can trust
> > * Predictions that I cannot trust that much
> > * No predictions (i.e., schedule = GTFS)
> 
> Then, we could go with:
> `enum PredictionCertainty` with `PREDICTION_CERTAIN` and `PREDICTION_UNCERTAIN`
> With this enum we're trying to create 2 bins to effectively replace the integer `uncertainty` field, and these names seem like a better fit than "precision".

This makes perfect sense to me - the information received by the rider is either a) from the schedule (not from GTFS-RT) or b1) from GTFS-RT but is reliable or b2) from GTFS-RT but unreliable. So within the GTFS-RT there are two enum values (essentially a boolean).

As I understand it there is interest from @mike-swiftly in an additional field (or additional enum value(s)) to say that GTFS-RT data are based on historical performance rather than currently observed location. But most or all other commenters seem to think this difference can be subsumed into the binary accurate/inaccurate or reliable/unreliable or certain/uncertain dichotomy.

gcamp:
Agree with @barbeau and @abyrd. In general I think there's use to change the schedule and mark them as such, modified schedule. However I think this should be resolved in 
https://github.com/google/transit/issues/113 instead of here.

barbeau:
@gcamp What are your thoughts on changing the name to `enum PredictionCertainty` with `PREDICTION_CERTAIN` and `PREDICTION_UNCERTAIN`?

gcamp:
I prefer current precise/imprecise, but certain/uncertain also works. It's really not that different in my opinion.

abyrd:
Well, if we've reached the point of debating the name of the field I guess that means there's converging opinion on the core idea :)  

Re-reading this, I find it odd to say a prediction is "certain", which means there is no chance it will be untrue. Ideally the word chosen should imply a continuum of values segmented into two regions, neither of which is ever truly certain. Precision has a technical meaning so some people are likely to find it confusing - arguably all times in GTFS/RT have the same precision.

Nonetheless "uncertainty" seems like the most appropriate word for the amount of error in a prediction (as in the original field that is being deprecated) even though its antonym "certain" is strange here.

How about a field uncertainty with values UNCERTAINTY_HIGH and UNCERTAINTY_LOW?

barbeau:
>How about a field called uncertainty with values UNCERTAINTY_HIGH and UNCERTAINTY_LOW?

Makes sense to me.  If we're providing a enum version of the integer `uncertainty` field, and the general intentions for the data are the same for the enum and integer, then I'd suggest keeping the same naming convention.

barbeau:
@abyrd Also, just to clarify - then the default would be `UNCERTAINTY_LOW`?

skinkie:
@barbeau Are we sure we want to have a negation **un**certainty in the standard, opposed to certainty_high / certainty_low?

mike-swiftly:
I strongly agree with Stefan that it would be better to not have a double
negative for the name. certainty_high / certainty_low are easier to
understand.

*Michael Smith*
*CIO, **Swiftly, Inc*
(415) 260-4700

barbeau:
@skinkie Good point.  So `certainty` with values `CERTAINTY_HIGH` and `CERTAINTY_LOW`, with `CERTAINTY_HIGH` being the default?

gcamp:
One counterpoint, with that naming, the option of having `SCHEDULED` goes away. See comment [here](https://github.com/google/transit/pull/111#issuecomment-435181453) and related discussion.

barbeau:
>One counterpoint, with that naming, the option of having SCHEDULED goes away. See comment here and related discussion.

@gcamp Correct, this is intentional, with the general thought being that schedule data shouldn't be conflated with prediction uncertainty.

abyrd:
> I strongly agree with Stefan that it would be better to not have a double negative for the name. certainty_high / certainty_low are easier to understand. 

I agree that the double negatives are awkward, which is why I also gave a positive version (CERTAINTY_HIGH). "Uncertainty" was present in the original spec and my above suggestions because "uncertainty" is often used when describing measurements, without it being a conscious negation of "certainty". Nonetheless, it does look strange when used with enum values.

A very practical argument is that there's already an integer "uncertainty" field which is going to be marked deprecated, so we need a new name for the enum field, hence "certainty".

>>One counterpoint, with that naming, the option of having SCHEDULED goes away. See comment here and related discussion.
>@gcamp Correct, this is intentional, with the general thought being that schedule data shouldn't be conflated with prediction uncertainty.

If we're intentionally excluding other dimensions from this enum (which I agree with), then it's worth asking if it should be an enum or a boolean. It seems to me that GTFS already uses enums for other two-valued fields so this matches GTFS conventions.

barbeau:
>If we're intentionally excluding other dimensions from this enum (which I agree with), then it's worth asking if it should be an enum or a boolean. It seems to me that GTFS already uses enums for other two-valued fields so this matches GTFS conventions.

Another argument for enum is that it allows expansion in the future if producers can provide more precision for uncertainty (e.g., `CERTAINTY_MEDIUM`).

Thinking more about the default - we should have another enum `CERTAINTY_UNKNOWN` as the default. This would match the current definition of int `uncertainty`:

>If uncertainty is omitted, it is interpreted as unknown. To specify a completely certain prediction, set its uncertainty to 0.

This would give us `certainty` with values:
* CERTAINTY_HIGH
* CERTAINTY_LOW
* CERTAINTY_UNKNOWN (default)

A question is if we also want an equivalent to `uncertainty = 0` for a "completely certain prediction" - elsewhere in the spec it references trains under computer control as a type of data that can be "completely certain".  Off the top of my head I'm personally not aware of any systems publishing `uncertainty = 0`.

abyrd:
Good points @barbeau. A default of UNKNOWN makes sense and matches the original system. I'm not a big fan of allowing for more discrete levels of uncertainty because once you get beyond reliable vs unreliable, it becomes very unclear what "medium" means and I think we'd be better off back with a quantitative field.

I also think we should exclude certainty=0. As an everyday user of trains under computer control on completely grade-separated tracks, I can tell you all it takes is someone getting their umbrella or backpack stuck in the door to make the prediction incorrect :) I think the real world only allows for CERTAINTY_HIGH, anything higher is wishful thinking.

gcamp:
Not sure how `UNKNOWN` will be semantically different than `CERTAINTY_HIGH`. For consumer, it will end up as having those two values to be essentially the same. 

My understanding was that a `uncertainty=unknown` prediction would basically currently be the same as `CERTAINTY_HIGH` would mean now. No idea how precise the prediction actually is, but we know it's precise enough to show in an UI as real time.  

Even worst, `uncertainty=unknown` could mean different things depending how you interpret it. Having the default to `CERTAINTY_HIGH` is clear and unambiguous.

barbeau:
`CERTAINTY_HIGH` means you have a high confidence in the accuracy of the prediction.  `UNKNOWN` means that you have no idea how accurate the prediction is.  This is the current default for all predictions in GTFS-rt feeds, unless you specify an `uncertainty` integer value.

We could assign numbers to LOW/HIGH values, like 68th and 95th confidence levels, but my feeling is that these values wouldn't mean much in practice.  But conceptually that's what they represent - buckets used to express confidence in a prediction.

gcamp:
@anatolip-rt can you approve again? I only merged with master, seems like this shouldn't change the CLA status

gcamp:
Updated the proposition with the `certainty` value as discussed previously. Sorry it took this long. 

Let me know if you have any comments.

gcamp:
Since there was no more comment, I'm calling a vote for this. Thanks for the comments @mike-swiftly @abyrd @barbeau! Since this is an experimental field, 80% of the votes need to be in favour, and the rest of the process is the same. 

The vote will close on Jan 17 at 23:59:59 UTC.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -221,6 +221,33 @@ message TripUpdate {
...
+    // This allows the consumer to decide how to display the real time information
+    // to the user depending on the precision.
+    // If HIGH and LOW certainty information are both shown to riders,
+    // consumers should represent these types of information differently in the
+    // user interface. LOW certainty values should not be represented in the same
+    // way as schedule data to the rider.
+    // RealTimeCertainty is ignored if schedule_relationship = SKIPPED or NO_DATA
+    enum RealTimeCertainty {
+      // Precision of the real time is unknown, default behavior
+      UNKOWN = 0;
```
barbeau: `UNKOWN` should be `UNKNOWN`

gtfs-realtime/spec/en/reference.md
```diff
@@ -177,6 +177,21 @@ The update is linked to a specific stop either through stop_sequence or stop_id,
...
+
+Experimental field, subject to change
+RealTimeCertainty represents the certainty of the real time information provided. This allows the consumer to decide how to display the real time information to the user depending on the precision. If HIGH and LOW certainty information are both shown to riders, consumers should represent these types of information differently in the user interface. LOW certainty values should not be represented in the same way as schedule data to the rider.
+RealTimeCertainty is ignored if schedule_relationship = SKIPPED or NO_DATA
+
+#### Values
+
+| _**Value**_ | _**Comment**_ |
+|-------------|---------------|
+| **UNKOWN** | Precision of the real time is unknown, default behavior |
```
barbeau: `UNKOWN` should be `UNKNOWN`

barbeau:
+1, although there are two typos that need to be fixed prior to merge (I've commented on them)

juanborre:
I'm OK with my commits being contributed to this project.

mike-swiftly:
I have to vote -1 on the current pull request because I think it does not address the goal of clarifying when the real-time icon should be displayed in a client application. It does address uncertainty, but it still assumes that predictions should always be based on GPS. The RealTimeCertainty value of LOW is described as:

> Update is based on information from a vehicle that is or was followed in real time but is now low certainty. Lower certainty can come from things like a bus losing connection after the start of the run or when the prediction is dependant on driver behaviour.

I think the above definition of LOW is good for handling those specific situations. But there are other situations that we were trying to handle, such as when there is no GPS for a vehicle because no vehicle has been assigned to particular block/route. For schedule based predictions the certainty would be LOW, and no real-time icon should be displayed as part of the associated predictions. But there are other situations where the certainty would be LOW, such as the prediction being real-time GPS based but also dependent on driver behavior.  For the driver behavior situation would still want to display the real-time icon since the predictions are based on real-time GPS info. This means that a certainty of LOW is not adequate to specify whether or not the real-time icon should be displayed.

But perhaps there is a very simple solution. I propose we add another possible value called something like NOT_REAL_TIME. Before I was proposing a separate field, but I'm trying to keep things simple here so that we might be able to move forward. If the certainty is HIGH or LOW then the real-time icon should be displayed. If it is NOT_REAL_TIME then the icon should not be displayed. If the value is UNKNOWN then most likely the icon should be displayed.

gcamp:
@mike-swiftly NOT_REAL_TIME would go against the rule that and the ongoing discussion about if trip updates should update schedule information or not. See https://github.com/google/transit/issues/113 

It was on purpose removed from this PR to let that discussion continue on that second PR. Yes, the current changes are not as feature rich as the original proposition, but at least we can get the ball rolling. Future updates can add the capability of editing schedule from the trip updates.

gcamp:
@mike-swiftly The goal of `LOW` was specifically to now show the real time indicator. I'm not sure I get what the difference between `LOW` and `NOT_REAL_TIME` is adding if it ends up being the same behaviour. 

>>>  But there are other situations where the certainty would be LOW, such as the prediction being real-time GPS based but also dependent on driver behavior.

The point of certainty was exactly to know if it's certain enough to show the prediction icon, not if the prediction is based on GPS or now. If an app should show the real time icon it should be `HIGH`, if it's not then it should be `LOW`. 

For changes to the schedule, that is not real time at all, then further discussion should be in https://github.com/google/transit/issues/113 since it was removed from this PR.

abyrd:
@mike-swiftly said:
> I have to vote -1 on the current pull request because I think it does not address the goal of clarifying when the real-time icon should be displayed in a client application. 

This reaction is understandable, considering the title of this ticket. I believe some commentators' understanding of the goal of this ticket has shifted considerably since that title was written. Maybe this fact hasn't been addressed clearly enough, but much of the discussion has shifted to communicating how accurate a prediction is, sidestepping the original goal of controlling a real-time icon in an end-user app. 

To me, the details of when a particular icon is displayed, and even the definition of "real-time" being used in the comments are specific to particular consumers and may not be relevant to a general-purpose firehose data specification like GTFS-RT. We should be careful not to assume that symbols and features in particular end-user apps are shared by all other consumers of GTFS-RT.

>It does address uncertainty, but it still assumes that predictions should always be based on GPS. 

Nothing here assumes any predictions are "always based on GPS". As already stated above in response to your proposed `gtfs_based` field, many systems for tracking vehicles do not use GPS. Many systems use different forms of continuous location tracking (odometers, GPS position matched to route or street topology) or discontinuous location cues (rail signaling systems, vehicle operator actions). They may integrate these with historical data on speed and variability.

By "based on GPS", do you instead mean "not solely based on schedules"? Do you consider predictions sent every few minutes as GTFS-RT messages, but based entirely or largely on schedules and historical data to be "not real-time"? 

> For schedule based predictions the certainty would be LOW, and no real-time icon should be displayed as part of the associated predictions. 

I think this is an example of a consumer-specific concept of "real time". Rather than discussing how  producers can directly control this particular app-specific UI element, it may be more productive to talk about which pieces of raw information you need from the feed to control this UI element.

@mike-swiftly, as I understand it you want information about the data sources used to produce the prediction, including whether GPS location or schedules were considered in the prediction. Is this correct?

> But perhaps there is a very simple solution. I propose we add another possible value called something like NOT_REAL_TIME. 

It is not clear to me what it means for a real-time message to be "not real-time". We will need a full definition to discuss clearly. But to me, this description of the prediction seems mostly orthogonal to the proposed accuracy enum.

Ideally we can include all the necessary information in GTFS-RT messages to allow you to control your app following your definitions and design. But I would prefer to include general-purpose pieces of information in the GTFS-RT messages, rather than items which allow controlling UI elements of one particular app according to its own definitions.

gcamp:
Vote is now close. Votes are +2, -1 and thus the proposition doesn't pass. 

@mike-swiftly can you detail your thinking a bit after the responses?

gcamp:
@mike-swiftly Can you please answer, this is blocking this current PR. I'll start an other vote next week if there's no response.

mike-swiftly:
My one single issue, as I have tried to previously state, is that the current proposal does not address the issue at hand for real-time information systems to indicate if predictions are based on real-time location information or not. I am completely in agreement with also specifying uncertainty values. But the proposed values are not adequate to also indicate whether predictions are based on real-time location info. 

Side note: we are at a disadvantage of not receiving feedback from a significant number of other producers or consumers of GTFS-realtime data. Hey Google, CityMapper, Moovit, NextBus, Clever Devices, Init, etc. are you out there???

Changing the title of the pull request does not change our initial goal. I think it only tries to circumvent it. So let's figure out how to move this forward in a concrete way.

But first, a bit of explanation to the questions that @abyrd asked. I see real-time information as being based on real-time vehicle location information. Of course there are also additional possible information sources, such as historical data. But the key thing is whether the system actually knows that there is a vehicle in service. The real-time location is usually GPS based, though as @abyrd pointed out, it can of course be from other sources, such as a train control system. 

Now let me try to define my previously used term NOT_REAL_TIME (if people have better suggestions for terms, that would be great. I'm just interested in the meaning). A prediction indicated to be NOT_REAL_TIME would mean that the system has not received location information for a vehicle that will serve the corresponding trip or block. The prediction information therefore is based on the expectation that a vehicle will be put into service. For some agencies this can be a good assumption, for others, a bad one. It depends on many trips an agency simply misses. I can tell you with certainty that some agencies like SFMTA miss quite a few. We don't really know the uncertainty. But we do know that it is not as certain as if there was real-time location information, and that can be useful information for the passenger.

I expect that most clients would want to show a real-time icon only if real-time location information is available. Otherwise I think it would be very misleading to the passengers. Of course clients might also want to somehow convey other forms of uncertainty as well. But I don't think that a radio wave icon is understood to indicate certainty. Of course clients can design any kind of app they want to. It is their app. But information providers, like myself, should provide the necessary information so that the app designers can do what they think is right. 

Therefore if we either add NOT_REAL_TIME as a possible value to the proposed uncertainty field, or if we add a second field true/false to indicate that whether or not real-time location information was used for the prediction I think we would cover all situations and accomplish our goal. Can we do so?

gcamp:
@mike-swiftly here's how I see it and the different cases. 

- Vehicle is running and is followed in real time : high certainty
- Vehicle is running but is not followed in real time : low certainty
- Vehicle has GPS at the start of run but is dependent on driver behaviour : low certainty
- No location information and best information we have is the schedule information : no prediction should be given, consumer will display schedule information.
- No location information but we have improved schedule information : not possible with this PR, should be done in an other PR with discussions from https://github.com/google/transit/issues/113 taken into account. It wasn't possible before this PR, it won't be possible after but it's not a regression.

I agree that the PR reduced in scope compared with the start, but it doesn't mean what we have now is not valid.

mike-swiftly:
Dividing predictions into simply two groups, high and low certainty, is simply not adequate when it comes to describing whether predictions are real-time based. When real-time location information is not available there are some situations where there is a somewhat but not completely high certainty (there isn't yet a live vehicle for the assignment yet, but perhaps this is common and there is a good change that a vehicle will be assigned and leave on time) and situations where there is a low certainty (who knows what is going on??). 

At least users can understand that a real-time icon indicates there is a tracked vehicle and a lack of the icon indicates there isn't one. 

@gcamp , what is your objection to making the change more descriptive by separating out the indication of whether the predictions are real-time or not?

barbeau:
@gcamp can we change to `PredictionCertainty` instead of `RealTimeCertainty`?  I think this is more accurate, as predictions don't need to be based on real-time data.

>When real-time location information is not available there are some situations where there is a somewhat but not completely high certainty (there isn't yet a live vehicle for the assignment yet, but perhaps this is common and there is a good change that a vehicle will be assigned and leave on time) and situations where there is a low certainty (who knows what is going on??).

So is this `MEDIUM` uncertainty? :)

>what is your objection to making the change more descriptive by separating out the indication of whether the predictions are real-time or not?

I'd prefer to see a more explicit indicator of when a vehicle isn't assigned to a trip, instead of abstracting this to a level of including "not real-time" information in the GTFS-realtime feed.

In theory vehicle assignment to a trip can already be represented in TripUpdate.VehicleDescriptor (and VehiclePosition.TripDescriptor), but the problem is that the consumer can't tell if the field was left blank because the producer didn't have this information or if the trip is known to be unassigned.

A related concept here is is ["TripUpdate.delay"](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-tripupdate), which is (emphasis mine) "the current *schedule deviation* for the trip. Delay should only be specified when the prediction is given relative to some existing schedule in GTFS." 
I think the intention here was to make this `delay` an objective data point based on an observed past vehicle location at a stop or timepoint - in other words, we know this bus passed this stop at noon - as opposed to stop_time_updates which are predictive and forward-looking into the future.  However, this intention isn't clearly defined in the spec and I don't think it's being used that way in practice for those that have adopted it, although it's still an experimental field that's open to change/clarification.  `TripUpdate.delay` relates to the vehicle assignment question as a TripUpdate with an assigned vehicle should have a schedule deviation, while trips without an assigned vehicle wouldn't have a schedule deviation.  But again, the issue of having an explicit flag stating that the schedule deviation isn't known is needed.

`VehiclePosition` has [`VehicleStopStatus`](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-vehiclestopstatus), which includes `INCOMING_AT`, `STOPPED_AT`, and `IN_TRANSIT_TO`.  Adding another enum to this and sticking it in TripUpdate seems like stretch to me.

So, we would need another enumeration to indicate vehicle assignment in TripUpdates.

We could add `VehicleAssignmentStatus`, with `UNKNOWN`, `ASSIGNED`, and `UNASSIGNED`.  Would it be useful to have additional enum values like `SERVING` too, indicating that not only has a vehicle been assigned to the trip, but it's actively serving the trip?

gcamp:
> @gcamp can we change to PredictionCertainty instead of RealTimeCertainty? I think this is more accurate, as predictions don't need to be based on real-time data.

@barbeau agree, changed that

> @gcamp , what is your objection to making the change more descriptive by separating out the indication of whether the predictions are real-time or not?

The proposition of `NOT_REALTIME` is to me the same as changing the schedule. At this point the difference between updated schedule information or low certainty real time information is fairly thin except for naming. 

And for adding the update of schedule as a feature, I wanted https://github.com/google/transit/issues/113 to be resolved before changing anything. For what it's worth, I agree that updated schedule should be done in the trip update format, but that's for the community to decide. 

@mike-swiftly would you agree to this current PR if service change would be able to update the schedule? 

Mobility Data is leading the service change proposition, maybe @LeoFrachet @barbeau have something to add into the current discussion regarding service changes.

gcamp:
Bump 😬

LeoFrachet:
[Disclaimer: I usually re-read the whole message thread before posting, but I didn’t because this thread is way too long. Guillaume (@gcamp), since you proposed it at the first place, could you update your *first* message with a sum up of the current discussion, scope, proposals? That would help new comers (and me) daring to answer without being off topic.]

**My understanding is that we need a replicable way to measure the « precision »** (or « uncertainty ») of a prediction, so that an uncertainty of `x` in NY MTA would have the same meaning than an uncertainty of `x` in PATH… and that an app can decide to flag the value as « real-time » when the uncertainty is less than `y`.

From the conversations I had with other stakeholders around GTFS-ServiceChanges and GTFS-Flex, I realized **we need two values: the prediction time, but also the spreading**. It can be a range: 10AM +/- 2 minutes (with the definition that in 99% of the time, the bus will pass between 9:58AM and 10:02). It would allow us to compare the two following prediction: 10AM +/- 2min vs 10AM +/- 15min, allowing the data consumer to decide when they want to flag the value as « real-time ».

Two remarks on that:

**First, time ranges are not symmetrical in public transit**, since passing early is the worst. It should never happened. I would be interested to see statistics of real passing time for a train scheduled at 10:00 for example. I’m expecting no train early, most trains on time, then a decreasing probability of being late. So instead of symmetrical timeframe, we should provide just the likely maximum delay. For example: 10AM +max 12min, defined as « the train will most likely be arrived at 10:00, but there is a 99% probability that at 10:12 it will be arrived. We could call such value « prediction_spread » for example. The unit could be the minute or the second, in both cases it would be replicable and well defined.

**Secondly, how can we calculate it?** Good question. But, like, how can we calculate a prediction? There is a naive but good way to start: just use historical data, and see retrospectively what those values should have been. 24 hours before, your prediction_spread is usually 15min, then 1h earlier it’s usually 10 min, and 10 min earlier it’s only 1min. Good. Just use that. It’s enough to be able to compare them between agencies. And yes, sometime it will be wrong, just like the predictions are sometime wrong. But at the end of the month, we’ll be able to see how accurate your `prediction_spread` were. And then, smart developers will work on improve those values, just like they work on improving the prediction time.

abyrd:
@LeoFrachet I agree with your comments on the fact that departure time distributions are skewed later than the scheduled departure time, and that it's complicated to calculate and define them. Even a value like "9:00 am + up to 5 minutes" is not really clear, because there's always some tail of the distribution out past 5 minutes, so what percentile is being reported?

This PR is basically a response to the fact that no one is providing this kind of detailed precision information, and we'd be better off with a discrete, qualitative description of accuracy: either it is accurate enough to trust or it's not.

This subject was found to be more complex because there's a second dimension that some people would like to fold into the same field: the source of the data (whether it is based on recently reported position or guessed from other sources). The conversation got long because some people just want a high-level judgement on whether the prediction is "good" or not, and others want to specifically know whether recently reported position data was used to calculate the prediction because they have a UI element that indicates this specific bit of information. The proposal was blocked by @mike-swiftly who it seems would really like a way to receive information about the data sources used to compute the prediction.

abyrd:
@gcamp said:
> @mike-swiftly would you agree to this current PR if service change would be able to update the schedule?

I think it's inadvisable to ask people to agree to a proposal based on the assurance that another (as yet unapproved) proposal is going to cover their use case. This has the potential to create a larger problem further down the line, a situation where the other proposal "must" be accepted because people have predicated past votes on its acceptance.

I know some organizations and people are counting on the service-changes proposal, but as I've commented elsewhere I have serious doubts that we will see widespread adoption of multi-layer GTFS patching by many consumer applications. This is asking every consumer in the world to carry out significant additional software development to understand a few feeds. This is unlikely to happen uniformly, and would basically fork the specification since any consumers that don't support service-changes could badly (and probably silently) misinterpret those feeds.

LeoFrachet:
Andrew (@abyrd) said:

>Even a value like "9:00 am + up to 5 minutes" is not really clear, because there's always some tail of the distribution out past 5 minutes, so what percentile is being reported?

I did define which percentile we should use. Defining the percentile is the key the replicability of my proposal. The definition I gave was "There is a likelihood of 99% that the vehicle will be passed at `PredictedTime + prediction_spread`". So the percentile I was offering to use is 99% (It can be 95%, it doesn't matter, what matters is that everybody uses the same).

> we'd be better off with a discrete, qualitative description of accuracy

Yup to you. I have no strong opinion. But I do think that without strict definition, everybody will use different definition and the "LOW" will be depending of the agency, and the "LOW" of one will be the "HIGH" of some other. But if producers and consumers see a value, I'm not against it.

abyrd:
@LeoFrachet said:
> I did define which percentile we should use. Defining the percentile is the key the replicability of my proposal.

Sorry, I know that was in your proposal - I should have been more clear. I was partly replying to your post and partly musing about the problems with the existing GTFS uncertainty concept. A while ago there was a discussion elsewhere about the fact that you need to specify a distribution and the parameters to that distribution. We could freeze some of those parameters but it's all quite arbitrary and I do wonder how much it would really be used, and how correctly.

I'm guessing these broader "certain/uncertain" categories would be more heavily and correctly used.

gcamp:
> > @mike-swiftly would you agree to this current PR if service change would be able to update the schedule?

> I think it's inadvisable to ask people to agree to a proposal based on the assurance that another (as yet unapproved) proposal is going to cover their use case. This has the potential to create a larger problem further down the line, a situation where the other proposal "must" be accepted because people have predicated past votes on its acceptance.

@abyrd That's probably right, but we could still eventually add schedule information to the enum proposed whatever happens with the service change. My point was more that we could still improve the GTFS-rt spec right now instead of getting stuck on adding more to this proposal. 

@LeoFrachet Sorry for getting you into the discussion, my point was more if there was any advancement on service changes that could direct the discussion on if we should add a `NOT_REAL_TIME` or not. I think @abyrd resumed well the discussion, I'll try to update the first post and re-call a vote.

abyrd:
@LeoFrachet said:
> I would be interested to see statistics of real passing time for a train scheduled at 10:00 for example. I’m expecting no train early, most trains on time, then a decreasing probability of being late.

I just did a quick informal literature review. As we'd expect, the distributions are skewed later than the scheduled arrival time, but up to 30-40% of buses arrive well before the scheduled time. Of course this is horrible for riders, but it's a fact visible in the data. And of course it depends heavily on mode of transport and location.

If we accept that this variability is just a fact of life in bus operations, rider experience can be greatly improved by showing them both the expectation of arrival time and a safe time to reach the stop. For example, showing an expected arrival time of 12:30 but also saying you must arrive at 12:25 to have a 95% chance of catching this bus, and have a 95% chance of being on the bus by 12:35. Providing the whole distribution to RT consumers (based on historical data and/or current conditions) would allow this. It would also allow providing a much more nuanced time window for arrival times and transfers downstream. Perhaps a sort of empirical CDF with N evenly spaced percentiles. But again I'm not sure who will produce such data (correctly). We might be able to make it happen in the Netherlands.

But of course this would be a separate proposal: instead of the current ill-defined uncertainty field in GTFS-RT, we would have one very simple enum field (as described in this PR) and a separate very detailed arrival probability field.

barbeau:
>Providing the whole distribution to RT consumers (based on historical data and/or current conditions) would allow this.

If provided, I think it need to be clear whether the agency is reporting a distribution based on historical data, or one created for the prediction given to the rider (which could be a mix of current and historical conditions).  The first is relative trivial to implement for producers if you have archived data, but less meaningful to riders if the current trip is an outlier - and these are the trips riders really want accurate information about.  The 2nd can be much harder to do accurately, at least from my experience dealing with location accuracy on mobile devices.  We found widespread inaccuracies in the "accuracy" value provided to apps by GPS hardware, and that was after producers had pushed to have the threshold be the 68% confidence level - it turns out they couldn't even meet that accuracy level.  I'm not sure if this experience directly translates to arrival/departure predictions yet, but so far it feels similar to me.  The other challenge is how the CDF is converted to something that riders understand - although that's outside the scope of the GTFS-rt exchange.  This is a very similar problem to [travel time reliability](https://ops.fhwa.dot.gov/publications/tt_reliability/ttr_report.htm) on the highway side of things, which I understand has been fairly heavily studied.

One last set of facts to drive home that the current `uncertainty` value isn't working.  We did a [quick analysis](https://github.com/MobilityData/transit/issues/18) of feeds from TransitFeeds.com, and found the following:

* 6 agencies use the existing TripUpdate uncertainty value (same for all stop_time_updates)
* 3 agencies are publishing uncertainty=0 for all predictions 😱

Feed | uncertainty value
-- | --
OCTA Trip Updates | 0
Thunder Bay Transit Trip Updates | 0
HART Trip Updates | 30
BART Trip Updates | 30
Kingston Transit Trip Updates | 0
MBTA Trip Updates | 300

lenaschimmel:
We are currently developing a sophisticated delay prediction for the [Dystonse](https://github.com/dystonse/dystonse) project that will generate arbitrary detailed probability distributions for delays based on historical data and real time information. _(Our repository is documented in English, but much of the extra information about our project is available only in German at this time.)_

The main purpose of those predictions will be in our own routing algorithm that takes the full distribution into account. We are looking into alternative ways to make our predictions available via APIs and/or feeds, so that other routing services or real time information displays can use them.

The current spec of the `uncertainty` field is not optimal for the task, so we've been looking for alternate proposals. This discussion is the best match we found so far.

As @LeoFrachet and @abyrd [discussed over a year ago](https://github.com/google/transit/pull/111#issuecomment-463217648), defining specific percentiles is mandatory to have data that is comparable across data providers.

Given our current data and algorithms, we could return a value for any percentile in theory, but _for most routes_, we don't have enough historical records to be really confident about the 1% or 99% percentile, but 5% and 95% should be fine. Even if detailed data was _available_ for any given percentile, different use cases might benefit from using different percentiles. So from our viewpoint, an ideal proposal would allow to attach any number of delay/percentile pairs to a `stop_time_update`. On the other hand, we realize that hardly anybody used the current `uncentainty` field, so maybe that proposal would be overly complex for everybody but us. Any thoughts on that?

Also, if you know of any other place where similar ideas are discussed, please leave a hint.

---

Here's a sample of our data, which coincidentally has very little skewing, although I know other examples that do have the heavy skewing that was mentioned before:

Bildschirmfoto 2020-06-08 um 20 04 10

skinkie:
On Monday, June 8, 2020 8:09:39 PM CEST, Lena Schimmel wrote:
> Given our current data and algorithms, we could return a value 
> for any percentile in theory, but _for most routes_, we don't 
> have enough historical records to be really confident about the 
> 1% or 99% percentile, but 5% and 95% should be fine. Even if 
> detailed data was _available_ for any given percentile, 
> different use cases might benefit from using different 
> percentiles. So from our viewpoint, an ideal proposal would 
> allow to attach any number of delay/percentile pairs to a 
> `stop_time_update`. On the other hand, we realize that hardly 
> anybody used the current `uncentainty` field, so maybe that 
> proposal would be overly complex for everybody but us. Any 
> thoughts on that?

I think it should not be forgotten that many transit agencies are using a 
first order prediction to model their time table. Hence it is not a 
summation of travel times, but the chance that given a departure time the 
arrival at a stop can be met in 75-85 percentile of the cases. I think this 
method has been described by the transit group of TU Delft in the early 
90ies. So I wonder, shouldn't the schedule not first be reversed engineered 
back, and then remodelled based on a historic estimation given the input 
schedule that *does* effect how drivers execute it?

If we could assist you regarding data. For The Netherlands we have a multi 
year historic real time dataset available for research purposes.

-- 
Stefan

barbeau:
>If we could assist you regarding data. For The Netherlands we have a multi
year historic real time dataset available for research purposes.

@lenaschimmel Same here - we've been collecting GTFS-RT data from 8-9 agencies over the last two years for related work going on at USF, with the goal of comparing multiple prediction techniques across different agencies. If you're interested in collaborating please contact me via LinkedIn - https://www.linkedin.com/in/seanbarbeau/.

barbeau:
>So I wonder, shouldn't the schedule not first be reversed engineered
back, and then remodelled based on a historic estimation given the input
schedule that *does* effect how drivers execute it?

@skinkie I think you're referring to something like [busbuzzard](https://github.com/bmander/busbuzzard)?

skinkie:
> @skinkie I think you're referring to something like [busbuzzard](https://github.com/bmander/busbuzzard)?

Not that specific code, but it looks fun. I have asked some people if they could come up with the original paper.

gcamp:
Unless there are big disagreement, I'm going to close this PR. Schedule change should probably be done in a potential [Service Change spec](http://bit.ly/gtfs-service-changes-v3_1). `uncertainty` is still not very well defined. If anyone wants to clarify it, I would be open to discuss further.

PR #120, opened 2018-11-20
Editorial and formatting changes for clarity

antrim:
**Motivation:** Inconsistencies in the GTFS make it more difficult for newcomers to understand the specification and build software. Our motivation is to make GTFS more clear and consistent.

**Summary:** This PR makes the following editorial and formatting changes to the GTFS specification. Except for one possible exception (discussed below), this pull request merely changes the language of the GTFS reference, but does not change any of its meanings.

**List of changes:**
* Standardized the choice of certain words
  * passenger/customer --> rider (Except in the case of “customer service”)
  * Blank --> empty
  * Feed vs dataset
  * Field vs field value
  * Transit Organization --> transit agency
   * Note: organization is still used for the group that publishes the feed
  * Itinerary --> "Rider Journey" [note: modified Dec 13]
* Made editorial changes to enhance clarity and legibility, including removing unnecessary and repetitive language. Major rewrites of:
  * `parent_station`
  * `stop_timezone`
  * `arrival_time` and `departure_time`
  * `Monday`,...,`Sunday`
  * `timepoint`
  * `exact_times`
* Formatting changes
  * Added code ticks to denote fields and other code values
  * Standardized Enum option formatting
  * Added hyperlinks to all .txt files and websites
  * Put field types in alphabetical order.
  * Different type setting for examples
* Defined or redefined the following terms:
  * Dataset
  * Field Value
  * Record
  * ID
  * Added referencing ID

**Background** discussion is in the original (@MobilityData) draft pull request: https://github.com/MobilityData/transit/pull/8

### DIFF
--- a/gtfs/CHANGES.md
+++ b/gtfs/CHANGES.md
@@ -50,6 +50,10 @@ Every new feature adds complexity to the creation and reading of feeds. Therefor
 
 ### Revision History
 
+#### January 17, 2019
+
+* Editorial and formatting changes for clarity.  See [pull request](https://github.com/google/transit/pull/120).
+
 #### August 22, 2018
 
 * Added `feed_contact_email` and `feed_contact_url` fields in the `feed_info.txt` file. See [discussion](https://github.com/google/transit/pull/31).
--- a/gtfs/spec/en/examples/GoogleTransitExamples.md
+++ b/gtfs/spec/en/examples/GoogleTransitExamples.md
@@ -1,0 +1,211 @@
+## Introduction
+The fare\_attributes.txt and fare_rules.txt files in the [Google Transit Feed Specification](https://github.com/MobilityData/transit/blob/master/gtfs/spec/en/reference.md) can be confusing. The following examples work through different kinds of fare structures and show how to represent them in a Google transit feed.
+
+### Fare Examples
+
+* Example 1: All trips have the same fare, unlimited transfers<br>
+* Example 2: All trips have the same fare, no transfers<br>
+* Example 3: All trips have the same fare, transfers allowed<br>
+* Example 4: Different pricing for local and express routes<br>
+* Example 5: Buying a transfer increases the fare<br>
+* Example 6: Fare depends on station pairs, how you get there doesn't matter<br>
+* Example 7: Fare depends on zones<br>
+
+### Example 1: All trips have the same fare, unlimited transfers
+
+
+Suppose Demo Transit Agency has the following fare structure:
+
+* riders pay $1.00 on boarding (`price`=`1.00`, `currency`=`USD`, `payment_method`=`0`) 
+* a ticket is good for all vehicles and doesn't expire (`transfers` is empty) 
+* passengers can ride as long as they like. (`transfer_duration` is omitted)
+
+Since all trips have the same fare, Demo Transit can omit fare_rules.txt.
+
+fare_attributes.txt
+
+| fare_id | price | currency_type | payment_method |transfers |
+|:-----|:-----|:-----|:-----|:-----| 
+| only_fare | 1.00 | USD | 0 | |
+
+#### Calculating an adult fare
+The trip planner calculates a fare of $1.00 for each leg of the itinerary that includes a change of vehicle. However, unlimited transfers are permitted, so the trip planner only displays the lowest charge. Adult fare: $1.00
+
+<hr>
+
+### Example 2: All trips have the same fare, no transfers
+Suppose Demo Transit Agency has the following fare structure: 
+
+* riders pay $1.00 on boarding (`price`=`1.00`, `currency`=`USD`, `payment_method`=`0`)
+* passengers can ride as long as they like. (`transfer_duration` is omitted) 
+* any change in vehicles requires a new fare. (`transfers`=`0`)
+
+Since all trips have the same fare, Demo Transit can omit fare_rules.txt.
+
+fare_attributes.txt
+
+| fare_id | price | currency_type | payment_method | transfers |
+|:--------------|:----------|:-------------------|:--------------------|:--------------|
+| only_fare | 1.00 | USD | 0 | 0 |
+
+#### Calculating an adult fare
+The trip planner calculates a fare of $1.00 for each leg of the itinerary that includes a change of vehicle. So an itinerary that requires a change of buses would be $2.00.
+
+<hr>
+
+### Example 3: All trips have the same fare, transfers allowed
+Suppose Demo Transit Agency has the following fare structure: 
+
+* riders pay $1 on boarding (`price`=`1.00`, `currency`=`USD`, `payment_method`=`0`) 
+* unlimited transfers are allowed within 90 minutes (`transfers` is empty,`transfer_duration`=`5400`)
+
+Since all trips have the same fare, Demo Transit can omit fare_rules.txt.
+
+fare_attributes.txt
+
+| fare_id | price | currency_type | payment_method | transfers | transfer_duration |
+|:--------------|:----------|:-------------------|:--------------------|:--------------|:-----------------------| 
+| only_fare | 1.00 | USD | 0 | | 5400 |
+
+#### Calculating an adult fare
+The trip planner calculates a fare of $1.00 for each leg of the itinerary that includes a change of vehicle. Then it calculates the time for the itinerary. If the itinerary time is less than 90 minutes, the fare is $1.00.
+
+<hr>
+
+### Example 4: Different pricing for local and express routes
+Suppose Demo Transit Agency has the following fare structure: 
+
+* riders pay $1.75 on boarding local buses (route 1) 
+* riders pay $5.00 on boarding express buses (routes 2 and 3) 
+* transfers aren't allowed.
+
+Since some trips cost more than others, Demo Transit must include fare_rules.txt, and each route must have an entry to associate it with a fare.
+
+fare_attributes.txt
+
+| fare_id | price | currency_type | payment_method | transfers |
+|:--------------|:----------|:-------------------|:----------------|:--------------|
+| local_fare | 1.75 | USD | 0 | 0 | | express_fare | 5.00 | USD | 0 | 0 |
+
+fare_rules.txt
+
+| fare_id | route_id |
+|:-------------|:--------------| 
+| local_fare | Route_1 |
+| express_fare | Route_2 | 
+| express_fare | Route_3 |
+
+#### Calculating an adult fare
+The $5.00 fare is only applicable if you ride routes 2 or 3. The $1.75 fare only applies on route 1. If an itinerary uses routes 1 and 2, the fare is $6.75.
+
+<hr>
+
+### Example 5: Buying a transfer increases the fare
+Suppose Demo Transit Agency has the following fare structure: 
+
+* riders pay $1.75 on boarding local buses 
+* riders can pay an extra $0.25 on boarding to purchase a transfer 
+* transfers purchased are valid for 90 minutes
+
+Since these rules apply to all trips, Demo Transit can omit fare_rules.txt.
+
+fare_attributes.txt
+
+| fare_id | price | currency_type | payment_method | transfers | transfer_duration |
+|:--------------|:----------|:-------------------|:--------------------|:--------------|:-----------------------| 
+| simple_fare | 1.75 | USD | 0 | 0 | | 
+| plustransfer_fare | 2.00 | USD | 0 | | 5400 |
+
+#### Calculating an adult fare
+Technically, both fares apply on itinerary that has no transfers. However, the trip planner always chooses the least expensive applicable fare: 
+
+* For an itinerary with one transfer, `simple_fare` is $3.50 vs. $2.00 when a transfer is purchased. So the trip planner will report $2.00 fare on all itineraries that require a change of vehicle. 
+* For an itinerary with no transfers, $1.75 fare is less than `plustransfer_fare` of $2.00. So if an itinerary doesn't require a change of vehicle, the fare is $1.75.
+
+<hr>
+
+### Example 6: Fare depends on station pairs, how you get there doesn't matter
+In this example only the entry and exit points from the system matter.
+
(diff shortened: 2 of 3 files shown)


antrim:
Earlier, we experimented with defining the term **feed** in relation to **dataset**:

> * **Dataset** - A complete set of files defined by this specification reference. Datasets should be published at a public, permanent URL, including the zip file name. (e.g., www.agency.org/gtfs/gtfs.zip). For more information see GTFS best practices.
> * **Feed** - Successive datasets comprise a feed.

The above definition of **feed** was not included in this pull request. The changes in this pull request removes ambiguous uses of the word **feed**.

Questions:
* What do you or your organization believe is the correct definition of **feed**? How do you use the term?
* Should we include a definition of **feed** in the GTFS reference? Or leave that out of the reference?

skinkie:
While I would personally rather see that documenation would be written in 
Transmodel terms, I can see that, this is not working due to the header 
choices in GTFS. Still nothing is wrong to do the documenation better than 
the headers. So from your list I would make for example;

> Feed vs dataset

Publication

> Transit Organization --> transit agency

Operator


For some changes I think this goes in the wrong direction, specifically:
Itinerary --> Journey (this matches Fares proposal)

This means two totally different things in my opinion, and would only 
confuse more people. I would like to see a better propesal for this that 
would not confuse Itinerary with ServiceJourney =/= GTFS Trip.

giocorti:
Recheck CLA

antrim:
Thanks for your review. Here are responses.

> Operator
I'm hesitant to use the word operator because "transit agency" has historically referred to service brand which is sometimes distinct from the operator. For example, RATP operates Capital MetroBus in Austin TX but we don't want to display "RATP" to riders in Austin because it doesn't mean anything to them. 

> Publication
I worry that "publication" has a very general meaning in English. As an alternative to dataset, what do you think of "feed iteration"?

>Operator
I'm hesitant to use the word operator because "transit agency" has historically referred to service brand which is sometimes distinct from the operator. For example the large french firm RATP operates Capital MetroBus in Austin TX but we don't want to display "RATP" to riders in Austin because it doesn't mean anything to them. 

>Publication
We're in the semantical weeds here but I worry that "publication" is such a general word (one can publish basically anything that can be recorded) that it almost doesn't mean anything. As an alternative to dataset, what do you think of "feed iteration"?

> Itinerary --> Journey
I am guessing that "trip" in GTFS is equivalent to "ServiceJourney" in Transmodel? Would "journey" confuse those who are familiar with Transmodel terms?

giocorti:
I'm the commit author and I'm okay with contributing.

skinkie:
> I'm hesitant to use the word operator because "transit agency" has historically referred to service brand which is sometimes distinct from the operator. For example, RATP operates Capital MetroBus in Austin TX but we don't want to display "RATP" to riders in Austin because it doesn't mean anything to them.

This discussion is really great. Because it shows you show that we are not interested in an operator or agency, but instead: the brand. Would you agree that we should use brand instead?


> > Publication
> 
> I worry that "publication" has a very general meaning in English. As an alternative to dataset, what do you think of "feed iteration"?

For NeTEx a delivery starts with:
```xml
<PublicationDelivery xmlns:mstns="http://www.netex.org.uk/netex" xmlns="http://www.netex.org.uk/netex" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:gml="http://www.opengis.net/gml/3.2" version="1.0">
```
I would say lets not try find synonyms or in betweeners.


> > Itinerary --> Journey
> 
> I am guessing that "trip" in GTFS is equivalent to "ServiceJourney" in Transmodel?

Exactly.

> Would "journey" confuse those who are familiar with Transmodel terms?

It again depends on what you are trying to communicate here. Do you mean all the legs that make up the entire journey the _rider_ takes to go from A to Z via B, C, D etc. or do you mean the leg that could be walking or a specific ServiceJourney/Trip that goes from A to B, B to C, etc.

This is why we need an unambigious shared vocabulary.

giocorti:
> How about we alter the text to reference "brand" where that will make sense? Addressing branding completely will require changes to the spec.

IMO "transit brand" is more technically correct than "transit agency" but, due to that fact that agency is written into many field names, I think we lose more than we gain by shoehorning "transit brand" in. I should also note that MobilityData plans to add transit branding as a fully featured GTFS extension. So, for now, I propose we go with a stopgap measure in which we amend the agency_id field description to read (my addition in bold): 

>Identifies a transit agency. A dataset may contain data from multiple agencies. **If a single transit agency runs multiple customer facing transit brands, this field should be used to identify the individual transit brands.** This field is required when the dataset contains data for multiple transit agencies, otherwise it is optional.

antrim:
Note @giocorti's earlier comment was edited. We propose to leave "transit agency" in for now, and later address questions of agency branding holistically. This follows the purpose of this spec modification -- to provide greater clarity without making substantive changes in meaning.

If there are no other comments, I'd like to call the vote tomorrow.

antrim:
I'd like to call the vote on this change. This PR makes editorial and formatting changes to the GTFS specification. See the opening comment for a summary and motivation.

The vote will close on Dec 26 at 23:59:59 UTC.

skinkie:
-1

As mentioned before, there is still too much unclarity and we are again introducing new terms for subjects that have been heavily standardised. Obviously this isn't all black/white, but some change are more controversal than others.

antrim:
@skinkie: Which changes can we remove or alter to get your support for the proposal?

skinkie:
@antrim as we have discussed before "transit agency" is at this moment what it is described. But the change of itinerary to "Rider Journey" just becomes _newspeak_.

antrim:
@skinkie If we reverted back to "itinerary" instead of "rider journey" would that solve your concern?

skinkie:
And add that _brands_ are at this moment encoded as different agencies: sure.

giocorti:
I've made the requested changes. "Itinerary" has been kept and "brand" vs "agency" is now discussed. 

I should also note that we've defined the term "service day" in this PR, and this was accidentally omitted from the initial change list.

giocorti:
@skinkie thanks for the feedback, and no offense taken at all!

I'm using the term "service day" because its already in the spec. I've specifically added it as a defined term because it appears in a number of places and it could be potentially confusing as it doesn't correspond to an actual day (service days can be greater than 24 hours).  So really the definition is just to make it explicit and obvious that a service day is not the same as a day. 

But you bring up a good point about "service day" vs "operational day". Unless someone else is opposed, I think that this a change that should be incorporated.

skinkie:
Please allow some time for feedback on the service day subject. If it is already used somewhere in the spec I am not opposed to use the term here. What I am eager to learn is where these terms originate from and where they started to deviate (etymology).

giocorti:
>What I am eager to learn is where these terms originate from and where they started to deviate (etymology).

My inspiration for these terms just comes from the general (American) english lexicon. In general I've just tried to use the most accurate and understandable word in order to minimize confusion for someone reading the reference. In some cases (such as "record") I've also consulted what is, IMO, the appropriate technical literature. I should explicitly state that I am not trying to mirror the terminology defined by some outside party or spec. Rather, I'm trying to make the reference as understandable and consistent as possible to a wide variety of readers. I admit that, as a native American english speaker and someone who is not well versed in other specs such as Transmodel , I do exhibit a linguistic bias towards vernacular American english. 

In some cases, such as empty vs blank, there is no real reason to chose one word over the other. I just wanted a single word to be consistently used. 

I'm more than happy to discuss my reasons for choosing specific words if there are any you're wondering about.

antrim:
How about we call the vote after the new year, since many people will be away for the holidays? This will let everyone weigh in and allow a discussion of the terminology. We have 30 days from the last vote to continue working on the proposal.

antrim:
I am calling a second vote on this change, which @giocorti updated (a48aa36c8f30667a7046e16802833e08a03064f2 Dec 20) after it failed to pass. This PR makes editorial and formatting changes to the GTFS specification. See the opening comment for a summary and motivation.

The vote will close on Dec 10 at 23:59:59 UTC.

skinkie:
+1

abyrd:
I completely agree that the spec language should be made internally consistent, and if possible consistent with other external usage.

However, re-reading this whole discussion I find it pretty confusing. I cannot grasp how changing the terms used in the GTFS documentation so they no longer match the field names in GTFS would make the documentation "better". Surely the GTFS documentation should, with absolute consistency, use the same terms with the same meanings as in the field names.

There are strong unstated assumptions underlying this discussion, perhaps most notably that a particular CEN norm (Transmodel) is inherently authoritative or correct, to the extent that it even supersedes common usage. At one point @skinkie questions a change from "Itinerary" to "Journey" saying that "This means two totally different things in my opinion". These are certainly two different words with different nuances, but I believe any native English speaker would find both are perfectly reasonable here and perhaps essentially interchangeable. If they mean two "totally different things", it is only within some external, strictly defined technical context.

I work on transit in Europe and hope to continue to do so, and am favorable to convergence between Transmodel/Netex and GTFS.  But I cannot agree with the idea that GTFS has somehow failed to adhere to Transmodel's authoritative definitions - this implies that Transmodel is predominant internationally and ignores the history of GTFS as a reaction against what were perceived as centralized, over-engineered specifications. I imagine most of the people who created GTFS in its current form had never even heard of Transmodel or Netex, and did not even consider adopting its ontology. They intentionally chose to sidestep all existing systems in the Transmodel vein. I would even wager that many or most people using GTFS today will never encounter Transmodel or Netex, and might even actively avoid it if they knew it existed.

I believe a structural equivalence between GTFS and Netex (especially its "European Profile") is both possible and desirable, but terminological differences will remain because they were created separately. I believe we should work toward structural equivalence, and then just include an annotated equivalence table to explain the vocabulary differences.

I don't think any of the expressions proposed above should be called "newspeak", which implies they are cynically simplifying or degrading language as a means of political control or limitation of thought. As @giocorti says, they are simply derived from common English usage and pieced together using standard grammar, which I'd say is the opposite of newspeak.

It is almost as if the definitions drawn from one specification (Transmodel) are taken to be the true or universal meaning of the words. I suspect this arises from the fact that outside the Anglosphere, English is a "work-language" employed in instrumental or strategic action, but rarely for communicative action which is carried out in a different local language (https://eight2late.wordpress.com/2013/02/07/the-system-and-the-lifeworld-a-note-on-the-gap-between-work-and-life/). Despite using English every day, someone may only encounter or use a word such as "itinerary" as a narrowly defined technical term, e.g. when reading and commenting on the Transmodel spec.

I do agree that we should favor usage patterns drawn from the transit operations context where possible and where this does not contradict the rest of the spec (e.g. "operational day" versus "service day").

Anyway, I'm saying all that as input to future discussion - I'd like to make less assumptions about "right" meanings of words and concentrate more on taking action to cooperate and converge.

This changeset looks reasonable to me in its current state so I give it a +1.

barbeau:
+1

prhod:
+1
Reading again this proposal, it seems that uniqueness of a service_id in the calendar.txt file has been removed. Does that implies that it is now possible to declare a service_id several times ? (for exemple a trip running from monday to friday from date1 to date2 and date3 to date4 ?)

LeoFrachet:
Good catch @prhod. Thanks.

We definitely do not intend to change this behavior. It has been discussed in the past and the conclusion was to keep it as it is. I don't want to re-open this discussion in this thread.

@giocorti Could you have a look on that? I see you define ID type as "A sequence of any UTF-8 characters which uniquely identifies an entity, but does not necessarily identify a specific record in a table." But is there a place where you define uniqueness?

skinkie:
@LeoFrachet while not intended I don't mind to give my support for this great idea of non-continuous service dates ;-)

giocorti:
Uniqueness is not explicitly defined anywhere in the spec. I've actually removed language that specified that IDs were dataset unique where dataset unique was defined as

>Dataset Unique - The field contains a value that maps to a single distinct entity within the column. 

I removed that language because it was factually incorrect in some cases. For example, shape_id in shapes.txt does not identify a distinct entity within a column.

Of course, uniqueness is an important concept in GTFS so we may want to define it. But we'd also need to be careful about the exact language we use to do as its easy to write something that entirely accurate.

skinkie:
@giocorti the point with service_id is that in calendar_dates.txt it is defined multiple times.

barbeau:
@LeoFrachet I agree - I think that change is significant enough it requires a re-vote.

LeoFrachet:
I am calling a third vote (!) on this change, which @giocorti update ([c7f18ba Jan 20](https://github.com/google/transit/pull/120/commits/c7f18baa63f521ee109907d1af8bde038f920e36)) during the second vote. This PR makes editorial and formatting changes to the GTFS specification. See the opening comment for a summary and motivation.

The vote will close on Jan 17 at 23:59:59 UTC.

For the four who had already casted they ballot, please revote: @skinkie, @abyrd, @barbeau, @prhod.

barbeau:
+1

skinkie:
+1

prhod:
+1  (Thanks for the clarification). Lets keep the non-continuous service dates for later ;-)

abyrd:
Yes, good catch @prhod. +1

dbabramov:
+1

antrim:
The vote closed on Jan 17 at 23:59:59 UTC. We have 5 votes in favor of the change, from both GTFS producers and consumers, and no votes against. So this change passes. We'll get this merged.

LeoFrachet:
Nice!

Aaron (@antrim), since all the commits are from Giovanni (@giocorti), may I let you handle the CLA issues?

Thanks

antrim:
@googlebot. The pull request author (@giocorti) confirmed they are ok with contributing in this comment: https://github.com/google/transit/pull/120#issuecomment-443283173

It looks as though that is the last step we need to pass the checks.

barbeau:
I just updated the revision history in https://github.com/google/transit/pull/120/commits/61c15c54dd02443ef03d591697cafbadb790a436 and it looks like that confused the CLA bot again.  I'm ok with my commits being contributed to this project (obviously).

PR #121, opened 2018-11-26
Clarify TripUpdate documentation with use of NO_DATA

barbeau:
The GTFS-realtime spec doesn't define "a delay of unspecified duration", so this patch clarifies how the ScheduleRelationship of `NO_DATA` should be used within StopTimeUpdates to implement this feature.

Announced on the GTFS-realtime Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/kATL9Q7Y1gI.

**EDIT** additional description supplied by @abyrd:

>This PR concerns an example of delay propagation. In this example, two StopTimeUpdates show a vehicle being late by a specific number of minutes. The third StopTimeUpdate is said to have "a delay of unspecified duration", but it's not entirely clear what such a message would look like. "Unspecified delay" must mean both the arrival time and departure time are missing in that StopTimeUpdate. The documentation on arrival and departure times says: "If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED. If schedule_relationship is NO_DATA, arrival and departure must be empty." So, if this third StopTimeUpdate has unspecified delay, the schedule_relationship must be SKIPPED or NO_DATA, so it is indeed clearer to change the example to use a NO_DATA TripUpdate.

### DIFF
--- a/gtfs-realtime/spec/en/trip-updates.md
+++ b/gtfs-realtime/spec/en/trip-updates.md
@@ -35,7 +35,7 @@ For the same trip instance, three [StopTimeUpdates](reference.md#StopTimeUpdate)
 
 *   delay of 300 seconds for stop_sequence 3
 *   delay of 60 seconds for stop_sequence 8
-*   delay of unspecified duration for stop_sequence 10
+*   [ScheduleRelationship](/gtfs-realtime/spec/en/reference.md/#enum-schedulerelationship) of `NO_DATA` for stop_sequence 10
 
 This will be interpreted as:
 


barbeau:
@IsaacEMoses That's certainly one way a "delay with unspecified duration" could be implemented, although I'd prefer not to debate that in this PR (we could discuss that separately).  For this proposal I'd like to focus on clarifying/fixing the documentation where propagation is described.  I'm about 95% sure that the intent here was to specify how `NO_DATA` works during propagation, but the current language is ambiguous.

harringtonp:
+1

abyrd:
I wasn't quite sure what this PR was about. The PR description says it clarifies how to use NO_DATA to "implement this feature" without stating what the feature in question is. This may be why there have been few votes on this issue.

Updating the description with the following information might lead people to accept this PR:

This PR concerns an example of delay propagation. In this example, two StopTimeUpdates show a vehicle being late by a specific number of minutes. The third StopTimeUpdate is said to have "a delay of unspecified duration", but it's not entirely clear what such a message would look like. "Unspecified delay" must mean both the arrival time and departure time are missing in that StopTimeUpdate. The documentation on arrival and departure times says: "If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED. If schedule_relationship is NO_DATA, arrival and departure must be empty." So, if this third StopTimeUpdate has unspecified delay, the schedule_relationship must be SKIPPED or NO_DATA, so it is indeed clearer to change the example to use a NO_DATA TripUpdate.

I give a +1 to this PR.

barbeau:
Thanks @abyrd, I agree that I could have been clearer in the initial description.  I've added your description to the first comment in this PR thread.

barbeau:
* Yes - 3

So it passes!  Thanks all!

PR #126, opened 2018-11-27
Define GTFS-realtime experimental field consensus process

barbeau:
As discussed in https://github.com/google/transit/issues/109, we need to define what the consensus process is for creating new GTFS-realtime experimental fields.

This proposal defines the consensus process as follows:

>Consensus is reached via a discussion and voting process that is the same as the below [Specification amendment process](#specification-amendment-process), but instead of unanimous consent only 80% yes votes are required for approval.

Announced on GTFS-realtime Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/yDDH75MpmPg.


gcamp:
@barbeau I think this is uncontroversial, potentially call for a vote now?

paulswartz:
+1

TomGoBravo:
+1

abyrd:
+1

barbeau:

* Yes - 7

So it passes! Thanks all!

Issue #131, opened 2018-12-21
Add three new optional fields for timestamps in Service Alerts feed: created_timestamp, last_modified_timestamp, closed_timestamp


ibi-group-team:
The current specification does not offer a straightforward way for feed producers to provide, and feed consumers to monitor the lifecycle of an alert based on information provided in the alerts feed. For example, there is no field related to when an alert was created, last modified, or closed. This means that consumers must monitor every incremental file to infer these values, which may cause errors. This can be especially important for alerts that are continuously modified by the producer to provide timely information related to the developing impacts of an unscheduled service disruption.

To solve this issue we propose that three fields are added called “created_timestamp”, “last_modified_timestamp”, and “closed_timestamp.” All three fields would be optional and comprised of uint64 data type containing POSIX time. The created_timestamp field would include the date and time for when the alert was first generated. The last_modified_timestamp field would be populated with the date and time when the alert is updated. The closed_timestamp field would include the date and time when the alert is closed. 

Producers will need to keep alerts that have been closed in their feed for a certain amount of time so that consumers can use the closed_timestamp field in their applications. Currently, feed consumers infer when an alert is active using the active_period start and end time fields. However, both of these fields are conditionally required. If the end time is not provided, consumers can use the closed_timestamp to know when the alert is closed. If the end time is provided, consumers should infer that an alert is no longer active when the closed_timestamp field is found in the feed because a feed producer may close an alert before the active period end time (for example, traffic delays subside earlier than expected). In this case, producers should remove the active time period from the feed.

Incorporating these three fields in the specification will improve consumer and transit customer confidence in the producer’s alert information, and help producers ensure they are publishing up-to-date alerts within their feed.

skinkie:
We materialise in GTFS-RT an entire valid situation. Can you refer to any version management on any current structure?

ibi-group-team:
@skinkie can you please clarify what you mean by "any version management on any current structure"?

skinkie:
I would like to introduce three attributes for reasons I do not understand. When an alert is removed from the feed it can be removed from the consumer, there is no ambiguity about that. So I would like to know, is there any location in GTFS-RT at this moment that would retain 'historic information' after the data has been expired, or announcing something in the future.

abyrd:
[...]

PR #134, opened 2019-01-10
Add three new optional fields for timestamps in Service Alerts feed: created_timestamp, last_modified_timestamp, closed_timestamp

ibi-group-team:
This is a proposal to add three new optional fields of type uint64 to the Service Alerts feed: created_timestamp, last_modified_timestamp, closed_timestamp.

See discussion here: https://github.com/google/transit/issues/131

Announced on the GTFS-realtime Google Groups here: https://groups.google.com/forum/#!topic/gtfs-realtime/dipoHhUt6aE

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -427,6 +427,15 @@ message Alert {
   // Full description for the alert as plain-text. The information in the
   // description should add to the information of the header.
   optional TranslatedString description_text = 11;
+  
+  //Time alert is created, in POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC)
+  optional uint64 created_timestamp = 12
+  
+  //Time alert is last modified, in POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC)
+  optional uint64 last_modified_timestamp = 13
+  
+  //Time alert is closed, in POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC)
+  optional uint64 closed_timestamp = 14
 
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime Specification in order to add and evaluate new features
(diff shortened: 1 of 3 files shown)


farah-machlab-ibigroup:
I signed it!

ibi-group-team:
This pull request has been open for more than one week, so we are calling for a vote.

Voting will be closed on Friday January 25 at 23:59:59 UTC.

jeffreykessler-keolis:
+1

jrsanbornjr:
+1

abyrd:
Unfortunately I have to vote -1 on this in its current form.

I agree that timestamp fields may be useful, but this proposal seems incomplete. The meaning of these fields needs some clarification before adoption. Comments are provided, but they reuse terms contained in the field names without providing definitions. What is the specific meaning of "created", "modified", and "closed"? How is "created" different than the beginning of the validity period? How is "closed" different than the end of the validity period? "Modified" implies that there are persistent alert entities which change over time - under what circumstances are separate entities received at different times identified as being the same entity even though their contents are different? Is the identity of such persistent alerts determined entirely by the identifier field? Or alternatively, can an alert be considered "modified" if it  replaces an alert with a different ID?

Some of these answers may seem obvious to people already using these fields, and I could venture guesses, but these things need to be stated clearly to allow other people to use and consume the fields consistently.

abyrd:
Additional things to clarify based on discussion over in #131: Do you distinguish between major changes (different meaning) which should notify end users, versus minor changes (typos, capitalization) which don't merit notifications? If a feed contains the modified timestamp field, does that field's value change if and only if the alert message changes? Is it a requirement that the modified timestamp must change if the alert's message or header text changes at all (even by one character or capitalization)?

barbeau:
-1 for me as well.  As @LeoFrachet says there are bigger issues regarding alert ID persistence as well as guidelines for how long an alert should remain in a feed after being "closed" that need to be addressed before fields like `closed_timestamp` can really be effectively utilized by consumers.

ibi-group-team:
Yes: 2
No: 3

This proposal is rejected so we will open a request for adding these fields as extensions.

Thank you all!

barbeau:
IBI Group extension request is at https://groups.google.com/forum/#!topic/gtfs-realtime/JfzgiKRRtOg.

PR #141, opened 2019-02-15
StopTimeUpdate additional field: sign_text

paulswartz:
Here @mbta, we've had an additional field in our Enhanced JSON TripUpdates feed which may be useful to others: `status`, attached to `StopTimeUpdate`. We use this value to represent text which appears on signs at the given stop/station, in order to provide that data to clients such as our website. 

You can see examples of this field at [North Station](https://www.mbta.com/stops/place-north?tab=departures), where it's used for boarding statuses on the Commuter Rail, like "Now Boarding" and "All aboard". We also use it for [messages](https://www.mbta.com/news/2018-08-27/stopped-train-messaging-begins-august-28-mbta-countdown-clocks) that trains are stopped unexpectedly.

Per some discussions in #109, if this interests other clients we can promote this into our actual Protobuf feed in addition to our JSON feed.

Google group: https://groups.google.com/forum/#!topic/gtfs-realtime/AmaiiCMLbjI

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -221,6 +221,9 @@ message TripUpdate {
     optional ScheduleRelationship schedule_relationship = 5
         [default = SCHEDULED];
 
+    // Optional text displayed on signs for this trip at the given stop.
+    optional string sign_text = 6;
+
     // The extensions namespace allows 3rd-party developers to extend the
     // GTFS Realtime Specification in order to add and evaluate new features
     // and modifications to the spec.
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -178,6 +178,7 @@ The update is linked to a specific stop either through stop_sequence or stop_id,
 | **arrival** | [StopTimeEvent](#message-stoptimeevent) | Conditionally required | One | If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED.  If schedule_relationship is NO_DATA, arrival and departure must be empty. |
 | **departure** | [StopTimeEvent](#message-stoptimeevent) | Conditionally required | One | If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED.  If schedule_relationship is NO_DATA, arrival and departure must be empty. |
 | **schedule_relationship** | [ScheduleRelationship](#enum-schedulerelationship) | Optional | One | The default relationship is SCHEDULED. |
+| **sign_text** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | Text to be displayed on a digital sign at the stop for this StopTimeUpdate. Examples: "Now Boarding", "All aboard", "Stopped 3 stops away" |
 
 ## _enum_ ScheduleRelationship
 


skinkie:
Could you make sure your patch only does a single thing. Not fix spaces? Next to this, why a status here, instead of in service alerts?

paulswartz:
> Could you make sure your patch only does a single thing. Not fix spaces? 

https://github.com/google/transit/pull/142 removes the whitespace.

>Next to this, why a status here, instead of in service alerts?

This is separate from service alerts. In the Commuter Rail example, a train being ready for passengers to board isn't an alert. In both cases, the text is different for each stop, and so implementing it at the Alert level would mean an alert for each stop on each trip.

skinkie:
> This is separate from service alerts. In the Commuter Rail example, a train being ready for passengers to board isn't an alert. In both cases, the text is different for each stop, and so implementing it at the Alert level would mean an alert for each stop on each trip.

This could still be implemented using serviceAlert.pb has an EntitySelector which includes trip and stop_id.

paulswartz:
> This could still be implemented using serviceAlert.pb as it has an EntitySelector which includes trip and stop_id.

This is technically true. As I write this though, we have 74 current or future service alerts in our feed, and 210 statuses. The alerts change much less frequently than the trip updates, so we'd be burdening clients who only care about alerts with this additional data as well as a dramatically increased update frequency.

barbeau:
@paulswartz Thanks for sharing this!  How much do the messages vary in type?  In other words, could they be an enumeration instead of free-text?

paulswartz:
Here are the current values in our feed:
- `On time`
- `Now boarding`
- `All aboard`
- `Cancelled`
- `Delayed`
- `Late`
- `Departed`
- `Info to follow`
- `See agent`
- `Stopped at station`
- `Stopped X stops away` (where X is an integer generally between 1 and 20)

We had others in the past as well, which we've since stopped using:
- `X stops away`
- `Approaching`
- `Boarding`

Some of these (On time, Cancelled, Departed) could be inferred from existing data. Others (Delayed for a specific amount of time vs. Late for an unknown amount) could be enumerated. The full set of values, however, are too large for us to enumerate. The commuter rail values used to be implemented as an enumeration. However, as we extended the usage of the field, it became unwieldy. Additionally, keeping it as text allows us to propagate changes to the wording down to clients automatically, rather than requiring each client to change their display of an enumerated value.

For us, the purpose of this field is to provide data that's available in stations (via the signs) to other clients, like apps and our website. An enumeration doesn't quite do that.

barbeau:
@paulswartz Ok, thanks for the info.  Maybe `sign_text` (or something similar) would be a better name for this field, then?  Do you have a length constraint?  

I think `status` in general is probably too broad in scope to be generally useful to clients, but if we narrow the scope more specifically so that consumers know what to expect in the field I think it could be generally useful.

abyrd:
I do see how this could be useful. I agree with @barbeau that the field name should be changed - something making it clear that this is text information for the rider shown on a variable message sign. `sign_text` or `sign_message` seems good. The comment describing the purpose of the field is kind of tautological though, mentioning a "status for this StopTime". It should be made clear that this is text intended for display at the physical stop location associated with this stop time, for riders waiting at that stop. For RT publishers using messages to drive arrival time signs at stops this field might be easy to understand, but it may be less obvious to people feeding data into trip planners etc.

paulswartz:
I went with `sign_text`. TBH, I'm not sure this is valuable for trip planners specifically; this is data that's targeted at the riders themselves, when they're at/near a station, rather than intermediaries.

sdjacobs:
+1 

Metro North has a similar field in their data, and might add it as an extension until this PR gets merged.

barbeau:
@sdjacobs Could you list the possible values that appear in Metro North's field?

sdjacobs:
I believe it's these: ONTIME, DELAYED, LATE, ARRIVING, ARRIVED, CANCELED, DEPARTED, BOARDING, BUS, FINAL_DESTINATION, NO_SERVICE
Though not all of these are used. And I think it's good that the type of this field is a string, since different systems may support different `sign_text` values.

PR #145, opened 2019-03-05
Fix link for example of block_id

gcamp:
Simple fix for the broken link of block_id example.


skinkie:
Approved. +1

barbeau:
Thanks @gcamp!

PR #146, opened 2019-03-05
Platform codes in stops.txt

LeoFrachet:
tldr: Google is using an unofficial GTFS extension to describe platform code, documented [here](https://developers.google.com/transit/gtfs/reference/gtfs-extensions#station-platforms). This proposal offers to add it to the official specification.

### The need
Stops in GTFS can represent two things: either isolated bus stops (when `parent_station` is empty), or platform within station (when `parent_station` is defined). When they describe platform within a station, they often have a name within the station, often digits or letters, that cannot be represented in GTFS, but which are need to inform the traveler on which platform the train is. Below are three example of such:

capture d ecran le 2019-03-05 a 15 49 58

(Sources: [Berlin](https://www.bahnhof.de/bahnhof-de/bahnhof/Berlin_Ostbahnhof-1029522), [Paris](http://www.garedelyon.fr/guide-gare/plan-orientation/), [Toronto](http://maps-toronto.com/map-of-union-station-toronto))

### The proposal
The proposal is exactly what Google is already using, so that the producers of the current Google extension won't have to change their code to produce the official version. It offers to add in `stops.txt` the following field:

> `platform_code` (Text, Optional) Indicates the platform identifier for a platform stop. This should be just the platform identifier (eg. "G" or "3"). Words like "platform" or "track" (or the feed’s language-specific equivalent) should not be included. This allows feed consumers to more easily internationalize and localize the platform identifier into other languages.


gcamp:
Can the git history be cleaned to only have one commit on this?

For the proposal itself I have nothing against.

skinkie:
+1 Implemented in The Netherlands.

abyrd:
This seems like a a good idea to me, and indeed it's a de facto existing feature that has producers and consumers.

harringtonp:
Are there any good examples of how platform_code should be used ?

I am looking through the MBTA stops.txt file where they have both platform_code and platform_name and see overlap in many cases between what is in these fields and the stop_name and stop_desc fields. There is not a clear delineation between the fields.

I guess the issue is that as searching is done on the stop_name field it cannot contain something like "A" even if the stop_name for its parent_station describes accurately where the station is.

skinkie:
@harringtonp in GTFS you could write:
stop_id = CS:A, stop_name = Central Station platform A

This would render this string in every client as platform A. Obviously you would be able via translations.txt to bring in a locale, but that gives more mess. So in this very generic case:
stop_id = CS:A, stop_name = Central Station, platform_code = A

Would allow the client to present "A" in a different shape, or translate the word "platform" to the locale.

abyrd:
@harringtonp I think the key idea is that this is a very short, frequently one or two character identifier without any language specific. So the contents of this field should be safe for use in the following way:

`[Word for track or platform]: [platform_id]`

So it could be rendered in any language like `Voie: F` or `Spoor: F` or `Track: F`.

harringtonp:
Thanks guys, clears that up

paulswartz:
To use an @mbta example:

| stop_id | stop_name | platform_code | platform_name |
| - | - | - | - |
| place-sstat | South Station | | |
| South Station-01 | South Station | 1 | Commuter Rail - Track 1 |

+1

jakluk:
Also used in Prague (PID) in both stations and bus stops on streets. +1

LeoFrachet:
Indeed, it would be worth adding guidelines about `stop_name` vs `platform_code`.

In my understanding, we all agree on what should be in:
- the station name (aka the `stop_name` of the parent station): "South Station" in MBTA example, or "Union Station" in Toronto;
- the `platform_code`: "1" in MBTA example, "12" in Toronto.

Something which is still unclear for me is the stop name. I see two options:
- Either, for backward compatibility reason (aka users not parsing station information nor platform code), we request both information in the stop name, which would be "South Station, platform 1" in MBTA example. This field should just be ignore by up-to-date consumers, which would find the same information in station name + platform code.
- Or... or what? MBTA seems to be offering to say that stop name should be equal to station name? Or we let the producer decide? (but what should the consumer do?)

@skinkie & @jakluk: what do you put in stop_name when you have a station name and a platform code?

skinkie:
| stop_id | stop_code | stop_name                    | stop_lat      | stop_lon      | location_type | parent_station | stop_timezone | wheelchair_boarding | platform_code | 
|---------|-----------|------------------------------|---------------|---------------|---------------|----------------|---------------|---------------------|---------------| 
| 103968  |           | "Den Haag, Centraal Station" | 52.080507     | 4.323319      | 0             | NL:S:32002601  |               | 1                   | H             | 
| 1542169 |           | Amsterdam Centraal           | 52.3793258368 | 4.89970386028 | 0             | stoparea:18188 |               |                     | 7a            | 
| 1542170 |           | Amsterdam Centraal           | 52.3786872838 | 4.90142047405 | 0             | stoparea:18188 |               |                     | 7b            | 
| 1542189 |           | Alphen a/d Rijn              | 52.1246262133 | 4.65733408928 | 0             | stoparea:18304 |               |                     | 3             | 


H being busplatform H,
7a being the front section of the train platform
7b being the back section of the train platform
3 being a single train platform without phases

harringtonp:
@LeoFrachet

It should at the very least be explicitly mentioned that the stop_name contain at least the same information as it's parent (i.e. station) stop_name. I guess there will be varying opinions as to whether or not the platform should be included. 

You would think it should be obvious that the station name be incorporated but the UK Network West Midlands example below illustrates this is not always the case:

stop_id,stop_code,stop_name,stop_lat,stop_lon,stop_timezone,location_type,parent_station,platform_code
0600CO210,CHEPWJG,"Bus Station (Stand 1)",53.163761,-2.211721,,0,060GCO124,1
0600CR287,CHEPWJD,"Bus Station (Stand B)",53.068733,-2.520081,,0,060GCR195,B
0600CR289,CHEPTPT,"Bus Station (Stand D)",53.068779,-2.520126,,0,060GCR195,D
0600CR666,CHEPWAD,"Bus Station (Stand 3)",53.098331,-2.443227,,0,060GCR194,3
0600CR667,CHEPWAG,"Bus Station (Stand 4)",53.098331,-2.443316,,0,060GCR194,4
0600CR669,CHEPWAM,"Bus Station (Stand 6)",53.098339,-2.443511,,0,060GCR194,6
0600CR673,CHEPWAW,"Bus Station (Stand 10)",53.098347,-2.443840,,0,060GCR194,10
0600MA6110,CHEPWJP,"Bus Station (Stand 9)",53.257961,-2.125498,,0,060GMA280,9
0600MA6148,CHEPWGP,"Bus Station (Bay 3)",53.258240,-2.125019,,0,060GMA280,3
0610CH19054,CHWTJPD,"Chester Bus Interchange (Stand C)",53.194275,-2.889155,,0,061GCHCBI,C
100000008699,dbsgdjwj,"Bus Station (Stop 3)",52.774445,-1.558618,,0,100GDSTC1371,3
100000015551,dbsgpjap,"Bus Station (Stop 2)",53.015377,-1.733605,,0,100GDATC0641,2
1000DSBS1636,dbsdjwtp,"Bus Station (Stop 4)",52.774410,-1.558871,,0,100GDSTC1371,4
1000DSBS1931,dbsdjwtw,"Bus Station (Stop 5)",52.774384,-1.559108,,0,100GDSTC1371,5
1000DSBSB818,dbsdagmd,"Bus Station (Stop 1)",52.774502,-1.559033,,0,100GDSTC1371,1
1000DSBSB961,dbsdagmg,"Bus Station (Stop 2)",52.774475,-1.559196,,0,100GDSTC1371,2
1090BSTN01,dbyptmad,"Bus Station (Bay 1)",52.921745,-1.472122,,0,109GDDCCBS01,1
1090BSTN23,dbyptmjd,"Bus Station (Bay 23)",52.922318,-1.471103,,0,109GDDCCBS01,23
1090BSTN24,dbyptmjg,"Bus Station (Bay 24)",52.922344,-1.471058,,0,109GDDCCBS01,24
1090BSTN29,dbyptmpa,"Bus Station (Bay 29)",52.922405,-1.470582,,0,109GDDCCBS01,29
2000G000101,wocgpdgj,"Crowngate Bus Station (Stand A)",52.193016,-2.225531,,0,200G000001,A
2000G000102,wocdwpwj,"Crowngate Bus Station (Stand B)",52.193062,-2.225371,,0,200G000001,B
2000G000103,wocdwpwm,"Crowngate Bus Station (Stand C)",52.193100,-2.225210,,0,200G000001,C
2000G000104,wocdwpwp,"Crowngate Bus Station (Stand D)",52.193153,-2.224976,,0,200G000001,D
2000G000105,wocdwpwt,"Crowngate Bus Station (Stand E)",52.193218,-2.224713,,0,200G000001,E

TomGoBravo:
stop_name: I like @LeoFrachet 's suggestion of backward compatibility for consumers that ignore the parent_id and platform_code. I think it is important that this change recommend consumers that display a platform_code ignore the stop_name on that row and use the parent stop_name as needed.

I'm concerned that finding the right word(s) to put next to platform_code will be annoying. Apps will need to find the route_type that use the stop to determine if it is a bus stand or rail platform then pick the right term in the local language or UI language. When signs in the station say `Voie: F` or `Spoor: F` it'd be nice for the UI to show exactly what is on the sign, not translated into the UI language. On the other hand, when that breaks it is doubly confusing; a French reader not knowing English is going to have an easier time getting to a place labeled `Platform: F` with a UI showing `Voie: F` than a UI showing `Track: F`. The problem is likely worse when navigating where the local script is unfamiliar.

abyrd:
> When signs in the station say `Voie: F` or `Spoor: F` it'd be nice for the UI to show exactly what is on the sign, not translated into the UI language. 

Real trouble comes when "Voie F", "Spoor F", _and_ "Platform F" are all used in the same place, but all that will not fit in a mobile UI. In multi-lingual cities this happens. I used French, Dutch, and English in the example thinking of Bruxelles / Brussel / Brussels.

harringtonp:
> About the stop_name for platforms: I would lean toward combining the parent station stop_name with the platform_code, perhaps without inserting any word for "track" or "platform". So a platform stop_name might be "Amsterdam Centraal 7a", which allows it to be useful when displayed in places where the platform_code is not displayed separately, and ensures it is clear that this entity does not represent the entire station. One argument for this is backward compatibility; another way to look at it is that it's better to have redundancy across fields of a single record than to have information that can be misinterpreted across many records (every platform looking like it might be the whole station).
>

I would support the above as it seems like the best compromise and as Andrew says does look after backwards compatibility. In yourstop a user can search for stops. As they type away matches are found for both stations and individual stops.  In the attached image, Worcester Crowngate is the station and the matches above it are all stops within the station. Because the internal stops or "stands" have an accurate description in the stop_name field, it should (hopefully) be clear to a user what they are looking at. There are other examples in this feed which would be a lot less clear to a user due to poor descriptions in the stop_name field. There are also cases where a user would never be able to find a stand through a text search as stop_name only has generic stand information and gives no indication of the parent station.

stopsearch

jakluk:
@LeoFrachet The platform is not included in the stop_name. It's mostly because the platform is never written in paper timetables. So when there is a parent station, its sub-stops have the same stop_name.

Our on-street bus and tram stops also use platform codes, but they are not inside a parent station group, because we treat stations only as buildings. Internally we call a group of close stops as a "node", stops in one node can even have different names, but it's convenient to have them marked by letter codes when they are around one intersection.

LeoFrachet:
Regarding the `stop_name`, the current definition is "Name of the location. Use a name that people will understand in the local and tourist vernacular." Maybe we can keep it like that for now. It's broad and simple. And apparently, you are all handling it well (Stefan @skinkie, Paul @harringtonp, Paul @paulswartz, Jakob @jakluk). But if you think we should, we can add something like:

> When the `stop` belongs to a station and has a `platform_code`, its `stop_name` should contain both the name of the station and the platform_code.

On the other subjects, I agree that two limits of the current proposal are:
- There is no way to know if those are tracks or platforms (or even "Bay" or "Stand" in a bus terminal)
- Some large stations like [Gare de Lyon](http://www.garedelyon.fr/wp-content/uploads/2013/10/plan-halls-gare-de-lyon.png) group tracks into "Halls". Tracks A to N belong to "Hall 1" and Tracks 5 to 23 to "Hall 2". 

But I don't see easy way to handle those issues. Google is already using `platform_code` as it is, so I would advocate to just consolidate the spec for now.

harringtonp:
> Regarding the `stop_name`, the current definition is "Name of the location. Use a name that people will understand in the local and tourist vernacular." Maybe we can keep it like that for now. It's broad and simple. And apparently, you are all handling it well (Stefan @skinkie, Paul @harringtonp, Paul @paulswartz, Jakob @jakluk). But if you think we should, we can add something like:
> 
> > When the `stop` belongs to a station and has a `platform_code`, its `stop_name` should contain both the name of the station and the platform_code.
> 

I would be in favor of this addition for the stop search reasons illustrated above

skinkie:
> > > When the `stop` belongs to a station and has a `platform_code`, its `stop_name` should contain both the name of the station and the platform_code.
> 
> I would be in favor of this addition for the stop search reasons illustrated above

I would be against that.

abyrd:
> Regarding the `stop_name`, the current definition is "Name of the location. Use a name that people will understand in the local and tourist vernacular." Maybe we can keep it like that for now. It's broad and simple. 

Actually, maybe we should rewrite this. What exactly is the "local and tourist vernacular", and is that really what feed producers should use for the stop_name? Shouldn't the stop_name be exactly the name shown on any station signage and on official timetables, with no regard at all for slang terms, nicknames, or expressions used by tourists? How the agency chooses that name is up to them.

abyrd:
> I would be against that.

Can you give your line of reasoning? 

Is it simply that copying the parent station stop_name into the child stations allows displaying the name and platform for that child station without referencing external entities (i.e. without looking at the parent station to see what the stem of the name is)?

skinkie:
> > I would be against that.
> 
> Can you give your line of reasoning?

The idea for introduction of _platform_code_ was to remove: "Centraal Station perron 1" to make it "Centraal Station" and have  platform_code = 1. Because it does not support localisation.

abyrd:
> > > I would be against that.
> > 
> > Can you give your line of reasoning?
> 
> The idea for introduction of _platform_code_ was to remove: "Centraal Station perron 1" to make it "Centraal Station" and have platform_code = 1. Because stop_name does not support easy localisation.

I completely agree that splitting out the platform code is good for localization. But it doesn't tell us what to put into the stop_name, that remains ambiguous. In fact the stop_name is redundant once you add the platform_code field to stops that have a parent station, hence the idea of using stop_name as a fallback for consumers that don't see the platform_code.

The other non-redundant alternative is to say the stop_name of a platform is its platform_code, and the encompassing station name is taken from the parent station's stop_name. This seems more clean and normalized to me, but I'm guessing it's not going to fly because Google already uses platform_code.

skinkie:
We should quit supporting so called "backwards support" for stuff that isn't following up with new standards.  If there are producers that are using non-supported software, they should produce "legacy GTFS" instead of forcing up to date producers and clients with stupid choices. The parent_station could have a complete different name than the stop, of which I also have Dutch examples.

abyrd:
> We should quit supporting so called "backwards support" for stuff that isn't following up with new standards. If there are producers that are using non-supported software, they should produce "legacy GTFS" instead of forcing up to date producers and clients with stupid choices.

I follow your thinking. I've also thought about the advantages of just abandoning backward compatibility in GTFS, and expecting consumer apps to be updated. However:
1. So far it's been a basic rule of GTFS to be backward compatible, so the decision to allow new versions of the format with breaking changes would have to be accepted first. Currently, there is no real concept of "new standards" in GTFS.
2. Updating consumers has a cost, which many organizations haven't budgeted for and don't know how to do (or are not even allowed to do due to licensing) on their own. They bought something from a supplier based on a spec that claimed it would remain backward compatible. 

GTFS was intentionally made to tolerate source data of varying quality, compensating with extra work by consumers, as a way of encouraging the adoption of open data practices. Maybe that means we should be more willing to break consumers and expect them to adapt.

> The parent_station could have a complete different name than the stop, of which I also have Dutch examples.

Please provide the examples, they would be decisive in convincing everyone. Especially if you have a case where the parent station has one name, but the child station has both a different name _and_ a platform code.

harringtonp:
> > > I would be against that.
> > 
> > 
> > Can you give your line of reasoning?
> 
> The idea for introduction of _platform_code_ was to remove: "Centraal Station perron 1" to make it "Centraal Station" and have platform_code = 1. Because stop_name does not support easy localisation.

Actually you guys have a good point. Yourstop does not yet support platform_code but if it did (and I expect it will) then there is always the questions of whether or not it should be appended to the stop_name field when showing matching stops to users during search. If stop_name includes the platform code and you append it you are showing it twice. If stop_name doesn't include the platform_code and you don't append it, then you will have a lot of matches with the same stop_name field.

A parallel example of how things can get messy with the standard is in MTA maryland where the headsign includes the route and to display both at the same time you need the following in code

```
      var routeName = route.route_name;
      var headsign = route.head_sign;
      if (headsign.indexOf(routeName) == 0) {
        // MTA maryland fix
        headsign = headsign.substring(routeName.length).trim();
      }
```

TomGoBravo:
Backwards compatibility is costly but I think giving implementers confidence that their system will continue working for a long time has been critical to the growth and success of GTFS.

We've certainly had problems with mapping route name and headsign fields to the UI in the past so lets try to be clear. How about this:
 - The stops.txt column stop_name must contain a name that makes sense within the region of GTFS coverage. For location_type=1 this identifies a station. For location_type=0 or unset this may be a station name OR producer formatted station_name + platform_code. Either provide backwards compatibility. I lean towards recommending producer formatted station_name + platform_code so the stop_name uniquely identifies the row of the stops.txt file.
- producer formatted station_name + platform_code uses the feed's local language. A translation may appear in https://developers.google.com/transit/gtfs/reference/gtfs-extensions#translations.
- Rows of stops.txt with parent_station set may include a platform_code. Consumers may ignore the stop_name of these rows and display platform_code in the context of the parent's stop_name.
- When formatting a list of stops made by a trip use the parent stop_name values. These should match announcements and signs telling riders when to disembark. 
- The parent stop_name is enough to get riders to the right station, where they can look for other attributes of the trip on physical departure boards to find the right platform within the station.
- platform_code is a short identifier for a specific track/platform/voie/spoor/stand/bay/via/... within a station. It is generally enough to get riders to the correct platform but a savvy rider might check the departure board to confirm that it hasn't been changed.

@abyrd : Report from my mother as she went through the domestic part of Brussels Midi last weekend: signs are Dutch+French and many track numbers appear without any adjacent word. :-)

There isn't a perfect answer to how consumers format the parent stop_name and platform_code.... It look like Google always translates "Platform" to the UI language, even for buses. See for example [this trip](https://www.google.com/maps/dir/42.3870657,-71.0048651/42.376094,-71.0358616/@42.380908,-71.0286933,15z/am=t/data=!3m1!4b1!4m3!4m2!3e3!5i1!5m1!1e2) at [MBTA stop 5880](http://transitfeeds.com/p/mbta/64/latest/stop/5880). Append `hl=fr` to the Google URL to see French `6 min (13 arrêts) · Voie C · Numéro d’arrêt : 5880`.

LeoFrachet:
So, **different things are being discussed**:
- Should we keep backward compatibility for future improvement of GTFS, or should we allow breaking changes
- Should we change the definition of `stop_name` to require the official name instead of allowing the vernacular name?
- If we put the platform code in `platform_code`, what should be in the `stop_name`:
  - the stop name alone
  - the stop name + the platform code for backward compatibility
- If we don't add a new field, should we put directly the platform code in the `stop_name` of the stop, and have the station name in the `stop_name` of its parent station?

I think **we should put the platform name in a distinct field**, otherwise I see no way for the data consumers to know what they should do with the `stop_name` field.

Regarding "track" vs "platform" vs "bay", if you think there is a need for it, we can defined it. For example with an ENUM field saying which one we speak about. Track vs platform seem obvious to define. platform vs bay is more mode related. I personally think **we can start by moving on without specifying track vs platform**. But if you think we should, it's doable.

Regarding the over conversations (changing the definition of `stop_name`, allowing breaking changes in the spec), maybe we could leave it aside for now (or open another conversation).

Some agencies are already producing `platform_code` without official definition (only a Google-specific one), adding it to the core spec would help normalize the GTFS datasets and would allow consumer to play with the same rules.

harringtonp:
In the interest of progress, simplicity and backwards compatibility my 2 pence worth would be to just run with the initial proposal which is being used by Google.

If additionally the stop_name field is changed slightly from

"The stop_name field contains the name of a stop, station, or station entrance. Please use a name that people will understand in the local and tourist vernacular."

to something like

"The stop_name field contains the name of a stop, station, or station entrance. Please use a name that is representative of where the stop is located and which people will understand in the local and tourist vernacular."

it may be enough (if read carefully by producers) to avoid the likes of 

stop_id,stop_code,stop_name,stop_lat,stop_lon,stop_timezone,location_type,parent_station,platform_code
0600CO210,CHEPWJG,"Bus Station (Stand 1)",53.163761,-2.211721,,0,060GCO124,1
0600CR287,CHEPWJD,"Bus Station (Stand B)",53.068733,-2.520081,,0,060GCR195,B
0600CR289,CHEPTPT,"Bus Station (Stand D)",53.068779,-2.520126,,0,060GCR195,D

LeoFrachet:
So, the conversation faded out since last week. I don't know if it's because we're reaching a consensus or if it's because people disagree.

To see which one it is, I'm tempted to open to vote on this proposal, which currently only included `platform_code`. I still think that the conversation should go on for the definition of `stop_name` (and also on the question "should we say if it's a platform or a track"), but maybe we can move step by step.

So I'm opening the vote. It will be closed on Friday March 22nd, at 23:59:59 UTC.

skinkie:
+1 I favor for the platform_code change.

Stefan de Konink
Stichting OpenGeo, Bliksem Labs B.V.
Bliksem Integration, KV4J, NetexConv2, rrrr
Producer, Consumer

harringtonp:
+1 from me too

paulswartz:
+1

aababilov:
+1 from Google

TomGoBravo:
I don't want to slow this down but.... as written consumers could display the platform code with the stop_name or if parent_station is set the parent stop_name. It'd be really nice to not have readers of the spec guessing how to display the new field. On the other hand that ambiguity makes it easier to produce valid data.

I looked for data mentioned in this PR at http://transitfeeds.com. I couldn't find any platform_code columns in http://transitfeeds.com/l/177-united-kingdom or http://transitfeeds.com/l/49-toronto-on-canada. I put stops.txt from providers mbta and ov in https://docs.google.com/spreadsheets/d/1E8rplgWe39VjAxPtaYMMu_68RjGfLztJ5K0zTI2GmQ0/edit

I had a theory that we can add the constraint that when platform_code not blank that location_type=0 and parent_station is not blank. mbta does conform to the additional constraint. ov is less consistent. Of the 3329 lines with platform_code set, all have location_type=0 but 275 have parent_station blank.

On the question of what to put in stop_name when platform_code is set: mbta seems to consistently put something that looks like the a parent station name, I didn't check it was matched the parent_id stop_name. mbta sometimes puts the platform_code in station_desc. In the ov stops.txt most platform_code non-blanks have stop_name just the station name or various formats of station name and platform_code. A few seem to have contradictory information such as stop_name="Spijkenisse, Spijkenisse Centrum perron D", platform_code=E which is pretty bad. If consumers display different information then errors like this could easily cause lots of rider frustration before getting fixed everywhere. Lack of a requirement that platform_code appear in stop_name makes it hard to add a validation check.

What is the long term benefit of allowing platform_code non-blank with location_type not 0? What about platform_code non-blank with parent_station blank? stop_name that doesn't contain platform_code? How hard is it for ov to produce more consistent data? mbta to put platform_code in stop_name?

skinkie:
@TomGoBravo would you rather have a property in feed_info.txt to tell what to expect?

TomGoBravo:
I want consumers to have an unambiguous way to display platform_code. I'd like the existing producers to modify their data so platform_code can be always displayed with parent_station stop_name. A feed_info property that removes the ambiguity might be easier for producers but creates a more complicated spec. I think ov has some data cleaning to do anyway so I don't feel very bad asking them to modify their data.

flocsy:
+1

aababilov:
+1 from Google.

Please feel free to merge this pull request.

LeoFrachet:
5 in favor (Stefan, Paul Harrington, Paul Swartz, Alexej & Gavriel)
0 against

We already have producers and consumers.

The proposal is therefore adopted.

LeoFrachet:
@aababilov: Since I resolved the conflict, the CLA bumped out. Could you re-force them.?

aababilov:
@LeoFrachet Done!

LeoFrachet:
Thanks!

harringtonp:
Been looking at this implementation wise the past week and now whenever I find a platform code and a valid parent station name I replace the stop name with these 2 items and a configurable word such as "Platform", "Stand" in between.  In cases where it kicks in, it would seem to improve the stop names and make UI usage easier

PR #151, opened 2019-03-15
Implement upgrade_translations.py

aababilov:
This script upgrades GTFS data from Google translations extension [1] to GTFS-Translations [2].

[1] http://developers.google.com/transit/gtfs/reference/gtfs-extensions#translationstxt
[2] http://bit.ly/gtfs-translations


davidlattimore:
LGTM

gcamp:
Not sure I understand why this is in the GTFS specs repo? It's also a script for something that's not part of the spec yet?

barbeau:
@gcamp I agree - these should probably be in https://github.com/google/transitfeed instead.

gcamp:
@barbeau Created an issue on this https://github.com/google/transit/issues/157

PR #152, opened 2019-03-18
Add a minimal .gitignore

aababilov:


skinkie:
+1

barbeau:
+1

Do we want to open to including other file types in the .gitignore?  I don't know what other editors people use, but I use IntelliJ, which has it's own files that could be added here.

aababilov:
We can add more files in the future. If you have any pattern to add, I will be glad to merge a pull request.

PR #155, opened 2019-03-22
Allow to choose particular stations for generating a pathway graph

aababilov:
The feed may be too large to fit on a single graph.


aababilov:
See https://github.com/google/transitfeed/pull/472

PR #156, opened 2019-03-22
Use doubleoctagon for entrances and hexagon for generic nodes

aababilov:
The new shapes are more compact and look better on the graph.


aababilov:
This is the picture for our test feed:
au-sydney-entrances

sharma-shail:
+1

PR #162, opened 2019-04-09
Fix typo: was using "pickup" instead of "drop off" and vice-versa

iliasbhal:
I think there is a typo. 

Screen Shot 2019-04-09 at 10 27 52 AM


aababilov:
Thank you for that change.
Please fix another typo as well. Vice versa cannot be spelled as "vis-versa". It sounds as /'wike 'wersa:/ in Latin, you can use this as a hint.

iliasbhal:
@aababilov done :)

aababilov:
Thanks! LGTM

barbeau:
Thanks @iliasbhal!

PR #163, opened 2019-04-09
Clarify TripDescriptor as referencing a single trip instance

barbeau:
In the past, it was unclear whether TripDescriptor could reference more than one trip at a time when used within Alert EntitySelector.

This change clarifies TripDescriptor as follows:
* TripDescriptor always refers to a single trip instances - it cannot be used to identify sets of trips. If a TripDescriptor doesn't resolve to a single trip instance, it's an error and consumers can drop the data.
* List examples of known cases where additional fields other than trip_id are required to resolve TripDescriptor to a single trip instance
* TripDescriptor.route_id cannot be used within an Alert EntitySelector - use EntitySelector.route_id instead
* ScheduleRelationship field is ignored by consumers if set within an EntitySelector.TripDescriptor

Basic descriptions for the fields in Alert EntitySelector are also added.

A big motivator for these clarifications is to make it easier to distinguish data errors - in some cases it hasn't been clear if data was missing or if agencies intended to reference multiple trips.

Announced on the GTFS-realtime Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/ZlqUyM9hiV4.

**EDIT** Following discussion, this proposal also adds EntitySelector.direction_id as an experimental field to avoid losing the ability to reference all trips in one direction for a route.


gtfs-realtime/spec/en/reference.md
```diff
@@ -351,18 +351,31 @@ A geographic position of a vehicle.
...
-| **route_id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | The route_id from the GTFS that this selector refers to. If trip_id is omitted, route_id must be provided. |
+| **route_id** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | The route_id from the GTFS that this selector refers to. If trip_id is omitted, route_id, direction, start_time, and schedule_relationship=SCHEDULED must all be set to identify a trip instance. TripDescriptor.route_id should not be used within an Alert EntitySelector to specify a route-wide alert that affects all trips for a route - use EntitySelector.route_id instead. |
```
skinkie: Should this be direction **_id** ?

gcamp:
With these changes, how would you have an alert affect a route on a single direction? Listing all trip_id on a direction doesn't seems realistic. 

That said, I agree with those changes, but we've seen producers not be compliant with these changes because they wanted to assign a direction to alerts.

gtfs-realtime/spec/en/reference.md
barbeau: @skinkie Thanks, good catch!  I just fixed this in https://github.com/google/transit/pull/163/commits/95b2192d8f94d4b9b7d500b7864ac82636666402

barbeau:
@gcamp Yes, that's a good point - with these changes it wouldn't allow a simple way to specify all trips on a route in a single direction (other than enumerating each trip_id individually in its own EntitySelector).  One way to allow this would be to add a `direction_id` field directly to `EntitySelector` that would be used in combination with `route_id`.

I should also mention that a big motivator for these clarifications it to make it easier to distinguish data errors as well - in some cases in the past it wasn't clear if data was missing or if agencies intended to reference multiple trips.  I'll update my original post with this info too.

skinkie:
Is there a markdown reason I don't know about that you enter double space after an end of sence?

barbeau:
@skinkie Nope, that's just an old habit that's hard to kill.  From a quick look it looks like single space after period is more common throughout the spec, so I'll change this PR to use single space.

paulswartz:
@barbeau 
> One way to allow this would be to add a `direction_id` field directly to `EntitySelector` that would be used in combination with `route_id`.

We do this in a JSON version of our alerts feed: we'd love it to be part of the full specification!

barbeau:
@paulswartz I've added `EntitySelector.direction_id` as an experimental field in https://github.com/google/transit/pull/163/commits/27df975511d3478bd232133a4ff10b737adede66 to address this use case.

@gcamp Would Transit App be a consumer for this?

skinkie:
I would like to point out that above direction_id change, while a potential improvement, is not backwards compatible.

barbeau:
@skinkie How is it not backwards compatible?

skinkie:
Because all current implementations would be using a _partial_ tripDescriptor.

barbeau:
@skinkie Well, yes - if a current producer is using EntitySelector.trip.direction_id to reference multiple trip instances, it would be considered out-of-spec following this proposal.  This isn't out of line with other GTFS-rt clarifications in the past, though, as we typically have the case where X producers/consumers are doing it one way and Y are doing it another, and we have to decide which way to move forward.  Someone is going to be out of compliance following the clarification, and then everyone can catch up to the agreed-upon "best" way of doing things going forward.

gcamp:
@barbeau sure we can support that `direction_id` quickly.

barbeau:

Vote will be closed on Thursday May 2nd at 23:59:59 UTC.

gcamp:
+1

paulswartz:
+1

jrsanbornjr:
+1

laurentg:
Just to make sure, the vote is about the "single trip" proposal, not adding `direction_id` selector ?

skinkie:
> Just to make sure, the vote is about the "single trip" proposal, not adding `direction_id` selector ?

I think this change must be holistic otherwise we cripple the publication?

laurentg:
Sorry about my last comment, my mistake, I guess the vote is about all the changes.
About adding `direction_id` in `EntitySelector`: our (Mecatran) Urbiplan alert editor allow the selection of a route direction, so we are interested in seeing this field added. For the record, we can propose a GTFS-RT service alert producer with this extension rather soon if this spec is accepted.

laurentg:
+1

barbeau:
Correct, this vote is in the entire content of the PR, which is single trip selection plus adding EntitySelector. direction_id. This way we don't lose functionality of being able to specify all trips in one direction for a route.

skinkie:
+1

JymDyer:
+1

harringtonp:
+1

In yourstop when implementing alerts, only the trip id from the descriptor is pulled out. This was partially because frequency based trips are not supported but also because of the complexity of the descriptor and overlap with the route field in entity selector. 

The changes here read better and are more understandable then the original wording. And as in other implementations direction_id in the entity selector could be processed with little change.

gtfs-realtime/spec/en/reference.md
```diff
@@ -393,15 +406,18 @@ Identification information for the vehicle performing the trip.
...
+| **route_type** | [int32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | The route_type from the GTFS that this selector refers to.
+| **direction_id** | [uint32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | The direction_id from the GTFS feed trips.txt file, used to select all trips in one direction for a route, specified by route_id. If direction_id is provided, route_id must also be provided. <br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future.<br>|
+| **trip** | [TripDescriptor](#message-tripdescriptor) | Conditionally required | One | The trip instance from the GTFS that this selector refers to. This TripDescriptor must resolve to a single trip instance in the GTFS data (e.g., a producer cannot provide only a trip_id for exact_times=0 trips). If the ScheduleRelationship field is populated within this TripDescriptor it will be ignored by consumers when attempting to identify the GTFS trip.
```
tleboulenge: (e.g., a producer cannot provide only a trip_id for **exact_times=0** trips)

I think this is true for all trips defined in frequencies.txt, not only exact_times=0. A start_time must always be provided (as stated above)

tleboulenge:
+1

gtfs-realtime/spec/en/reference.md
barbeau: @tleboulenge Yes, that is correct, all trips defined in frequencies.txt must include a `start_time` to be valid.  The example I gave here wasn't intended to be exclusionary (i.e., wasn't intended to imply that `exact_times=1` trips don't require `start_time`) - if you think this should be clarified in the text let me know and we can restart the vote.

tleboulenge: I see my comment as just a detail, akin to a typo, wouldn't require restarting the vote for this. I'll leave it up to you to decide to go ahead as it is, or update that wording.

barbeau:

Yes - 8

So it passes! Thanks all!

PR #164, opened 2019-04-19
Add accessibility issue alert effect

gcamp:
I've seen that a big amount of alert available in GTFS-rt alerts feed are to give the status of elevator that are out of service. However, there was no Effect available in the enum to mark them as such. 

This adds that value in the Effect enum.

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -416,7 +416,7 @@ message Alert {
     UNKNOWN_EFFECT = 8;
     STOP_MOVED = 9;
     NO_EFFECT = 10;
-    OUT_OF_SERVICE_ELEVATOR = 11;
+    ACCESSBILITY_ISSUE = 11;
   }
   optional Effect effect = 7 [default = UNKNOWN_EFFECT];
 
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -416,6 +416,7 @@ message Alert {
     UNKNOWN_EFFECT = 8;
     STOP_MOVED = 9;
     NO_EFFECT = 10;
+    OUT_OF_SERVICE_ELEVATOR = 11;
   }
   optional Effect effect = 7 [default = UNKNOWN_EFFECT];
 
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -416,7 +416,7 @@ message Alert {
     UNKNOWN_EFFECT = 8;
     STOP_MOVED = 9;
     NO_EFFECT = 10;
-    ACCESSBILITY_ISSUE = 11;
+    ACCESSIBILITY_ISSUE = 11;
   }
   optional Effect effect = 7 [default = UNKNOWN_EFFECT];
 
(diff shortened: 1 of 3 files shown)


paulswartz:
What about escalators, movable ramps, or other accessibility issues? Maybe ACCESS_ISSUE?

gcamp:
@paulswartz I'm open to that, but we would need to specify more clearly what that means. Is an out of service escalator really an access issue? An out of service elevator is clearly an access issue on the other hand.

paulswartz:
That may be technically true (since you can't take a wheelchair up/down an escalator), but we group escalator and elevator alerts together under Access: https://www.mbta.com/alerts/access

jakluk:
How should this effect be interpreted, only as an advisory or should transit planners take it into account when searching for an accessible path? Sometimes only one of more elevators may be out of service.

gcamp:
@jakluk I don't think this should affect trip planning because there's clearly missing information to interpret those. Also, this was never defined for other effects, but I know there's plan to clarify this soon (see https://github.com/google/transit/pull/137#issuecomment-457255246). I would say that clarifying the behaviour is outside of this PR scope. 

Longer term, I can see EntitySelector to include a `pathway_id` from pathways.txt to identify exactly which elevator is out of service, at this moment the trip planners could take that info account. But I would also say this is outside of the scope of this PR.

gcamp:
@paulswartz the more I'm thinking about it, I think it makes sense to not have a general ACCESS_ISSUE type. Consumer will probably want to have different behaviour depending on the type of access issue. If you think that would be better I can add more types in this PR (`OUT_OF_SERVICE_ESCALATOR`)

ritesh-warade-ibigroup:
We are in favor of adding either one generic effect named ACCESS_ISSUE or multiple specific effects like OUT_OF_SERVICE_ELEVATOR, OUT_OF_SERVICE_ESCALATOR, etc.

paulswartz:
@gcamp do you have a sense of how the additional type would be used by consumers? Thinking of our applications, we have an additional field `facility_id` which represents the elevator or escalator that's affected. I'm not sure what having an explicit `OUT_OF_SERVICE_ELEVATOR` type provides in addition to that.

gcamp:
For our purpose, we have an option to show wheelchair accessible information. In this case, I would show more prominently elevators alerts, but not escalators.

gcamp:
After discussion with @paulswartz, I updated the proposition to change out of service elevator to `ACCESSBILITY_ISSUE` that applies only for wheelchair users (or any users that would need step-free access), but that encompass more than just elevators.

gcamp:

Vote will be closed on Tuesday May 7nd at 23:59:59 UTC.

laurentg:
Does `ACCESSIBILITY_ISSUE` implies that all wheelchair accessible stops referred by this alert become non-wheelchair accessible?

gcamp:
@laurentg alerts so far didn't have any change of behaviour in the consumer application per the official spec. This PR doesn't change anything about this, see https://github.com/google/transit/pull/164#issuecomment-485132005

paulswartz:
+1

jrsanbornjr:
+1

gcamp:
@ritesh-warade-ibigroup @ibi-group-team vote is about to expire without enough votes, would you be able to vote?

kurtraschke:
+1

skinkie:
+1 for the accessibility_issue, as producer in NL, we will add this.

laurentg:
+1

ibi-group-team:
+1

harringtonp:
+1

barbeau:
Hi all! I opened a new issue with a question of how Alerts with effect of ACCESSIBILITY_ISSUE are being used:
https://github.com/google/transit/issues/209

Please comment in that thread! Thanks!

PR #165, opened 2019-06-10
Arrival and departure time best practices

antrim:
This is one of 5 PRs that incorporates a recommendation from the [GTFS best practices](https://gtfs.org/best-practices/).

This PR recommends including estimated arrival and departure times where possible.

Related pull requests:
- [Add best practice: stop_id is not rider facing](https://github.com/google/transit/pull/169)
- [Add best practice: In seat transfer specifics](https://github.com/google/transit/pull/168)
- [Headsign best practices](https://github.com/google/transit/pull/167)
- [Add best practices: Headsign and route name examples](https://github.com/google/transit/pull/166)

PR #166, opened 2019-06-10
Add best practices: Headsign and route name examples

antrim:
This is one of 5 PRs that incorporates a recommendation from the GTFS best practices.

This PR adds examples detailing proper headsign and route name conventions. Examples are illustrations that add meaning and clarity to the Spec.

Related pull requests:
- [Add best practice: stop_id is not rider facing](https://github.com/google/transit/pull/169)
- [Add best practice: In seat transfer specifics](https://github.com/google/transit/pull/168)

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -146,14 +146,25 @@ File: **Required**
 |  ------ | ------ | ------ | ------ |
 |  `route_id` | ID | **Required** | Identifies a route. |
 |  `agency_id` | ID referencing `agency.agency_id` | **Conditionally required** | Agency for the specified route. This field is required when the dataset provides data for routes from more than one agency in [agency.txt](#agency), otherwise it is optional.  |
-|  `route_short_name` | Text | **Conditionally required** | Short name of a route. This will often be a short, abstract identifier like "32", "100X", or "Green" that riders use to identify a route, but which doesn't give any indication of what places the route serves. Either `route_short_name` or `route_long_name` must be specified, or potentially both if appropriate. |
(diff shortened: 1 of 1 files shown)


antrim:

abyrd:
+1

skinkie:
@antrim I would like to see a rail example. At this moment it seems to be common practice (at least on Google Maps) that the route_short_name is also used as product formula. Hence "route_short_name" = "Slow Train"  or "Intercity". I would like you to come up with these examples as well, or make it a poor practice.

abyrd:
@skinkie I have seen that before, and have always found it confusing to have multiple routes with the same route_short_name. This seems to be misusing the route_short_name as some kind of service type field, so I see it as a practice to avoid.

skinkie:
@abyrd again a good reason to detail how this should be denoted and have consumers to implement that.

TomGoBravo:
+1
I agree it'd be nice to have an example where route_short_name is not sethttp://transitfeeds.com/p/sncf/1069/latest/routes has a bunch of examples. And maybe note the existence of trip_short_name.

mgilligan:
+1

paulswartz:
+1

antrim:
@TomGoBravo:
> I agree it'd be nice to have an example where route_short_name is not set.

Right, and there are cases where only the route_long_name should be set.

In response to this feedback, I am going to cancel this vote to add more examples. Then we'll return to discussion and vote.

harringtonp:
Typo

" If waypoint(s) are removed from the headsign show to passengers after the vehicle passes those waypoints"

to

 "If waypoint(s) are removed from the headsign show**n** to passengers after the vehicle passes those waypoints"

Trivial I know but caused me to read it twice to understand

PR #167, opened 2019-06-10
Headsign best practices

antrim:
This is one of 5 PRs that incorporates a recommendation from the GTFS best practices.

This PR adds explains that stop_headsign does not carry down.

Related pull requests:
- [Add best practice: stop_id is not rider facing](https://github.com/google/transit/pull/169)
- [Add best practice: In seat transfer specifics](https://github.com/google/transit/pull/168)
- [Add best practices: Headsign and route name examples](https://github.com/google/transit/pull/166)
- [Arrival and departure time best practices](https://github.com/google/transit/pull/165)

PR #168, opened 2019-06-10
Add best practice: In seat transfer specifics

antrim:
This is one of 5 PRs that incorporates a recommendation from the [GTFS best practices](https://gtfs.org/best-practices/).

This PR adds details about properly implementing in-seat transfers. 

Related pull requests:
- [Add best practice: stop_id is not rider facing](https://github.com/google/transit/pull/169)
- [Headsign best practices](https://github.com/google/transit/pull/167)


MuckT:
https://github.com/google/transit/pull/168#issuecomment-500629619 @paulswartz

antrim:

abyrd:
It looks like this PR only adds the text "If in-seat (block) transfers are allowed between trips, then the last stop of the arriving trip must be the same as the first stop of the departing trip." It seems strange to me that this detail is the first thing mentioned for the transfer_type field, before even the values the field can take on. No new values are defined, so it's not clear how this text relates to transfer_type. There has been discussion in the past about introducing an in-seat transfer_type, was the intent to add that value in this PR?

antrim:
@abyrd Good point. Since @MobilityData is planning to return to https://github.com/google/transit/pull/32 soon, I think you're right that we should leave this change aside for now and wait for the the changes around "transfer rules for routes and trips, and in-seat transfers". I'll close this.

skinkie:
@antrim Addionnaly the last stop is not always the first stop. For example an exit stop is the last stop in the sequence but does not pick up new passengers, but it is still considered the same block.

PR #169, opened 2019-06-10
Add best practice: stop_id is not rider facing

antrim:
This is one of 5 PRs that incorporates a recommendation from the [GTFS best practices](https://gtfs.org/best-practices/).

This PR explains that stop_id is not rider facing.

The 4 related PRs can are:
- [Add best practice: In seat transfer specifics](https://github.com/google/transit/pull/168)
- [Headsign best practices](https://github.com/google/transit/pull/167)


paulswartz:
Can we add this note about other IDs as wel? We periodically need to ask clients not to parse the `service_id` or `trip_id` values and it would be great to point them to the documentation.

barbeau:
I agree with @paulswartz that this should apply to all IDs.

Note that in the ["Field Types"](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#field-types) section, under the "ID" type we already say:

>An ID field value is not aimed to be displayed to the user, and is a sequence of any UTF-8 characters, but using only printable ASCII characters is recommended. 

`stop_id`, `service_id`, `trip_id`, and others are all categorized as ID field types, so this definition would apply to them.

This PR says:

>`stop_id` is an internal ID, not intended to be shown to riders.

So it's pretty close in language.

@antrim I think the Best Practice exited prior to the addition of the Field Types into GTFS?  If so, maybe the goal of this PR is already accomplished?  Or do you think we should call this out specifically in the description context of a few ID fields that are hotspots?

antrim:
@barbeau I favor adding this to the definition of the ID field type because it's more concise. Any different opinions?

barbeau:
@antrim So new sentence in ID definition would be:

>An ID field value is is an internal ID, not intended to be shown to riders, and is a sequence of any UTF-8 characters, but using only printable ASCII characters is recommended.

HeidiMG:
I've changed the wording to add the sentence Sean proposed and removed the now-redundant language that was in `stop_id`.

antrim:

abyrd:
+1

gtfs/spec/en/reference.md
```diff
@@ -123,7 +123,7 @@ File: **Required**
 
 |  Field Name | Type | Required | Description |
 |  ------ | ------ | ------ | ------ |
-|  `stop_id` | ID | **Required** | Identifies a stop, station, or station entrance. The term "station entrance" refers to both station entrances and station exits. Stops, stations or station entrances are collectively referred to as locations. Multiple routes may use the same stop. |
+|  `stop_id` | ID | **Required** | Identifies a stop, station, or station entrance. `stop_id` is an internal ID, not intended to be shown to riders. <br><br> The term "station entrance" refers to both station entrances and station exits. Stops, stations or station entrances are collectively referred to as locations. Multiple routes may use the same stop. |
```
skinkie: Is the <br><br> allowed in mark down? I would expect two spaces.

gtfs/spec/en/reference.md
```diff
@@ -48,7 +48,7 @@ This section defines terms that are used throughout this document.
...
-- **ID** - A sequence of any UTF-8 characters which uniquely identifies an entity, but does not necessarily identify a specific record in a table. IDs defined in one .txt file are often referenced in another .txt file. An ID field value is not aimed to be displayed to the user, and is a sequence of any UTF-8 characters, but using only printable ASCII characters is recommended. <br> *Example: The `stop_id` field in [stops.txt](#stopstxt) is a ID. The `stop_id` field in [stop_times.txt](#stop_timestxt) is an ID referencing `stops.stop_id`.*
+- **ID** - An ID field value is is an internal ID, not intended to be shown to riders, and is a sequence of any UTF-8 characters, but using only printable ASCII characters is recommended. IDs defined in one .txt file are often referenced in another .txt file. An ID field value is not aimed to be displayed to the user, and is a sequence of any UTF-8 characters, but using only printable ASCII characters is recommended. <br> *Example: The `stop_id` field in [stops.txt](#stopstxt) is a ID. The `stop_id` field in [stop_times.txt](#stop_timestxt) is an ID referencing `stops.stop_id`.*
```
TomGoBravo: This change introduces an almost duplicate sentence "An ID field value...not .... shown to riders... only printable..." and removes "uniquely identifies an entity, but does not necessarily identify a specific record in a table." which I'm guessing is not intentional.

gtfs/spec/en/reference.md
giocorti: The \<br>\<br> tag is allowed in markdown. It appears like <br><br> this.

ibi-group-team:
We agree with the intent of this change but also agree with @TomGoBravo that the wording isn't quite right. If the wording is updated then we would vote yes on this.

HeidiMG:
Duplicative ID field language flagged by @TomGoBravo and @ibi-group-team has been removed.

antrim:
Since the latest commit is a correction, not a change of meaning, I propose that we continue with the current voting timeline -- which isn't to the letter of the voting process but follows it closely.

The process risk I see here is that we are changing text after some have voted. We don't want to set that precedent. @abyrd : You are the only current "+1" vote. Do you maintain the +1 vote?

abyrd:
Yes, still +1.

TomGoBravo:
+1

gcamp:
+1

thomastrillium:
+1

paulswartz:
+1

ibi-group-team:
+1

antrim:
7 votes are in favor, including at least one consumer and producer. 0 votes against. The vote passes.

AlaaAhmed1993 (commented):
Hello, I work as a food delivery driver. I work on a program called talabat Rider
The application of the driver and I can't get requests. Is there a solution?

derhuerst:
> I work on a program called talabat Rider. The application of the driver and I can't get requests. Is there a solution?

Sorry, this is an entirely different project. It is about a data form for public transportation.

PR #170, opened 2019-07-01
Fix #94 - Clarify that GTFS-rt extension field numbers are permanent

### LINKED ISSUE #94: Clarify that .proto field numbers within GTFS-rt extensions are permanent
**Summary:** 

Currently, the GTFS-realtime extension process:
https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#extensions

...doesn't cover protocol buffer best practices.

Under the Protocol Buffers Language Guide:
https://developers.google.com/protocol-buffers/docs/proto

...in the "Assigning Field Numbers" section it says:

>Each field in the message definition has a unique number. These numbers are used to identify your fields in the message binary format, and should not be changed once your message type is in use.

We should add this text to the GTFS-realtime extensions section to make sure that producers don't break their custom extensions (even if the field is just defined for a proposal) when they are in use.

For example, if Agency X uses field number `1` for an extension for a GTFS-rt proposal, then it can never use `1` again in the future for another field, even if the proposal was accepted and the field `1` is no longer in use.

We should also add a link to the "Assigning Field Numbers" section in the GTFS-realtime extensions section too:
https://developers.google.com/protocol-buffers/docs/proto#assigning-field-numbers

...as there are other guidelines there for selecting field number values.

barbeau:
As discussed in https://github.com/google/transit/issues/94, some producers have mistakenly re-used field numbers in GTFS-realtime extensions - for example, when an extension was officially adopted, the producer would recycle that field number in their extension for a new extension field.

This proposal adds guidance that currently exists in the [Protocol Buffers Language Guide](https://developers.google.com/protocol-buffers/docs/proto) directly in the GTFS-realtime extension documentation in regards to not re-using extension field numbers to make it more obvious to producers.

Announced on the GTFS-realtime Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/BIh8jXEm_Vc.

### DIFF
--- a/gtfs-realtime/CHANGES.md
+++ b/gtfs-realtime/CHANGES.md
@@ -85,6 +85,20 @@ extend transit_realtime.TripDescriptor {
 }
 ```
 
+When creating extensions, developers should follow the [Protocol Buffers Language Guide](https://developers.google.com/protocol-buffers/docs/proto). A common mistake is re-using an extension field number. In the [Assigning Field Numbers section](https://developers.google.com/protocol-buffers/docs/proto#assigning-field-numbers), the Language Guide says:
+
+> Each field in the message definition has a unique number. These numbers are used to identify your fields in the message binary format, and should not be changed once your message type is in use.
+
+For example, in the first example `some_string` was assigned field number `1`. When the developer no longer wishes to use `some_string`, or when `some_string` has been adopted in the official GTFS Realtime spec and there is no need for the extension, the developer cannot re-use field number `1` for a new field. Instead, the developer should deprecate the field and use a new number for the new field:
+```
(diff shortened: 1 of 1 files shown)


abyrd:
Looks good.

barbeau:

Vote will be closed on Wednesday July 17th at 23:59:59 UTC.

paulswartz:
+1

gcamp:
+1

saraemarcus:
+1

barbeau:
The vote on this issue is closed with the following results:
* Yes - 3
* No - 0
* Abstain - 0

So it passes!  Thanks all!

PR #177, opened 2019-08-02
Added Transit extension id

juanborre:
We would like to add a new extension_id for internal usage.

### DIFF
--- a/gtfs-realtime/CHANGES.md
+++ b/gtfs-realtime/CHANGES.md
@@ -85,6 +85,20 @@ extend transit_realtime.TripDescriptor {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


barbeau:
@juanborre I agree that private extension IDs can be useful.  I'd like to alternately propose that we designate a range of private extension IDs that could be used by anyone (rather than allocating private IDs to specific organizations).

Google uses extension IDs >= 9000 for internal usage, and I'd like to propose that we designate this same range for private use by any organization.

One important item of note - messages using private extension IDs in this range *MUST* remain private.  If these messages are exchanged publicly with other organizations you run the risk of mixing data fields.

devadvance:
@barbeau I support a designated block of extension IDs for private use, similar to blocks like 10.0.0.0/8 in the IPv4 space. Is it worth encouraging parties to seek an interoperable, standardized field whenever possible, however, to offset the mixed data fields concern?

In any case, the outcome of this thread will be helpful guidance for similar things we're doing at Intersection.

abyrd:
This makes a lot of sense to me.

juanborre:
Both of your comments make a lot of sense. 

I will open a new PR with the clean git history to reflect that: https://github.com/google/transit/pull/178

PR #178, opened 2019-08-06
Reserve extension IDs range 9000-9999 for internal use by any organization

juanborre:
This is a clean Pull Request coming from: https://github.com/google/transit/pull/177

Reserve the range of extension IDs ~~>= 9000~~ 9000-9999 for internal use. 

Instead of asking for an extension ID, producers can use that range to convey information internally to their organization or for testing purposes.

We encourage asking for an official extension ID if the producers need to publish that data out of their organization.

### DIFF
--- a/gtfs-realtime/CHANGES.md
+++ b/gtfs-realtime/CHANGES.md
@@ -68,6 +68,8 @@ To allow producers to add custom information to a GTFS Realtime feed, we will ta
 
 When possible we should avoid extensions and add fields useful to many agencies to the main specification to avoid fragmentation and extra work for consumers to support various extensions to the spec.  Before requesting an extension id, producers should propose adding the field to the specification (see [Adding new fields to GTFS Realtime](#adding-new-fields-to-gtfs-realtime))
 
+The extension IDs >= 9000 are reserved for private use by GTFS-rt producers. These IDs should only be used to convey information internally to your organization. If you need to publish information in the GTFS-rt you produce, it is encouraged to ask for a new extension ID (see below). 
+
 To create a new extension, we will assign a producer the next available extension id, picked incrementally from a list of numbers starting at 1000 and going up and documented in the Extension Registry section found below.
 
 These assigned extension ids corresponds to the tag ids available in the "extension" namespace for each GTFS Realtime message definition. Now that the developer has an assigned extension id, they will use that id when extending any and all GTFS Realtime messages. Even if the developer only plans to extend a single message, the assigned extension id will be reserved for ALL messages.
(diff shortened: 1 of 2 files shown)


gtfs-realtime/CHANGES.md
```diff
@@ -68,6 +68,8 @@ To allow producers to add custom information to a GTFS Realtime feed, we will ta
 
 When possible we should avoid extensions and add fields useful to many agencies to the main specification to avoid fragmentation and extra work for consumers to support various extensions to the spec.  Before requesting an extension id, producers should propose adding the field to the specification (see [Adding new fields to GTFS Realtime](#adding-new-fields-to-gtfs-realtime))
 
+The extension IDs >= 9000 are reserved for private use by GTFS-rt producers. These IDs should only be used to convey information internally to your organization. If you need to publish information in the GTFS-rt you produce, it is encouraged to ask for a new extension ID (see below). 
+
```
barbeau: Could you please add an entry to the extension registry table at the bottom of the page? 

>|>= 9000|RESERVED - INTERNAL USE ONLY|[GTFS community](https://groups.google.com/forum/#!forum/gtfs-realtime)|[Group discussion](https://github.com/google/transit/pull/178/)|

juanborre: Sure, it is done 👍

LeoFrachet:
Looks good.

@barbeau Technical question: are we limited to 4 digits by the system, or by convention? Should we say 9000-9999 instead of >= 9000 of future-proof-ness?

skinkie:
Can we add something that an extension > 9000 in a public feed, invalidates the feed for public usage?

barbeau:
I agree with @skinkie that we should also explicitly say that these messages cannot be used in public feeds.

So after:

>The extension IDs >= 9000 are reserved for private use by GTFS-rt producers. These IDs should only be used to convey information internally to your organization.

...I suggest adding:

>Extensions in this range **must not** be used in public feeds.

barbeau:
@LeoFrachet Yes, having an open range is bothering me too, as extension IDs aren't limited to 4 digits.  In fact, looking more at https://developers.google.com/protocol-buffers/docs/proto#choosing-extension-numbers, we can't use overlap 19000 though 19999 because that is reserved for the Protocol Buffers implementation.

So, yes, I would prefer to restrict this private extension ID range from 9000-9999.

@tleboulenge would the range of 9000-9999 as private extension IDs for GTFS-realtime work for Google?

juanborre:
I have updated the PR to reflect the last changes suggested.

Waiting for @tleboulenge to confirm the range works for Google.

juanborre:
It's been 9 days since the last comment and we don't have an answer yet. 

Is there anyone else capable of providing that information? @LeoFrachet @barbeau ?

tleboulenge:
Sorry for the delay, yes the range works for us.

juanborre:
Excellent. Thanks everybody for the feedback and the amendments.

As there seems to be consensus and the PR has been open for more than a week, per the Official Process I'm calling for a vote.

Vote will be closed on Thursday, August 29th at 23:59:59 UTC.

barbeau:
+1

juanborre:
While working on something else in the spec, I found out that we need to define the extension range in the main `gtfs-realtime.proto`, like this:

```
  // The following extension IDs are reserved for private use by any organization.
  extensions 9000 to 9999;
```

I added that in the last commit.

As it does not change the essence of this PR, the vote is still ongoing.

tleboulenge:
+1

abyrd:
+1

juanborre:
Voting is closed on this proposal and here are the results:

- YES: 3
- NO: 0

So it passes! Thanks everybody!

Issue #138, opened 2019-01-29
Attributions in GTFS


LeoFrachet:
Hi GTFS community!

Some data producers told us they'd like to provide attribution of the data, and that the existing fields `feed_publisher_url` and `agency_name` weren't enough, because other stakeholders were implied... and were asking to be given credit.

Given the broad variety of such stakeholders (French "Autorité Organisatrice", German "Verbund", the SF Bay Area MTC which is "planning, financing and coordinating", aggregators like Danish Rejseplannen...), we ended up drafting a proposal allowing any kind of attribution (aka free text).

The draft is here: [**GTFS-Attributions** bit.ly/gtfs-attributions](http://bit.ly/gtfs-attributions)

Since this question was closely knitted with the question of translation in GTFS, we also drafted a proposal for translations in GTFS. It is inspired of the Google's one, but stricter and improved.

The draft is there: [**GTFS-Translations** bit.ly/gtfs-translations](http://bit.ly/gtfs-translations)

As usual, I'm posting it as Issues to gather your impressions, and if we get closer to a consensus I'll move it as proposal.

Thanks!
Leo

LeoFrachet:
Based on community feedback, I've updated the [GTFS-Attributions bit.ly/gtfs-attributions](GTFS-Attributions bit.ly/gtfs-attributions) proposal to remove the bit-encoded values and to replace them by distinct fields.

abyrd:
I definitely agree that we shouldn't have positional bit fields in GTFS. Multiple boolean fields makes more sense. However the granularity of the permissions seems somewhat arbitrary. Why is there one field for "network data (files agency.txt, routes.txt and entities with location_type=0 in stops.txt)", but one completely separate field for attribution of `feed_info.txt`? Might it make more sense to allow the name of any txt file as a column name, so `routes`, `agency`, `feed_info`, etc. are the columns, with 1s in the columns where the attribution applies? It also seems convenient to have some way to specify that attribution applies to the whole contents of the feed without having to exhaustively enable attribution on each file. I imagine a common use case will be a single attribution record applying to the entire feed.

LeoFrachet:
@abyrd, you're suggesting two things.

1) Remove the category, and just allow to specify table names (and field names?)

My intention was to make it both simpler (by grouping fields and files). Allowing only file name wouldn't cover some edge cases like, for example, enhancing a dataset by adding pathways information in stops.txt (which doesn't affect the stops themselves, only the other entities in the file).

[...]

PR #192, opened 2019-12-02
GTFS-Attributions (core only)

timMillet:
### Context

Some data producers would like to provide more data attributions. The existing fields `feed_publisher_url` and `agency_name` aren't enough, because many other stakeholders may be implied and they should be given credit.

Given the broad variety of such stakeholders (French "Autorité Organisatrice", German "Verbund", the SF Bay Area MTC which is "planning, financing and coordinating", aggregators like Danish Rejseplannen...), the proposal allows to name them freely and to classify them as necessary either as producers, operators, or authorities.

The Danish data producer Rejseplanen is already producing the file attributions.txt according to this proposal. Datasets are available [here (new version)](http://www.rejseplanen.info/labs/GTFS_New.zip) or [here (old version)](http://www.rejseplanen.info/labs/GTFS_Old.zip) — both versions include attributions.txt.

Google is also consuming the attributions.txt file.

### Pull request

This PR is related to the issue [Attributions in GTFS](https://github.com/google/transit/issues/138) and concerns the proposal [GTFS-Attributions (core only)](http://bit.ly/gtfs-attributions-core) according to the current (2019-12-02T18:08:00-05:00) state of the Google Doc.

This PR suggests the addition of:
- attributions.txt (new file)

Compared to [GTFS-Attributions full version]( http://bit.ly/gtfs-attributions), this PR _does not_ include:
- attribution_scopes.txt (new file)
- translations.txt (extended file - still at the proposal step)

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -82,6 +82,7 @@ This specification defines the following files:
 |  [pathways.txt](#pathwaystxt)  | Optional | Pathways linking together locations within stations. |
 |  [levels.txt](#levelstxt)  | Optional | Levels within stations. |
 |  [feed_info.txt](#feed_infotxt)  | Optional | Dataset metadata, including publisher, version, and expiration information. |
+|  [attributions.txt](#attributionstxt)  | Optional | Dataset attributions. |
 
 ## File Requirements
 
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -359,6 +359,7 @@ Describe the different levels of a station. Is mostly useful when used in conjun
 |  `level_index` | Float | **Required** | Numeric index of the level that indicates relative position of this level in relation to other levels (levels with higher indices are assumed to be located above levels with lower indices).<br><br>Ground level should have index 0, with levels above ground indicated by positive indices and levels below ground by negative indices.|
 |  `level_name` | Text | Optional | Optional name of the level (that matches level lettering/numbering used inside the building or the station). Is useful for elevator routing (e.g. “take the elevator to level “Mezzanine” or “Platforms” or “-1”).|
 
+
 ### feed_info.txt
 
 File: **Optional**
@@ -389,7 +390,7 @@ The file defines the attributions applied to the dataset.
 |  `route_id` |  ID referencing `routes.route_id`  | Optional | Functions in the same way as `agency_id` except the attribution applies to a route. Multiple attributions can apply to the same route. |
 |  `trip_id` |  ID referencing `trips.trip_id`  | Optional | Functions in the same way as `agency_id` except the attribution applies to a trip. Multiple attributions can apply to the same trip. |
 |  `organization_name` | Text | **Required** | Name of the organization that the dataset is attributed to. |
-|  `is_producer` | Enum | Optional | The role of the organization is producer. Valid options are:<br><br>`0` or empty - Organization doesn’t have this role.<br>`1` - Organization does have this role.<br><br>At least one of the fields `is_producer`, `is_operator`, or `is_authority` should be set at `1`. If none of them is set to `1` (e.g. because they are non-official fields defining other types of attribution), the data consumer is allowed to ignore them. |
+|  `is_producer` | Enum | Optional | The role of the organization is producer. Valid options are:<br><br>`0` or empty - Organization doesn’t have this role.<br>`1` - Organization does have this role.<br><br>At least one of the fields `is_producer`, `is_operator`, or `is_authority` should be set at `1`. |
 |  `is_operator` | Enum | Optional | Functions in the same way as `is_producer` except the role of the organization is operator. |
 |  `is authority` | Enum | Optional | Functions in the same way as `is_producer` except the role of the organization is authority. |
 |  `attribution_url` | URL | Optional | URL of the organization. |
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -26,6 +26,7 @@ This document defines the format and structure of the files that comprise a GTFS
     -   [pathways.txt](#pathwaystxt)
     -   [levels.txt](#levelstxt)
     -   [feed\_info.txt](#feed_infotxt)
+    -   [attributions.txt](#attributionstxt)
 
 ## Term Definitions
 
@@ -358,7 +359,6 @@ Describe the different levels of a station. Is mostly useful when used in conjun
 |  `level_index` | Float | **Required** | Numeric index of the level that indicates relative position of this level in relation to other levels (levels with higher indices are assumed to be located above levels with lower indices).<br><br>Ground level should have index 0, with levels above ground indicated by positive indices and levels below ground by negative indices.|
 |  `level_name` | Text | Optional | Optional name of the level (that matches level lettering/numbering used inside the building or the station). Is useful for elevator routing (e.g. “take the elevator to level “Mezzanine” or “Platforms” or “-1”).|
 
-
 ### feed_info.txt
 
 File: **Optional**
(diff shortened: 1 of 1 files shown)


timMillet:
Announced in the Google Group https://groups.google.com/forum/#!topic/gtfs-changes/6YrQYmAL3A0

gtfs/spec/en/reference.md
```diff
@@ -358,7 +359,6 @@ Describe the different levels of a station. Is mostly useful when used in conjun
 |  `level_index` | Float | **Required** | Numeric index of the level that indicates relative position of this level in relation to other levels (levels with higher indices are assumed to be located above levels with lower indices).<br><br>Ground level should have index 0, with levels above ground indicated by positive indices and levels below ground by negative indices.|
 |  `level_name` | Text | Optional | Optional name of the level (that matches level lettering/numbering used inside the building or the station). Is useful for elevator routing (e.g. “take the elevator to level “Mezzanine” or “Platforms” or “-1”).|
 
-
```
skinkie: Should not be in this pull request.

skinkie:
I would like to know why we should facilitate this _show off_. Attribution on a dataset is something that does not make sense to the end user. If the feed producer is not the the origin of data, it is likely that this organisation should be held accountable for the data quality. If the data is from multiple parties, why is an aggregator in between in the first place?

gtfs/spec/en/reference.md
```diff
@@ -375,3 +375,23 @@ The file contains information about the dataset itself, rather than the services
...
+|  `route_id` |  ID referencing `routes.route_id`  | Optional | Functions in the same way as `agency_id` except the attribution applies to a route. Multiple attributions can apply to the same route. |
+|  `trip_id` |  ID referencing `trips.trip_id`  | Optional | Functions in the same way as `agency_id` except the attribution applies to a trip. Multiple attributions can apply to the same trip. |
+|  `organization_name` | Text | **Required** | Name of the organization that the dataset is attributed to. |
+|  `is_producer` | Enum | Optional | The role of the organization is producer. Valid options are:<br><br>`0` or empty - Organization doesn’t have this role.<br>`1` - Organization does have this role.<br><br>At least one of the fields `is_producer`, `is_operator`, or `is_authority` should be set at `1`. If none of them is set to `1` (e.g. because they are non-official fields defining other types of attribution), the data consumer is allowed to ignore them. |
```
prhod: I don't understand the `data consumer is allowed to ignore them`. What is the goal of this sentence ?
Does allowing / forbidding something to the data consumer should be stated in the spec ? I think it depends on the chosen license attached to the feed, and it's not included in the GTFS (yet ?).

flocsy:
I get "Web Page Blocked!" for both the old and new version of Rejseplanen.

timMillet:
@flocsy I do not face this issue, but these datasets are available on my cloud if you need them:
[Rejseplanen (new version)](https://drive.google.com/open?id=1GnWRut8c-nc3jWIP0r6qZ1Gqc6Chy0_T)
[Rejseplanen (old version)](https://drive.google.com/open?id=1gypo0F_dFyGRv5-8ryBpT1F4zfG8jD6N)
I hope this will work.

gtfs/spec/en/reference.md
LeoFrachet: My phrasing was bad, and my intend wasn't the one you're describing. But I think you're right on the point that the spec can just stay silent on that case.

LeoFrachet:
The pull request is open for more than 7 days, so as per the official process, I'm opening the vote. It will be closed on Friday 20th, 23:59:59 UTC.

skinkie:
-1
This does not enhance travelinformation as end-user experience. Is not in line with the CC-0 direction many data providers are moving into. It isn't missed.

gcamp:
@skinkie It's true that it doesn't enhance travel information but if it makes some data available because this is available that otherwise wouldn't I consider this a overall win. Not sure how important that was for producers you talked to @LeoFrachet @timMillet

abertuca:
I vote +1

@skinkie This spec extension is necessary for a couple of key reasons:
(1) It is a pre-requisite by some data producers to release data; as such, consumer applications (such as Google Maps) will support the feature and we are actively trying to reduce shadow specifications of GTFS
(2) GTFS shouldn't require any specific license, and should allow licenses other than open ones. Google Maps for example consumes a substantial amount of travel information that is private.
(3) Even in the open data world, we want to do whatever possible (and within reason) to encourage data producers to continue producing data. If that means providing attribution to recognise their contributions to enhancing travel information, so be it.

skinkie:
@abertuca 
1. If it is a pre-requisite, organisize it in a SLA, which is a contract.
2. Great to know, but we are talking about factual information, which has no copyright
3. I disagree, since no **producer** has reported here' or in broader sense the standardisation community, that they want such extension.

abertuca:
@sinkie - thanks for your contributions here. 

To respond:
(1) Google Maps does include attribution in its contracts -- but we still need a spec to exchange the information.
(2) Not sure I understand your point
(3) Google Maps has had producers express their interest in the Attributions spec and I expect at least one producer to vote positively on this spec extension. I fully respect that you (as a producer) do not have this need, but others do.

skinkie:
1. This seems to be an internal requirement. 
2. I do not see any legal obligation to give attribution on any timetable published. So why facilitate this?
3. I would love to see these producer(s) to participate in the standardisation proces. Just expressing interest to a consumer is not participating in this ecosystem.

abertuca:
@skinkie - it seems you are opposed to this spec extension because you do not find it useful as a data producer. I would kindly ask that you rescind your (-1) if the voting confirms that this is useful for other data producers and consumers.

skinkie:
I don't find attribution useful as data producer, as data consumer, nor as end-user.

nighthawk:
I vote +1 (SkedGo)

There are different cases of GTFS data being provided with a condition "Yes, you can use this, but only if you include the following attribution". So as a data consumer, this addition can be a useful if this can be standardised in this way.

rp-pli:
I vote +1

For us at Rejseplanen in Denmark the attributions feature fulfils our policies concerning use of data on other platforms etc.

skinkie:
@abertuca The vote of @rp-pli triggered me to get in contact with the European Commission on this subject. Below is the legal framework, hence I'll +1.


Article 8 Requirements for service provisions reuse of travel and traffic data and linking of travel information services

1. The travel and traffic data listed in the Annex and the corresponding metadata including information on the quality thereof shall be accessible for exchange and reuse within the Union on a non-discriminatory basis, through the national or common access point and within a time-frame that ensures the timely provision of travel information services. They shall be accurate and up to date.

2. The data referred to in paragraph 1 shall be reused in a neutral manner and without discrimination or bias. Criteria used for ranking travel options of different transport modes or combinations thereof, or both, shall be transparent and not be based on any factor directly or indirectly relating to the user identity or, if any, the commercial consideration related to the reuse of the data and shall be applied on a non-discriminatory basis to all participating users. The first principle travel itinerary presentation shall not mislead the end-user.

3. **Where reusing the static and dynamic travel or traffic data, the source of those data shall be indicated.** The date and time of the last update of the static data shall also be indicated.

4. The terms and conditions for the use of the traffic and travel data provided through the national access point may be determined through a licence agreement. Those conditions shall not unnecessarily restrict possibilities for reuse or be used to restrict competition. Licence agreements, whenever used, shall in any event impose as few restrictions on reuse as possible. Any financial compensation shall be reasonable and proportionate to the legitimate costs incurred of providing and disseminating the relevant travel and traffic data.

5. Terms and conditions of linking travel information services shall be defined in contractual agreements between the travel information service providers. Any financial compensation of the expenses of linking travel information services incurred shall be reasonable and proportionate.

dekaikiwi:
Hi all, I would like to offer my perspective working with a few Japanese producers. I have invited their participation in this vote but I fear that the language barrier may be a problem here. Therefore I would like to offer context in place of this..

One particular use case I can provide from a producer in Japan was their concern that the current standard didn't give a full picture of who was responsible for the production of the final GTFS feed. The current GTFS format only gives the ability to provide contact details for the transit agency and data publisher (in this case the publisher was not producing GTFS so the producer implemented this conversion), therefore it would be reasonable to expect that the end-user might assume that the final (GTFS) feed is provided directly by one of these entities as opposed to this producer.

timMillet:
The voting period has ended and the `attributions.txt` file is now adopted!

Producer: Rejseplanen
Consumer: Google

4 votes in favor:
@abertuca from Google  
@nighthawk from SkedGo
@rp-pli from Rejseplanen
@skinkie from Bliksem Labs B.V. 

No abstentions and vetos.

I'll merge.

Issue #198, opened 2020-02-02
StopTime dependent bike allowance


hbruch:
Currently, GTFS supports bike_allowed as attribute for trips.

However, very often, bike allowance is either time-dependent (e.g. disallowed on workdays between 6am and 8:30am) or stop-dependent (e.g. boarding only allowed at first stop, unboarding only at the last).

I'd like to suggest to extend StopTimes.txt by two new attributes "bikePickupType" and "bikeDropoffType", with the following values:

For bikePickupType:
```
NO_BIKES_PICKUP_INFORMATION = 0
BIKES_PICKUP_ALLOWED = 1
NO_BIKES_PICKUP = 2 
```

For bikeDropoffType:
```
NO_BIKES_DROPOFF_INFORMATION = 0
BIKES_DROPOFF_POSSIBLE = 1
NO_BIKES_DROPOFF = 2
BIKES_DROPOFF_MANDATORY = 3
```

If this is a need felt by others as well, I'd prepare a PR to discuss this further. If there are already ongoing efforts I'm not aware of, please let me know.

skinkie:
I would support this.

paulswartz:
@mbta is also in favor.

scmcca:
Any developments on turning this into a PR?

PR #199, opened 2020-02-02
Support stop times dependent bike carriage

hbruch:
This PR proposes a more fine granular modeling of bike carriage. It addresses #198.

Currently, the availability of bike carriage can be specified via trips.bike_allowed.
However, many transport agencies (dis)allow bike carriage during specific time windows.
Or there are routes, where bicycles are allowed to (un)board only at specific stops.

This PR aims to support these policies by introducing a new optional field `stop_times.bikes_allowed` with the following options:

```
`0` or empty - No bike information for this stop. In this case `trip.bikes_allowed` defines, if bikes are allowed.
`1` - Mounting/unmounting with bicycle is allowed.
`2` - No bicycles are allowed beyond this stop. Bicycles on board must unmount at this stop.
`3` - Bicycles are allowed to stay on board, but are not allowed to mount/unmount at this stop.
```

Note however, that this straight forward modeling maps time policies to static stop_times. It's disadvantage is loosing the specific time window semantics, which could be problematic in realtime situations when delays occur. I'm unsure, how transport agencies handle such situations. If they'd require cyclists to adhere strictly to the official time windows, a proper file (e.g. `bike_allowance.txt`) with fields `trip_id`, `start_time`, `end_time`, `bikes_allowed` perhaps would be necessary.
Such a file additionally could provide notes regarding fare policies when carrying a bike.

N.B.: I'd like to announce this PR via gtfs-changes, but can't access it. Could anyone grant me access rights?


antrim:
* BART restricts carrying bikes onboard by certain hours (https://www.bart.gov/guide/bikes/bikeRules). BART also restrict based on certain train cars, so there might be an eventual tie in with GTFS-Vehicles (#200).
* @MBTA as well https://www.mbta.com/bikes (ping @paulswartz)

What other examples are there of agencies with these restrictions?

Bertware:
Stockholm's public transport authority has similar restrictions:

> Bicycles are generally not allowed on the SL services. During off-peak periods you may take your bicycle on board the commuter trains and on the Saltsjöbanan train. If you are taking the commuter train, you may not board or disembark the train at the stations Stockholm City or Arlanda if you have a bicycle with you. You may also take your bicycle on board the commuter ferries if space is available.

https://sl.se/en/in-english/getting-around/carry-on-luggage/

They consider every weekday 06:00-09:00 and 15:00-18:00 as rush hour (this is mentioned on the Swedish part of their site).

prazska-integrovana-doprava:
Is this still to be merged? We would use it.

In the Prague integrated transport system we needed to represent the bike allowance per stop time, because trams and particular buses carry bicycles only at some parts of their track. Also we needed to add two more options not present in this scenario.

`4` = mounting with bicycle is allowed, present bicycles are allowed to stay on board, but you can't unmount the bicycle
`5` = dismounting with bicycle is allowed, present bicycles are allowed to stay on board, but you can't mount the bicycle

hbruch:
When writing this PR, I had issues announcing it via gtfs-changes, but these should be resolved now. I'd announce it there to get some more feedback from possible producers/consumers. As this would be my first gtfs-changes PR, I'd be happy if there'd be a shepherd guiding me through, should I miss anything :-)

skinkie:
I think it would be good (in general) to have a discussion about the future of attributes. The proposal done now is certainly an option, but I could also imagine going for the HAFAS approach where a table exists for a trip, stop sequencenumber (and stop id) with the present features. Which is certainly more compact.

ValorNaram:
> but I could also imagine going for the HAFAS approach where a table exists for a trip, stop sequencenumber (and stop id) with the present features. Which is certainly more compact.

To understand you correctly @skinkie . The HAFAS approach
- has `trips.txt` which contains a field `bikes_allowed` and
- has `stop_times.txt` which contains also `bikes_allowed`?

to allow/disallow boarding with bike per `trips.txt` **and** per `stop_times.txt`?

Issue #117, opened 2018-11-08
GTFS-BoardingRestrictions


LeoFrachet:
Hi everybody!

MobilityData received the request to improve the representation of the boarding & alighting restrictions in GTFS.

### Issue
Some bus or train intercity agencies are allowed to sell tickets only for intercity trips, but not for intra-city trips. GTFS currently cannot represent such limitation.

Example of trip which cannot be described using pickup_type & drop_off_type:
- Washington DC: pick up allowed, drop off forbidden.
- New York City Downtown: pickup allowed, drop off allowed.
- New York City Uptown: pickup allowed, drop off allowed **if has boarded in DC**.
- Boston: pickup forbidden, drop off allowed. 

### Proposal
The proposal is to normalize an existing practice: the `stop_time_limitations.txt` extension defined by Bileto (@jspetrak). Here is the proposal: [bit.ly/gtfs-boarding-restrictions](http://bit.ly/gtfs-boarding-restrictions).

Thoughts?

skinkie:
On donderdag 8 november 2018 19:28:26 CET, Leo Frachet wrote:
> GTFS currently cannot represent such limitation.

It can, by implementing multiple trips, having boarding restrictions on all 
other stops than the stop the limitation is for. Fully backwards 
compatible.

> The proposal is to normalize an existing practice: the 
> stop_time_limitations.txt extension defined by Bileto 
> (@jspetrak). Here is the proposal: 
> bit.ly/gtfs-boarding-restrictions.
>
> Thoughts?

Unless an algorithm is included to convert the limitation into something 
that a journey planner can work from (positively defined, instead of 
negatively defined) this idea will not get my support.

-- 
Stefan

abyrd:
@LeoFrachet this sounds like the infamous (among journey planner developers 😁) French rules disallowing local transport to certain operators. The document from @jspetrak shows this also exists in the Czech Republic. I am curious though, does such a thing actually exist in most other places? This seems like a construct that would only be created by ad-hoc regulations. Even though I have a personal interest in covering French transit data well, I would be hesitant to add an entire new file to GTFS to allow such a special rule to be represented concisely.

I know there's a push to represent every detail in GTFS, but we should consider whether such details will ever be handled by the majority of consuming software, as well as the implied workload of revising every consuming application. We should be cautious of GTFS drifting into a less organized approximation of Netex.

[...]

PR #205, opened 2020-02-21
GTFS-TravelRestrictions

timMillet:
### Context

Currently, the GTFS does not handle extra pickup and drop-off limitations between some predefined groups of stop times. However, such limitations exist in several countries (e.g: CZ, DE, FR, NL, US), where some bus or train agencies are allowed to sell tickets only for intercity trips but not for intra-city trips, even though their trips make several stops within the same city.

Example of trip which cannot be described using `stop_times.pickup_type` and `stop_times.drop_off_type`:

- Washington DC: pick up allowed, drop off forbidden.
- New York City Downtown: pickup allowed, drop off allowed.
- New York City Uptown: pickup allowed, drop off allowed if has boarded in DC.
- Boston: pickup forbidden, drop off allowed.

MobilityData drafted the GTFS-BoardingRestrictions proposal to add this use case to the GTFS specification. This draft was discussed in the [proposal document](http://bit.ly/gtfs-boarding-restrictions) as well as in the issue #117.  Some modifications were brought according to the feedback of several contributors, and a consensus has been reached on the option B.

### Pull Request

This PR is made from option B for practical reasons:
- It allows many travel restrictions to a single stop time (as opposed to option A);
- It does not create duplicated trips (as opposed to option C).

This PR suggests the addition of an optional file `travel_restrictions.txt` that contains 3 fields.

The initial proposal was mentioning “boarding restrictions”. The term “travel restrictions” was preferred by several contributors, in order to reflect both boarding and alighting aspects.

Don’t hesitate to provide me with your feedback on the PR!


skinkie:
What currently has been described is a drop_off restriction. Could you elaborate in the example how you envision the extra pickup restriction and how to differentiate between them? In addition to that, since this is an extremely advanced feature we need QA-tests for this.

paulswartz:
> Boston: pickup forbidden, drop off allowed.

We currently do this in our GTFS using the `pickup_type` and `drop_off_type` fields. Do you have an example where we'd instead be using a `travel_restrictions.txt` file?

jspetrak:
@paulswartz You need to interpret the four points together. Travel Restrictions deal with situation when drop off/pickup rule of a stop is different depending on second pickup/drop off stop of one passenger journey.

E.g. On route with stops A->B->C, A can be boarded if travelling to C, but A can’t be boarded if desired to travel to B.

darylweinberg:
This sounds more like a fare structure discussion. How, operationally, do you stop someone from getting off at B?

Thank you,

Daryl Weinberg
Transit Systems Architect
o: 512-369-6216
|
|

jspetrak:
You are forbidden to sale tickets on such route. And you do not promote travel on such section (A-B) in journey planner results. That is what law requires. If someone buys A-C ticket for A-C price and gets off at B, it just something not addressed in law.

But because you want to not show A-B trips, just A-C and B-C in journey planner results, you need to have the data covered in GTFS, not in fare description. Mainly if there are journey planners that do not work with fare informationa at all.

timMillet:
@skinkie: Thank you for raising this question, I wasn’t explicit enough. GTFS-TravelRestrictions is for both pickup and drop_off restrictions: they are two sides of the same coin. When two stop_times A & B are grouped together with the same `restriction_id`, then you cannot be picked up at A if you want to be dropped off at B, and you cannot be dropped off at B if you have been picked up at A. Below is an example that better reflects these two sides. I will also update the example included in the PR.

> The trip starts in Washington DC and ends in Boston, MA. It has several stops in between.
> - [...]
> - Trenton, NJ: pick up allowed, drop off allowed.
> - New York City Downtown, NY: pickup allowed _if the rider will not alight in New York City Uptown_, drop off allowed.
> - New York City Uptown, NY: pickup allowed, drop off allowed _if the rider has not boarded in New York City Downtown_.
> - New Haven, CT: pickup allowed, drop off allowed. 
> - [...]

@paulswartz: To complement what @jspetrak said, maybe the example I gave in my first comment wasn’t the best. Above is another example, more precise, where all stop_times have `pickup_type=0` and `drop_off_type=0`: you can see that the `pickup_type` and `drop_off_type` fields are not enough to describe such a possibility. A real-life example was shared by Bileto, in the section “Example: Moravia Intercity Bus” within our [Gdoc proposal](bit.ly/gtfs-boarding-restrictions). @smsm1 also mentioned examples from the UK [in the issue related to this PR](https://github.com/google/transit/issues/117#issuecomment-594622663).

@darylweinberg I agree with @jspetrak, this PR is made for trip-planner purposes. It allows more accurate trip results without the need for fare implementation on the GTFS-consumer side.

skinkie:
@timMillet we still have the issue that we must prevent alighting and reboarding on the same or future trip, which is technically something that could happen.

timMillet:
@skinkie Good question. If boarding at A and alighting at C is forbidden but there are no limitations from A to B or B to C, then a trip-plan request from A to C may return a result from A to B then B to C, with a transfer at B, as long as it is not the same trip from A to C (i.e: it has to be 2 different trips). GTFS-TravelRestrictions is made for reflecting non-competitive obligations that may include this kind of (absurd?) situations. I can make this use case clearer in the PR. Does this accurately answer your concern?

skinkie:
@timMillet this will happen, there is no question about that, especially in situations where no other operators will have the same stop/station. This is the reason why in my opinion automatic trip expansion (option C) seems to be a reasonable consumer implementation. An alternative would invalidate each board/alight pair **after** it has been computed. This will require an additional sequential scan for each extra travel restriction. Without a some pseudo code on how this effectively could be implemented, I would not like to see voting.

jspetrak:
@skinkie FYI, Bileto has implemented TravelRestrictions on top of both SCA and Raptor and we use it for country-wide end-to-end trip planner in Czechia.

prhod:
Navitia is also using Raptor and implementing a subset of the proposal since 2014 (non overlapping restrictions). Implementing the full spec is not planned, but should not be a problem.

skinkie:
@jspetrak @prhod Could you both share you you have prevented checking the entire travelrestriction list for every passing time?

prhod:
Hello,
I took some time to check with our team expert, here is a summary of our discussions :

- If there is a better way to describe what a rider can and cannot do, we'd like to have the right modelization in the data (Option B chosen). The modelization should not be modified for the sake of technical implementation (in our point of view, but we understand the point).
- Our data exchange model (NTFS) uses an equivalent modelization of the "Option A".
- Our Raptor implementation have trouble with some edge cases of the "Option A" (for example this test is failing with the A implementation : https://github.com/CanalTP/navitia/blob/3099c82a921121791469ed8934fbefb802dfd8de/source/routing/tests/raptor_test.cpp#L2557)
- To achieve better itineraries, we may have to implement an internal deduplication of trips (but we think it's a technical matter and should not be a reason to amend the data exchange format)
- There is also the concern with GTFS and GTFSRT id consistencies when they are provided by different companies for the same operator.

I hope this message could help with the actual discussion.

timMillet:
To ease the implementation of this extension, some pseudo code is available in the proposal document, at the section [“Pseudocode”](https://docs.google.com/document/d/1E0RdoFTNKWEGVjRiaXM92Q1W1p-FbPm_KVt-cZOD4s8/edit#bookmark=id.j15x7cyrmfo1). Also, as previously announced, I updated the example given in the field definition of `restriction_id` to make the use case clearer.

2 GTFS stakeholders are producing and consuming `travel_restrictions.txt` by now. Therefore, I’m calling for a vote!

- Producer: Bileto [[Datasets](https://drive.google.com/open?id=10Og6507UA8ybYPQT58JVgnXFEXcUBOHS)]
- Consumer: Apple

The vote will be open until Monday, April 20th at 23:59:59 UTC.
(The voting period will cover 7 full calendar days and 5 full Swiss business days, Good Friday and Easter Monday being holidays in almost all Swiss cantons).

skinkie:
@timMillet so cloning is at this point the only 'smart' implementation?

dudebrains01:
+1 for [Transcollines](https://transcollines.ca/)

saraemarcus:
+1 from Apple

jspetrak:
+1 Bileto

skinkie:
-1 until we have an efficient runtime example (not require preprocessing, effectively implementing option C) for this feature.

GeorgoSK:
+1 from [CENDIS](https://www.cendis.cz/nase-projekty/system-jednotneho-tarifu/)

For context, we are a state-owned IT taskforce founded by the Czech Ministry of Transportation. We are in the process of opening up schedule data for the entirety of public transportation in CZ, and we are counting on GTFS as one of the supported input/output formats. In order for this to work long-term, we need a reliable way to represent licensing restrictions for the carriage of passengers, as this is a widely used concept. Our use case is closely tied to a nationwide ticketing scheme, which needs to provide reliable information to the passenger—and the (in)ability to make certain trips independently from the actual timetable is vital here.

prhod:
+1 from Kisio

gcamp:
+1 from Transit

thomastrillium:
+1 from Trillium

timMillet:
The vote ended and the proposal did not pass.

- 7 votes for the proposal: Transcollines, Apple, Bileto, CENDIS, Kisio, Transit, Trillium.
- 1 vote against the proposal: Bliksem Labs B.V.

The proposal is not abandoned. I’ll work on addressing the feedback and come up with a solution. Thank you for your votes!

abyrd:
I did not participate in this vote, but did contribute significantly to the extensive discussion in #117 and on draft documents. In a new vote, I would support this (as it aligns well with my position in all discussions).

@skinkie I understand that you want to delay official adoption of changes that may have problematic technical implications for consumers. This perspective also motivates some of my responses to GTFS change proposals. But in this case, a lot of time was invested in extensive multilateral discussion. If you review #117 I think you'll find very clear arguments for option B, which went through a series of well thought-out revisions.

It seems to me that you are allowing your detailed knowledge of consumer implementations to influence your ideal data representation too heavily. This option B very concisely represents a "natural" domain entity that's easy to reason about: a restriction on traveling between certain groups of stops. Other representations lose that clarity, bending data production to match somewhat arcane details of one possible use case of GTFS data (routing algorithms). The authors have already agreed to include pseudocode in the specification to help producers understand how this new data can be interpreted by routing code. There are already several producers of the new data, and several consumers which successfully bridge from the new GTFS travel_restriction abstraction to data structures suitable for routing.

It seems like all requirements are met here, except your caution about ensuring existence of a consumer implementation that uses an approach different than the one in the pseudocode. But consider that we have statements from several consumers that they have chosen this implementation and have no problem with it. I would say the same for OTP2: we will probably consume this new table, and prefer to perform the pattern expansion transformation in the consumer so the source data is more natural, concise, and clear.

It's definitely worth slowing down and carefully thinking through the impact of changes before they're made final. But in this case, the discussion and experimentation have already gone on for well over a year, so I'd ask you to reconsider your position.

skinkie:
@abyrd I have posted a reply to the gtfs-changes mailing list. To me presenting the expansion as only solution is inadequate. We are at a cross road of enabling a **very important** new feature within GTFS. As both being producer and consumer (and we are currently expanding) I don't see this feature being easily picked up as a new presentation attribute. If my only request is having a performant algorithm that would run under runtime conditions, that should not be too far fetched for all the bright minds here in this standardisation process.

skinkie:
@timMillet I have just spoken to @abyrd he just made an attempt to rerail this proposal by a possible lite weight implementation without trip duplication to be discussed in the Google Group.

timMillet:
Hi @abyrd @skinkie
Would you have any news on this possible lightweight implementation without trip duplication?
Please tell me if I can help with anything. Thank you!

antrim:
Hi @abyrd and @skinkie : Pinging on this. I don't think there was more discussion on GTFS-Changes, unless I missed it. [Trillium](http://trilliumtransit.com) voted for this and has implemented some version of this through workarounds in the existing spec (trip duplication). We're interested in seeing a fix adopted in the specification so that we can implement it in our product and data.

skinkie:
@antrim I am all for standardising travelrestrictions, but it must a method to describe outcome in the main context of the GTFS usage: a journey planner; with the key argument that a GTFS feed with travel restrictions is **not** backwards compatible. If there is already a consumer implementing this, the ecosystem would certainly benefit if that solution is shared and does not mean a performance degradation.

abyrd:
My comment on April 22 is still valid: "I did not participate in this vote, but did contribute significantly to the extensive discussion in #117 and on draft documents. In a new vote, I would support this (as it aligns well with my position in all discussions)." 

I saw the gtfs-changes mailing list post. It mentions "a major computation requirement happening at every consumer" and "the massive computational impact of this feature". I have not measured or observed this impact myself so had no response to offer. It is true that the implementation being discussed has poor asymptotic complexity, but it also seems likely that the quantity of restrictions in most or all real-world feeds would be small. @skinkie have you measured large changes in computational requirements from this change? Can you provide details of specific cases or feeds that are problematic?

In any case, I'm not sure it matters whether a high efficiency implementation of routing is known and publicly documented. These restrictions are important enough for regulatory and user experience reasons that several GTFS consumers and producers _have_ implemented restricted routing, and find it worthwhile to pay for the extra compute capacity to handle it (which I suspect is almost unmeasurably small for most real-world feeds). 

I am very supportive of the requirement that changes to GTFS and GTFS-RT be implemented and used in practice before they are officially adopted. It prevents the spec from expanding massively into theoretical territory, or requiring all consumers to be capable of processing inputs that don't exist in the wild.

But in this case, several organizations are already producing and consuming data in the proposed format. They seem to have found routing implementations efficient enough for use in production applications, and have shared those concepts with the community. Modern table-scanning routing algorithms are fast enough to deal with tens or hundreds of expanded stop patterns on a few routes. The fact that the example approach has poor asymptotic complexity certainly deserves a footnote, but I don't see why this implementation detail should block us from sharing information about what travel restrictions exist.

Implementation of routing (or any other use of GTFS) is a separate concern from data interchange design. Here, we are designing and standardizing the description of available transportation services. We should be aiming to express those services in a clean and concise way.

I also don't see there being a deal-breaking backward compatibility issue here, because the travel restrictions table functions as a supplement to existing schedule tables. Output from a non-conforming consumer would (on certain specific itineraries) contain legs that are not allowed, but it would not completely break the consumer.

skinkie:
>  Can you provide details of specific cases or feeds that are problematic?

It seems that the German (but also Norwegian) cases have the biggest impact. They are typically not small. Hence we are talking for every trip on a line if they have multiple stops in a city. 

> But in this case, several organizations are already producing and consuming data in the proposed format.

Who are? And why aren't they stepping into this discussion regarding complexity?

> I also don't see there being a deal-breaking backward compatibility issue here, because the travel restrictions table functions as a supplement to existing schedule tables. Output from a non-conforming consumer would (on certain specific itineraries) contain legs that are not allowed, but it would not completely break the consumer.

Here I disagree. You can't do a backwards compatible implementation, and have this proposal in a single feed, and have legacy consumer naively using the data.

abyrd:
> > Can you provide details of specific cases or feeds that are problematic?
> 
> It seems that the German (but also Norwegian) cases have the biggest impact. They are typically not small. Hence we are talking for every trip on a line if they have multiple stops in a city.

This seems like a theoretical concern. I'd need to run the numbers for a specific existing data set, and/or see actual profiling results showing an impact to have an opinion. How many restrictions are there on how many routes in Germany? I suppose it's reasonable to ask someone to provide performance measurements before officially adopting the change, as it would prove that some usable implementation actually exists out there. But it's unusual: other GTFS changes have been adopted with no requirements of implementation efficiency.

I worked through this once with @t2gran in the context of OTP2. I believe we came to the conclusion that we could implement it by just repeatedly scanning any routes with restrictions (once per boarding group), effectively performing the pattern expansion on demand, incurring only a small speed hit but no increase in space complexity. These would be repetitive, linear access patterns on contiguous arrays so taking full advantage of cache prefetch; I would expect it to be very fast. I'm not sure if @t2gran has anything to add here.

> > But in this case, several organizations are already producing and consuming data in the proposed format.
> 
> Who are? And why aren't they stepping into this discussion regarding complexity?

In the above comments, @jspetrak says they've implemented it, and @prhod says Navitia has implemented a subset and sees no difficulty implementing the whole proposal. I've heard of these restrictions being applied in French trip planners for years (due to regulatory requirements). Kisio produces their own extended GTFS that has some features (like this one) in advance of official adoption. There might be others, and I'd invite them to clarify whether they found the implementation straightforward and performant enough for widespread adoption.

> > I also don't see there being a deal-breaking backward compatibility issue here, because the travel restrictions table functions as a supplement to existing schedule tables. Output from a non-conforming consumer would (on certain specific itineraries) contain legs that are not allowed, but it would not completely break the consumer.
> 
> Here I disagree. You can't do a backwards compatible implementation, and have this proposal in a single feed, and have legacy consumer naively using the data.

If you want to prevent older consumers from naively misinterpreting new data, then we need a mechanism to ensure routes with TravelRestrictions are perceived as invalid by those old consumers. The first thing that comes to mind is a new pickup/dropoff type meaning "refer to TravelRestrictions", which would be present in a long-existing table (stop_times.pickup) but use a value that was undefined. On #117 @LeoFrachet suggested requiring all pickup and dropoffs to be set to 1 (no pick up/drop off) which is another option. 

However, I believe there are other additions to GTFS that would be unseen by older consumers and naively misinterpreted, e.g. continuous_pickup. All consumers should flag/log/warn about the presence of unrecognized fields and tables, which should tip off the people deploying the system that they're missing some nuances of the data. Or should we ensure that all extensions are perceived by older consumers as deactivated?

skinkie:
> This seems like a theoretical concern. I'd need to run the numbers for a specific existing data set, and/or see actual profiling results showing an impact to have an opinion. How many restrictions are there on how many routes in Germany? 

I have only the data from FlixBus available, in corona time those are 15109 different travel restrictions from a single stop to multiple stops that are now allowed. For us this is rather problematic, because we cannot keep on make multiple feeds because of the majority of consumers not supporting a feature that affects the majority of routes.

> > Here I disagree. You can't do a backwards compatible implementation, and have this proposal in a single feed, and have legacy consumer naively using the data.
> 
> If you want to prevent older consumers from naively misinterpreting new data, then we need a mechanism to ensure routes with TravelRestrictions are perceived as invalid by those old consumers. The first thing that comes to mind is a new pickup/dropoff type meaning "refer to TravelRestrictions", which would be present in a long-existing table (stop_times.pickup) but use a value that was undefined. On #117 @LeoFrachet suggested requiring all pickup and dropoffs to be set to 1 (no pick up/drop off) which is another option.
> 
> However, I believe there are other additions to GTFS that would be unseen by older consumers and naively misinterpreted, e.g. continuous_pickup. All consumers should flag/log/warn about the presence of unrecognized fields and tables, which should tip off the people deploying the system that they're missing some nuances of the data. Or should we ensure that all extensions are perceived by older consumers as deactivated?

Lets not forget that we _still_ do not have a GTFS version identifier in the feed, or metadata with what features are used. A consumer can only know something new is added by a new file or a new column.

e-lo:
Hi all – not to sidetrack the current discussion about adoption, but I am wondering what the motivation was behind using the `stop_sequence` field rather than `stop_id` field to identify the restrictions? There are two reasons I feel like `stop_id` is better:
1. `stop_sequence` is unique to **`stop_times.txt`** and therefore less direct to join to **`stops.txt`** to see geographically.  
2. `stop_id` is likely more stable over time whereas `stop_sequence` would require updating every value in a trip, say, if you add a preceding stop.  Ideally the travel restrictions shouldn't need to change if the restrictions themselves aren't.

skinkie:
> Hi all – not to sidetrack the current discussion about adoption, but I am wondering what the motivation was behind using the `stop_sequence` field rather than `stop_id` field to identify the restrictions? There are two reasons I feel like `stop_id` is better:
> 
> 1. `stop_sequence` is unique to **`stop_times.txt`** and therefore less direct to join to **`stops.txt`** to see geographically.
> 2. `stop_id` is likely more stable over time whereas `stop_sequence` would require updating every value in a trip, say, if you add a preceding stop.  Ideally the travel restrictions shouldn't need to change if the restrictions themselves aren't.

I think _circle lines_ might be an argument as well for this specific case, for example to pick up passengers within a city, but in general we should avoid using stop_id to join towards stop_times.txt. It becomes a mess if a stop_id occurs twice during a trip.

e-lo:
> It becomes a mess if a stop_id occurs twice during a trip.

In general, yes. But I can't think of any use cases where you would. be prohibited from boarding the first time the bus stops at at `stop_id` but would be the second. 

Willing to set this aside, but I really want to avoid changes to files where changes haven't actually occurred (issue 2 I raised)

skinkie:
> > It becomes a mess if a stop_id occurs twice during a trip.
> 
> In general, yes. But I can't think of any use cases where you would be prohibited from boarding the first time the bus stops at at `stop_id` but would be the second.

I don't think boarding is the problem, the alighting is the problem. But we can think of a exceptional case...


> Willing to set this aside, but I really want to avoid changes to files where changes haven't actually occurred (issue 2 I raised)

I don't think that is a justifiable reason. GTFS is serialisation, even if everyone would change their stop_ids every time, this is still a valid GTFS file. That some of the players are suggesting fixed identification is something that is never formalised. Or in your case, your management database with prohibited stop pairs is likely something different than the serialisation towards the travel restrictions file.

antrim:
Trillium is currently duplicating trips in order to represent travel restrictions (described as Option C in the [proposal document](http://bit.ly/gtfs-boarding-restrictions)). It's non-ideal because it makes GTFS datasets less useful for real-time, timetable generation, and other applications. It is also somewhat more complex to produce GTFS data by duplicating trips.

We'd like to know what option will ultimately be adopted so that we can build support for GTFS-TravelRestrictions it in our GTFS Manager product!

Let me know if there is anything that I can do to get this ready for a modification or re-vote:
- Could we set up a phone call or time a Slack meeting to discuss further?
- Would it be useful to list a few more real examples where this feature is needed?

> Hi all – not to sidetrack the current discussion about adoption, but I am wondering what the motivation was behind using the stop_sequence field rather than stop_id field to identify the restrictions?

On `stop_id` vs `stop_sequence`: it doesn't matter to us. I doubt this would make a significant different for most data producers, except `stop_id` could be a tad easier for producing GTFS by hand. This could be a coin toss sort of situation.

prhod:
Hello everyone, 
In my understanding of the situation, **stop_time** is a better choice than **stop**. There may be cases where the same stop pole is used by several agencies, some providing local travels (no restrictions) and others long distances (with restrictions).  But I don't have a real example of that in mind !
I'd prefer stop_times, but I won't go against the other.

peetjvv:
> 
> 
> Hello everyone,
> In my understanding of the situation, **stop_time** is a better choice than **stop**. There may be cases where the same stop pole is used by several agencies, some providing local travels (no restrictions) and others long distances (with restrictions). But I don't have a real example of that in mind !
> I'd prefer stop_times, but I won't go against the other.

Just to aid you with some real world examples.
In Johannesburg, South Africa the municipal bus system shares stop infrastructure with some privately run bus operators (both short distance and long distance operators) in the city. And here in Cape Town the municipally run BRT will sometimes share their infrastructure with a privately run bus operator in the city.

abyrd:
The last few comments refer to `stop_time`, but the question raised by @e-lo was about `stop_id` versus `stop_sequence`. I think this should be clarified before we draw any conclusions. The fact that multiple operators use the same stop infrastructure does mean we'd want to select specific `stop_times` rather than every trip passing through a given stop. But that would only be an issue if you removed the `trip_id` reference from the travel restrictions. I believe what @e-lo was suggesting was to keep the `trip_id` field and combine it with a `stop_id` instead of a `stop_sequence` for selecting `stop_times`. The main difference between the two approaches is that the first always unambiguously selects a single `stop_time`, while the latter can select multiple `stop_times` (but always from the same trip). The latter may actually more closely match regulatory requirements (forbidding travel within geographic zones) but the former is more exact and flexible, and can represent every situation that the latter can.

That said, I tend to agree with the responses that GTFS data should ideally be machine-written from a database. Sequence number changes should be handled automatically, in fact each export to GTFS should be "from scratch" and not be compromised by any change in IDs or sequences. We want to enable smaller operators to share data sets, but should do that by providing them with good open source software rather than catering to hand-editing tables in Excel. 

We would err on the side of greater precision and flexibility using `stop_sequence`, but my sense is that over-selecting trips from other operators is not a problem even if `stop_id` were used its place.

prhod:
Sorry if I misunderstood the question ! I agree with @abyrd.  
If the question to keep the `trip_id` and use the `stop_id` instead of the `stop_sequence`, creating a GTFS will be much simpler and ease the adoption of this extension, and side effects shouldn't occur that much IMHO.

antrim:
I agree with what @abyrd wrote on [Sep 15](https://github.com/google/transit/pull/205#issuecomment-693175889). Here's my own take, with some elaboration.

The aim of the GTFS backwards compatibility principle is to enable new features in applications while not breaking other applications. It is not a requirement of the GTFS process that every application should offer every feature and use all data elements.

Travel restrictions are already being implemented in feeds:
* through a largely undocumented workaround (duplicated trips)
* through the new travel_restrictions.txt table

What we've seen before with GTFS is that waiting too long to formalize new spec features or leaving ambiguities open too long leads to confusion and splintering practice. GTFS allows experimentation in new features, which is one of its strengths, and then at a certain point it is time to agree, formalize, and pursue broader adoption. I think we're at that point with GTFS-TravelRestrictions, and so I'd urge that we vote again. I don't see a spec-based solution to the issue of added complexity in routing graphs.

I also agree with @abyrd comments on [Sep 17](https://github.com/google/transit/pull/205#issuecomment-694653484) -- to "err on the side of greater precision and flexibility using `stop_sequence`."

skinkie:
>  I don't see a spec-based solution to the issue of added complexity in routing graphs.

There is actually. The "workaround" (iff documented) is explicit. The "proposal" is implicit. Hence every application remains to work with a documented workaround, while the proposal requires either a precomputation that likely to effectively create the workaround, or a change in a journey planner algorithm that would filter results based on this file. The main reason for me voting no is that the method to integrate this extra file into a restricted_stop_times.txt is not made explicit.

_In my perspective this documentation lacks for frequencies.txt too, but we are now at a point that frequencies.txt and travelrestrictions may be combined. If it **such** an effort for the people that propose this to write some documentation how this can be used in a journey planner, to me this means that it is not a priority or the desire to create applications that cannot be matched easily. Try to implement HAFAS TEXT from scratch._


> It is not a requirement of the GTFS process that every application should offer every feature and use all data elements.

I think our common aim should be to be as clear as possible and prevent ambiguities or plain wrong information. Making the workaround by hand is virtually impossible. Not able to create the workaround results directly into invalid journey planner results.

abyrd:
@skinkie I regularly caution against haphazard expansion of GTFS, or adding features that may shift unknown amounts of effort to the authors of GTFS consumer systems. So I appreciate your concern that we should be careful and patient, and ensure ideas are technically sound before finalizing them. 

However, in this case I just can't follow your line of reasoning. I think that line of reasoning will need to be clarified to convince anyone. I have made comments inline below, hopefully identifying what can be clarified so we can understand each other and reach a consensus.

> > I don't see a spec-based solution to the issue of added complexity in routing graphs.
> There is actually. The "workaround" (iff documented) is explicit. The "proposal" is implicit. 

This reaction seems to be from the single perspective of routing with particular algorithms. The purpose of GTFS is not to perform routing, it is to exchange descriptions of transportation services. When you say the workaround and proposal are explicit and implicit respectively, you seem to mean that the data structure or method to be used in routing is explicit or implicit. I perceive this as a category mistake: the workaround and proposal are ways of describing services. They belong to a separate logical category than the process of finding paths through a network.

The proposal for describing travel restrictions is not implicit, it is clearly and completely specified. On the other hand, the data structure or algorithm in which this information is used are not even implicit: the proposal does not prescribe any method whatsoever, nor does it even prescribe using GTFS data for routing.

> Hence every application remains to work with a documented workaround

My interpretation of this statement is that "with a documented workaround, all applications will continue to work". This does not seem right to me. Once trips are expanded into multiple trips representing the same vehicle at the same time, some uses of GTFS would require inferring the fact that these resulting multiple trips in fact represent a single vehicle making a single trip. The trip-splitting workaround makes one use case with a certain family of algorithms easier, but only by making other use cases harder and breaking the useful, intuitive assumption that no two trips are simultaneously representing the same vehicle at the same time.

> while the proposal requires either a precomputation that likely to effectively create the workaround, or a change in a journey planner algorithm that would filter results based on this file

That is correct: in order to make full use of the new information, a consuming application must either make changes to its algorithm, or transform its inputs in a way equivalent to changing the algorithm. But isn't that true of any new information added to GTFS? It is not clear why it's a problem that you need to add code or change data structures to use new information.

> The main reason for me voting no is that the method to integrate this extra file into a restricted_stop_times.txt is not made explicit.

It is not clear what restricted_stop_times.txt is, and what it means to "integrate this extra file" into it. We can guess that you are referring to tables of data being transformed to allow a certain routing process, but the .txt file name format implies you are talking about a file in a GTFS feed, whereas most people might expect transformations of input data into routable data structures to happen in a consumer application's internal domain model, after the GTFS feed has already been loaded.

> _In my perspective this documentation lacks for frequencies.txt too

It is not clear what documentation is lacking. I'm guessing you're referring to exact_times, which genuinely is shorthand for expanding one example trip into many. That has always seemed clear to me, but is there something specific you think is missing? Or are you referring to routing on frequency-based lines without exact_times? That on the other hand is a great example of something that clearly describes a service but has no obvious single implementation from a routing perspective.

> we are now at a point that frequencies.txt and travelrestrictions may be combined.

How would they be combined, and why? They seem to be completely unrelated concepts. Are you referring to the fact that it might be efficient to represent both of these GTFS concepts in a similar way in the internal model of certain routing algorithms? 

> If it **such** an effort for the people that propose this to write some documentation how this can be used in a journey planner, to me this means that it is not a priority or the desire to create applications that cannot be matched easily. 

Your "if" clause here sets up a false premise. No one thinks it's a problem to document methods for routing with restrictions. Everyone already agreed it was reasonable, and I think most agree that the proposal should not be adopted until we can point to one or more working open source implementations of travel restrictions. 

It is hard to understand the syntax of your last clause, but I think you mean people want to create proprietary GTFS-consuming software with capabilities others cannot achieve, and by making the GTFS format ever more complex, push competitors off the scene. This is a reasonable concern - it's not unheard of for powerful businesses to use anti-competitive strategies. But in this case, we're looking at a feature that's implemented by multiple organizations, some of which fought long and hard to open source their core systems. It seems clear that those organizations welcome anyone looking at their implementation and learning from it. 

>Try to implement HAFAS TEXT from scratch._

I expect most of your readers will perceive this as a non sequitur or simply find it incomprehensible. A Google search for the expression "HAFAS TEXT" yields a total of three results: a forum comment written by you, a page of random words including "kurt cobain" and "fashion design", and a PhD thesis containing a single instance of the expression "HAFAS text-file". The reader might sense that this is a rhetorical question, but will then be wondering: Who are you asking to implement it? Why would they be implementing it? What exactly does it mean to implement a data exchange format, does this mean reading and writing files in this format? And finally: How is reading and writing this format related to frequencies or travel restrictions in GTFS?

> > It is not a requirement of the GTFS process that every application should offer every feature and use all data elements.
> I think our common aim should be to be as clear as possible and prevent ambiguities or plain wrong information. Making the workaround by hand is virtually impossible. Not able to create the workaround from this proposal by a machine results directly into invalid journey planner results.

This doesn't appear to be a response to Aaron's point. He's saying that people do not have to implement routing on travel-restricted lines. At least two ways have been proposed to ensure simpler or legacy software does not misinterpret the new data or produce misleading results. 

Finally, as far as I know no one is advocating the workaround (early expansion of trips in the GTFS domain), by hand or by machine. They're specifically making proposals on how to avoid this workaround. There is not agreement that such a workaround will be as "clear as possible". For me this is because it breaks intuitive assumptions about a one-to-one relationship between the common-language concept of a "trip" and a single vehicle at any given time.

skinkie:
> > > I don't see a spec-based solution to the issue of added complexity in routing graphs.
> > > There is actually. The "workaround" (iff documented) is explicit. The "proposal" is implicit.
> 
> This reaction seems to be from the single perspective of routing with particular algorithms. The purpose of GTFS is not to perform routing, it is to exchange descriptions of transportation services. When you say the workaround and proposal are explicit and implicit respectively, you seem to mean that the data structure or method to be used in routing is explicit or implicit. I perceive this as a category mistake: the workaround and proposal are ways of describing services. They belong to a separate logical category than the process of finding paths through a network.
> 
> The proposal for describing travel restrictions is not implicit, it is clearly and completely specified. On the other hand, the data structure or algorithm in which this information is used are not even implicit: the proposal does not prescribe any method whatsoever, nor does it even prescribe using GTFS data for routing.

I fail to see how this could **only** be applied to the _most common_ use case of routing. This does apply to any consumer that presents stop times, for example some thing trivial as an arrival / departures application, or more legacy the creation of departures display prints (to put into analogue stop signage) from GTFS. Everyone that is using the stop_times.txt concept must be aware that there is now a change that goes beyond pickup_type. Since the pickup_type is not changed for a stop_times.txt line any legacy user might suggest a pickup that may be prohibited. The workaround explicitly expresses the restriction in _legacy_ terms. I am **absolutely not against** the proposal of GTFS-TravelRestrictions as a separate file, but I must guard that this standard may never result wrongful interpretation. And to be honest, that wrongful interpretation starts with every party that did not implement the latest GTFS specification. In this thread an alternative pickup_type as been proposed, I think that should be added to guard for legacy clients.

> > Hence every application remains to work with a documented workaround
> 
> My interpretation of this statement is that "with a documented workaround, all applications will continue to work". This does not seem right to me. Once trips are expanded into multiple trips representing the same vehicle at the same time, some uses of GTFS would require inferring the fact that these resulting multiple trips in fact represent a single vehicle making a single trip. The trip-splitting workaround makes one use case with a certain family of algorithms easier, but only by making other use cases harder and breaking the useful, intuitive assumption that no two trips are simultaneously representing the same vehicle at the same time.

This statement in principle shows that since I am not aware as engineer of consumers that will have more problems with method A we should go for method B. The sole thing I asked for "document how this can be used in consumers" still holds, I am basically the only one asking for clarification how this should be handled in one specific use case. 


> > while the proposal requires either a precomputation that likely to effectively create the workaround, or a change in a journey planner algorithm that would filter results based on this file
> 
> That is correct: in order to make full use of the new information, a consuming application must either make changes to its algorithm, or transform its inputs in a way equivalent to changing the algorithm. But isn't that true of any new information added to GTFS? It is not clear why it's a problem that you need to add code or change data structures to use new information.

No, this is not the case. Virtually all the new information we have added to GTFS are complementary and create a higher level of detail.  I would make the distinction between attributes and objects. We have added many new attributes that in no way effect existing implementations. And we have added certain objects (for example paths) that would increase the level of detail within a parent_station, but if it was ignored, nobody would get hurt. As I explained, above the fact that pickup_type is equal for a restriction and without a restriction a legacy implementation would not use the restriction. Negations are problematic by nature. 


> > The main reason for me voting no is that the method to integrate this extra file into a restricted_stop_times.txt is not made explicit.
> 
> It is not clear what restricted_stop_times.txt is, and what it means to "integrate this extra file" into it. We can guess that you are referring to tables of data being transformed to allow a certain routing process, but the .txt file name format implies you are talking about a file in a GTFS feed, whereas most people might expect transformations of input data into routable data structures to happen in a consumer application's internal domain model, after the GTFS feed has already been loaded.

I think it should be not to far fetched to have OneBusAway transformer to create a feed that is free of travelrestrictions, in a similar style that calendars can be optimised within the same application. I'll continue below.


> > _In my perspective this documentation lacks for frequencies.txt too
> 
> It is not clear what documentation is lacking. I'm guessing you're referring to exact_times, which genuinely is shorthand for expanding one example trip into many. That has always seemed clear to me, but is there something specific you think is missing? Or are you referring to routing on frequency-based lines without exact_times? That on the other hand is a great example of something that clearly describes a service but has no obvious single implementation from a routing perspective.

I actually mean both. The fact that it is for you and me clear how exact times work is that we have a significant background in transit modelling and normalisation. The frequency aspect without exact_times is a thing that could also made visual by something like the probability that a vehicle arrives at a certain time and approaches how a more naive view could be created, especially for _transfers_ between services. For anyone else that does not have this background the least we can provide is the the documentation what the transformation towards a denormalised datastructure results in, purely as documentation. In this line of thinking I would like to see documentation or at least someone that took the effort to explain how trips would be expanded from the combination of stop_times.txt and travelrestrictions.txt. Hence I don't care that I understand how travelrestrictions could work both as producer and consumer, I care about the next guy/girl that has to implement it.

> > we are now at a point that frequencies.txt and travelrestrictions may be combined.
> 
> How would they be combined, and why? They seem to be completely unrelated concepts. Are you referring to the fact that it might be efficient to represent both of these GTFS concepts in a similar way in the internal model of certain routing algorithms?

No, my point here is that frequencies (+ stop times) can be combined with travel restrictions and we have a situation that everyone is able to describe the restriction, only a select group of people is able to oversee how that should be implemented.
 
> > If it **such** an effort for the people that propose this to write some documentation how this can be used in a journey planner, to me this means that it is not a priority or the desire to create applications that cannot be matched easily.
> 
> Your "if" clause here sets up a false premise. No one thinks it's a problem to document methods for routing with restrictions. Everyone already agreed it was reasonable, and I think most agree that the proposal should not be adopted until we can point to one or more working open source implementations of travel restrictions.
>
> It is hard to understand the syntax of your last clause, but I think you mean people want to create proprietary GTFS-consuming software with capabilities others cannot achieve, and by making the GTFS format ever more complex, push competitors off the scene. This is a reasonable concern - it's not unheard of for powerful businesses to use anti-competitive strategies. But in this case, we're looking at a feature that's implemented by multiple organizations, some of which fought long and hard to open source their core systems. It seems clear that those organizations welcome anyone looking at their implementation and learning from it.

We decided as group that both a producer and consumer should exist before we are standardising something. So you have a consumer in mind that did implement **this** proposal? And if you have, is there any reason why they have are motivated to standardise this proposal, but months on wards nobody is technically capable to explain to the ecosystem how it should be implemented? My concern is that $$$ pays MobilityData, but MobilityData does not have the staffing that is technically capable of this style of documentation work. Hence the difference between Information Science (design) and Computer Science (programming), while their members certainly have (to) implement(ed) this.


> > Try to implement HAFAS TEXT from scratch._
> 
> I expect most of your readers will perceive this as a non sequitur or simply find it incomprehensible. A Google search for the expression "HAFAS TEXT" yields a total of three results: a forum comment written by you, a page of random words including "kurt cobain" and "fashion design", and a PhD thesis containing a single instance of the expression "HAFAS text-file". The reader might sense that this is a rhetorical question, but will then be wondering: Who are you asking to implement it? Why would they be implementing it? What exactly does it mean to implement a data exchange format, does this mean reading and writing files in this format? And finally: How is reading and writing this format related to frequencies or travel restrictions in GTFS?

HAFAS Text (HRDF, https://transportdatamanagement.ch/content/uploads/2020/04/HRDF.5.20.39-Guidelines-e.pdf ) is a specification by Hacon / Siemens Mobility that has been the defacto transit exchange format for all big railway operators in Europe. It includes virtually all travel styles that one can imagine. It describes how it can be loaded in a specific journey planner. Now in this specific instance this journey planner would be bought by a commercial party, so implementing this as consumer sounds strange. Why would you want to implement this proprietary format as consumer? The same is actually now happening with GTFS. We are introducing features where the concepts can be explained to a producer (like with HAFAS), but it is not obvious how to implement it. In an open forum I ask this question especially since we posed to requirement that both a producer and a consumer understand how they should handle it. If we skip the consumer part, who (which consumers) are we building this standardisation format for?


> > > It is not a requirement of the GTFS process that every application should offer every feature and use all data elements.
> > > I think our common aim should be to be as clear as possible and prevent ambiguities or plain wrong information. Making the workaround by hand is virtually impossible. Not able to create the workaround from this proposal by a machine results directly into invalid journey planner results.
> 
> This doesn't appear to be a response to Aaron's point. He's saying that people do not have to implement routing on travel-restricted lines. At least two ways have been proposed to ensure simpler or legacy software does not misinterpret the new data or produce misleading results.

I know about the pickup_type, what is the other one?

> Finally, as far as I know no one is advocating the workaround (early expansion of trips in the GTFS domain), by hand or by machine. They're specifically making proposals on how to avoid this workaround. There is not agreement that such a workaround will be as "clear as possible". For me this is because it breaks intuitive assumptions about a one-to-one relationship between the common-language concept of a "trip" and a single vehicle at any given time.

Lets explicitly document how to avoid this, GTFS.org seems to be a superb place for this.

antrim:
I've read through this discussion carefully. I think this gets at a core assumption about what GTFS is. GTFS has been presented as a format to describe transit service, not a prescription for how to implement transit features in an application. Over time, some data producers and consumers have created guidance on what constitutes good GTFS, and a worldwide community of practice has developed. But the format itself is meant to describe offered transit service (and increasingly a few operational details).

**GTFS-TravelRestictions** could serve many applications: routing, timetable generation, arrival displays, real-time system configuration, analysis of transit service and access. Data with GTFS-TravelRestictions gives these applications a clear description of the offered service. Those applications can choose how to represent that service.

Implementation notes, multiple reference implementations (especially open-source), how-to resources, and best practices all help people understand and use GTFS. It would be helpful to have a body of practice (instructions, open-source reference applications) to guide implementation of GTFS-TravelRestrictions, but I think we lose more than we gain by waiting. We don't have precedent to wait: with previous changes to GTFS, we've only required at least one reference dataset, one reference implementation and a vote.

I urge that we should move towards adding this to the specification to facilitate adoption and develop the community of practice. On the question of the relationship of "GTFS trips" to real vehicle trips, that does seem like it could be useful to define in the spec or in a best practice. Adoption of GTFS-TravelRestrictions would support that -- making the travel restriction workaround involving duplicated trips unnecessary.

skinkie:
> we've only required at least one reference dataset, one reference implementation and a vote.

Is Apple supporting this at this moment in production?

abyrd:
From @skinkie's response, it seems to me that we have a consensus from a technical point of view on a good way to represent travel restrictions. The problem seems to be procedural: questioning whether the vote on April 9 was premature, due to lack of an implementation.

As @antrim pointed out, the requirement has long been one producer (ideally a public example feed) and one consumer, not necessarily open source. It may be a good idea to make provisions for, or even require open source examples or pseudocode, but that's not currently part of the process and amendments to procedure should be addressed separately. 

When I said I'd be in favor of this PR, that was based on comments above stating Bileto is both producing and consuming this proposed extension, and Apple is consuming it. I checked out the public feed examples which looked fine, and I was giving the benefit of the doubt about the producers.

Could @saraemarcus and @jspetrak please confirm that you have production systems consuming this GTFS extension to resolve this point definitively?

At least part of the confusion here is due to different sets of assumptions. As @antrim pointed out, the (to my mind very intuitive) assumption that only one GTFS trip corresponds to a particular vehicle at a particular time is not stated anywhere, but it could eliminate future misunderstandings to make it more official. Again, that's something that can and should be done separately later, but I think it's a good idea.

I believe the only technical thing remaining to clear up is the relationship to the existing pickup/dropoff codes: is this proposal is layered on top of them, or does it totally override them? Or is it enabled by a new value in those fields? However, I don't see much harm in clarifying this detail later if necessary, it seems to layer easily on top of the contents of this PR.

I also took a look at the pseudocode in the proposal documents explaining trip expansion. It is good that we already have some pseudocode there, but it seems to be expressed in terms of the most extreme case, making the problem look worse than it is for the average case. It should be possible to perform the expansion in terms of unique sets of restrictions on stop_times. I have some draft pseudocode and should soon have an alternative to propose.

abyrd:
> > At least two ways have been proposed to ensure simpler or legacy software does not misinterpret the new data or produce misleading results.
> 
> I know about the pickup_type, what is the other one?

Both involve pickup_type. One is to set them all to "no pickup" (old consumers ignore the trip since it never picks anyone up), the other is to set them all to some new value (old consumers see the trip as broken).

skinkie:
I would like to add to the post of @abyrd that we want to be one of the supporting **producers**. Hence I am more then willing to produce a conforming feed. But it should be tested with a consumer, which I am trying to find in this issue.

> Both involve pickup_type. One is to set them all to "no pickup" (old consumers ignore the trip since it never picks anyone up), the other is to set them all to some new value (old consumers see the trip as broken).

My personal preference is to have a _new value**s**_, because it it is likely that you want to retain the pickup or dropoff value for the non restricted part of the trip. I would not be voting against this if not proposed, but I think this kind of addition would guard for invalid usage better than just a new file.

antrim:
Just a note that @e-lo opened a discussion following on my earlier notes 2 days ago https://github.com/google/transit/pull/205#issuecomment-696357304 about assumed purposes of GTFS.

The discussion is more general than GTFS-TravelRestrictions. Hopefully discussion can help future spec conversations be more focused. Discussion prompts about the scope of GTFS are on the MobilityData slack: https://mobilitydata-io.slack.com/archives/C0J4TJY8L/p1600720624002700. (If you're not yet a member of the MobilityData Slack, you can sign up here: https://mobilitydata-io.herokuapp.com)

jspetrak:
@abyrd I am sorry for late response. To your question

> Could @saraemarcus and @jspetrak please confirm that you have production systems consuming this GTFS extension to resolve this point definitively?

Yes, we had a production-running implementation since 2017. The export part (an extension of [tulsawebdevs/django-multi-gtfs](https://github.com/tulsawebdevs/django-multi-gtfs)) was gradually updated according to changes in this proposal. The import part in our two different journey planning systems (one proprietary internal development, second is custom version of [Umotional AnyRoute](https://umotional.com/cs/anyroute/)) are running the first iteration of the proposal but can be updated easily.

PR #206, opened 2020-02-28
Proposal: Recommend providing TripUpdate predictions for the next trip in block

barbeau:
Previous discussions (1) (2) have raised the question of whether more than one TripUpdate can be included in a GTFS-realtime feed for a vehicle in the case where the vehicle is serving more than one trip in the same block. 

These discussions have concluded that multiple TripUpdates per vehicle are clearly beneficial in this case to avoid prediction "pop-in" for riders as the vehicle transitions from one trip to another and also give riders advance notice of delays that impact downstream trips (e.g., when the known delay exceeds planned layover times between trips).

However, this issue currently isn't addressed anywhere in the spec.

This proposal adds language to the spec recommending that multiple TripUpdates are included in a GTFS-realtime feed for vehicles running more than one trip in the same block. It also normalizes trip-updates.md and reference.md language related to blocks (future work may combine these files).

1. GitHub PR proposal - "Define rules for propagating delays across trips in same block" - https://github.com/google/transit/pull/110
1. Google Groups post - "trip is in TripUpdates but not in VehiclePositions feed" - https://groups.google.com/forum/#!topic/gtfs-realtime/lVPOmi9A5vQ

Announced on the GTFS-realtime Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/69b__hYzKNc.


gtfs-realtime/spec/en/reference.md
```diff
@@ -133,7 +133,9 @@ Depending on the value of ScheduleRelationship, a TripUpdate can specify:
...
+If a vehicle is serving multiple trips within the same block (for more information about trips and blocks, please refer to [GTFS trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt)):
+* the feed should include a TripUpdate for the trip currently being served by the vehicle as well as a another TripUpdate for the next trip in the block. Including multiple TripUpdates for the same vehicle avoids prediction "pop-in" for riders as the vehicle transitions from one trip to another and also gives riders advance notice of delays that impact downstream trips (e.g., when the known delay exceeds planned layover times between trips).
```
juanborre: This sentence is limiting the propagation to the next trip.
We could open the door to including more than one continuation trip. 
If the trips are very short, that will be useful.

gtfs-realtime/spec/en/trip-updates.md
```diff
@@ -4,6 +4,10 @@ Trip updates represent fluctuations in the timetable. We would expect to receive
 
 There should be **at most** one trip update for each scheduled trip. In case there is no trip update for a scheduled trip, it will be concluded that no realtime data is available for the trip. The data consumer should **not** assume that the trip is running on time.
 
+If a vehicle is serving multiple trips within the same block (for more information about trips and blocks, please refer to [GTFS trips.txt](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#tripstxt)):
+* the feed should include a TripUpdate for the trip currently being served by the vehicle as well as a another TripUpdate for the next trip in the block. Including multiple TripUpdates for the same vehicle avoids prediction "pop-in" for riders as the vehicle transitions from one trip to another and also gives riders advance notice of delays that impact downstream trips (e.g., when the known delay exceeds planned layover times between trips).
```
juanborre: Same as above.

gtfs-realtime/spec/en/reference.md
barbeau: @juanborre Yes, I think this is a key part of this proposal to decide on. We need to balance suggesting providing more than one TripUpdate, without encouraging producers to share too many TripUpdates for future trips to the point where predictions are meaningless.

The current text was an attempt to explicitly suggest providing predictions for the next trip (Trip B, which usually makes sense), without explicitly addressing additional trips - so my intent wasn't to prohibit Trips C, D etc. if they make sense in a certain context.

Do you have any suggestions for different language that strikes this balance?

mpaine-act: If past trips in block continue to be included as TripUpdate entities, is this allowed?  Purpose for this is for block-based systems to deterministically verify conclusion of full service.

gtfs-realtime/spec/en/trip-updates.md
```diff
@@ -4,6 +4,10 @@ Trip updates represent fluctuations in the timetable. We would expect to receive
...
+* the respective TripUpdate entities are not required to be added to the feed in the same order that they are scheduled in the block. For example, if there are trips with `trip_ids` 1, 2, and 3 that all belong to one block, and the vehicle travels trip 1, then trip 2, and then trip 3, the `trip_update` entities may appear in any order - for example, adding trip 2, then trip 1, and then trip 3 is allowed.
```
skinkie: Is there a reason why not encourage this production? I think if we would suggest tripUpdates should be sorted in departure order we will cover all situation, won't we?

barbeau: @skinkie Good question - although this particular PR isn't intended to address block ordering of TripUpdates (the 2nd bullet point shows up in the diff because I re-organized the text). There was previous discussion on the block ordering of TripUpdates in https://github.com/google/transit/pull/176, and some pushback on encouraging they be sorted. If you'd like to discuss further I suggest we open another issue focused just on that.

gtfs-realtime/spec/en/reference.md
barbeau: @mpaine-act Yes, my new text is not intended to prohibit the inclusion of past TripUpdates in any way. In fact, there is text just above and below the text I added that says:

>The updates can be for future, predicted arrival/departure events, or for past events that already occurred. 
>...
>Note that the update can describe a trip that has already completed... Note that the feed provider is allowed, but not required, to purge past updates

juanborre: I understand. On the same line of thinking, providing a prediction for the immediate next trip can be as meaningless as providing a prediction for the trip N+3. It will depend on the duration of the trips and how far in the future we are.

How about including reliability in the phrasing, like:
> The feed should include a TripUpdate for the trip currently being served by the vehicle. If the producer is able to provide reliable real time information for the next trips that the vehicle is doing in the block, they are encouraged to include TripUpdates for those trips. The producer should only provide updates which they are sure of.

gtfs-realtime/spec/en/trip-updates.md
skinkie: +1 lets do that.

gtfs-realtime/spec/en/reference.md
ibi-group-team: We like the changes in the revision from @juanborre , notably that multiple TripUpdates "are encouraged" rather than "should" be included and the emphasis on providing more than one only if producer is confident of the predictions for future trips. Here is a suggested revision:
 

> The feed should include a TripUpdate for the trip currently being served by the vehicle. Producers are encouraged to include TripUpdates for one or more trips after the current trip in this vehicle's block if the producer is confident in the quality of the predictions for these future trip(s).

barbeau: @juanborre and @ibi-group-team Thanks for the suggestions! I've revised the text as suggested in @ibi-group-team's [latest post](https://github.com/google/transit/pull/206#discussion_r385863796) in https://github.com/google/transit/pull/206/commits/bb8a727cbb060af23bb72e94e62c97a32f8f5b68.

barbeau:

Vote will be closed on Thursday March 19th at 23:59:59 UTC.

paulswartz:
+1

darylweinberg:
+1 as long as it is optional
Currently Swiftly only provides predictions for current time +45 minutes. We will need to work closely with consumers if implementing as this will break at least one of my consumers.

barbeau:
@darylweinberg Yes, including more than on TripUpdate per vehicle is not strictly required by this change, although it is recommended. It's at the discretion of the producer, with the goal of providing predictions in the feed as far ahead as the producer is confident in to provide better information to the rider.

>We will need to work closely with consumers if implementing as this will break at least one of my consumers.

Do you mean that existing consumers aren't currently showing information for future trips in a block (i.e., they are only showing information for the trip the vehicle is currently serving)?

darylweinberg:
Generally speaking, my consumers have worked around this by identifying the block a vehicle is serving and carrying the last prediction, allowing for a layover to “catchup” on any lateness, into the next trip,

Block A, Trip 123 is +5 minutes late at the last prediction for the trip
There is a 3 minute layover before the next trip on the block, Trip 456.
The prediction displayed for the departure from the first stop of Trip 56 is +2 minutes late.

One of my concerns, if receiving updates for multiple trips served by the same vehicle, will drop the first update. This would result in losing the information for the current trip. It has ben my experience that many consumers code to what we are producing at the time the contract was signed, not to the specification as it exists at the time let alone future iterations.

In future procurements we are trying to change this.

Thank you,

Daryl Weinberg
Transit Systems Architect
o: 512-369-6216
|
|

jrsanbornjr:
+1

harringtonp:
+!

skinkie:
I still dislike the fact that the any order is allowed almost encouraged, while this is trivial to get sorted on the producer side.

barbeau:
>I still dislike the fact that the any order is allowed almost encouraged, while this is trivial to get sorted on the producer side.

@skinkie That's fair, although that's not what we're voting on in this PR (that text shows as green because I combined the two block-related rules into one paragraph). The vote in this PR is only for the first bullet point for the recommendation that TripUpdates for subsequent blocks should be included in feeds - in other words, this PR doesn't change the existing statement that producers aren't required to order TripUpdates within the feed by block-order.

I plan to open another issue to discuss the ordering of TripUpdates by block-order.

barbeau:
@harringtonp Just to clarify, that `+!` is `+1`?

harringtonp:
Oops sorry +1      From my perspective (I don't propagate predictions to future trips in a block) the change is a no brainer

ibi-group-team:
+1

barbeau:
* Yes - 5
* No - 0
* Abstain - 0

So it passes! Thanks all!

gtfs-realtime/spec/en/trip-updates.md
barbeau: Issue opened for discussion at https://github.com/google/transit/issues/207.

PR #212, opened 2020-04-09
Occupancy v2

gcamp:
Update : new thinking available here https://github.com/google/transit/pull/212#issuecomment-669402378

OccupancyStatus was added 6 years ago as an experimental field to the GTFS-rt spec. Multiple producers and consumers have adopted the value but the implementation has currently some problems. 

The enum of `occupancy_status` has 7 values, however we've seen that the range of values is not used in practice. We've seen consumer mostly use 3 values to represent the entire range of occupancy. For example : 
- New South Wales buses only use the `MANY_SEATS_AVAILABLE`, `FEW_SEATS_AVAILABLE` and `STANDING_ROOM_ONLY`. 
- Some New South Wales trains uses all the values available except `EMPTY` and `NOT_ACCEPTING_PASSENGERS`. However, the agency suggest to never display more than 3 textual values for the consumer facing message. More granularity can be provided with colors. 
- MTA NYC has limited availability of the feature in their feed but do use `FEW_SEATS_AVAILABLE`, `STANDING_ROOM_ONLY`, `FULL`. 

For this reason I would suggest to only give 3 values in the enum. `SEATS_AVAILABLE`, `STANDING_AVAILABLE`, `FULL`. The naming changed slightly to reflect the same 3 values available in SIRI. 

~However, we do lose a lot of granularity of how fine the occupancy value can be provided. To alleviate this problem I would suggest to add a second value `occupancy_percentage` to give a more accurate value of occupancy.~

~It would be percentage value representing the degree of passenger occupancy of the vehicle. The values are represented as an integer without decimals. 0 means 0% and 100 means 100%. The value 100 should represent the total maximum occupancy the vehicle was designed for, including both seated and standing capacity. It is possible that the value goes over 100 if there are currently more passengers than what the vehicle was designed for.~

~It was decided to have both values because they fulfill two different purpose. `occupancy_status` is provided to give a understandable value for the user (`STANDING_ROOM_ONLY` is way more understandable for a user than 50% full). `occupancy_percentage` is provided to give more granular data.~

The `occupancy_percentage` is now part of a separate PR : https://github.com/google/transit/pull/213


skinkie:
To me these are two separate issues.
Reducing the enum: -1
Adding the percentage: +1

I would even suggest to add:
free seats available (int)
mobility impaired seats available (int)

juanborre:
@googlebot I consent

gcamp:
@skinkie you don't think that having producer independently using 3 values of the enum but not always the same 3 values is a problem?

nighthawk:
As a consumer and end user of this data, I think this is good.

As an end user (in Sydney), the existing seven values very often were of the nature "precise, but not accurate", i.e., they have all these different cases but what they *say* rarely matches what a bus looks like inside and what the actually availability of seats or standing room is, or whether the bus was in fact still accepting passengers or not. Limiting the enum cases moves this more towards "less precise, but more accurate".

The enum cases were more useful as a data consumer and having more of a "fill state" where it's about a percentage rather than whether the enum string value makes sense, but this PR addresses this by having a percentage.

So +1.

nighthawk:
(And, agreed, having a separate indicator for "mobility impaired seats available" would be fantastic.)

skinkie:
@gcamp this is a problem, but an implementation / validation detail.

mike-swiftly:
I agree with skinkie that there are two separate issues here. A hearty +1 for including percentage (and allowing it to be above 100%), but a -1 for reducing the enumerated types since it appears that some agencies could already be using the other values.

stevenmwhite:
I'm with @nighthawk on the "precise but not accurate" description. As a producer on behalf of agencies, we've seen disparities within our customer base on what all of these mean to them. In addition to that, it's difficult to predict whether riders are going to sit or stand and being so precise on whether seats are available is misleading. Certain types of services are more likely to have riders standing even if seats are available (circulators like LADOT's DASH) than others where very few people stand and sometimes aren't even allowed to (commuter routes). We've found that you may map a percentage to "FEW SEATS AVAILABLE" on a service where the bus is pretty full, but because most people are standing there's actually _many_ seats available in reality. The precision here is very often not _accurate_.

I'm not quite convinced that the three proposed enums are the right ones, particularly because they makes what appears to be a clear statement about sitting vs standing.

I'm an absolute +1 on adding the percentage, and I agree with it being the total seated+standing capacity.

barbeau:
BTW, if anyone wants to check out the Transit App feature based on this data, there is a Medium article here:
https://medium.com/transit-app/you-can-avoid-crowds-on-public-transit-with-new-real-time-crowding-info-b61e60f5502

I think the "precise but not accurate" argument mentioned by @stevenmwhite and @nighthawk is a good reason to reduce the number of enum values.

>I'm not quite convinced that the three proposed enums are the right ones, particularly because they still make what appears to be a clear statement about sitting vs standing.

IMHO, one strong argument for adopting the proposed enum values are that they exactly match the SIRI values for occupancy. In the long term, this will make it much simpler to develop compatible systems that can easily output the occupancy data in either format, rather than try to do mapping between slightly different semantics.

>but a -1 for reducing the enumerated types since it appears that some agencies could already be using the other values.

OccupancyStatus is still an experimental field, mostly because we weren't sure exactly how this would be used in practice when it was initially defined. So the fact that some agencies are using values that are deprecated in this PR should not alone be a reason not to make a change (also, the deprecation here is soft, meaning that agencies could continue publishing the values they are without any problem until they update their systems).

>you don't think that having producer independently using 3 values of the enum but not always the same 3 values is a problem?
>* this is a problem, but an implementation / validation detail. 

I disagree that this is an implementation / validation detail. 

Based on the current spec, different agencies can pick and choose which enum values they want to use. One agency may use `STANDING_ROOM_ONLY` and `CRUSHED_STANDING_ROOM_ONLY`, while another may not. This means as a consumer, for each system you don't know how many buckets are being used, and what the max value is. 

This significantly limits the way this information can be presented to the end user - to be consistent across providers (e.g., for Sydney riders transitioning from bus to train), you're basically forced to show the text description of the enum. If you show a graphic representation of "fullness" (for example, from Transit's article above:

image

..it's not clear what "fullness" you should be showing for any given producer.

And, if we have too many enum values where accuracy is an issue (as mentioned above), then the text descriptions of the enums aren't meaningful either.

darylweinberg:
Capital Metro plans to implement OccupanyStatus this summer. Our current plan is to use all the enum values for granularity. Our contracted partners (consumers) will then be displaying Low, Medium, High or N/A to our customers via our mobile application and dynamic message signs with the below guidance:


     *   Empty: Occupancy – Low
     *   Many_seats_available: Occupancy – Low
     *   Few_seats_available: Occupancy – Medium
     *   Standing_Room_Only: Occupancy – Medium
     *   Crushed_Standing_Room_Only: Occupancy – High
     *   Full: Occupancy – High
     *   Not_accepting_passengers: Occupancy – High
     *   No Real-Time or no information: Occupancy – N/A

Third party consumers will then have the ability to display as they see fit in their applications.

+1 to adding a percentage.

Thank you,

Daryl Weinberg
Transit Systems Architect
o: 512-369-6216
|
|

skinkie:
> IMHO, one strong argument for adopting the proposed enum values are that they exactly match the SIRI values for occupancy. In the long term, this will make it much simpler to develop compatible systems that can easily output the occupancy data in either format, rather than try to do mapping between slightly different semantics.

If I recall correctly SIRI has recently adopted all the fields from GTFS-RT.

> > you don't think that having producer independently using 3 values of the enum but not always the same 3 values is a problem?
> > 
> > * this is a problem, but an implementation / validation detail.
> 
> I disagree that this is an implementation / validation detail.
> 
> Based on the current spec, different agencies can pick and choose which enum values they want to use. One agency may use `STANDING_ROOM_ONLY` and `CRUSHED_STANDING_ROOM_ONLY`, while another may not. This means as a consumer, for each system you don't know how many buckets are being used, and what the max value is.
>
> This significantly limits the way this information can be presented to the end user - to be consistent across providers (e.g., for Sydney riders transitioning from bus to train), you're basically forced to show the text description of the enum. If you show a graphic representation of "fullness" (for example, from Transit's article above:

Lets go into this argument. A consumer should translate and present not conclude anything from a bucket value. You may disagree with that, but no operator wants a redaction of the information that has been supplied. Do you want to go into a debate about that? 


> And, if we have too many enum values where accuracy is an issue (as mentioned above), then the text descriptions of the enums aren't meaningful either.

If a system integrator only picks three values. That sounds like the system integrator is not able to provider a higher level of detail, isn't it?

barbeau:
>Lets go into this argument. A consumer should translate and present not conclude anything from a bucket value. You may disagree with that, but no operator wants a redaction of the information that has been supplied. Do you want to go into a debate about that?

I'm not sure I understand exactly what you're saying.

But, I think @darylweinberg post [above](https://github.com/google/transit/pull/212#issuecomment-611667724) is a good illustration of the challenge I'm trying to describe with the current enums.  

CapMetro has a clear intent on how they expect their contracted consumers to display these occupancy values in an ordinal way, mapping them to `Low`, `Medium`, and `High` values (which again, maps to 3 values as @gcamp mentioned earlier):

>* Empty: Occupancy – Low
>* Many_seats_available: Occupancy – Low
>* Few_seats_available: Occupancy – Medium
>* Standing_Room_Only: Occupancy – Medium
>* Crushed_Standing_Room_Only: Occupancy – High
>* Full: Occupancy – High
>* Not_accepting_passengers: Occupancy – High

But unless this intent is communicated via side-channels to all consumers, it can be hard to infer (again, beyond just showing a plain text version of the enumeration).

If we're trying to share data that can more directly map to graphic UI elements, then it seems like having more of a LOW, MEDIUM, HIGH ordinal representation would make this clearer.

>If I recall correctly SIRI has recently adopted all the fields from GTFS-RT.

@skinkie Do you have a reference for this? It looks like v2.0 at https://www.vdv.de/siri.aspx still has 3 values.

barbeau:
@darylweinberg For `Not_accepting_passengers` - is this data coming from your APCs, or is the driver somehow indicating that they are not accepting any more boardings due to load?

gcamp:
Adding on what @barbeau 's arguments, I would ask to people that are against removing values in the enum : what is the purpose of having that amount of granularity? 

If it's to display the value visually with levels then the percentage value fulfill that function and everybody seems to agree it's also needed. 

If it's to show the value textually and make some subtle differences between levels – for example standing room only and crushed standing room – it seems no consumer actually uses it that way. But if a consumer has that need I could reconsider. 

The one value I think could have value is not_accepting_passengers but I never saw producer actually produce it until @darylweinberg mentioned it here.

skinkie:
> But, I think @darylweinberg post [above](https://github.com/google/transit/pull/212#issuecomment-611667724) is a good illustration of the challenge I'm trying to describe with the current enums.
> 
> CapMetro has a clear intent on how they expect their contracted consumers to display these occupancy values in an ordinal way, mapping them to `Low`, `Medium`, and `High` values (which again, maps to 3 values as @gcamp mentioned earlier):
> 
> > * Empty: Occupancy – Low
> > * Many_seats_available: Occupancy – Low
> > * Few_seats_available: Occupancy – Medium
> > * Standing_Room_Only: Occupancy – Medium
> > * Crushed_Standing_Room_Only: Occupancy – High
> > * Full: Occupancy – High
> > * Not_accepting_passengers: Occupancy – High
> 
> But unless this intent is communicated via side-channels to all consumers, it can be hard to infer (again, beyond just showing a plain text version of the enumeration).
>
> If we're trying to share data that can more directly map to graphic UI elements, then it seems like having more of a LOW, MEDIUM, HIGH ordinal representation would make this clearer.

Capmetro should not invent their own values, this is a validation/implementation detail. This brings the argument back to a low-medium-high discussion. I am not supporting a change in this direction. I am happy to kill the enumeration and have a percentage instead, but even then the 100% should be defined as all seats are taken, and not "you can't fit an extra passenger".
 
> > If I recall correctly SIRI has recently adopted all the fields from GTFS-RT.
> 
> @skinkie Do you have a reference for this? It looks like v2.0 at https://www.vdv.de/siri.aspx still has 3 values.

It was a long search:
https://github.com/SIRI-CEN/SIRI/pull/13

skinkie:
> Adding on what @barbeau 's arguments, I would ask to people that are against removing values in the enum : what is the purpose of having that amount of granularity?

In my opinion the purpose of an enumeration is a value that can be unambigiously translated / localized and creates a path to clear communication. Anyone that just wants to go for a discrete value, should propose a discrete numeric value like a percentage, but especially for occupancy that percentage is ambigious in consumerusage.

An additional reason is that operators do not want to give away numeric values because they are absolute and factual. If the value does not want to match the actual free seats available people get angry, therefore try to use fuzzy descriptions.

barbeau:
>In my opinion the purpose of an enumeration is a value that can be unambigiously translated / localized and creates a path to clear communication.

I think the root of the problem here is that it's not clear to both producers and consumers if the current OccupancyStatus enumeration is ordinal, which is causing confusion. Producers are mapping a numeric % value to the enumerations, and at least some producers clearly state in documentation that they are treating them as ordinal. And most consumers are trying to extract ordinality from the enums, but without knowledge of the complete set of enums being used, it's difficult.

>I am not supporting a change in this direction. I am happy to kill the enumeration 

Instead of killing the enum, I'm in favor of trying to make it match what both existing producers and consumers seem to be trying to communicate and graphically represent - which appears to consistently be 3 ordinal values.

@gcamp There doesn't seem to be any objection to adding a field for the numeric value so far - should that be moved to it's own PR so it could become experimental?

barbeau:
@skinkie Thanks for the [SIRI link](https://github.com/SIRI-CEN/SIRI/pull/13/files)! The way I read the diff, the GTFS_RT enum values were added to the existing SIRI values? 

So the full allowed SIRI set is now:
* empty
* manySeatsAvailable
* seatsAvailable
* fewSeatsAvailable
* standingAvailable
* standingRoomOnly
* crushedStandingRoomOnly
* full
* notAcceptingPassengers

Do you have any examples for how these are represented in UIs to riders?

skinkie:
> @skinkie Thanks for the [SIRI link](https://github.com/SIRI-CEN/SIRI/pull/13/files)! The way I read the diff, the GTFS_RT enum values were added to the existing SIRI values?

The VDV454 values were extended with the GTFS-RT values, so SIRI could always exchange GTFS-RT information.

> Do you have any examples for how these are represented in UIs to riders?

I don't. I think - if the values of VDV454 - are ever used, it is likely so in the OJP/DJP/VDV Trias interfaces, but no concrete examples of GUI's. @sgrossberndt do you maybe have some context?

gcamp:
Hi everyone, I moved `occupancy_percentage` to a [different PR](https://github.com/google/transit/pull/213).

wafisher:
From the Long Island Rail Road:

The values we're considering for our new data feed of realtime occupancy are EMPTY, MANY_SEATS, SOME_SEATS, SRO, FULL and LOCKED_OUT.

At a minimum, we do consider SRO and FULL to be different. In one case you can get on the car and squeeze in. In the other case, you physically can't get on the car. The different might be represented between a royal red and dark red, a color distinction that Google Maps has made on its traffic layer for a while now and thus one we expect users will grok.

LOCKED_OUT is when a crew will specifically lock the doors of a married pair, which they are allowed to do mid-day when there's low ridership. It is effectively full though if you color a car dark red and people see the cars are empty, they might think the sensors are faulty.

gcamp:
One more piece of information. MTA NYC themselves use only 3 value visually in their app. `FEW_SEATS_AVAILABLE` is also reported as "Seating Available" which is obviously not in line with what we show in Transit. 

| MTA  | Transit |
| ------------- | ------------- |
| image  | Screenshot_20200504-191149 (1)  |


I've also started [a discussion on the MTA Google group](https://groups.google.com/forum/#!topic/mtadeveloperresources/ZGhcW192VLw).

barbeau:
@gcamp I believe MTA uses SIRI internally to represent these values, which they've then mapped to GTFS-RT OccupancyStatus. So IMHO that would be another reason to align with SIRI, as we'll continue to have these discrepancies for any SIRI feed that gets converted into GTFS-RT OccupancyStatus.

antrim:
While we are figuring out the future of occupancy in GTFS, which may take some time, I've proposed guidance on how to use OccupancyStatus while Covid-19 social distancing practices are in effect -- Issue #222. This is not meant to be a permanent change to the specification, just a clarification of practice during the global pandemic and while we are working out occupancy v2. Please comment if you have thoughts.

barbeau:
Hi all! For everyone that participated in this discussion for improving OccupancyStatus - could you please take a look at https://github.com/google/transit/issues/223?

Please comment there for which you prefer - OccupancyStatus vs. occupancy_percentage.

flocsy:
-1 removing enums
I don't think we should remove enums. I don't understand why a consumer would think that enums should be "distinct". It is ok, that the standard has possibly "overlapping" values like: FEW_SEATS_AVAILABLE, SEATS_AVAILABLE, MANY_SEATS_AVAILABLE. I would recommend to set some relation between the values, like:
FULL <= STANDING_ROOM_ONLY <= FEW_SEATS_AVAILABLE <= SEATS_AVAILABLE <= MANY_SEATS_AVAILABLE <= EMPTY
The point is that the producer should chose the values it feels it can accurately provide, but a consumer should not assume that from EMPTY it'll transfer to MANY_SEATS_AVAILABLE, etc. It is possible the producer only uses some of the enum values.

+1 adding percentage
The same goes for the percentage: a consumer can't assume how precise it is. Maybe it's jumping by 2, or 10, or 25 or 33% steps. It's up to the producer.

willwong430:
I had commented on #223 expressing the MTA's preference to use occupancy_percentage as our primary indicator for our realtime APC feed.  While this is in motion, I agree with reducing the enumeration to match SIRI’s.

CleverDevices-gtfs:
Clever Devices believes some enum values should be combined because they are confusing or open to broad interpretations:
o	MANY_SEATS_AVAILABLE and FEW_SEATS_AVAILABLE should be combined into SEATS_AVAILABLE.  There is no information about what kind of seats are available and calculating this as a producer is hard (ie, if the bus is carrying its full complement of wheel chairs, number of available seats will drastically drop)
o	CRUSHED_STANDING_ROOM and STANDING_ROOM_ONLY removed in preference for “FULL”.  From a customer decision making perspective, standing room only and crushed standing room will not make a determination for a customer to board the vehicle if they have to get somewhere, “FULL” is a status that a customer will pass on a vehicle because he or she cannot board the vehicle.

TCAT-UW:
I'm first and foremost concerned about the non-specificity of "mobility impaired seats available." There is a distinction between a passenger who is mobility impaired and requires preferential seating near the front of the bus and a passenger who requires wheelchair tie-downs (that are typically coincident physically with several preferential seating spots). 

Many buses I'm thinking of, including King County Metro buses, have the wheelchair tie downs and over them is a pulldown bench that seats 3 mobility-limited passengers with no wheelchairs. The distinction "mobility impaired seat available" can be ambiguous in the case of a passenger requiring a wheelchair tie-down, when 2 of the 3 mobility-limited preference seats are occupied by passengers. If the person using a wheelchair was to board, the driver is not likely to move the 2 passengers seated over the bench in order to stow it away and reveal the tie-down.

I contend that there ought to be a "number of available tie-down number" specific tag that would clearly demarcate for users who would be completely unable to board the vehicle without it. This is in contrast to passengers who may have a preference for MANY SEATS AVAILABLE, even though I see the enumerated types to be problematic as well due to the subjective nature of the descriptors and the reliance on bus drivers? or crowds? to report these data. 

Having precise counts allows any downstream data consumer and mobility app to provide its users whatever enumerated type they want, while removing drivers and passengers from the data loop would provide actual countable measures for furnishing such information.

gcamp:
I want to reopen the discussion here since a lot happened since. 

MBTA released a their implementation of occupancy with 3 buckets with different labels than what is currently in the occupancy spec : "Not crowded", "Some crowding", "Crowded". 

I'll summarize their approach but you can also listen to MBTA's part of the webinar that we did here : https://youtu.be/5Yai8oV1Y3Q?t=192. In summary, using 3 buckets with allow to be accurate even if less precise than the current spec (as discussed above). In addition, those 3 labels are labels that users understand and that is actionable. 

At Transit we recently decided to take a very similar approach, in our UI we now use the 3 buckets as the main interface for crowding as shown below

IMG_5282

I would also add that since this issue was created we've seen more agencies use more than 3 values for their feed, but the vast majority of producers still use 3 buckets.

stevenmwhite:
We use three values as well -- though I believe they are different than the three used by MBTA. 

The terminology of "**Not crowded**", "**Some crowding**", "**Crowded**" is much better than indicating whether or not _seats_ are available (which is not necessarily knowable simply based on the number of people on the bus) and we'd support standardizing around those three.

barbeau:
>The terminology of "Not crowded", "Some crowding", "Crowded" is much better than indicating whether or not seats are available (which is not necessarily knowable simply based on the number of people on the bus) and we'd support standardizing around those three.

@stevenmwhite I agree. If the purpose of the `OccupancyStatus` enum is to define buckets in a range of values, which seems to be how most consumers and producers are using them, then IMHO "**Not crowded**", "**Some crowding**", "**Crowded** is a good, generalizable solution to all vehicle types.

If the purpose of `OccupancyStatus` enum is to provide a text description of occupancy on a vehicle (e.g., if seats are available, standing room only), then the existing enums work (but need clarification that they should only be used this way). But it doesn't seem like anyone (?) is actually using the enums this way - producers and consumers are instead mapping numeric ranges to the enums, which causes problems when the vehicle doesn't have seats or standing room or producers choose to publish only a partial subset of the enum values.

mike-swiftly:
I have seen other agencies use a very different set of values from the 3
emphasized here. I think the info provided by client apps should not be
limited to the enumerated types in the GTFS-realtime spec. Client apps can
likely come up with better wording. But that doesn't mean the spec should
be changed to match what a particular client provides.

*Michael Smith*
*CIO, **Swiftly, Inc*
(415) 260-4700

gcamp:
@mike-swiftly can you share which feed you have seen with more than 3 values in the enum if they are public or if they used differently? I have seen new ones since my first message of this thread but they are all in private beta.

mike-swiftly:
The ones I was referring to are still not public.

*Michael Smith*
*CIO, **Swiftly, Inc*
(415) 260-4700

jxeeno:
Auckland Transport has 4, `EMPTY` in addition to the typical three buckets.

scmcca:
@gcamp FWIW the [STM](http://www.stm.info/en/press/press-releases/2020/stm-introduces-occupancy-levels-for-metro-network-customers) also uses 4, as described by @jxeeno. 

I think that `EMPTY` in addition to `NOT_CROWDED`, `SOME_CROWDING`, `CROWDED` is useful as it completes the list of possibilities. `NOT_CROWDED` does not distinguish `EMPTY`, albeit nested. 

A viable option could be to make `NOT_CROWDED`, `SOME_CROWDING` and `CROWDED` required, and `EMPTY` optional.

barbeau:
Prior to the above two mentions IIRC no one was using `EMPTY` in the last survey of existing producers we did for `OccupancyStatus`. I'm not sure how useful it is in practice as due to count errors it may be hard to definitively say that a vehicle is empty. It's easy to be wrong if you publish `EMPTY` and you're off by a few counts. I'm guessing publishing `NOT_CROWDED` would be preferred for most producers. I'd be curious to know how often these producers are actually publishing `EMPTY` values and how confident they are in that data.

@paulswartz Do you have any wisdom from the user studies that MBTA did on this? Was there a specific reason MBTA chose not to use `EMPTY`?

I'd be against making one of the enum values optional in a new scale as IMHO that's one of the primary challenges with the current enums - I'd prefer to have a fixed required scale for consistency.

paulswartz:
We didn't find rider benefit to the `EMPTY` vs `NOT_CROWDED` distinction. Once a vehicle isn't crowded, it doesn't matter to them exactly how not crowded it is.

rcbouchard:
Hi, MobiltyData invited us to discuss about our experience regarding Occupancy from STM. 
We have chosen the levels that we think are the clearest for the customer
1) Level 1: Distancing possible (or empty) - this is specific to the pandemic situation and could disappear when the pandemic is behind us (to be reviewed at that time). we have a very low ridership in the buses.
2) Level 2: the customer can sit down
3) Level 3: Customers can expect to be standing
4) Level 4: the bus / train is almost full

At the moment, 3 levels would be too restricted, because we could not communicate the situation of physical distancing. I also find that the levels ((NOT_CROWDED, SOME_CROWDING, CROWDED) are ambiguous and can be interpreted differently depending on the client. This is why we decided to be a little more specific. Obviously, all of this could be adjusted if we realize that customers do not fully understand these levels.

René-Claude Bouchard
Société de transport de Montréal (STM)

flocsy:
Regarding the pandemic: IMHO it can be mapped to the existing values. In places where the distancing is enforced (they don't let more people to board) then "distancing not possible" == "full", and where it's not enforced, there "distancing not possible" == "crowded"

francispeixoto:
@flocsy In response to the pandemic, we've implemented in our ITS a ratio system that allows to differenciate between maximal occupancy per vehicle type from desired occupancy. 

Say a given bus model has 45 seats, but to respect the two meter rule, you want to limit occupancy to 20 seats, our users can set a ratio (in %) of maximal occupancy that will drive the GTFS-RT occupancy status, coupled with our people counter data. 

In this case, the 45 seat bus would report "standing room only" if 20 people were on the bus at that time.

scmcca:
It seems that opinions vary greatly on changing `OccupancyStatus`. 

The original problem points to difficulties for consumers to consistently interpret qualitative uses of `OccupancyStatus` on a quantitative scale, with the goal of distilling occupancy descriptions to generalized and meaningful information for riders. Some agencies and consumers have adopted the practice of using 3 enumerations for consistency (cc: @gcamp), backed by MBTA's research (cc: @paulswartz) that these are the most meaningful to riders. This standardization makes it easy to consistently interpret `OccupancyStatus` across parties. 

At the same time, it has become clear in this thread and issue #223 that the ability to qualitatively describe occupancy levels with `OccupancyStatus` is distinct from mapping occupancy on a quantitative scale. This is characterized by explicit -1's from stakeholders in this conversation on reducing enums (@skinkie, @mike-swiftly, @flocsy), suggestions to add qualitative descriptors (@skinkie, @darylweinberg, @TCAT-UW), and the prevailing support for adding `occupancy_percentage` (#213) as experimental to address the need.

Meanwhile, among those who advocate a reduction of the enums, there is not a consistent agreement on how many there should be. @gcamp, @paulswartz, @barbeau, and @stevenmwhite advocate for 3 enums, while Auckland Transport (cc: @jxeeno) and STM (cc: @rcbouchard) use 4.

To summarize, it seems that the addition of `occupancy_percentage` poses a productive solution to an otherwise unproductive conversation.

Does `occupancy_percentage` address the need we are contending with without disrupting historical and varying qualitative use of `OccupancyStatus`? If so, what are everyone's thoughts on making `occupancy_percentage` official?

Figuring this out is a dependency for #240, a proposal for providing expected or usual occupancies in static. It should not be adopted with occupancy descriptors that are experimental or contested in GTFS Realtime. We would like consistency between the two so that they are representatively interchangeable, and that descriptors adopted in static do not cement anything that is experimental in GTFS Realtime. 

Welcoming different opinions!

RTL-Gilles-Ferland:
As an agency. We prefer the occupancy status enum as it directs the data consumers on what we want to publish to the clients, it also leaves a little room for interpretation (i.e. error) which is favorable in the static estimated occupancies. It also allows us to control the value published according to the vehicle size. We prefer a 4 level enum. (Empty, some seats available, standing room only, full)

We feel that occupancy percentage leaves the data consumers / clients with the responsibility to understand what 64%, for example, actually means. Since the transit vehicles vary in size, number of seats and standing room.

flocsy:
I think that the standard has to address 2 issues that came up:

1. producers will need to be able to use it "conveniently". What I mean by that is that there might be technical difficulties giving accurate percentage, so a producer should be able to decide if they can/want to give percentage or enums.

2. consumers will need to be able (with the help of the standard) understand/interpret the data. They might decide to display it different from the original data. They might map the percentage to enums, they might translate the enums different during a pandemic, etc. They might even do the opposite: "translate" enums/percentage to a scale and display it that way (like green/yellow/orange/red)

What's important is to give enough flexibility to the producer on one hand and enough confidence to interpret it in the correct way for the consumer on the other hand. So maybe the enums should not have a meaning hard coded in their name. Maybe there should be a configuration that the producer publishes with their enums. For example one producer would decide to use 3 enums, another to use 5. So we could call them: occupancy0...occupancyN, where N is given by the producer in the config. The standard would state that occupancy(M) < occupancy(M+1), and that it is recommended that they're "spaced" equally to make it easier to map them on a scale.

scmcca:
@flocsy 
> there might be technical difficulties giving accurate percentage

I agree. A good practice would be to use percentage buckets, such as at increments of 20 points (0-20, 20-40, 40-60, 60-80, 80-100). Providing fine-grained percentages would surely be subject to false promises, and would be unnecessary for rider understanding.

> consumers will need to be able (with the help of the standard) understand/interpret the data. They might decide to display it different from the original data. They might map the percentage to enums, they might translate the enums different during a pandemic, etc. They might even do the opposite: "translate" enums/percentage to a scale and display it that way (like green/yellow/orange/red)

Perhaps this could be governed by a best practice instead of making hard changes to the existing `OccupancyStatus` enum list that are in use by others.

> So we could call them: occupancy0...occupancyN, where N is given by the producer in the config. The standard would state that occupancy(M) < occupancy(M+1), and that it is recommended that they're "spaced" equally to make it easier to map them on a scale.

A challenge here is that depending on the seating and standing space available in a given vehicle, the existing `OccupancyStatus` enums are not always spaced equally. This was echoed by many in #223. However, I think it is a pertinent point that there needs to be some reference to interpret qualitative or discrete enumerations, especially when used differently across producers and consumers. It seems that  `occupancy_percentage` can be used to accomplish this (i.e., we already have a solution). The producer is able to map what qualitative markers of occupancy mean on a continuous-scale, for as fine-grained or coarse as the number of `OccupancyStatus` enums are provided. Whether producers actually provide both in this way may not be governable, but again, could be a best practice for the disambiguation of occupancies between producers and consumers, and more importantly, to riders.

@gcamp @barbeau I'm wondering if we could reframe the PR to not deprecate experimental enums that are in use by others, but simply to make official 3 of them?

barbeau:
@scmcca which 3 enums would you make official?

scmcca:
Based on the PR for 3 enums, I propose to make official:
- `MANY_SEATS_AVAILABLE`
- `STANDING_ROOM_ONLY`
- `FULL`

We could also advocate (before or after) that `MANY_SEATS_AVAILABLE` be changed to `SEATS_AVAILBLE`, and that `STANDING_ROOM_ONLY` be changed to `STANDING_AVAILABLE`. The meanings would be essentially the same within themselves and within the context of the other listed experimental enums, while avoiding false promises brought on by "only" language.

Over time, there is the opportunity to reduce this list even further if stakeholders come to find that `FEW_SEATS_AVAILABLE` can be consolidated to `SEATS_AVAILABLE`, and `CRUSHED_STANDING_ROOM_ONLY` can be consolidated to `STANDING_AVAILABLE`.

RTL-Gilles-Ferland:
@scmcca We agree with your proposal.

As for changing names to SEATS_AVAILBLE, we think it works. But for STANDING_AVAILABLE, we think preferable to keep the "only" since standing is by default always available. What we want to communicate by that value is that there is no more seats available.

devadvance:
One of the aspects of percentages that I found thought-provoking: a percentage does not assume information about the type of accommodation available, which is dependent on both the vehicle and the rider's needs/preferences.

Is there an opportunity in this enum proposal to be more inclusive of individuals for whom `SEATS_AVAILABLE` or `STANDING_AVAILABLE` are not sufficient or purely scalar information for making decisions about their journey? E.g., while `SEATS_AVAILABLE` implies more space than `STANDING_AVAILABLE`, only the latter may be useful for someone with a mobility aid.

stevenmwhite:
> One of the aspects of percentages that I found thought-provoking: a percentage does not assume information about the type of accommodation available, which is dependent on both the vehicle and the rider's needs/preferences.

This is also the issues I've had with the current enums.

Unless your technology is actively reporting on individual seats, you can not guarantee that "only" standing room is available. Many people on transit will stand even if a seat is a available, and thus there is often more people on board than the seats accommodate, and yet there are still seats available.

While a percentage does not give information about the types of accommodation available, at least it is universally understood as a scale where 0 is always < 50 < 100. The same cannot necessarily be said for the enums that specifically call out seating vs standing.

Of course, if the technology used _is_ actively tracking the occupancy of seats, that's a different story, but this is very few and far between if used at all.

scmcca:
@stevenmwhite I'm wondering if instead of changing the meaning of enums that are in use and are being advocated for by other stakeholders, that we instead clarify in the spec some guidelines around `occupancy_percentage` to leverage general descriptions of `Not crowded`, `Some crowding`, `Crowded`. Something like:
- `0-50` - `Not crowded`
- `50-75` - `Some crowding`
- `75-100+` - `Crowded`

barbeau:
I'd be concerned that officially adopting a subset of the current `OccupancyStatus` enum doesn't really solve the problems that currently exist with it. Along with the issues raised above, another problem with making a subset of the current `OccupancyStatus` official and interpreting it as a range is that we won't know which producers are using the "official" interpretation vs. the experimental one when looking at the data.

I think there are 3 primary occupancy use cases that people are discussing:
1. Discrete occupancy state (`EMPTY`, `MANY_SEATS_AVAILABLE`, `FEW_SEATS_AVAILABLE`, `STANDING_ROOM_ONLY`, `CRUSHED_STANDING_ROOM_ONLY`, `FULL`, `NOT_ACCEPTING_PASSENGERS`) - These aren't necessarily continuous across a range or generalizable across all vehicle types, but instead they define labeled states the vehicle can be in. Best represented in a UI as the text version of the enumeration and should NOT be used as a scale. `OccupancyStatus` is currently defined this way in the spec.
2. Coarse bucketed range (could be `NOT_CROWDED`, `SOME_CROWDING`, `CROWDED`) - These are several generic, continuous, ordered values that represent a scale. Generalizes well across all types of producers and can be represented in a UI as a graphic element (e.g., filled person icons) or the text version of the enumeration. `OccupancyStatus` is currently being used this way by some producers and consumers, but doesn't work well because of the above-mentioned issues with not fitting to a fixed scale.
3. Fine numeric range (95%) - A fine grained version of (2) with actual numeric values. Best represented in a UI as a number. `occupancy_percentage` in the spec (although as @scmcca said, it sounds like many producers will bucket these values, making it more like (2) ).

One option is that we could keep `OccupancyStatus` to represent the vehicle state for (1) if producers really want the ability to represent when there is `STANDING_ROOM_ONLY` for example, but with the caveat that consumers shouldn't interpret these values as a scale - that would fall to a new field `OccupancyScale` (or better name) for (2). So a producer could publish `CROWDED` via `OccupancyScale` as well as `STANDING_ROOM_ONLY` via `OccupancyStatus` if they really wanted to do that and had technology to ensure all the seats were actually full.

The obvious downside with this approach is that we would have 3 different ways to represent occupancy :(. Although as mentioned above, if producers plan to bucket values in `occupancy_percentage` anyway, maybe we could get rid of that field and use `OccupancyScale` instead.

stevenmwhite:
> @stevenmwhite I'm wondering if instead of changing the meaning of enums that are in use and are being advocated for by other stakeholders, that we instead clarify in the spec some guidelines around `occupancy_percentage` to leverage the generalized descriptions of `Not crowded`, `Some crowding`, `Crowded`. 

I think this would be a good idea. A guideline to help define the percentage and crowding descriptions would be useful for producers and consumers alike.

On the other hand, should we also add guidelines indicating that the other enums are not necessarily a linear scale but represent actual communication about the state of the accommodations available and should be used specifically to represent those accommodations?

I certainly have no issue keeping something already in and that is helpful to other stakeholders and am not necessarily advocating to get rid of any mention of seats vs standing (these are certainly valuable if they are true!)... it'd probably be good to define their use specific to those accommodations, however, so that it's clear this is not a linear scale and rather represents the state of particular accommodations aboard the vehicle. Right now it seems many producers and consumers are interpreting them as a linear scale and that's what's unclear.

barbeau:
>should we also add guidelines indicating that the other enums are not necessarily a linear scale but represent actual communication about the state of the accommodations available and should be used specifically to represent those accommodations?

💯 @stevenmwhite Our posts crossed :). Yes, I think no matter what the outcome is, this needs to be clarified if `OccupancyStatus` continues to be used.

scmcca:
@barbeau 
> I'd be concerned that officially adopting a subset of the current OccupancyStatus enum doesn't really solve the problems that currently exist with it. 

It certainly does not solve the problem wholesale but is a step in the right direction and maybe a viable alternative to adopting all enums that we still legitimately are not sure about and should therefore remain experimental, or by unviably removing enums (which received 3 pre-vote -1's). Meanwhile, it seems that we are confident with at least 3 enums, so phasing the solution may be okay in this context. 

Notwithstanding that `OccupancyStatus` has been experimental for 6+ years when the [specification amendment process](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#experimental-fields) allows fields to be experimental for only 2 years (albeit grandfathered in), one of the goals right now is to adopt an occupancy description to unblock GTFS-Occupancies (#240).

> we won't know which producers are using the "official" interpretation vs. the experimental one when looking at the data.

This is true but should not matter in the interim. Since `OccupancyStatus` has been experimental for over 6+ years, there would be little change short-term in how producers and consumers are interpreting them. Long-term, it should ideally converge the industry on the fewest enums possible, which was the main issue highlighted by @gcamp as the motive for this PR:
> The enum of occupancy_status has 7 values, however we've seen that the range of values is not used in practice.

To accomplish this I think we need a positive, rather than negative, framing. Right now the PR is proposing to take experimental enums away that are in use. Instead, we ought to adopt the ones we are confident about.

@stevenmwhite
>A guideline to help define the percentage and crowding descriptions would be useful for producers and consumers alike.

> On the other hand, should we also add guidelines indicating that the other enums are not necessarily a linear scale but represent actual communication about the state of the accommodations available and should be used specifically to represent those accommodations?

Very well said! I'm all for these clarifications. As discussed at length, `OccupancyStatus` and `occupancy_percentage` are distinct. I think we already have the fields that we need, but we need to emphasize how they should be used.

RTL-Gilles-Ferland:
> Right now it seems many producers and consumers are interpreting them as a linear scale and that's what's unclear.

We are indeed using the `OccupancyStatus` enum as a linear scale. If a new enum is added to represent `crowding` on a linear scale we will probably move to that enum and abandon the `OccupancyStatus` one.

Why not change the meaning of the different enum values if producers, like us, are already using it as a linear scale? It will make everything clearer and not break the current implementations of the enum. And also work with GTFS-Occupancies ([#240](https://github.com/google/transit/pull/240)).

scmcca:
@RTL-Gilles-Ferland 
> Why not change the meaning of the different enum values if producers, like us, are already using it as a linear scale?

The challenge is that the existing `OccupancyStatus` enum values describe nominal states of vehicle occupancy levels, such as "many seats available" or "standing room only". Accepting that seating and standing capacities are different for different vehicle types, ascribing the existing `OccupancyStatus` enum values on a fixed linear scale would not work universally.

@barbeau @stevenmwhite On the topic of disambiguation between intended uses of `OccupancyStatus` and `occupancy_percentage`, please see the newly opened #259 and feel free to give feedback on the topic there. Keep in mind that it is a basic proposal that should be taken as a starting point for sorting these out.

scmcca:
Hi everyone. Please find a continuation of the broader realtime occupancy discussion in [#223 ](https://github.com/google/transit/issues/223#issuecomment-771146817). Welcoming engagement over there, thanks!

scmcca:
I made a [PR on this branch](https://github.com/TransitApp/transit/pull/3) to reform this proposal to add a CrowdLevel enum in place of deprecating values in OccupancyStatus. Inviting reviews and feedback [here](https://github.com/TransitApp/transit/pull/3/commits/6e5663dba57a3b0bc1ab32e083ece80e1a404f27)!

scmcca:
Please see a continuation of this proposal at #261. Thanks!

PR #213, opened 2020-04-09
Add occupancy_percentage

gcamp:
The current enum for occupancy doesn't provide enough granularity, for this reason I would suggest adding `occupancy_percentage`.

It would be percentage value representing the degree of passenger occupancy of the vehicle. The values are represented as an integer without decimals. 0 means 0% and 100 means 100%. The value 100 should represent the total maximum occupancy the vehicle was designed for, including both seated and standing capacity. It is possible that the value goes over 100 if there are currently more passengers than what the vehicle was designed for.

This field is experimental at the moment.


skinkie:
I like the approach where over 100% is exceeding the designed safety limit.

+1

stevenmwhite:
All for this -- exactly as described.

+1

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -393,6 +393,15 @@ message VehiclePosition {
   }
   optional OccupancyStatus occupancy_status = 9;
 
+  // A percentage value representing the degree of passenger occupancy of the vehicle.
+  // The values are represented as an integer without decimals. 0 means 0% and 100 means 100%.
+  // The value 100 should represent the total maximum occupancy the vehicle was designed for,
+  // including both seated and standing capacity.
+  // It is possible that the value goes over 100 if there are currently more passengers than what the vehicle was designed for.
+  // The precision of precision should be low enough that you can't track a single person boarding and alighting for privacy reasons.
```
barbeau, line 401:
```suggestion
  // The precision of occupancy_percentage should be low enough that you can't track a single person boarding and alighting for privacy reasons.
```

lauramatson:
We are interested in making this information available in light of COVID and asking customers to wait for less crowded vehicles. If the current "safe" capacity of a bus is 10 passengers to allow for social distancing, would having 10 passengers mean occupancy_status = 100? Or would it be tied to the physical seating/standing capacity of the vehicle so occupancy_status = 20?

barbeau:
@lauramatson That's a great question. I would say if the "safe" capacity is 10, meaning that drivers are instructed to only allow 10 riders board, then 10 passengers would be `occupancy_status = 100`.

We might have to tweak the language in the description just a bit for this - maybe instead of:

>The value 100 should represent the total maximum occupancy the vehicle was designed for

...it could be:

>The value 100 should represent the total maximum occupancy the vehicle was designed for and current operating regulations allow

Or something similar. Suggestions for better text welcome.

mike-swiftly:
This will be great addition.
+1

nighthawk:
> I would say if the "safe" capacity is 10, meaning that drivers are instructed to only allow 10 riders board, then 10 passengers would be `occupancy_status = 100`.

This would be problematic for data consumers. If you get the data for two operators, one using the vehicle's actual capacity and one using the currently recommended capacity, then how do you know which is which and what to show to users in light of @lauramatson's use case? In the first case a value of 50 might mean "This is pretty fully, so maybe consider taking another bus" and in the second case it would mean "All good, plenty of space".

barbeau:
I would think that in both cases you would want a user to look at it as "This is pretty fully, so maybe consider taking another bus". But I'd certainly welcome other perspectives on this.

gcamp:
@nighthawk It's true that it's not a perfect solution but I way prefer agencies adjusting their percentage to reflect possible load. It's way better for an agency to put 100% if they don't accept passengers even if in practice there's "plenty of space".

paulswartz:
For producers: are you planning to report an exact percentage, or some amount of bucketing? We're currently looking at 20% buckets, so reporting 0/20/40/60/80/100%.

stevenmwhite:
> For producers: are you planning to report an exact percentage, or some amount of bucketing? We're currently looking at 20% buckets, so reporting 0/20/40/60/80/100%.

While our internal APIs report an exact percentage, we generalize it to those same buckets when putting the information into our white labeled rider-facing smartphone apps. We show it visually (rather than the numbers), but we use a 5-step visualization for capacity.
image

To make it sufficiently "fuzzy" I think it's likely we'd report the same buckets out on GTFS-RT feeds.

gcamp:
This has been open for over a week and seems there's a consensus, so per the official process I'm calling for a vote. Reminder that this is an experimental field. 

Vote will be closed Monday April 27, 23:59:59 UTC.

skinkie:
+1 OpenGeo

paulswartz:
+1

barbeau:
+1 OneBusAway

stevenmwhite:
+1 GMV Syncromatics

drewda:
+ 1 for @interline-io as a consumer

mike-swiftly:
+1 for Swiftly as consumer and producer

abyrd:
+1 Conveyal

Bertware:
+1 for Trafiklab as a producer

gcamp:
Hi all, the voting has concluded. The result is 8 votes for the proposal and 0 against. The proposition is accepted!

barbeau:
Hi all! For everyone that participated in this discussion for `occupancy_percentage` - could you please take a look at https://github.com/google/transit/issues/223?

Please comment there for which you prefer - OccupancyStatus vs. occupancy_percentage.

PR #217, opened 2020-04-27
Mention pathways.txt and attributions.txt in translations.txt

aababilov:
Fields in pathways.txt and attributions.txt  also may be translated.


aababilov:
Thanks!

PR #219, opened 2020-04-29
Proposal: Support stop assignments in GTFS-RT

barbeau:
### Background
Several producers and consumers have expressed the need to reflect platform assignments within stations in real-time. However, the GTFS-RT spec currently does not officially support this use case. The spec is unclear whether changing `stop_id` to a value other than what is defined in GTFS stop_times.txt for a given `stop_sequence` is an error [1].

### Proposal

This proposal adds a new message and field ~~`StopTimeProperties.platform_id`~~ `StopTimeProperties.assigned_stop_id` (within `StopTimeUpdate`) which can be used to specify ~~assignments to platform stop_ids within the same parent station as the stop originally defined in static GTFS~~ changes to the stop within the trip (see **Oct 27, 2020 Update** below).

Because platform assignments are currently implemented by a few producers and at least one consumer by changing the StopTimeUpdate `stop_id` [2], this feature could be adopted prior to the rest of the entire ServiceChanges spec. with minimal changes to existing feeds.

This proposal uses the terminology ~~`platform_id`~~ `assigned_stop_id` for a new field, instead of (re)using `stop_id` for several reasons:
1. ~~The only targeted use case for this new field is changing stop_ids within a station - producers cannot put any stop_id in this field and expect consumers to be able to process this change to the trip geometry quickly. If the field is named `stop_id`, producers and consumers may start using the field for other use cases of changing stops within a trip (e.g., detours). Therefore, the field is narrowly named `platform_id` instead of `stop_id` to make the use case very clear, even if the spec documentation isn't read.~~ (see **Oct 27, 2020 Update** below)
1. ServiceChanges defines a more robust solution for detours that can change/add stops within the trip by adding a new `Trip` object with new `StopTime` objects to the real-time feed. However, it takes consumers longer to process these new trips due to necessary pre-processing, and therefore this approach is not suitable for platform assignments that must be processed in the same order-of-magnitude of time as arrival predictions. Therefore, platform assignments should be implemented separately from more complex detours.
1. Re-using the existing StopTimeUpdate `stop_id` field to assign platforms causes issues with feed quality and validation - it's not clear whether the change from the static GTFS stop_id is intentional or an error. Using a new field ~~`platform_id`~~ `assigned_stop_id` makes the intent to assign a platform explicit and unambiguous. (see **Oct 27, 2020 Update** below)

~~To further clarify the spec, this proposal also adds documentation saying StopTimeUpdate stop_id and stop_sequence pairings must match static GTFS - in other words, it clarifies that these fields are selectors to specify the stop, and cannot be modified in real-time.~~ **EDIT** Per further discussion in this thread, enforcing stop_id/stop_sequence pairing matching GTFS is no longer proposed in this PR - it can be discussed along with the larger set of ServiceChanges v3.1.

Future proposals may add other fields defined in GTFS stop_times.txt to the `StopTimeProperties` message  (e.g., pickup_type, drop_off_type, stop_headsign, shape_dist_traveled) so they can be changed in real-time. See the [ServiceChanges proposal](https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit#bookmark=id.m61cxqvifvmg) for more info.

@paulswartz @skinkie I believe you are both producers of the existing unofficial method of platform assingments. Thoughts on the above? Could you switch to using ~~`StopTimeProperties.platform_id`~~ `StopTimeProperties.assigned_stop_id` instead?

cc @tleboulenge

Announced on the Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/3sTrsi2Sw5U.

**Oct 27, 2020 Update**

Changes based on community feedback. See https://github.com/google/transit/pull/219#issuecomment-717524976 for details. Summary:

* Field name change of `platform_id` -> `assigned_stop_id` - The feature is no longer constrained to just platforms within stations (see next point).
* Use more generic language to define when `assigned_stop_id` can be used (versus requiring changes to be only within a station). This broadens potential uses cases of the field without being as prescriptive. There is intentional use of the word "should" vs. "must" in some cases to allow for deviation from the platform assignments use case.
* Add that you can assign a stop without providing real-time predictions by using `StopTimeUpdate.schedule_relationship = NO_DATA` (based on @sunny-lirr question [above](https://github.com/google/transit/pull/219#issuecomment-679181070))
* Require that StopTimeUpdate.stop_sequence be used as the selector (based on @skinkie feedback above in moving towards using `stop_sequence` as the selector in https://github.com/google/transit/pull/219#issuecomment-675767164)
* Don't block use of `StopTimeUpdate.stop_id` being populated with other values in case others want to have unofficial GTFS-realtime feeds that changes `stop_id` to represent a different stop (e.g., @skinkie with Bliksem Labs B.V. wants to be able to define detours by populating TripUpdates with the entire stop sequence list, aka "GTFS-realtime full"). Therefore, this proposal doesn't block the current use or potential future adoption of a "GTFS-realtime full" if the community would desire this.

[1] https://github.com/google/transit/pull/81, https://github.com/CUTR-at-USF/gtfs-realtime-validator/issues/297
[2] https://groups.google.com/d/msg/gtfs-realtime/BZOfsVeI2Cc/gIUjGbkCBAAJ


skinkie:
-1 it has previously been indicated that changes in stop_id for tripUpdates are allowed if the stop_id falls within the same parent_station. Currently a stop has a both a stop_id and platform_code field. Introducing an extra field platform_id creates even more complex processing, because the stop_id cause should be handled anyway.

paulswartz:
We could provide this, but we'd need to continue to also change `stop_id` until larger consumers like Apple and Google commit to supporting a new field.

gcamp:
> Currently a stop has a both a stop_id and platform_code field. Introducing an extra field platform_id creates even more complex processing, because the stop_id cause should be handled anyway.

Adding` platform_id` actually simplifies the processing for consumer because it makes the intent very clear about what is happening and it avoid additional processing if the `stop_id` doesn't match. 

Currently Transit doesn't support that unofficial extension since, well, its unofficial.

skinkie:
> Adding` platform_id` actually simplifies the processing for consumer because it makes the intent very clear about what is happening and it avoid additional processing if the `stop_id` doesn't match.

It doesn't simplify anything if you also want to implement a detour in the same software. Why an if-then-else case if you could just replace a stop. We both know why planners using transfer patterns can't use a trip pattern that falls outside of a parent_station. Hence it does not make sense to explicitly support it, if the software that already supports it, added the restriction in the first place.

gcamp:
> It doesn’t simplify anything if you also want to implement a detour in the same software. Why an if-then-else case if you could just replace a stop.

It’s also about making the spec understandable. Even if this is specified in the documentation changing the stop_id is far from obvious to a newcomer. My point was also that `stop_id` sometime doesn’t match due to errors, and this avoid processing for if this wasn’t intentional.

> We both know why planners using transfer patterns can’t use a trip pattern that falls outside of a parent_station.

I’m not sure what this means. I don’t think it really matters in the context since the both unofficial spec and the proposition restrict to inside the parent_station.

skinkie:
> It’s also about making the spec understandable. Even if this is specified in the documentation changing the stop_id is far from obvious to a newcomer. My point was also that `stop_id` sometime doesn’t match due to errors, and this avoid processing for if this wasn’t intentional.

A newcomer shouldn't have to worry about implementation details of dinosaurs. Even @barbeau is adding the restriction "platform changes must fall within the same parent_station" as something that is completely understandable, which is not the case. For a newcomer it should be as easy as "you can replace a stop within a trip by changing its stop_id". This proposal adds complexity because it forces the consumer and producer to check both changes in stop_id and changes in platform_id, where the latter actually references to an existing stop_id which then falls within the same parent_station but has a different platform code. "Just replace the stop_id" makes a lot more sense. And if there are systems that can't handle this... too bad for them, it's 2020, update your code.

> > We both know why planners using transfer patterns can’t use a trip pattern that falls outside of a parent_station.
> 
> I’m not sure what this means. I don’t think it really matters in the context since the both unofficial spec and the proposition restrict to inside the parent_station.

Please educate yourself why this proposal and the unofficial spec enforce this limitation. Otherwise ask @barbeau to add it to the documentation. We could again start a discussion in the same terms as the _Travel Restrictions_ where it was kind of clear that the producer should be able to describe something simple in simple terms, and the consumer should do the hard work.

thomastrillium:
> Please educate yourself

Hi @skinkie -- I don't know you and I suspect you're a nice fellow in many contexts. However, in this community, to be frank, you are often coming across as insensitive and unhelpful to me, and I suspect to others. I need to take this moment to call you out and ask for you to change your behavior, as a fellow community member that wants to work together on the best solutions.

This is not a helpful response you've given to @gcamp . If you're here to shut ideas down, then you need to be here to communicate why. "Please educate yourself" is a rude and antisocial reply. If there is education needed, and you are the one saying no, it *is* your job to do the education--and it's *not* education it is advocacy for your own point of view. Others have their own and they are just as valid as yours.

So take a step back, apologize, provide an a real answer to Guillaume's question, and then take another step back and think about how to implement this advice in the future. We are all hurting and we are all trying to save the world. Please grant others some compassion--it is absolutely critical to community in this time. 

Thanks for your consideration and efforts.

skinkie:
image

@thomastrillium if @gcamp is making an argument in favor for a change that @barbeau is proposing. But does not know why @barbeau is using the wording as he does. The kind reply is "please educate yourself" not: "you are stupid for not knowing this". In a reply https://github.com/google/transit/pull/219#issuecomment-621452709 I have exactly laid out what the problem is for the current providers, which he does not quote in his reply. So did you miss it as well?

We are here in a discussion forum where everyone is invited to participate. Sadly I am under the impression that _some_ of the ideas proposed here are not proposed to make this standard any better, or more useful, but rather introduce complexity to justify problems that some paying MobilityData members have.

Bertware:
> It’s also about making the spec understandable. Even if this is specified in the documentation changing the stop_id is far from obvious to a newcomer. My point was also that `stop_id` sometime doesn’t match due to errors, and this avoid processing for if this wasn’t intentional.

@gcamp agreed that the spec is a little unclear if you don't know of this feature, I always thought stop_id had to match the original one. This is probably due to the wording. From the spec:

> Must be the same as in stops.txt in the corresponding GTFS feed. Either stop_sequence or stop_id must be provided within a StopTimeUpdate - both fields cannot be empty.

This suggests that stop_id and stop_sequence serve the same purpose of identifying the stop in the journey. However, as I understood just now, we could simply use sequence_id as the identifier (as Trafiklab is already doing today) and change stop_id. 

> A newcomer shouldn't have to worry about implementation details of dinosaurs. Even barbeau is adding the restriction "platform changes must fall within the same parent_station" as something that is completely understandable, which is not the case. For a newcomer it should be as easy as "you can replace a stop within a trip by changing its stop_id". This proposal adds complexity because it forces the consumer and producer to check both changes in stop_id and changes in platform_id, where the latter actually references to an existing stop_id which then falls within the same parent_station but has a different platform code. "Just replace the stop_id" makes a lot more sense.

Agreeing on this one as well. I do not see the benefit of having a separate field, since, in this proposal, stop_id would become static and always match the static feed. In that case it feels like it's a more limited alternative to stop_sequence, as it can't identify a stop on a trip in case the same stop is visited twice.

When looking at both options, they would cause the following situation on client side:
- using stop_id: Simply look if it has changed compared to the static feed.
- using platform_id:  Simply look if platform_id has changed compared to the static feed. Don't ever look at stop_id.

As a producer, I'd ask why the stop_id is still there in case plaform_id is used. It doesn't seem to serve any purpose in this case, except for those who want to save a tiny bit of work client side as stop_sequence doesn't need to be looked up in order to get the original stop_id. It could also serve as a way to identify the stop on a trip, but stop_sequence is better at that anyway. 

Regarding the original arguments:

> 1. Therefore, the field is narrowly named platform_id instead of stop_id to make the use case very clear, even if the spec documentation isn't read.

In my opinion this is a matter of clearly declaring this in the documentation.

> even if the spec documentation isn't read.

The spec needs to be read. I'd argue that there are more fields that aren't 100% clear if you don't read the spec.

> 3. It's not clear whether the change from the static GTFS stop_id is intentional or an error.

If stop_id is defined as able to change, it's intentional. If stop_id can include errors, platform_id can as well. In both cases the client expects them to be correct to the truth.

At this point I'd rather improve the documentation for stop_id than to introduce a new field.

skinkie:
> As a producer, I'd ask why the stop_id is still there in case plaform_id is used. It doesn't seem to serve any purpose in this case, except for those who want to save a tiny bit of work client side as stop_sequence doesn't need to be looked up in order to get the original stop_id. It could also serve as a way to identify the stop on a trip, but stop_sequence is better at that anyway.

I initially thought this trick would work (6 years ago) and more recently made the same mistake. But sadly there is no "good" way to overcome to go back from stop_sequence to a stop_id. The assumption that a stop_id is present only once in a trip does not hold. GTFS supports this, Google human feed validation even marks it as an issue if the headsign does not change at this point. This suggest that stop_sequence _is_ the way to go. It could be argued that it is an efficient way of doing this (if the producer would guard gapless stop sequences and won't introduce logic bombs that would fill a preallocated array).

But imagine that you want to change your add a stop to your trip in the realtime feed, _this is outside of the scope of this discussion_, because the intercity train now also calls at a local station. The only way to resolve this in a consumer system would be to trust the order of StopTimeUpdates in the feed and that the position of the newly introduced stop can be established at all. This is also not enforced in the standard.

Bertware:
> Feed quality is an issue industry-wide, and generally speaking IMHO having overloaded fields and implicit data changes that can't easily be differentiated from errors is part of this problem. Hence the proposal here to explicitly define a new field platform_id to handle dynamic changes, and restricting stop_id and stop_sequence to act only as identifiers of stops, not containers for new values.

I still don't see how it would be easier to find errors in the proposed `platform_id` field compared to the `stop_id` field. stop_id is only conditionally required, so you could even say that you only need to include a `stop_id `in case of a changed stop when using `stop_sequence` to identify the stop in the trip. This is already possible today with the current spec (which would only need some minor adjustments to clarify that stop_id in fact can point to a different platform).

As you're talking about simplifying things to prevent errors, I'd say the stop_sequence documentation is prone to cause issues as "stop_sequence is required for trips that visit the same stop_id more than once (e.g., a loop) to disambiguate which stop the prediction is for." might cause issues with companies thinking they'll never face loops until an agency which uses their software starts using loops. Simply requiring the stop_sequence and making stop_id optional (to handle dynamic changes) would resolve this. While this is a different issue I feel it's relevant to keep in mind, as using stop_sequence as the identifier in all cases would simplify this issue.

I might not be aware of some issues/downsides of stop_sequence, feel free to correct me in case this doesn't work for all operators/cases.

stevenmwhite:
As a producer, this was the case for us at GMV Syncromatics. @barbeau, I don’t know if we’re the one you were referring to, but we did not originally include stop sequence, and now we do.

I’d say this is much less of a problem today than it was in past years and any AVL provider who is actively investing in their product should be able to handle it now.

________________________________

barbeau:
@stevenmwhite I actually wasn't referring to Syncromatics, so make that "at least two" 😉

paulswartz:
@barbeau 
> To clarify, part of this proposal is locking stop_id so it can't be changed in real-time - stop_id/stop_sequence pairing must match static GTFS. So consumers only need to check platform_id. Any mismatch of stop_id/stop_sequence gets thrown out as an error.

This make it harder for producers to migrate, as we'll break functionality if any of our consumers are still using the old specification. I'd prefer to have both supported, so we can continue to provide the old `stop_id` change for a while along with the new `platform_id` as consumers update to the new specification.

barbeau:
@paulswartz If this proposal is adopted we'll work with existing consumers to have a deprecation period of their existing behavior to allow current implementers of the unofficial method to switch over. The existing spec can be interpreted either way so I see this as a clarification of behavior.

jxeeno:
Echoing some of the thoughts here.  I too, would like to see a more explicit way of identifying when platform / track changes are provided.   As a consumer, it's not always clear to us whether it's an intended platform change or some other reason e.g. GTFS-R being out of sync with GTFS (static).

But for backwards compatibility reasons, I don't think `platform_id` is the way to go necessarily.
 Since many consumers already handle the platform change with `stop_id` change, this will break implementations on both consumer and producer sides.  I'm not a producer, but I suspect there'll be hold out in implementing this change simply because some agency's internal consumers would also need to be updated.  And not everyone builds things in-house :)

I'd prefer seeing the inverse -- a new field for the `schedule_stop_id` and make both `stop_id` and `schedule_stop_id` mandatory when there is a platform change.  That way, we aren't breaking existing implementations whilst still being explicit about both the platform change and the stop time it affects.

skinkie:
> I'd prefer seeing the inverse -- a new field for the `schedule_stop_id` and make both `stop_id` and `schedule_stop_id` mandatory when there is a platform change. 

The problem with the scheduled_stop_id is that it can occur more than once within a trip. Hence is it not an unique property to identify a StopTimeUpdate, but rather an attribute that is likely to be resolved by a consumer when it is not provided.

jxeeno:
> The problem with the scheduled_stop_id is that it can occur more than once within a trip. Hence is it not an unique property to identify a StopTimeUpdate, but rather an attribute that is likely to be resolved by a consumer when it is not provided.

Yeah, agree.  The way I see it is the problem exists either way with `platform_id`, `schedule_stop_id` or the current implicit platform change using `stop_id`.  But that should already be resolved by `stop_sequence` being a required field in the spec when there's a multi visit.

> stop_sequence is required for trips that visit the same stop_id more than once (e.g., a loop) to disambiguate which stop the prediction is for.

Both consumers and providers should already be handling loops / multi visits by using `stop_sequence` as the unique identifier when sending/applying existing well-defined attributes like delays or schedule relationships.  So when a theoretical `schedule_stop_id` is supplied, a consumer would basically interpret it as a flag for a platform change rather than using it as the identifier for the scheduled stop.

skinkie:
>  So when a theoretical `schedule_stop_id` is supplied, a consumer would basically interpret it as a flag for a platform change rather than using it as the identifier for the scheduled stop.

This part is something I don't get. Why would a consumer need a "flag" to know there is a change? The consumer is likely to have the schedule _loaded_. We currently don't have such _flag_ because we assume that a schedule is available at the consumer.

tleboulenge:
+1
IMO the main benefit of this change is spec clarity, as expressed in (3) of the original message. It looks like we more or less all agree the current meaning of stop_id here is very ambiguous and confusing, and most of us certainly understood it wrongly at some point. Therefore it seems a good idea to suggest a change to clarify this, and the suggested change does just that.

There is no hidden agenda behind this change, and it doesn't enable or modify a feature that didn't exist before.

It looks to me that a field that can be at times a selector and at times a modifier (with no way to distinguish between those cases except by looking up both values in the most current GTFS) is a major problem.
The solution to that problem is to have fields that are only selectors (in the proposed change, that would be stop_sequence and stop_id) and fields that are only modifiers (platform_id).

This would result in the message being more explicit, clearly express the intent, easier to validate and therefore would lead to better feature adoption and less broken data.

@skinkie Do you disagree this is a problem to start with, or do you disagree with the suggested solution to that problem?

skinkie:
@tleboulenge given the GTFS datamodel, there is no difference between a station/stop and a platform. Therefore modelling a field that in GTFS-RT that suggests that only mutations within the same station are allowed in this field, while effectively a stop is mutated (not a platform) does not make sense to me. Especially since the mutation of a stop (any) should be allowed anyway. Some current consumers that are able to have mutations do not allow mutations beyond parent station mutations (ex. Google) have a technical limitation for this others (ex. OTP, RRRR) do not. A client must have a GTFS loaded (otherwise something trivial as stop labels cannot be shown) therefore a client is always be able to compare if a stop_id mutation is effectively a platform change or a station change. So the answer is both.

Bertware:
>It looks like we more or less all agree the current meaning of stop_id here is very ambiguous and confusing, and most of us certainly understood it wrongly at some point. Therefore it seems a good idea to suggest a change to clarify this

Agreed.

> The solution to that problem is to have fields that are only selectors (in the proposed change, that would be stop_sequence and stop_id) and fields that are only modifiers (platform_id).

I agree with the concept of separating selectors and modifiers. However, I strongly disagree with having `platform_id` to achieve this. The problem is `stop_id` being a selector and modifier. If we're going to change the behaviour of this field (this PR proposes to only use it as an identifier) i suggest doing the opposite and removing the identifier function and using `stop_sequence` as the only identifier. A less drastic approach would be to simply rewrite the `stop_id` documentation to make it clear that this field can be used for platform changes.

I furthermore agree with @skinkie that we should not limit changes to platform changes only. There have been cases in Sweden where trains were redirected to different stations. Distance between the stations is only 500m for one change and 0 meters for the other (tunnel vs surface station). There should be no reason to block these changes in the spec, while platform changes in large railway stations can be a 400m walk as well. I understand that there are clients who might not be able to handle this information due to technical reasons, but this should not be blocked from the spec just because there are a few parties that have trouble handling it, when others can handle it without problems.

gcamp:
> The problem is stop_id being a selector and modifier. If we're going to change the behaviour of this field (this PR proposes to only use it as an identifier) i suggest doing the opposite and removing the identifier function and using stop_sequence as the only identifier.

Effectively this would make `stop_sequence` required instead of conditionally required. It would be the first version of GTFS-rt that is not backward compatible. Even GTFS-rt 2.0 was mostly a clarification release even if technically it could have rendered feeds incompatible I don't think it happened in practice. 

Does anybody know how widespread that would make feeds invalid? If not I can do some research on my side. 

If the backward compatibility problem is not a blocker because no or a few feeds specify it, I think this would be the best solution. 

One note, I'm not sure how what Google did was ever considered valid. Reminder that the spec says 
> The update is linked to a specific stop either through stop_sequence or stop_id, so one of the fields below must necessarily be set.

I'm not sure how this could be compatible with "stop_id changes the stop affected if stop_sequence is defined" like Google defined.

barbeau:
>The problem is stop_id being a selector and modifier.

💯 

If `StopTimeUpdate` were to follow the design of the rest of the spec where `*Descriptor` messages function purely as selectors (`TripDescriptor`, `VehicleDescriptor`), we should have a `StopDescriptor`, ideally with just `StopDescriptor.stop_sequence`, which would function purely to select the stop.

skinkie:
"Updates can be supplied for both past and future events. The producer is allowed, although not required, to drop past events." is an additional problematic choice. If one would state that a producer should provide an integral update of the trip, then all StopTimeUpdates could form a trip on their own, it would be self describing. It is not self describing if stop_sequence is exchanged and it may be self describing if stop_id is exchanged.

But I fully agree with you @barbeauon on the StopDescriptor semantic. It is also problematic for VehiclePosition where the _current_stop_sequence_ attribute is introduced, different attribute naming. But even with the StopDescripter semantic, what would it have to _select_ to? The selection property only is required when a _partial update_ is presented for example in a delta update. Given that Google is moving more and more towards VehiclePosition updates and do their own predictions, for the remainder of the trip, why not explicitly request for the _special cases_ to exchange all StopTimeUpdates for at least **all** future events. In that case stop_id it is a pure modifier.

gcamp:
To answer my own question, here are the list of feeds that Transit supports that don't have `stop_sequence` provided : 

- Tri Delta Transit (SF Bay Area)
- Hamilton Street Railway
- Washington State Ferries
- GO Transit (Toronto, Ontario)
- Sydney Trains
- Burlington Transit
- Société de transport de l'Outaouais
- MTA NYC - Metro-North Railroad
- TANK (Cincinnati)
- Rhode Island Public Transit Authority
- San Diego Metropolitan Transit System
- Metro St. Louis
- Caltrain (SF Bay Area)

tleboulenge:
> I furthermore agree with @skinkie that we should not limit changes to platform changes only.

I also agree the spec is lacking a way to represent e.g. a detour, where you'd replace the scheduled stop by another that is located further away. This is actually the subject of another part of the Service Changes spec, which allows you to re-draw the path (and stops) of a trip. The reason this requires a new spec and not a simple "stop #3 is now moved to some-new-stop" is that there are certainly cases where they won't match 1:1, and there is no way to insert new stops in the sequence (e.g. you're replacing 2 stops by 3 new ones).

Platform changes (which is often actually "platform assignment" because they are not known in the GTFS static and the assigned stop is the parent station) is a common enough case that can benefit from this much simpler syntax rather than having to redefine the whole (portion of) trip with any arbitrary set of new stops.

> One note, I'm not sure how what Google did was ever considered valid. 
This is one reason we'd like to create now a valid way to represent that case. It helps nobody to keep this useful feature hidden and looking the way it does today.

barbeau:
Based on feedback in this thread, I've updated the PR to remove the text that required that `stop_id` and `stop_sequence` must match static GTFS - we can table that discussion to be part of the larger change set in ServiceChanges v3.1 (https://bit.ly/gtfs-service-changes-v3_1) that also handles detours, etc.

So, this proposal now focuses only on defining a very specific use case for platform changes, using the `StopTimeUpdate.stop_sequence` as the selector and a new `StopTimeProperties` container to specify the new `platform_id`.

I'd like to move forward with this updated proposal - any specific thoughts before we call for a vote?

barbeau:

Vote will be closed on Wednesday June 3rd at 23:59:59 UTC.

skinkie:
-1 (OpenGeo)
I wish not to support platform changes, I rather allow all to change any stop_id.

barbeau:
@skinkie Thanks for the feedback! Just to clarify - in response to your earlier feedback I made changes to the proposal so it no longer blocks the use case of changing any stop_id (see https://github.com/google/transit/pull/219#issuecomment-631601533 for details).

skinkie:
> @skinkie Thanks for the feedback! Just to clarify - in response to your earlier feedback I made changes to the proposal so it no longer blocks the use case of changing any stop_id (see [#219 (comment)](https://github.com/google/transit/pull/219#issuecomment-631601533) for details).

Thanks for explicitly pointing me at this. I overlooked it. Sadly my main objective stays from a model point of view. Adding a platform_id to facilitate the changes would defeat my own argumentation why it would be a _good_ idea. I would obviously support a stop change, where my reasoning is, if I can change a stop_id, I can change a platform. If I would like to retain the stop_id to match to the original schedule and have something like changed_stop_id to have both, that would not be called platform. So I am happy to discuss this with you outside of this "issue tracker".

I understand what you are trying to do with the StopTimeProperties, separating the matching part with the modification part. But wouldn't changing platform_id to stop_id not something that would be another option?

tleboulenge:
@skinkie @barbeau If there is no strong objection to the idea/ semantic of the proposal, could you provide (both of you) an example of a message that would translate in a syntax that you feel is appropriate this following statement:

The train "trip-1" will now stop at "stop-id:platform 23" for its 3rd stop (stop-sequence: 3) [optionally: ... instead of the previously planned "stop-id: platform 15"]


... That would help me (us?) better understand where the disconnect exactly is.

Thanks

skinkie:
@barbeau if the "global" consensus would be that it is a good idea to have a Descriptor (matching static) and Properties (updateble) I would like to see this:

```
trip_update {
    trip: {
      trip_id: "trip-1"
    }
    stop_time_update {
      stop_sequence: 3
      arrival {
        time: 1560267700
      }
      stop_time_properties {
        stop_id: "id_23"
      }
    }
}
```
But to be critical as well. This object type is called StopTime**Update**. I would find it strange that "arrival" in this case is **not** part of stop_time_properties, and **is** updated with respect. So I the structure below is _currently_ working for me. So I would love to hear why it wouldn't work for them, while the example above does.
```
trip_update {
    trip: {
      trip_id: "trip-1"
    }
    stop_time_update {
      stop_sequence: 3
      stop_id: "id_23"
      arrival {
        time: 1560267700
      }
    }
}
```

barbeau:
@skinkie Just to clarify - with your proposal of this data structure:

```
      stop_time_properties {
        stop_id: "id_23"
      }
```

Is `stop_id` required to be a platform, or could it be any stop (e.g., a detour)?

skinkie:
> @skinkie Just to clarify - with your proposal of this data structure:
> 
> ```
>       stop_time_properties {
>         stop_id: "id_23"
>       }
> ```
> 
> Is `stop_id` required to be a platform, or could it be any stop (e.g., a detour)?

Any update for an update **at this specific** stop_sequence. You couldn't do this for an arbitrary detour, only for detours that replace a stop.

abyrd:
As someone just joining this discussion, after reading the above comments it seems that no consensus has been reached, and that this needs to be rethought a bit and restated before a vote is possible. As several people have noted, a clear distinction is needed between selectors and modifiers. My initial impression is that it's very strange to add a field called platform_id that in fact references a stop_id, alongside another stop_id field. I suspect this will be perceived as confusing to anyone newly approaching the spec. 

The RT message under discussion is an update message - if any element within it can clearly identify the entity to be updated, it seems reasonable for the rest of the fields to be interpreted as updates to the selected entity. It seems viable to use the stop_sequence as the only selector field. Although the majority of TripUpdates would not change the stop visited (only the time), stop_id could still technically be considered a modifier in those cases, but retain the scheduled value, redundantly confirming the stop_id rather than changing it. But I do get the backward compatibility argument that not all producers include stop_sequence.

So if stop_id must continue to be a selector (for backward compatibility) then it is indeed the right idea to have another field specifying the new stop_id. Perhaps the only problem is the naming of the modifier field? What if stop_id was the selector, and the modifier was simply called new_stop_id or diverted_stop_id? I think this is structurally identical to the original proposal, but by simply changing names I would find it less confusing, and it leaves open the possibility of changing to stop_ids outside the same station (should that constraint ever be relaxed).

I know I've mentioned this before, but I think it's worth mentioning again here: many of these problems would be alleviated if GTFS-RT required full trip updates for every trip, i.e. if you're updating any stops in a trip, you must supply one update for every stop from the beginning to the end of the trip. Some level of redundancy in the data can be beneficial. That seems unlikely to happen, but a smaller amount of redundancy here (both the stop_id and stop_sequence as selectors) could be considered beneficial.

Bertware:
-1 From Trafiklab/Samtrafiken for the currently proposed implementation

While we support the possibility to have platform changes, we are opposed to the idea of immediatly limiting this to platforms changes only. We'd be happiest with the solution where stop_sequence becomes the single selector and stop_id becomes a modifier, allowing re-routing to both platforms within the same stop and platforms on another stop. 

If we want to keep stop_id as-is, we could also agree to having another field such as updated_stop_id, similar to the proposed idea in this PR but not being limited to platforms in the same parent stop (similar to what @abyrd describes above).

tleboulenge:
Thanks for the insight on this proposal, I see that we at least all agree on the need to split selector and modifier and that the field "stop_id" has a problem in the way it's defined right now.

On the question of what is an acceptable "replacement stop" (in this proposal: "restricted to a platform in the same station"; general ask: "any stop"): It can't realistically be **any** stop, it has to be a stop that is somewhat close to the original stop (e.g. not a stop on the other side of the city). Would you bake such a restriction in the spec, as a rule, as a comment, as a guideline, or just leave it to common sense?

Bertware:
> Would you bake such a restriction in the spec, as a rule, as a comment, as a guideline, or just leave it to common sense?

I'd leave it to common sense, but it could be worded as "to another stop serving the same area" or something similar to make sure that it's still relevant for travellers going to the original stop. The case described in #226 would fit this requirement, it's one block further but still serving the same area, passengers for the original stop would get off at the new stop.

tleboulenge:
Note: the main limitation of this syntax to describe detours is that you are forced to replace one stop with one other. So it falls short of a detoured trip where 2 stops are cancelled and replaced by 3 new ones. The new Service Changes spec (another discussion for later) covers detours in a more flexible way to address these cases as well.

Platform changes are obviously simple enough candidates for the 1 stop replaced/moved to one other. But there are obviously other similar cases where this notation could still provide very useful, with this understanding of "stop serving the same area".

skinkie:
@tleboulenge the GTFS-RT FULL posts would also cover the detour solution, in my opinion in a more sustainable way.

tleboulenge:
On the topic of why to single out platform changes, here are a few thoughts:

From a pure data syntax, those 2 changes can seem equivalent and be modelled using the same syntax:
[A] The train A will now stop at platform 23 instead of platform 15
[B] The train B from Paris to Frankfurt will run via Strasbourg instead of Bruxelles

However, from all practical aspects, they are a world apart.
Case [A] is just business as usual as far as train operators and passengers are concerned. This change can be decided, broadcasted to and understood by passengers in a matter of minutes, and most of them will easily manage to catch their intended train.

Case [B] is about as unusual as it gets, the change needs to be planned for and broadcasted to the public hours or even days in advance, and will severely impact many passengers (to the extent that they will chose an alternative service).

In my opinion, case [A] fits the bill for a Trip Update (it's just a routine update about the real-time status of that trip), while case [B] is squarely a Service Change (the service is breaking previous expectations the public had about it).

In terms of processing time / advance notice, this also has very different implications: case [A] needs to be processed in a very short time (say at most 5 minutes), while it's totally OK that case [B] should take hours to be processed, and it should be expected that this is published well in advance.

The way the information is presented to passengers is also different: case [A] would probably not even be presented as a "change", just a matter of fact on a departure board: "Train-A 10:15 (+2min) pl. 15". Case [B] will almost certainly be accompanied with a notice of changed service of some kind.

So the suggestion behind the original proposal was to recognize they are 2 different cases, and provide a way to specify them in a way that fits into the spirit of their respective messages and recommended processing times:

1. Change of platform: This is expected. This belongs a TripUpdate, needs to be processed and published very fast (5 minutes)
2. Moved stop: This is unexpected. This belongs to a Service Change, can be processed in a longer time (say 2 hours) and needs to be published hours in advance (Note: the details and times in this sentence might be debated, but are out of the scope of this particular discussion). 

And in doing so, allowing us all to move forward with cases (1) soon, as we still discuss what cases (2) might look like.

Note that some producers and consumers are actually **already** doing cases (1), using the undocumented, ambiguous and error-prone syntax we all (?) agree should be clarified / avoided.

barbeau:
Sorry all, given all the discussion I just realized that I forgot to update this PR on the status of the vote. There were two -1s (@skinkie and @Bertware) that then generated a lot of discussion, so at this point I'd like to pause the vote until we can further discuss a path forward.

To summarize, there don't seem to be major objections to the general structure of the data with a new field `stop_time_properties.x` that defines the stop from `stops.txt` that is being referenced, and using `stop_sequence` as the selector for where this new stop is updated within the trip:

~~~
    stop_time_update {
      stop_sequence: 3
      ...
      stop_time_properties {
        x: "id_23"
      }
    }
~~~

It seems like the major objections here from @skinkie and @Bertware are instead of restricting this field to only modify platforms within a station (as proposed in this PR), the desire is to instead allow it to modify any stop on a trip (e.g., to model a detour).

@skinkie and @Bertware have you taken a look at the full GTFS-ServiceChanges v3.1 draft spec? 

I think it would help the discussion for you to get a high-level view of a comprehensive solution (which includes detours) that we've spent a lot of time on evolving with @tleboulenge and other producers and consumers. 

Early drafts of ServiceChanges had solutions that looked more like what you're proposing for detours, but we ended up splitting detours out into new `Trip`s for many reasons, including those that @tleboulenge mentions in https://github.com/google/transit/pull/219#issuecomment-638691463. The problem with trying to model some detours within `TripUpdates` and others in new `Trips` is that it's not clear based on any given set of circumstances that change a trip when producers should put updates in one entity vs the other. We had feedback from producers that it was confusing.

Please comment in the ServiceChanges Google Doc or further in this thread after you've reviewed the document so we can figure out how to agree on a platform changes spec - given platform changes data is currently being exchanged, I'd really like to find a way to formalize it. Thanks!!

barbeau:
@skinkie and @Bertware - just wanted to circle back with you both. Have you had a chance to review the [GTFS-ServiceChanges v3.1 draft](https://bit.ly/gtfs-service-changes-v3_1) to see how detours are communicated there?

Bertware:
Sorry for the late reply, I have been rather busy lately. I quickly looked through the ServiceChanges proposal (made some comments there). I believe that it'd be easiest to allow any change in the new stop_id here (although limited to a 1:1 change), given the relative complexity of the alternative. However, we can settle with the proposal to limit the change to platforms within the same station, assuming that support for detours as described in the ServiceChanges proposal will be implemented as well.  

One question that comes to mind right now is whether a producer is allowed to use the upcoming detour functionality to publish platform changes. This would provide two ways to achieve the same thing.

tleboulenge:
One side remark this proposal raises is whether the Vehicle Positions feed needs to be kept in sync when a platform is assigned (or changed) for the upcoming stop.

In a case where trip-1 is declared (in GTFS) to make a stop at:
stop-sequence: 1 => stop-id "platform-10"

Then in realtime (in GTFS-TU), it's assigned/changed to be:
stop-sequence: 1 => stop-id "platform-18"

I assume the Vehicle Positions feed should be kept in sync and therefore, if it provides these fields, they should match:
current_status: IN_TRANSIT_TO
current_stop_sequence: 1
stop_id: "platform-18" (and no longer "platform-10")

I'm not sure this should be part of this proposal or a distinct one, but probably should be mentioned somewhere.

barbeau:
>I assume the Vehicle Positions feed should be kept in sync and therefore, if it provides these fields, they should match

@tleboulenge Good point! I've added some language in `StopTimeProperties` and `VehiclePosition` that explicitly states this. I've also changed the terminology for "platform changes" to "platform assignments" as I believe it better reflects the concept.

Per the above discussion, it looks like @Bertware and @tleboulenge are both on-board with the current proposal.

I'd like to re-open the vote on this proposal! Vote will be closed on Monday Aug 17th at 23:59:59 UTC.

paulswartz:
+1

skinkie:
-1 i am unpleased how this (specifically platform_id) paves the way so GTFS-ServiceChanges will force producers to use GTFS-NewTrip for minor detours with an extra call. I disagree with @tleboulenge that an extra call (for a train) justifies creating new trips. Hence my opposition is mainly that I would like to exchange the new state of the timetable, where the proposals here want to be able to include inline relations to the static data.

tleboulenge:
+1

@skinkie I'm all for NOT forcing producers to use the verbose GTFS-NewTrips syntax if there's a convenient and unambiguous way to do this somehow else. Would you like to elaborate what you have in mind for such a case? We could see how much this could be either bundled as part of this change, or be subject to another small change in TripUpdates (or as a further extension of this one),...

But your vote should probably reflect whether you think that the current proposal serves a purpose and advances the spec in a constructive manner, rather than doesn't go far enough -- there's always room to add a follow-up proposal to extend it later.

skinkie:
@tleboulenge would you be available for a more 'interactive' discussion regarding the subject? I think @abyrd and myself would both prefer GTFS-RT "Full". We would have to understand why others would prefer more complexity over this.

gcamp:
+0

I'm mixed on this proposition. 

If I see it as a "bug fix" where we just want to standardize an existing practice, I think this is a good proposition. It clarifies the current Google extension with the same constraints as before. 

However, if I see the proposition as a new feature, modifying stops via Trip-Update might be a more elegant solution than using GTFS-NewTrips to add a new trip with a modified stop and cancelling the existing stop.

lauramatson:
+1 (Metro Transit, Minneapolis/St. Paul)

barbeau:
The vote is closed, and here are the results:
* Yes (+1)
    * Google (@tleboulenge)
    * Metro Transit, Minneapolis/St. Paul (@lauramatson)
    * MBTA (@paulswartz)
* No (-1)
    * Bliksem Labs B.V. (@skinkie)
* Abstain (+0)
    * Transit App (@gcamp)

We need at least 80% yes for experimental GTFS-RT features [per the official process](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#experimental-fields), so this proposal does not pass.

@skinkie @gcamp Thoughts on where we go from here?

skinkie:
@barbeau StopTimeProperties has been added to this proposal with the suggestion that this structure would be more clear that values inside it could change, opposed to the selector properties of the current StopTimeUpdate. In addition that it would allow consumers to match the static data with a realtime update in order to explicitly communicate a change (the old information is available in the message). This is in opposition to the "GTFS-RT Full" proposal where the message can fully replace a trip, including changes such as platform assignments, allowing journey planners to directly consume information, rather than postprocessing it.

**Standardisation part A**

Given that a stop sequence is mandatory is GTFS static is mandatory. It could be said that marking the stop sequence in GTFS-RT mandatory is a no regret. It directly disambiguates every communication even _future_ additions. Hence changing the wording towards:

> Must be the same as in stop_times.txt in the corresponding GTFS feed. stop_sequence must be provided within a StopTimeUpdate.

Full stop: standardise this part.

**Document part B**

Track change messages should end up where they are supposed to be: Alerts. If it is actually a track change (rather than an _assignment_) an alert with trip_id, stop_id would be good. Adding cause/effect for a track change (STOP_MOVED?). Deciding when to send an alert should be up to the agency running. My rationale here is that communicating track changes in apps don't make sence if there is no ground truth. Hence showing an hour before time that the track will not be at track 1 but instead track 7 is a distraction. Unless a user subscribes to the trip planning result, the information presented is volatile by nature, and would never get updated unless the user queries the information again. If the user queries the information again, the latest information would be available (hence the need for explicitly marking a track change is void).   

Full stop: document best-practise for this part.

**Standardisation part C**

I would be ambivalent regarding the introduction of the StopTimeProperties as a structure, as I at this point do not see the benefit of introducing it. It does not bring more clarity in my perspective, and blocks future use, in addition it increases the feed size. Standardise the fact stop_id can be updated for switching platform within the parent station. Hence the current "unofficial" way of doing so at this moment. 

Full stop: standardise this part.

**Standardisation part D**
_I would be in favor of a feed feature versioning before this goes live, since it starts assuming at the client side can handle arbitrary updates._

For this we would need to update the stop_sequence description again:

> Must be in trip order. stop_sequence must be provided within a StopTimeUpdate.

Any trip stop sequence can now be defined to introduce arbitrary possible sequences. But do require the entire trip (until termination) will be communicated. Consumers can directly use the predictions made by the server, and expect that the stop_sequence is positively ordered, in addition the arrival and departure times must increase.

Journey Planner algorithms that can handle arbitrary updates (RAPTOR, CSA) directly benefit. Others (Transfer Patterns) could either ignore the transfers at ADDED stops, or recompute the transfers on the fly.

abyrd:
Looking at recent discussion, I am no longer sure what's being proposed or voted on. The changeset  itself includes a new nested StopTimeProperties structure, and the new field called platformId that in fact references a stopId.

But people voting (such as @gcamp) seem to be referring to this as mere standardization of existing practice, or a small clarification that can be built upon incrementally. @gcamp says this "clarifies the current Google extension" and some comments from @tleboulenge seem to show the stop sequence as a selector, with stop locations specified as stop_id (without mention of StopTimeProperties or platformId).

This is confusing to me, and may be to others. People may be hesitant to vote without more clarity. Is there indeed a current, widely used extension featuring a nested StopTimeProperties with a platformId field? My understanding is that the current widely used extension simply defines the stop sequence as the selector and allows the stop ID to be changed to that of another platform within a station.

It is worth noting that only three people voted on this, so even with 80% or unanimous support there may be some concern about legitimacy due to a (perhaps unstated) notion of quorum. Maybe low participation is due to lack of clarity. Another factor is vacation season in the northern hemisphere. Many people take a month or more off work in July and August and barely even look at their email. In many places this is not seen as a good time to get feedback and obtain popular support for new rules.

sunny-lirr:
If this were adopted, would it no longer be mandatory to have either arrival or departure specified for a `schedule_relationship` of `SCHEDULED`? At LIRR it is common for us to specify track/platform changes without the train having departed yet, thus it is not necessarily accurate to say it is on time.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -245,10 +245,19 @@ message TripUpdate {
...
+      // Supports real-time stop assignments. Refers to a stop_id defined in the GTFS stops.txt.
+      // The new assigned_stop_id should not result in a significantly different trip experience for the end user than
+      // the stop_id defined in GTFS stop_times.txt. In other words, the end user should not view this new stop_id as an
+      // "unusual change" if the new stop was presented within an app without any additional context.
+      // For example, this field is intended to be used for platform assignments by using a stop_id that belongs to the
+      // same station as the stop originally defined in GTFS stop_times.txt.
+      // To assign a stop without providing any real-time arrival or departure predictions, populate this field and set
+      // StopTimeUpdate.schedule_relationship = NO_DATA.
+      // If this field is populated, StopTimeUpdate `stop_sequence` must be populated and StopTimeUpdate `stop_id`
+      // should not be populated. Platform assignments should be reflected in other GTFS-realtime fields as well
```
skinkie, line 257: Why shouldn't stop_id be populated?

barbeau:
I'd like to thank everyone for leaving feedback on this proposal. I've updated the proposal in https://github.com/google/transit/pull/219/commits/5f003fa5741ddbe555f2097ac8077c99bbdfdf15 based on this feedback to try and address everyone's concerns.

I'd encourage you to check out the diff above, but here is a high-level summary:
* Field name change of `platform_id` -> `assigned_stop_id `- The feature is no longer constrained to just platforms within stations (see next point).
* Use more generic language to define when `assigned_stop_id` can be used (versus requiring changes to be only within a station). This broadens potential uses cases of the field without being as prescriptive. There is intentional use of the word "should" vs. "must" in some cases to allow for deviation from the platform assignments use case.
* Add that you can assign a stop without providing real-time predictions by using `StopTimeUpdate.schedule_relationship = NO_DATA` (based on @sunny-lirr question above)
* Require that StopTimeUpdate.stop_sequence be used as the selector and that StopTimeUpdate.stop_id be empty when using this new field `assigned_stop_id ` (based on @skinkie feedback above in moving towards using `stop_sequence` as the selector)
* Don't block use of StopTimeUpdate.stop_id being populated with other values in case others want to have unofficial GTFS-realtime feeds that changes `stop_id` to represent a different stop (e.g., @skinkie with Bliksem Labs B.V. wants to be able to define detours by populating TripUpdates with the entire stop sequence list, aka "GTFS-realtime full"). Therefore, this proposal doesn't block the current use or potential future adoption of a "GTFS-realtime full" if the community would desire this.

Feedback is welcome!

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: This was based on your feedback that we should move towards using `stop_sequence` as the selector for stops - see https://github.com/google/transit/pull/219#issuecomment-717524976.

paulswartz:
>     * Require that StopTimeUpdate.stop_sequence be used as the selector and that StopTimeUpdate.stop_id be empty when using this new field `assigned_stop_id` (based on @skinkie feedback above in moving towards using `stop_sequence` as the selector in [#219 (comment)](https://github.com/google/transit/pull/219#issuecomment-675767164))

This seems like it breaks backwards compatibility with the existing behavior (changing `stop_id` to the platform stop ID). Maybe something like "if a `stop_id` is provided, it should match the `assigned_stop_id`"?

barbeau:
>This seems like it breaks backwards compatibility with the existing behavior (changing stop_id to the platform stop ID). Maybe something like "if a stop_id is provided, it must match the assigned_stop_id"?

@paulswartz Yes, that's a good point. What about if we change the language to the following to still express the preference towards leaving `stop_id` empty for producers that haven't been using the current unofficial legacy method:

>If `StopTimeProperties.assigned_stop_id` is populated, it is preferred to omit `stop_id` and use only `stop_sequence`. If `StopTimeProperties.assigned_stop_id` and `stop_id` are populated, `stop_id` must match `assigned_stop_id`.

barbeau:
Based on @paulswartz 👍  above, I've updated the proposal to reflect the above language for `stop_id`.

barbeau:
I'd like to call for a vote on adding this proposal as an experimental field in GTFS-realtime.

tl;dr - this feature allows producers to assign a stop for a given StopTimeUpdate in real-time, with the primary intended use case being platform assignments within a station. However, some flexibility in assignments is allowed for additional uses of real-time assignments as long as new stop does not result in a significantly different trip experience for the end user than the stop_id defined in GTFS stop_times.txt (see the diff in the PR for full description!).

Vote will be closed on Tuesday Nov 10th at 23:59:59 UTC.

@tleboulenge, @lauramatson, @paulswartz, @skinkie, @gcamp - please be sure to vote again, thanks!

gcamp:
+1 for Transit

harringtonp:
Coming to this late, haven't read through all the arguments but have looked at the changes.

I would have thought that the assigned_stop_id would have to be different to the stop_id and wouldn't have seen any reason to change the existing mechanism of supplying either the stop_id or stop_sequence or both in the StopTimeUpdate

Correct me if I'm wrong but the stated reason for the addition seems to be to allow for last minute changes from the usual stop to another close by stop. Presumably both stops are sufficiently close (as cited in the platform change example) that they can be walked in a couple of minutes. 

Taking a stop centric application, a user will typically bookmark a stop they frequently use. When looking for next departures all trips that use that actual stop or listed. If a trip that regularly uses that stop has moved to an alternative close by stop the user will undoubtedly want to know about it and it should still show in the listing. By providing both the original and assigned stop id,  developers have a possibility of making the user aware of the new location. Without any reference to the original stop id there is no chance of this. This would likely happen if the original stop id was replaced with the assigned stop id. 

I realise the original stop id can be derived as usual using the trip id and stop sequence but why would you prevent a producer from placing the stop id  in StopTimeUpdate too. I get adding StopTimeProperties but don't get why any changes are needed to the existing StopTimeUpdate fields.

barbeau:
>Correct me if I'm wrong but the stated reason for the addition seems to be to allow for last minute changes from the usual stop to another close by stop. Presumably both stops are sufficiently close (as cited in the platform change example) that they can be walked in a couple of minutes.

@harringtonp Correct - although the primary use case that this proposal addresses is platform assignments in real-time, which is an expected change that users don't consider to be "unusual".

>By providing both the original and assigned stop id, developers have a possibility of making the user aware of the new location. Without any reference to the original stop id there is no chance of this....I realise the original stop id can be derived as usual using the trip id and stop sequence but why would you prevent a producer from placing the stop id in StopTimeUpdate too

One issue with requiring that the RT `stop_id` always match the static `stop_id` is that Google currently has an unofficial "extension" to RT that allows producers to indicate platform assignments via the `stop_id` field - so the `stop_id` and `stop_sequence` fields already don't match static GTFS for some producers. The goal of this proposal is to standardize the platform assignments feature in a more predictable way that doesn't break backwards compatibility for early adopters of the custom extension. If we require that RT `stop_id` always matches static, there isn't a migration path for existing producers of the unofficial extension. I realize that it creates an extra step for consumers to retrieve the static stop_id using `stop_sequence`, but it seems a reasonable trade-off given that `stop_sequence` can already be used by itself (without `stop_id) to key a StopTimeUpdate on a stop. Generally speaking, in the future I'd like to move towards using `stop_sequence` as a required key for StopTimeUpdates because it's unambiguous, versus `stop_id` which can appear several times in a trip (e.g., in loop/lasso routes).

>When looking for next departures all trips that use that actual stop are listed. If a trip that regularly uses that stop has moved to an alternative close by stop the user will undoubtedly want to know about it and it should still show in the listing. 

The new language is intentionally loose enough to handle nearby moved stops, although it's really to the producer's discretion at how it's used. The guidance in the proposal is that "The new `assigned_stop_id` should not result in a significantly different trip experience for the end user than the `stop_id` defined in GTFS `stop_times.txt`. In other words, the end user should not view this new `stop_id` as an "unusual change" if the new stop was presented within an app without any additional context."

The field is intentionally experimental so it can be revised as producers and consumers learn more as it's implemented.

gtfs-realtime/proto/gtfs-realtime.proto
paulswartz: This language seems to conflict with the language below about using both `stop_id` and `assigned_stop_id` below. I think you could drop the part about "and StopTImeUpdate `stop_id`..." and only keep the part about `stop_sequence` being required?

paulswartz:
One small comment on the text: +1 from MBTA.

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: @paulswartz Ah, sorry, my mistake - thanks for point it out. I made the change in the spec language but missed this comment in the .proto file - I've updated it in https://github.com/google/transit/pull/219/commits/81e8804fdf88bd3b9512f9b33616ec45fa6f1340 to match the spec language.

@gcamp Just a heads up that this changed since you already voted.

tleboulenge:
+1 for Google.

I think @harringtonp brings an interesting case there (ability to map original_stop_id to assigned_stop_id for all trips), I'm curious to see if we can bring an elegant solution to that (and other existing constraints).

john-burge-sys:
+1 for Trapeze Group

barbeau:
The vote is closed, and here are the results:
* Yes (+1)
    * Google (@tleboulenge)
    * MBTA (@paulswartz)
    * Transit App (@gcamp)
    * Trapeze Group (@john-burge-sys)

So with 4 "yes" votes this proposal passes as an experimental feature! 🎉  Thanks to everyone who participated in the discussion.

PR #220, opened 2020-04-29
Move pathways, levels and attributions to the list of translatable ta…

aababilov:
…bles

The following standard fields in those tables can be translated:

* pathways.txt - signposted_as, reversed_signposted_as
* levels.txt - level_name
* attributions.txt - organization_name, attribution_url, attribution_phone

Previously, the documentation placed them in a list of tables whose standard fields should not be translated.


paulswartz:
Does it make sense to translate 'signposted_as'? If it shows up as translated in an application, that likely does not match a physical sign and doesn't help that rider navigate.

I can go either way on 'level_name', and the others seem fine.

aababilov:
Certain signboards show text in multiple languages (e.g., Singapore uses 4 languages, Finland - 2 etc). GTFS can model that by giving one language in `signposted_as` and the other languages in translations.txt.

paulswartz:
Right, but I don't see that in the diff. Might be worth calling out explicitly if it isn't somewhere already?

aababilov:
What needs to be called out in the diff? Please could you give a sample text?

timMillet:
I agree with this PR.

To address @paulswartz 's concern, I'd edit the field definition of `pathways.signposted_as`. Currently, it's written:
> String of text from physical signage visible to transit riders. The string can be used to provide text directions to users, such as 'follow signs to '. The language text should appear in this field exactly how it is printed on the signs - it should not be translated."

I'd suggest making this change:
> String of text from physical signage visible to transit riders. The string can be used to provide text directions to users, such as 'follow signs to '. The language text should appear in this field exactly how it is printed on the signs.
> When the physical signage is multilingual, this field may be populated and translated following the example of `stops.stop_name` in the field definition of `feed_info.feed_lang`.

I hope this may help.

aababilov:
Thank you, Tim. I updated the field definition of `pathways.signposted_as`.

Issue #216, opened 2020-04-24
Add duplicated trips in GTFS-realtime


ritesh-warade-ibigroup:
Case: An added trip is created in a CAD/AVL system by duplicating an existing trip or block and then assigning a schedule offset (so the duplicated trip operates XX mins before or after the original trip)

Question: What is the best way to represent such added trips in GTFS-realtime?

Two possible approaches are (1) assigning a unique entity_id and unique trip_id to the duplicated trip or (2) assigning a unique entity_id with the same trip_id from the schedule but a different start_time than the scheduled trip. 

Approach 2 seems better since then consumers would know which trip (and so its shape and stop_times) was duplicated, but 1) can two trips in the same GTFS-realtime feed have the same trip_id? 2) If yes, does this cause any issues for consumers?

Is there a better approach?

skinkie:
@ritesh-warade-ibigroup the approach that you lay out might even depend on the case. The added trip that you describe is a reinforcement trip. For that I could imagine using 2 _concurrently_. But for really added or modified trips (different route) I would say that 1 is preferable.

barbeau:
@ritesh-warade-ibigroup In short, there isn't an official way to duplicate trips in the current GTFS-RT spec. Google supports (2) with `ADDED` schedule_relationship, but we actually want to move away from this to a more robust solution (one reason being that others use `ADDED` differently).

Goining forward, the "right" way to duplicate a trip in GTFS-realtime is the new `DUPLICATED` schedule_relationship in ServiceChanges v3.1:
https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit#bookmark=id.46iq4ddo589j

I'm actually in the process of working on a new PR that's a subset of Service Changes, pulling out the fields to support `DUPLICATED`, so we could move this forward and deprecate `ADDED`. Is this something that you could use?

skinkie:
@barbeau I hope we can still use ADDED for the other purpose :)

barbeau:
@skinkie `ADDED` will be soft-deprecated in the .proto, meaning that you could still consume and produce it if you like for your own purposes. But no promises on a community consensus on what it actually means :).

sunny-lirr:
[...]

Issue #113, opened 2018-10-22
GTFS-ServiceChanges vs extending GTFS-TripUpdates


LeoFrachet:
Hi GTFS Community,

A decision has to be made regarding service changes, and the feedback of everybody is heavily needed.

Currently, many things cannot be done in real-time, including:
- Case A: Updating time of scheduled arrival/departure
- Case B: Updating trip_headsign, trip_short_name, wheelchair_accessible or bikes_allowed
- Case C: Adding a new stop, a new shape, a new route or a new transfer.
- Case D: Have unactivated trips in the GTFS, that we trigger when needed.

What we currently have on the table are:
- Existing GTFS-TripUpdates (allowing to provide real-time update for arrival & departure).
- A proposal to extend GTFS-TripUpdates (#111 , including UPDATE_SCHEDULE) by @transitapp, handling case A.
- A proposal to define a new feed, called GTFS-ServiceChanges, using the CSV GTFS structure, handling cases A, B, C & D (bit.ly/gtfs-service-changes).

So what do we do?

### Option 0: Use only the existing formats. Update static GTFS more often.
Will require overhaul of GTFS consumer GTFS pipeline to be able to process GTFS every hour or more.

### Option 1: We beef up GTFS-TripUpdates to support the cases A, B, C & D (Cc @stefan)
We already have a proposal for case A (#111 ), we could easily extend the TripUpdate object to handle case B, but no proposal so far to handle case C.

#### Advantages:
- Small change to existing pipeline for case A & B.

#### Disadvantages:
- Handling cases C & D forces us to redefine almost every GTFS concepts slightly differently, therefore duplicating the spec.

### Option 2: We keep GTFS-TripUpdates for real-time update only, and use GTFS-ServiceChanges to change schedule data (aka cases A, B, C & D)
This is what I had in mind when I drafted GTFS-ServiceChanges, and this is the current state of the GTFS-ServiceChanges proposal.

#### Advantage:
- It makes a clear distinction between schedule updates and real-time updates
- It keep TripUpdate implementation as they are
- As branding, it pushes agency to adopt this new feed as a new feature for their rider
- It doesn’t increase the size of the TripUpdates feed (but it may be not an issue @stefan @gcamp)

#### Disadvantage
- Some overlapping for adding feed with GTFS-TripUpdates
- It’s a distinct feed, and force to open a new endpoint, with a new URL

### Option 3: Middle ground proposed by Transit (Cc @gcamp & @juanborre)
[...]

Issue #106, opened 2018-09-24
Anyone using GTFS-rt trip.schedule_relationship=ADDED?


barbeau:
Is anyone currently using the GTFS-rt TripDescriptor.ScheduleRelationship = `ADDED` value?

It's defined here:
https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-schedulerelationship-1

...as:

>An extra trip that was added in addition to a running schedule, for example, to replace a broken vehicle or to respond to sudden passenger load.

And text above that says:

>If a trip is done in accordance with temporary schedule, not reflected in GTFS, then it shouldn't be marked as SCHEDULED, but marked as ADDED.

Also posted to Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/W6bm2Xj3p-Q.

dmaltzan:
At the MBTA we use the `ADDED` value for certain trips our subway routes. We primarily use it because the underlying data we get from our signaling system on some subway lines does not contain any way of matching up live trips to scheduled trips. On other lines, we do have schedule data, but there's no real guarantee that it'll match up with our GTFS schedule because it comes from a system that does not reflect shuttle-bus diversions etc. (whereas that information _is_ put into our GTFS file).

Because of this, we use our own algorithm to search for a GTFS trip_id that matches up with each subway trip as soon as we need to make predictions for it. Currently we look for a trip that starts within plus or minus ten minutes of the time that we're predicting this trip will start, and has the same pattern of service. If no GTFS trip exists matching those criteria (or if one exists but has already been used) then we will create an `ADDED` trip.

In particular, this means that certain patterns of service will _never_ match up with a GTFS trip. For example, we run some Green Line trips from Kenmore to Park Street on a "run as directed" basis during Red Sox games, but these are not scheduled so they will never match and will always show up as an `ADDED` trip.

Hope this helps.

barbeau:
Thanks everyone for your responses.  I'm closing this issue, but have referenced it in the issue on the larger debate of how we provide full support for added trips going forward - https://github.com/google/transit/issues/113#issuecomment-436384798.

barbeau:
@skinkie Do you currently have an implementation of ADDED trips?

From your recent comments (https://github.com/google/transit/issues/216, https://github.com/google/transit/pull/221) it sounds like you do.  We're trying to collect existing implementations of ADDED trips in this issue, given that ADDED trips are not well-described in the current GTFS-RT spec. 

[...]

PR #221, opened 2020-05-04
Proposal: Support DUPLICATED trips in GTFS-RT 

barbeau:
### Background

Producers and consumers have identified the need for a lightweight method to take an existing trip defined in static GTFS, and then in real-time say that its pattern (e.g., the stop_times.txt definition) is running at a new service date and/or time [1].

The GTFS-RT spec does not currently define a official way to duplicate trips. The `schedule_relationship` of `ADDED` is a potential candidate to convey this type of information. However, the GTFS-RT spec has not clearly defined what exactly `ADDED` is and how it should be used. As a result, multiple consumers and producers have different interpretations of what an `ADDED` trip is [2][3].

### Proposal

This proposal adds a new trip `schedule_relationship` of `DUPLICATED`, which can be used to duplicate an existing trip from GTFS (CSV) but then run that trip pattern at a new time (`TripProperties.start_time`) and/or service date (`TripProperties.start_date`). The existing field `TripDescriptor.trip_id` is used to identify the trip to be duplicated in GTFS trips.txt (CSV), and `TripProperties.trip_id` defines the new `trip_id` for the new trip.

~~This proposal also deprecates the existing trip `schedule_relationship` of `ADDED`, which has be used differently by various producers and consumers with varying meaning due to lack of clarify in the current spec [2][3]. `ADDED` is soft-deprecated, which means that if existing producers and consumers want to continue using it for their own purposes it will remain usable in the .proto file.~~ **EDIT** Per discussion below, the deprecation of `ADDED` is being removed from this proposal for future discussion separately.

A few producers and at least one consumer currently implement this use case via ADDED, so it should be relatively straightforward to change the implementation to using `DUPLICATED` instead. Therefore, this feature could be adopted prior to the remainder of the ServiceChanges proposal.

Future proposals may add other fields defined in GTFS trips.txt to the `TripProperties` message  (e.g., route_id, trip_headsign, trip_short_name, block_id, shape_id) so they can be changed in real-time. See the [ServiceChanges proposal](https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit#bookmark=id.rne8sft785w0) for details.

**EDIT** July 9th, 2020 - Added a [migration guide](https://github.com/google/transit/pull/221#issuecomment-656365202) for existing consumers and producers publishing duplicated trips using the `ADDED` enumeration so there is a standardized transition to the `DUPLICATED` enumeration.

**EDIT** July 14th, 2020 - Updated the [migration guide](https://github.com/google/transit/pull/221#issuecomment-658257550) to allow linking trips via 2nd option to accommodate MBTA interpretation of `ADDED`.

Announced on the Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/XL96r9g3W-8.

[1] https://github.com/google/transit/issues/216
[2] https://github.com/google/transit/issues/106
[3] https://github.com/google/transit/issues/113#issuecomment-436384798


skinkie:
I noticed that you explicitly mention the "soft-deprecation" again, which I don't know the purpose of. Hence I don't see the difference of an ADDED trip as in "a fully new defined trip is added" versus "a reinforcement trip is started" I fail to see why two different semantics are be required.

This proposal by MobilityData is again trying to achieve multiple things at once, but fail to deliver a complete solution, documented solution for the changes proposed. First of all the introduction of a new concept, and in the background the removal of something else, but not solving the use case that the removal or "soft removal" or what new speak you want to add to it would justify it. Either splitting DUPLICATED from ADDED (as being more specific) wouldn't have to touch the definition of ADDED at all. It should be a separate pull request.

To conclude:
 -1: I don't like how the trip duplication introduces TripProperties it does not make sense. A reinforcement trip uses the same information as the trip it reinforces with typically one extra number to identify the number of reinforcements. We don't have such reinforcement number in GTFS-RT but we can have a different start_time, this is in line with the frequency based definition. I think it is reasonable to assume there there always will be a second difference between two departing trips.
 -1: I don't like how this deprecates ADDED
 -1: I think this could once and for all be solved with a good definition of a (generic) ADDED of which (duplication) is a specific instance. I see you rather introduce something new, than force organisations to converge their implementations to a single functional description.

barbeau:
>I noticed that you explicitly mention the "soft-deprecation" again, which I don't know the purpose of.

* Soft-deprecated - Field is removed from docs and marked as `[deprecated = true]` in .proto. When you re-generate the new bindings from the `gtfs-realtime.proto` file, you can continue to use the deprecated field in your own systems if you'd like if you already have an implementation for your own purposes. But there is no community consensus on the definition of the field.
* Hard-deprecated - Field is removed from docs and marked as `reserved` in .proto, or removed entirely from the .proto. When you re-generate the new bindings from the `gtfs-realtime.proto` file, you CANNOT use the deprecated field in your own systems. You would need to stop using the field in any existing implementations (e.g., change to using an extension instead).

tleboulenge:
> Hence I don't see the difference of an ADDED trip as in "a fully new defined trip is added" versus "a reinforcement trip is started" I fail to see why two different semantics are be required.

I can see 2 distinct cases here where you have a trip-A defined in GTFS and you want to run a vehicle on a trip-B that isn't defined there:

1. Trip-B looks a lot like trip-A (best case is a reinforcement, where it's identical), and you want to define trip-B as trip-A + some changes (required change: assign a new trip-id; optional changes: a new start time, skip some stops, other usual things you can do to alter a trip in realtime; perhaps change headsigns or other in the future)

2. Trip-B really doesn't look like trip-A and it's not useful to use trip-A as a template to define trip-B. You then have to fully describe a new trip from scratch: list of stops, times, headsigns, etc.

This pull request is exclusively about (1). There is a whole section in the Service Changes proposal about (2), but it's another level of complexity.
Of course, once both proposals exist, you can always use the verbose syntax of (2) to define the simple case of (1), but I see that as needlessly complex and confusing. 
The case of reinforcement buses (or slightly modified trips) is IMO common enough that it warrants a shortcut notation that makes it very explicit what it really is (roughly an extra service on that route), and very readable (e.g. the new 3:21 bus runs as the 3:20 except for the last 5 stops).

Irrespective of the suggested syntax and deprecation, do you agree with this need?

skinkie:
@tleboulenge my point is, that I can model example 1 in your example 2. Doing so as producer and consumer I have to maintain fewer lines of code, that is _less complex_ and in my humble opinion less confusing.

I do not agree with the shortcut. Because it would allow to _bypass_ a shortcut later on with the full specification, having to handle both situations in both code paths. But you ask do I agree with the need? I do, and therefore it is already implemented using ADDED.

tleboulenge:
> I do, and therefore it is already implemented using ADDED.

ADDED in its current form is very far from what I called (2) above, i.e. a full-featured solution to define a new trip in Realtime (e.g. giving a whole new sequence of stops and times, plus other attached information such as headsigns).

It _is_ the shortcut you disagree with, and moreover various producers understand its syntax in a different way:
a. is trip-id referring to an existing trip-id from GTFS, or 
b. is it a new unique id that can be used to refer uniquely to this new trip?

and consumers handle or ignore different usages of it (e.g. Google reads this as a reinforcement trip, but won't read a detour in it - I'm sure others do something different), and so whether we like it or not, we're already in the world where we have an incomplete solution and we will need to create a new (most likely incompatible) way towards the full solution (define a whole new trip). 

In the current situation for ADDED to be able to represent a reinforcement trip, you need its trip-id to refer to an existing GTFS trip-id (solution (a) above, not (b)), and therefore it's impossible to assign this duplicate trip a unique trip-id and to refer to it uniquely in subsequent messages (or in the other sibling feeds). There are of course many hacks around this... Have some kind of naming convention in the trip-id itself, assign it a slightly different start time,... but I'd really wish for this spec to become more explicit and standard across all the actors, and avoid those kind of semi-documented partial ad-hoc stopgaps.

barbeau:
In addition to @tleboulenge's points above, another big problem with using the current spec fields for reinforcement trips is that it violates the design conventions of using `TripDescriptor` as a selector from static GTFS and not a modifier. The only way to specify a different starting time is to use `TripDescriptor.start_time`. There seems to be consensus in the discussion in https://github.com/google/transit/pull/219 that mixing selectors and modifiers in the same field causes problems, so we don't want to introduce that problem here.

This is also the rationale behind the design decision to create a new message `TripProperties` to hold the trip attributes that can be modified within a `TripUpdate`. `TripDescriptor` remains the selector, and `TripProperties` is the modifier.

sam-hickey-ibigroup:
The proposed addition of DUPLICATED looks good, but we are in favor of not soft-deprecating ADDED. We see the value in keeping ADDED so producers can provide real-time info for added/unscheduled trips. As an example, we saw yesterday that MBTA had 49 Green Line trips in their VP feed, and 9 of those were ADDED (see https://cdn.mbta.com/realtime/VehiclePositions.pb and https://cdn.mbta.com/realtime/TripUpdates.pb). It may be worth having a discussion about deprecating ADDED once the full Service Changes proposal is finalized, but it may also be worth keeping ADDED once the Service Changes proposal is in place for the same simplicity/shortcut reason it makes sense to keep DUPLICATED in the long term.

botanize:
Since our CAD/AVL software produces messages with ADDED I would like to see consensus on how those should be consumed, and I would like relevant applications to actually consume them. 

However, I sense that there is more momentum behind ServiceChanges, so it's probably not worth trying too hard to clarify ADDED unless it is part of the ServiceChanges implementation. I think that puts me in agreement with @sam-hickey-ibigroup.

barbeau:
Based on feedback in this thread, I've updated the PR so that the `ADDED` enumeration is no longer deprecated in this proposal - we can table that discussion to be part of the larger change set in ServiceChanges v3.1 (https://bit.ly/gtfs-service-changes-v3_1) that also handles detours, etc.

So, this proposal now focuses only on defining a very specific use case for `DUPLICATED` trips, using the `TripDescriptor` as the selector for the trip to duplicate from static GTFS and a new `TripProperties` container to specify the new start service date and time.

I'd like to move forward with this updated proposal - any specific thoughts before we call for a vote?

jamespfennell:
My main concern with adding extra unconstrained fields to the TripDescriptor like this is that inevitably some producer will misuse them. For example, I can see a producer publishing a `SCHEDULED` trip with a valid trip ID but _also_ providing a new `TripProperties` to indicate that the scheduled start time has been updated. The current documentation on the TripProperties type ("Defines updated properties of the trip") seems to suggest that this is a valid use case.

I'm on the consumer side, and all these edge cases have to be accounted for.

To that end, at the very least I would personally like the `TripProperties` message to be renamed `DuplicateTripProperties` and for it to be made more clear in the comments that it will be ignored unless the schedule relationship is `DUPLICATED`.

barbeau:
@jamespfennell That's certainly a valid concern. I've updated the text description for the `TripProperties` fields in this proposal - previously we said:

>Required if schedule_relationship=DUPLICATED, forbidden otherwise.

Now that's better explained as:

>Required if schedule_relationship=DUPLICATED, otherwise this field must not be populated and will be ignored by consumers.

Hopefully that helps.

As far as renaming `TripProperties`, if you look at the larger Service Changes v3.1 proposal draft (https://bit.ly/gtfs-service-changes-v3_1), we do plan to use `TripProperties` to allow modifying other trip attributes (e.g., headsign, short name) in real-time - see examples here:
https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit#bookmark=id.46iq4ddo589j

Those features are outside the scope of this specific PR, but we do want to use naming here that can accommodate those future changes.

barbeau:

Vote will be closed on Wednesday June 3rd at 23:59:59 UTC.

skinkie:
While I firmly believe that ADDED would be sufficient, I don't mind to have a more explict DUPLICATED for reinforced trips. The TripProperties future use should like a direction that the standard builds upon further, allowing to retain a specific relationship with the schedule.

+1 (OpenGeo)

sam-hickey-ibigroup:
We don't see an explicit mention that stop times on duplicated trips are adjusted based on the offset between the original trip start time and the duplicated trip's start time, but this is how we are interpreting this addition to the spec. Is this the correct interpretation?

prhod:
Sorry for asking a question this late, but I see in the PR that, for VehiclePosition: 
`When 'schedule_relationship' is 'DUPLICATED', the 'trip_id' identifies the trip from static GTFS to be duplicated.`
If I understand correctly, you could have both cases:
- if the trip A (used for duplication) is DELETED, the VehiclePosition will apply to trip B (the duplicated one) 
- if the trip A  is not DELETED, the VehiclePosition will apply to trip A (or both of them ?)

Shouldn't we use the new trip_id to reference the position of the trip B ?

barbeau:
@sam-hickey-ibigroup and @prhod Great feedback, thanks! I agree that we should clarify both of these items in the spec language. I'm going to pause the vote and work on some updated text for both cases.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -655,7 +655,8 @@ message TripDescriptor {
     // (in calendar.txt or calendar_dates.txt) is operating within the next 30 days. The trip to be duplicated is
     // identified via TripUpdate.TripDescriptor.trip_id. This enumeration does not modify the existing trip referenced by
     // TripUpdate.TripDescriptor.trip_id - if a producer wants to cancel the original trip, it must publish a separate
-    // TripUpdate with the value of CANCELED. The VehiclePosition.TripDescriptor.trip_id for the new trip must contain
+    // TripUpdate with the value of CANCELED. Trips defined in GTFS frequencies.txt with exact_times that is empty or
```
skinkie, line 658: Why not? A frequency based trip can also have a reinforcement right?

barbeau: @skinkie Yes, a true frequency-based trip (exact_times=0) can have a reinforcement trip, but IMHO the `DUPLICATED` mechanism isn't the right way to express it.

I had to think through this a bit, but here are my thoughts - for RT, using the existing GTFS-RT spec you can just add another TripUpdate with the `trip_id`, `start_date`, and `start_time` and that represents a new frequency-based vehicle. RT frequency-based trip instances are materialized on-the-fly (i.e., there isn't a 1:1 match to static GTFS, unlike normal scheduled trips or exact_times=1 trips), so you could run as many vehicles as you want on the route, including re-enforcement trips  (two buses running bumper-to-bumper), and existing consumers of frequency-based RT data should show them.

If you're trying to express a single new re-enforcement frequency trip without real-time info, then you really don't need to do anything as of today - the UI should still read "arriving every 10 minutes", and that should make sense to the user. The [static spec](https://gtfs.org/reference/static/#frequenciestxt) also says you can't have overlapping headways, so this interpretation seems to align.

If you wanted to change the static headway for the entire service period to say buses were expected to come twice as frequently ("arriving every 5 min" instead of 10 min), then the FrequencyUpdates that we've defined in the larger [GTFS-ServiceChanges v3.1](https://bit.ly/gtfs-service-changes-v3_1) draft would work.

barbeau:
Hi all - I've updated the proposal to further clarify it based on @sam-hickey-ibigroup and @prhod's comments. Nothing substantive changed in concept:
1. In https://github.com/google/transit/pull/221/commits/a772563cf1229d0d41cf9550e6dfca340371d809, I've explictly defined how the duplicated trip schedule and RT values are calculated as specified by @sam-hickey-ibigroup, including an example, so there is no ambiguity.
1. In https://github.com/google/transit/pull/221/commits/853eb4a21a0c7ab3d640d73b019406b6b483a882, I've added more clarify about the relationship between DUPLICATED in VehiclePositions and in TripUpdates. This uses @prhod interpretation that the `trip_id` in a VehiclePosition for a duplicated trip would be the same `trip_id` as `TripUpdate.TripProperties.trip_id`. Below is an example of of a TripUpdate that defines a new duplicated trip, and then two VehiclePositions for the original trip and the duplicated trip:

```proto
entity {
  id: "ei0"
  trip_update {
    trip: {
      trip_id: "TripId0"
      schedule_relationship: DUPLICATED
    }
    trip_properties {
      trip_id: "AddedTripId1"
      start_date: "20190821"
      start_time: "11:30:00"
    }
    stop_time_update {
	...
    }
  }
}

entity {
  id: "vehicle_position_0"
  vehicle {
    trip {
      trip_id: "TripId0"
	  schedule_relationship: SCHEDULED
    }
    position {
      latitude: 28.01125
      longitude: -82.9832
    }
    timestamp: 1592515839
    vehicle {
      id: "0"
    }
  }
}

entity {
  id: "vehicle_position_1"
  vehicle {
    trip {
      trip_id: "AddedTripId1"
	  schedule_relationship: DUPLICATED
    }
    position {
      latitude: 28.066446
      longitude: -82.42988
    }
    timestamp: 1592513933
    vehicle {
      id: "1"
    }
  }
}
```

In https://github.com/google/transit/pull/221/commits/fe8f4844463693a3deaf7810ec3306c64fe47c3b, I've also added further clarification that `DUPLICATED` can't be used in true frequency-based trips (exact_times=0). Re-enforcement frequency-based trips can be added in RT with the existing spec `TripUpdates` - see https://github.com/google/transit/pull/221#discussion_r443041176.

Comments are welcome! I'll wait a few days for any comments before opening the vote again.

barbeau:
I'd like to re-start the vote for this proposal to be added as experimental fields in GTFS-RT. Vote will be closed on Tuesday June 30th at 23:59:59 UTC.

paulswartz:
@barbeau what about backwards compatibility? If we start providing a feed which uses the `DUPLICATED` schedule relationship, I believe old clients will not see the new trip. Can we provide both a `DUPLICATED` trip and an `ADDED` trip with the same ID to support both clients of clients at the same time? Or should we consider something like increasing the GTFS-R version number?

barbeau:
@paulswartz Good question! You're right, if you immediately hard-switch to using `DUPLICATED`, any existing consumers that have implemented this use case of `ADDED` but hasn't yet transitioned to `DUPLICATED` wouldn't see these trips.

Off the top of my head, assuming you different time windows where multiple consumers would be transitioning to `DUPLICATED`, I would say providing both `DUPLICATED` and `ADDED` trips for the same ID would make sense. Having multiple TripUpdates for the same trip_id is technically out-of-spec, but then again so is this use of `ADDED`, so it seems to be reasonable 
 given the circumstances (🤷 ?). We could also try to reach out to existing consumers of `ADDED`, if you know of them, and try to coordinate a hard-switch date if there are only a few. I don't think adding a version to GTFS-RT helps here, as you still have the same issue with legacy consumers not seeing the new trips (unless you offer multiple feeds in parallel).

I'd welcome comments from the existing consumers (or producers) of `ADDED` for the use case of duplicating trips if anyone has other ideas for transitioning.

gcamp:
Except knowing from MBTA that these would be duplicated, how can we know they are? I would prefer a version increment because adding both added and duplicated would show both departure in what will be our implementation

barbeau:
@paulswartz @gcamp Could you expand on how exactly do you see a spec version increment being used here?

gcamp:
What I would imagine is there's a 2.1 release, and MBTA creates two feeds 
1) version 2.1 with DUPLICATED
2) 2.0 with ADDED

skinkie:
@gcamp wouldn't this just stipulate that DUPLICATED is a change that is too big of a change, but on its own is still to incomplete to justify a release?

barbeau:
Just to back up a step, I'd like to give data examples of the migration that we're talking about for the use case of the "copy/paste" trip we're trying to address with `DUPLICATED` - taking an existing static GTFS trip and running a duplicate of it at a new date/time. 

*Note that `ADDED` trips are currently not well-defined in GTFS-RT and are used many different ways - we're talking about standardizing just one use case of `ADDED` here.*

Here's an example of how duplicated trips are currently being hacked into `ADDED` trip_updates:

~~~
entity {
  id: "ei0"
  trip_update {
    trip: {
      trip_id: "TripId0" // <-- trip_id from static GTFS to copy
      schedule_relationship: ADDED
      start_date: "20190821" // <-- This would normally point to static, but instead defines the new trip date
      start_time: "11:30:00" // <-- This would normally point to static, but instead defines the new trip time
    }
    stop_time_update {
	...
    }
  }
}
~~~

...and here's how it would look with the new `DUPLICATED` proposal:

~~~
entity {
  id: "ei0"
  trip_update {
    trip: {
      trip_id: "TripId0" // <-- trip_id from static GTFS to copy
      schedule_relationship: DUPLICATED
    }
    trip_properties {
      trip_id: "AddedTripId1" // new trip_id unique to this trip
      start_date: "20190821" // new trip date
      start_time: "11:30:00"   // new trip time
    }
    stop_time_update {
	...
    }
  }
}
~~~

As I mentioned above, one relatively simple migration strategy here is to just have both messages side-by-side in the feed. As long as `trip.trip_id` points to the same static trip_id in both cases, legacy consumers will ignore `DUPLICATED` trips and new consumers can ignore `ADDED` trips that have the same `trip_id` as `DUPLICATED` trips. This avoids producers from having to set up an entirely new production GTFS-RT endpoint for switching over. 

I would suggest explicitly adding this migration guide to the spec, like:

>Migration Guide: For producers that have implemented duplicated trips via ADDED, when DUPLICATED trips are added to the feed the corresponding ADDED trip should have the same `trip.trip_id`. Consumers will then ignore any ADDED trips that have the same `trip.trip_id` as a DUPLICATED trip.

@paulswartz Thoughts on this as a producer?

paulswartz:
This isn't how we use `ADDED` trips, even in the situation where we're actually duplicating a trip with a different start time.

For some trips, it's not based on anything from static GTFS (other than stops + routes, I suppose). These get an ID based on a timestamp so that they're unique, but otherwise don't refer to a trip from GTFS.

For overloaded trips, where we're creating a new trip by adjusting the start time (and also running the original trip), it looks like this:
```
entity {
  id: "overload1"
  trip_update {
    trip: {
      trip_id: "TripId0-Overload1" 
      schedule_relationship: ADDED
      start_date: "20190821" 
      start_time: "11:30:00"
    }
    stop_time_update {
	...
    }
  }
}
```

The issue with the `DUPLICATED` trip is that now the `trip_id` is the static trip ID, not the new trip ID, and I worry about consumers being confused:
```
entity {
  id: "overload1"
  trip_update {
    trip: {
      trip_id: "TripId0"
      schedule_relationship: SCHEDULED
    }
    trip_properties {
      trip_id: "TripId0-Overload1"
      start_date: "20190821"
      start_time: "11:15:00" // originally scheduled at 11:15
    }
    stop_time_update {
	... // these are predicted arrival/departure times for the original trip
    }
  }
}
entity {
  id: "overload1"
  trip_update {
    trip: {
      trip_id: "TripId0"
      schedule_relationship: DUPLICATED
    }
    trip_properties {
      trip_id: "TripId0-Overload1"
      start_date: "20190821"
      start_time: "11:30:00" // extra trip running 15 minutes later
    }
    stop_time_update {
	... // QUESTION: are these predicted arrival/departure times? or the static times, shifted by 15 minutes?
    }
  }
}
```

The other possibility I can see here is a separate `ServiceChanges` feed which can include DUPLICATED trips, as well as any of the other future ServiceChanges updates. 

@gcamp what would happen for you if you got both an `ADDED` and `DUPLICATED` TripUpdate for the same trip ID?

darylweinberg:
And as a follow on question, if there is no relation to something in the static GTFS because it is a new TripID will consumers display it? Without a correlation to the static GTFS, how would it be matched to a trip headsign for example?


Daryl Weinberg
Transit Systems Architect
o: 512-369-6216
|
|

tleboulenge:
> This isn't how we use `ADDED` trips, even in the situation where we're actually duplicating a trip with a different start time.
> 
> For some trips, it's not based on anything from static GTFS (other than stops + routes, I suppose). These get an ID based on a timestamp so that they're unique, but otherwise don't refer to a trip from GTFS.
> 
> For overloaded trips, where we're creating a new trip by adjusting the start time (and also running the original trip), it looks like this:
> 
> ```
> entity {
>   id: "overload1"
>   trip_update {
>     trip: {
>       trip_id: "TripId0-Overload1" 
>       schedule_relationship: ADDED
>       start_date: "20190821" 
>       start_time: "11:30:00"
>     }
>     stop_time_update {
> 	...
>     }
>   }
> }
> ```
> 

@paulswartz That syntax doesn't match what we currently expect (at Google), and these entities are probably just dropped with a message like "trip id TripId0-Overload1 does not exist"

Which I assume is another good example of why we DO need the current proposal.

For the discussion on how to handle the transition (as suggested by @barbeau or another mean), it would be great to hear from other consumers how they handle (or not) ADDED trips today. Perhaps they don't, and there is no transition to care about?

skinkie:
> it would be great to hear from other consumers how they handle (or not) ADDED trips today. Perhaps they don't, and there is no transition to care about?

We handle them as consumer via the assumption they are provided in "GTFS-RT full" the entire trip is materialised with all departures/arrivals, and they will introduce a new trip in the journey planner.

juanborre:
The way we currently implement ADDED trips at Transit is that we duplicate another trip on the same route (and direction if it is provided) whose list of stopTimeUpdates more closely matches the list of stopTimeUpdates of the new ADDED trip.

As a consequence of this:
- If there is no route_id in the ADDED trip, we just ignore it (this is not clearly defined in the spec today but, without the route_id, how can you know where to display the trip?)
- We ignore the tripId completely and just create a new trip.

@paulswartz Without any special logic (like ignore the ADDED if there is a DUPLICATED with the same tripId), we would just create two trips. There is no way one can know they are the same trip unless it makes part of the protocol.

About the logic to handle the transition: it feels obviously _wrong_, but it can be a good compromise to ease the transition.

barbeau:
I've added a migration guide to this proposal in https://github.com/google/transit/pull/221/commits/6de5f97685a58a7c3003940fb5ea4f5d622d6e4e as [discussed above](https://github.com/google/transit/pull/221#issuecomment-649784469). This is only for producers and consumers currently using `ADDED` for duplicated trips who will transition to `DUPLICATED`.

Google (@tleboulenge) and Transit App (@gcamp, @juanborre) are in favor of this approach.

@paulswartz does this adequately address the concern for MBTA you raised in https://github.com/google/transit/pull/221#issuecomment-648848631?

If so, I'll restart the vote on this proposal.

skinkie:
@tleboulenge this means Google Maps supports after this is merged?

paulswartz:
@barbeau since we don't create the `ADDED` trip with an existing ID, we also wouldn't migrate in the way you describe. I think we'd do something like this:

- create a new `ServiceChanges.pb` feed
` TripUpdates.pb` continues to use `ADDED`, with the new trip IDs
- `ServiceChanges.pb` has the `DUPLICATED` trip, with the `trip_properties.trip_id` being the same as the trip ID in `TripUpdates.pb`, and `trip_update.trip.trip_id` being the trip ID from the static GTFS

This seems like it would allow new consumers to upgrade to ServiceChanges, while not forcing a change to existing consumers. @tleboulenge @gcamp @juanborre would you be able to consume a 4th feed from MBTA?

gcamp:
Having two gtfs-rt feeds is not an impossible task but it's certainly not ideal for us. I'd largely prefer the earlier proposition of two `TripUpdates.pb` with different versions.

tleboulenge:
> @tleboulenge this means Google Maps supports after this is merged?

Yes, with a bit of time to actually code the new syntax.

@paulswartz I'd rather keep this particular change into one TripUpdates feed (since in the end, it's also where it's going to stay), but we may amend the migration guide to allow both your current syntax and the new one (where your ADDED.trip-id would match the new DUPLICATED.trip-properties.trip-id). I think this should be sufficient to avoid duplicating the duplicate trips, if I may say so...

barbeau:
>@paulswartz I'd rather keep this particular change into one TripUpdates feed (since in the end, it's also where it's going to stay), but we may amend the migration guide to allow both your current syntax and the new one (where your ADDED.trip-id would match the new DUPLICATED.trip-properties.trip-id). I think this should be sufficient to avoid duplicating the duplicate trips, if I may say so...

@tleboulenge Thanks, that's a good idea! I just updated the migration guide in https://github.com/google/transit/pull/221/commits/426610d777bc3e7e0750acabaf564189194cedce to allow for this 2nd option for linking the two entities (`trip.trip_id` of the `ADDED` trip is the same as the `DUPLICATED` trip `trip_properties.trip_id`)

@paulswartz @gcamp @juanborre Does this work for you?

juanborre:
Yes, that works for Transit 👍

paulswartz:
That looks like it'll work for @mbta as well.

barbeau:
Great! Looks like Google (@tleboulenge) and Transit App (@gcamp, @juanborre) and MBTA (@paulswartz) are now all in favor of the current proposal and migration guide.

I'd like to re-start the vote for this proposal to be added as experimental fields in GTFS-RT. Vote will be closed on Tuesday July 28th at 23:59:59 UTC.

@tleboulenge (Google), @sam-hickey-ibigroup or @ibi-group-team (IBI Group), @gcamp or @juanborre (Transit App), @paulswartz (MBTA), and @skinkie (OpenGeo) you've all expressed support for the proposal in this thread. Could you please vote with a "+1" again?

cc @timMillet

juanborre:
+1 for Transit 🎉

paulswartz:
+1 for @mbta

tleboulenge:
+1 for Google

sam-hickey-ibigroup:
+1 for IBI Group

timMillet:
The voting period ended and the proposal is now adopted as experimental!

4 votes in favor:
@juanborre for Transit
@paulswartz for MBTA
@tleboulenge for Apple
@sam-hickey-ibigroup for IBI Group

No abstentions and vetos.

I'll merge!

barbeau:
Hi all! I just wanted to check back in and see if there were any implementations of DUPLICATED trips by producers or consumers yet. This field is still experimental now, but once we have producers and consumers it can be officially adopted.

In particular, @juanborre, @paulswartz, @tleboulenge, @sam-hickey-ibigroup - any movement towards implementations on your end?

juanborre:
👋 We haven't implemented it yet but we will when there is a producer. It won't be long.

sam-hickey-ibigroup:
Sorry for the delayed response. We also haven't implemented it yet as none of the transit agencies we work with are interested in using this at this point.

leonardehrenfried:
Just an update for all interested parties: it looks like OpenTripPlanner will start implementing this at some stage as an agency is interested in this feature.

wassimbenaissa:
Hello everyone,

First of all, thank you for all the content you produced on this topic :) 

I'd like to follow up on Barbeau's question: Have you come across any implementations of the "DUPLICATED" use case?

leonardehrenfried:
I have been working with GTFS-RT for > 4 years and have not encountered this once.

skinkie:
> I'd like to follow up on Barbeau's question: Have you come across any implementations of the "DUPLICATED" use case?

Fortifications (running an extra vehicle to combat full buses) is something defined in the Dutch standards, and was implemented as well. But in the last years, such instances have been very scares. So from an operational context, finding an example is not very easy.

wassimbenaissa:
Thank you for your response. 

I've been looking for examples of "DUPLICATED" on the [French open data](https://transport.data.gouv.fr/datasets?type=public-transit&filter=has_realtime) but I haven't found anything for now. 

And what about the "ADDED" use case? How often do the agencies you're in contact with utilize it?

jfabi:
At the @MBTA, we’ve been interested in the possible use of `DUPLICATED` (as well as `DELETED`) and may be taking a look at it later this year; for our subway lines, we fully produce the VehiclePositions and TripUpdates ourselves.

We currently make heavy use of the `ADDED` relationship whenever we’re unable to match a train's trajectory to a GTFS-static trip, and/or when the departure doesn't closely match a scheduled time, such as when we add unscheduled trips. What we've found, however, is that while we make this work well enough for our in-house applications, the spec is unclear about whether new trip IDs can be used, which is our practice (see also the references in [the top post of this PR](https://github.com/google/transit/pull/221#issue-612133479)), and so most third-party apps do not support predictions for our `ADDED` trips. See also @abyrd's [great summary of the issue in the OpenTripPlanner PR linked directly above](https://github.com/opentripplanner/OpenTripPlanner/pull/5726#issuecomment-2004044536).

For us, `DUPLICATED` would be an opportunity to replace some **but not all** of the trips we currently mark as `ADDED`.

Issue #215, opened 2020-04-13
versioning the static GTFS spec


drewda:
It's great to have a more regular cadence of additions to the static spec. However, this does make it more of a challenge for tool-makers to track which additions we support and for users to compare compatibility across different tools.

Currently, the best way to do this is to include a commit hash from this repo. However, that's not that easy for others to read.

It would be helpful if the static spec in the master branch had a version number, stored either in the Markdown file itself or as a release/tag on the repo.

Just a `YYYY-MM-DD` date of the last change merged into master would probably work fine.

scmcca:
There is currently a "Revision History" kept in CHANGES.md and a date indicating the last revision at the top of reference.md. However, these have not been routinely updated as there is no obligation or agreement for any party to do so. I see a couple of options:
1. MobilityData takes responsibility for updating the revision history anytime there is a major change (excluding things like minor editorial changes).
2. Updating the revision history becomes an obligation under the specification amendment process.

Thoughts whether either of these address the versioning problem?

scmcca:
We could also borrow from GBFS to have an easy and structured way to indicate the version within the documentation: https://github.com/NABSA/gbfs#specification-versioning

scmcca:
I just opened #267 for an update to the Revision History, and included the date of last revision at the top of reference.md. 

Is this a satisfactory versioning method?

e-lo:
> Is this a satisfactory versioning method?

skinkie:
> I just opened #267 for an update to the Revision History, and included the date of last revision at the top of reference.md.
> 
> Is this a satisfactory versioning method?

The problem with the version method is that it does not enforce usage. Hence stating to support version x.y.z does not mean anything. I wonder if a feature based version mechanism would be better.

e-lo:
> Is this a satisfactory versioning method?

In the opinion of [Cal-ITP](calitp.org) it is necessary, but not sufficient to use the revision history as a versioning method.
A few thoughts as to why a rigorous version/release system is critical:

1. Breaking Changes
I believe the "easy" reason why GTFS has not used a versioning system is because it doesn't (hardly ever?) add any breaking changes.

[...]

PR #225, opened 2020-05-28
Updated revision history for CHANGES.md

scmcca:
Updated GTFS revision history for [transit/gtfs/CHANGES.md](https://github.com/MobilityData/transit/blob/master/gtfs/CHANGES.md) as of May 28, 2020. The revision history was last updated January 2019, with some historical entries missing. 

If you have feedback or if I missed anything major, please let me know.

### DIFF
--- a/gtfs/CHANGES.md
+++ b/gtfs/CHANGES.md
@@ -50,9 +50,77 @@ Every new feature adds complexity to the creation and reading of feeds. Therefor
 
 ### Revision History
 
-#### January 17, 2019
+#### May 28, 2020
 
-* Editorial and formatting changes for clarity.  See [pull request](https://github.com/google/transit/pull/120).
+* Updated major GTFS changes under "Revision History" in `CHANGES.md`. Last revised in January 2019 with some historical entries missing.
+
+#### May 13, 2020
+
+* Added `continuous_pickup` and `continuous_drop_off` to `routes.txt` and `stop_times.txt`.  See [discussion](https://github.com/google/transit/pull/208).
+* Changed `shape_id` from "Optional" to "Conditionally required". See [discussion](https://github.com/google/transit/pull/208).
+
+#### March 24, 2020
+
+* Defined text-to-speech field and added `tts_stop_name` to `stops.txt`. See [discussion](https://github.com/google/transit/pull/49).
+
+#### March 20, 2020
+
+* Added recommendation for handling TripUpdate predictions for multiple trips in the same block. See [discussion](https://github.com/google/transit/pull/206).
+
+#### February 5, 2020
+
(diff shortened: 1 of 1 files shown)


gtfs/CHANGES.md
```diff
@@ -50,25 +50,116 @@ Every new feature adds complexity to the creation and reading of feeds. Therefor
...
+
+* Added `pathways.txt` and `levels.txt` to GTFS-static. See [discussion](https://github.com/google/transit/pull/143).
+
+#### February 6, 2019
+
+* Editorial and formatting changes for clarity.  See [discussion](https://github.com/google/transit/pull/120).
+
+#### October 2, 2018
+
+* Factorized field types for concision.  See [discussion](https://github.com/google/transit/pull/104).
```
barbeau, line 114: `concision` -> `conciseness`

gtfs/CHANGES.md
```diff
@@ -50,25 +50,116 @@ Every new feature adds complexity to the creation and reading of feeds. Therefor
...
+* Added `continuous_pickup` and `continuous_drop_off` to `routes.txt` and `stop_times.txt`.  See [discussion](https://github.com/google/transit/pull/208).
+* Changed `shape_id` from "Optional" to "Conditionally required". See [discussion](https://github.com/google/transit/pull/208).
+
+#### March 24, 2020
+
+* Defined text-to-speech field and added `tts_stop_name` to `stops.txt`. See [discussion](https://github.com/google/transit/pull/49).
+
+#### March 20, 2020
+
+* Added recommendation for handling TripUpdate predictions for multiple trips in the same block. See [discussion](https://github.com/google/transit/pull/206).
```
barbeau, line 68: This file has historically been used only for static GTFS revision history. 

However, unless I'm missing something, it looks like the Google website that has the GTFS-RT revision history:
https://developers.google.com/transit/gtfs-realtime/guides/revision-history

...never made it into the GitHub repo.  There is an equivalent CHANGES.md for GTFS-RT:
https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md

But it's missing the revision history.

So we need to decide if we keep all GTFS and GTFS-RT revisions in the same document, or if we split them across two files.

In any case, the Google website revision history should be migrated into this GitHub repo somewhere.

drewda:
@scmcca thanks for helping with this housekeeping

@barbeau I for one think that it would be most useful to track static and RT changes in separate places (still important to have both)

related #215

gtfs/CHANGES.md
scmcca: Good catch. I meant to include only the updated revision history for static GTFS only. 

I noticed as well that there was no GTFS-RT revision history at https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md

I would be for adding it there.

scmcca:
@googlebot I signed it!

PR #234, opened 2020-07-14
fix: Fix a typo in occupancy_percentage description

barbeau:
An editorial change from `The precision of precision` to `The precision of occupancy_percentage`


prhod:
indeed :) 
+1

PR #236, opened 2020-07-22
Update definition of stop_id

aababilov:
Now we have generic nodes and boarding areas. Description of the location_type field is the canonical listing of all supported types.


tleboulenge:
If you want to keep to wording generic at that place and more specific close to the location_type field, I suggest moving the sentence "The term "station entrance" refers to both station entrances and station exits" over there.

The word "location" seems a bit vague (this word is used for moving vehicles in the RT spec). I assume to common denominator of these objects is that they can be the origin or destination of a routing path (and can be walked to or from)?

scmcca:
I'd like to suggest the exhaustive enumeration of all supported types to define what a "location" is. This would leave less ambiguity as to what a `stop_id` can represent, while removing the awkward sentence about what a "station entrance" is. 

Something like:

> Identifies a location: stop/platform, station, entrance/exit, generic node, or boarding area (see `location_type`).
> 
> Multiple routes may use the same `stop_id`.

timMillet (commented):
I agree with @scmcca 's proposition. @aababilov How does it sound to you?

The word "location" may be vague; however, it is used many times (80+) in stops.txt, stop_times.txt, transfers.txt, pathways.txt to refer to a stop/platform, station, entrance/exit, generic node, or boarding area. So if "location" is removed or changed in the definition of `stops.stop_id`, it must be removed or changed anywhere else.

scmcca:
@timMillet IMHO it should be ok. The proposal is to generalize `stop_id` to match the generic use of the word "location" throughout the documentation.

In regards to the definition of "station entrance", I agree with @tleboulenge that, if defined as its own term, it should be defined once in the definition of `location_type=2`. 

Alternatively, any reference to "station entrance" could be replaced with "`location_type=2`" to avoid describing the same thing in separate ways (i.e., limiting the interchange between informal naming and precise enum value... which makes the reader do more work). I think "station entrance" deserves some standardization throughout the documentation along with updating `stop_id` to the suggestion above. Thoughts?

aababilov:
Hi all,

Sorry, I have somehow missed that thread. Sure, the suggestion by @scmcca works for me and I have updated this PR.

@timMillet

scmcca:
@aababilov Great! I just proposed a patch to this PR to also generalize the dataset file definition of stops.txt to make it consistent with updates in `stop_id`. It would read:

> Stops where vehicles pick up or drop off riders. Also defines other locations related to station navigation.

Open to review. See the proposed patch here: https://github.com/aababilov/transit/pull/3

timMillet (approved):

aababilov:
Thanks!

PR #200, opened 2020-02-11
GTFS-Vehicles (core version)

timMillet:
## Context

Currently, the GTFS specification does not provide any vehicle information besides `routes.route_type`, `trips.wheelchair_accessible`, and `trips.bike_allowed`. However, most transit agencies have data on vehicles they operate, which could enhance passenger information provided in GTFS format. 

The following needs could be addressed with vehicle data:
- Suggesting to somebody in a wheelchair where to board the vehicle. Refining accessibility information per carriage or door is the last key information for generating exhaustive step-free trip plans (`pathways.txt` already provides step-free trip plans through stations only).
- Suggesting to anybody where to board the vehicle according to vehicle features or station components at destination. Looking for optimal vehicle boarding is a common behavior of subway and train riders. Optimal vehicle boarding is sometimes displayed on station posters or screens, so GTFS data should represent the same information.
- Informing (more precisely) about vehicle crowdedness. The GTFS real-time specification supports crowdedness, but the field `OccupancyStatus` in `VehiclePosition` is still experimental, and the scale of crowdedness is relatively subjective.
- Informing about vehicle features and amenities. Wheelchair spots, bike racks, AC, wifi, carriage class, dining availability, and other information may be useful in certain circumstances for making travel decisions.

## Pull Request 

This PR is a core version of a broader GTFS-Vehicles extension proposal.
The goal of the current PR is:
- Describing vehicles and their composition (e.g: a subway train composed of 7 carriages) with `vehicle_categories.txt` and `vehicle_couplings.txt`;
- Assigning one or more vehicles to routes or trips, using `routes.txt`, `trips.txt`, and/or `vehicle_allocations.txt`;
- Providing information on optimal vehicle boarding, within `vehicle_boardings.txt`.

This PR suggests the extension or the addition of:
- `routes.txt` (1 new field);
- `trips.txt` (1 new field);
- `vehicle_categories.txt` (new file);
- `vehicle_allocations.txt` (new file);
- `vehicle_couplings.txt` (new file);
- `vehicle_boardings.txt` (new file).

Compared to GTFS-Vehicles full version (still at the discussion stage), this PR _does not_ include, mainly:
- Vehicle features and amenities in `vehicle_categories.txt`;
- Door information in `vehicle_doors.txt` and `vehicle_boardings.txt`;
- Vehicle allocation and orientation in `stop_times.txt`;
- Crowdedness information in message `OccupancyDescriptor`.

## Links
[...]

PR #237, opened 2020-07-27
Proposal: Support multi-car crowding in GTFS-RT

lfontolliet:
### **Background**

GTFS provides a way to surface vehicle crowding data as part of the VehiclePosition message using [`occupancyStatus`](https://developers.google.com/transit/gtfs-realtime/reference#enum-occupancystatus). An experimental field [`occupancy_percentage`](https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L396) is also available.

These fields only provide crowding data at vehicle level. In case a vehicle is composed of several cars (like a subway or train), there is no way to provide details per car. This proposal offers a way to provide multi car crowding data in GTFS-RT.

### Proposal
To handle the case of vehicles that provide crowding data at car level, we would like to propose the addition of a new repeated `multi_car_details` field in the `VehiclePosition` message. This field would be based on a `CarDetails` message containing two fields: `occupancy_status` and `occupancy_percentage`.

`CarDetails.occupancy_status` would follow the values of [`OccupancyStatus`](https://developers.google.com/transit/gtfs-realtime/reference#enum-occupancystatus) enum. `CarDetails.occupancy_percentage` would follow [the same rules](https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L396) than the ones defined for `occupancy_percentage` field (see Special Cases section below for details about unavailable data).

The first occurence of the repeated field would represent the first car of the vehicle, **given the current direction of travel**. The number of occurrences of the `multi_car_details` field represents the number of cars of the vehicle.

In a given occurrence of the repeated `multi_car_details` field, the `CarDetails.occupancy_status` and `CarDetails.occupancy_percentage` fields are not mutually exclusive.

### Special Cases
**Case when data is unavailable for some cars of the vehicle**
The structure also needs to accurately handle cases where the data is not available for all the cars of the vehicle. For `CarDetails.occupancy_status`, we propose to use a new `DATA_NOT_AVAILABLE` value in `OccupancyStatus` enum. For `Car.occupancy_percentage`, as we are using an `int32`, we could use `-1` to determine that data is not available for this specific car.

_Example for `CarDetails.occupancy_status`_: Vehicle is composed of three cars. Data is available only for first and third cars. To handle correctly the "padding", the repetition of `CarDetails.occupancy_status` would be the following: `STANDING_ROOM_ONLY`, `DATA_NOT_AVAILABLE`, `MANY_SEATS_AVAILABLE`.

_Example for `CarDetails.occupancy_percentage`:_ Vehicle is composed of three cars. Data is available only for first and third cars. To handle correctly the "padding", the repetition of `CarDetails.occupancy_percentage` would be the following: `15`, `-1`, `54`.

**Case when data is unavailable for the last cars of the vehicle**
The repeated `multi_car_details` field should contain a value for all the cars of the current vehicle.
_Example_: A vehicle is composed of 4 cars. The last two cars don't have crowding data available, due to faulty sensors.
The repetition of `CarDetails.occupancy_status` would be the following: `STANDING_ROOM_ONLY`, `MANY_SEATS_AVAILABLE`, `DATA_NOT_AVAILABLE`, `DATA_NOT_AVAILABLE`.

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -390,7 +390,7 @@ message VehiclePosition {
     // The vehicle is not accepting additional passengers.
     NOT_ACCEPTING_PASSENGERS = 6;
 
-    // Used in context of multi car crowding. There is no data available for this car in this vehicle
+    // There is no data available for this car in this vehicle. Used in context of multi car crowding. 
     // Useful for padding, in case a vehicle composed of several cars has data available only for some of those
     NO_DATA_AVAILABLE = 7;
 
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -500,11 +500,9 @@ message Alert {
   optional TranslatedString description_text = 11;
 
   // Text for alert header to be used in text-to-speech implementations. This field is the text-to-speech version of header_text.
-  // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional TranslatedString tts_header_text = 12;
 
   // Text for full description for the alert to be used in text-to-speech implementations. This field is the text-to-speech version of description_text.
-  // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional TranslatedString tts_description_text = 13;
 
   // Severity of this alert.
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -390,6 +390,10 @@ message VehiclePosition {
     // The vehicle is not accepting additional passengers.
     NOT_ACCEPTING_PASSENGERS = 6;
 
+    // Used in context of multi car crowding. There is no data available for this car in this vehicle
+    // Useful for padding, in case a vehicle composed of several cars has data available only for some of those
+    NO_DATA_AVAILABLE = 7;
+
   }
   optional OccupancyStatus occupancy_status = 9;
 
@@ -401,6 +405,32 @@ message VehiclePosition {
   // The precision of occupancy_percentage should be low enough that you can't track a single person boarding and alighting for privacy reasons.
   // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional uint32 occupancy_percentage = 10;
+
+  // Car specific details, used for vehicles composed of several cars
+  message CarDetails {
+
+    // Occupancy status for this given car, in this vehicle
+    optional OccupancyStatus occupancy_status = 1;
+
+    // Occupancy percentage for this given car, in this vehicle
+    // Follows the same rules as "VehiclePosition.occupancy_percentage"
+    // In addition, value -1 will be used to determine that crowding percentage is currently not available for this given car
+    // -1 is useful for padding, in case a vehicle composed of several cars has data available only for some of those
+    optional int32 occupancy_percentage = 2;
+
+    // The extensions namespace allows 3rd-party developers to extend the
+    // GTFS Realtime Specification in order to add and evaluate new features and
+    // modifications to the spec.
+    extensions 1000 to 1999;
+
+    // The following extension IDs are reserved for private use by any organization.
+    extensions 9000 to 9999;
+  }
+
+  // Details of the multiple cars of this given vehicle.
+  // The first occurence represents the first car of the vehicle, given the current direction of travel. 
+  // The number of occurrences of the multi_car_details field represents the number of cars of the vehicle.
+  repeated CarDetails multi_car_details = 11;
 
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime Specification in order to add and evaluate new features and
@@ -470,9 +500,11 @@ message Alert {
   optional TranslatedString description_text = 11;
 
   // Text for alert header to be used in text-to-speech implementations. This field is the text-to-speech version of header_text.
+  // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional TranslatedString tts_header_text = 12;
 
   // Text for full description for the alert to be used in text-to-speech implementations. This field is the text-to-speech version of description_text.
+  // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional TranslatedString tts_description_text = 13;
 
   // Severity of this alert.


skinkie:
Shall we start by changing Car to Coach?

gcamp:
@lfontolliet thanks for the proposition. 

A few questions : 
- Are you aware of GTFS-vehicles extension? https://github.com/google/transit/pull/200 Do you see any consequences or things that you would be different if that proposition would be accepted? 
- Do you have a producer and a consumer using this extension? If not, you'll most likely want to mark the new fields as experimental. See the change process : https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#adding-new-fields-to-gtfs-realtime

wafisher:
We (MTA LIRR) could produce using this as soon as it's ratified. Not sure about the consumer, though.

lfontolliet:
Thanks for the quick comments.

> Shall we start by changing Car to Coach?
@skinkie Yes that's a possibility. Could you describe the advantages you see of using Coach instead of Car?

> * Are you aware of GTFS-vehicles extension? #200 Do you see any consequences or things that you would be different if that proposition would be accepted?
@gcamp I wasn't aware of it, thanks for pointing out. From what I understand, GTFS-vehicles proposal is more related to static vehicle data and would enable to surface vehicle amenities. The `CarDetails` message in this current proposal is useful for data which is specific to Real-Time, like the crowding details (as it varies during a given trip), and wouldn't be meant to convey data such as amenities. We will review the GTFS-vehicles extension in more details today and provide additional details if we conclude that it would change anything from our proposal.

> * Do you have a producer and a consumer using this extension? If not, you'll most likely want to mark the new fields as experimental. See the change process : https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#adding-new-fields-to-gtfs-realtime
@gcamp We (Lyft) do not have a producer yet, we can indeed follow the experimental path.

> We (MTA LIRR) could produce using this as soon as it's ratified. Not sure about the consumer, though.
@wafisher We'd be happy to collaborate with you to discuss usage of this field!

skinkie:
> > Shall we start by changing Car to Coach?
> 
> @skinkie Yes that's a possibility. Could you describe the advantages you see of using Coach instead of Car?

Not inventing terms, keeping it in line with common terms from the UIC.
https://en.wikipedia.org/wiki/UIC_passenger_coach_types

wafisher:
We're currently using [this extension](https://github.com/OneBusAway/onebusaway-gtfs-realtime-api/blob/master/src/main/resources/com/google/transit/realtime/gtfs-realtime-MTARR.proto) to produce our GTFS-RT, hosted [here](http://staging-backend.mylirr.org/gtfs).

Happy to work together to come up with a more official standard that we can use.

lfontolliet:
Hey everyone,

We have been working with LIRR and MobilityData to update the original proposal. LIRR and MobilityData are aligned with the changes being proposed here. I will shortly change the original description of the PR to reflect the latest changes.

Here are the main takeaways of these new changes:
- We would use the "carriage" terminology instead of "cars".
- The `carriage_status` repeated field now contains two additional optional fields: `label` and `carriage_sequence`.
- A new value `NOT_BOARDABLE` has been introduced in `OccupancyStatus` to reflect vehicles or carriages that are not boardable (like an Engine, or a maintenance car for instance). 
- `occupancy_percentage` and `occupancy_status` provide default values to [avoid values automatically added by protobuf](https://developers.google.com/protocol-buffers/docs/proto3#default) that do not reflect the ground truth.

LIRR would be the producer of this proposal and Lyft would be the consumer.
Thanks to @barbeau and @wafisher for contributing to the changes.

davidlewis-ito:
@lfontolliet this looks like a really valuable addition.  I have two minor questions :

i) carriage_sequence:  The description suggests that there is no restriction on the values used - ie entries of 4,10 and 105 are valid.  I would suggest that using instead 1,2 & 3 provides for a degree of validation (ie "1,2,4,5" is missing carriage 3.).   

ii) Could you clarify the intent of the OccupancyStatus  of NOT_ACCEPTING_PASSENGERS .  Would this be a carriage that is out of service ?

wafisher:
I can answer the second: an example of `NOT_ACCEPTING_PASSENGERS` is when certain cars which have seats are locked out i.e. deadheading. Crews on our system (and others) can do that when the load is very light and they want to consolidate passengers into fewer cars. In this case, they will lock out a couple cars so pax can't board those cars.

One reason they're allowed to this is because we have plenty of short platforms in our system. If we run 8-car trains mid-day along certain routes, they'll come along some 6-car platforms. If the crew can lock out those cars, they can guarantee that nobody is in the 2 cars that won't platform. Obviously, though, when it's busier (or at all during COVID), they'll want to keep the entire consist open and make sure pax are in the right car to alight.

barbeau:
@lfontolliet Thanks for pulling this together!

I believe there are two more points to address to complete the proposal:

1. *What fields are required/optional* - due to [protocol buffer quirks](https://developers.google.com/protocol-buffers/docs/proto), new experimental fields in `gtfs-realtime.proto` will always be labeled as `optional`. We need to define field requirements separately in `reference.md` that are semantic GTFS-realtime spec. requirements. I would expect `carriage_sequence` to be required, and the rest of the fields to be optional. Any additional thoughts on this?
1. *Update `VehiclePosition.OccupancyStatus` and `VehiclePosition.occupancy_percentage` definitions* - these fields both need to be updated to reflect how they interact with per-carriage occupancy info.  I propose this language addition to both fields, respectively:

>If `multi_carriage_status` is populated with per-carriage `occupancy_percentage`/`OccupancyStatus`, then this field should describe the entire vehicle with all carriages accepting passengers considered.

lfontolliet:
> i) carriage_sequence: The description suggests that there is no restriction on the values used - ie entries of 4,10 and 105 are valid. I would suggest that using instead 1,2 & 3 provides for a degree of validation (ie "1,2,4,5" is missing carriage 3.).
@davidlewis-ito Yes that's a good point, I have modified the description of the field to start carriages at `1`, with an increment of one for each carriage.

@barbeau I have modified the `reference.md` file and added your comments. Let me know how that looks.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +407,56 @@ message VehiclePosition {
   // including both seated and standing capacity, and current operating regulations allow.
   // It is possible that the value goes over 100 if there are currently more passengers than what the vehicle was designed for.
   // The precision of occupancy_percentage should be low enough that you can't track a single person boarding and alighting for privacy reasons.
+  // If multi_carriage_status is populated with per-carriage occupancy_percentage, 
+  // then this field should describe the entire vehicle with all carriages accepting passengers considered.
   // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional uint32 occupancy_percentage = 10;
 
+  // Carriage specific details, used for vehicles composed of several carriages
```
barbeau, line 415: The `CarriageStatus` message and all fields within (except extensions) need the following experimental label, accordingly:

```
// This message/field is still experimental, and subject to change. It may be formally adopted in the future.
```

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +407,56 @@ message VehiclePosition {
...
+  }
+
+  // Details of the multiple carriages of this given vehicle.
+  // The first occurrence represents the first carriage of the vehicle, 
+  // given the current direction of travel. 
+  // The number of occurrences of the multi_carriage_details 
+  // field represents the number of carriages of the vehicle.
+  // It also includes non boardable carriages, 
+  // like engines, maintenance carriages, etc… as they provide valuable 
+  // information to passengers about where to stand on a platform.
```
barbeau, line 457: Same - needs experimental label:

```
// This field is still experimental, and subject to change. It may be formally adopted in the future.
```

gtfs-realtime/spec/en/reference.md
```diff
@@ -251,12 +253,28 @@ The degree of passenger occupancy for the vehicle.
...
 | _**FULL**_ | _The vehicle is considered full by most measures, but may still be allowing passengers to board._ |
-| _**NOT_ACCEPTING_PASSENGERS**_ | _The vehicle can not accept passengers._ |
+| _**NOT_ACCEPTING_PASSENGERS**_ | _The vehicle or carriage is not accepting passengers. The vehicle or carriage usually accepts passengers for boarding._ |
+| _**NO_DATA_AVAILABLE**_ | _The vehicle or carriage doesn't have any occupancy data available at that time._ |
+| _**NOT_BOARDABLE**_ | _The vehicle or carriage is not boardable and never accepts passengers. Useful for special vehicles or carriages (engine, maintenance carriage, etc…)._ |
+
+
+## _message_ CarriageStatus
+
+Carriage specific details, used for vehicles composed of several carriages
```
barbeau, line 268: Same here for CarriageStatus, label, and carriage_sequence - all need experimental label (occupancy_status and occupancy_percentage do have it here).

gtfs-realtime/spec/en/reference.md
```diff
@@ -251,12 +253,28 @@ The degree of passenger occupancy for the vehicle.
...
+
+#### Fields
+
+| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
+|------------------|------------|----------------|-------------------|-------------------|
+| **label** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | User visible label that may be shown to the passenger to help identify the carriage. Example: "7712", "Car ABC-32", etc... |
+| **occupancy_status** | [OccupancyStatus](#enum-occupancystatus) | _Optional_ | One | Occupancy status for this given carriage, in this vehicle. Default is set to `NO_DATA_AVAILABLE`.<br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future.|
```
barbeau, line 275: `Optional` is italicized here but not for the other fields in the table:

```suggestion
| **occupancy_status** | [OccupancyStatus](#enum-occupancystatus) | Optional | One | Occupancy status for this given carriage, in this vehicle. Default is set to `NO_DATA_AVAILABLE`.<br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future.|
```

gtfs-realtime/spec/en/reference.md
```diff
@@ -251,12 +253,28 @@ The degree of passenger occupancy for the vehicle.
...
+| **occupancy_percentage** | [int32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | Occupancy percentage for this given carriage, in this vehicle. Follows the same rules as "VehiclePosition.occupancy_percentage". Use -1 in case data is not available for this given carriage.<br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future. |
+| **carriage_sequence** | [uint32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | Identifies the order of this carriage with respect to the other carriages in the vehicle's list of CarriageStatus. The first carriage in the direction of travel must have a value of 1. The second value corresponds to the second carriage in the direction of travel and must have a value of 2, and so forth. For example, the first carriage in the direction of travel has a value of 1. If the second carriage in the direction of travel has a value of 3, consumers will detect an issue with the sequencing. Even if protobuf keeps sequencing, it will help with feed validation |
```
barbeau, line 277: In the interest of keeping the spec concise, I'd suggest eliminating the last sentence of `carriage_sequence`:

```suggestion
| **carriage_sequence** | [uint32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | Identifies the order of this carriage with respect to the other carriages in the vehicle's list of CarriageStatus. The first carriage in the direction of travel must have a value of 1. The second value corresponds to the second carriage in the direction of travel and must have a value of 2, and so forth. For example, the first carriage in the direction of travel has a value of 1. If the second carriage in the direction of travel has a value of 3, consumers will detect an issue with the sequencing. |
```

barbeau:
@lfontolliet Thanks, looks good! I left a few suggestions for formatting, and the experimental message/field labels are needed on everything new. Other than that, though, 👍.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +407,56 @@ message VehiclePosition {
...
+    // Identifies the order of this carriage with respect to the other
+    // carriages in the vehicle's list of CarriageStatus.
+    // The first carriage in the direction of travel must have a value of 1.
+    // The second value corresponds to the second carriage in the direction
+    // of travel and must have a value of 2, and so forth.
+    // For example, the first carriage in the direction of travel has a value of 1.
+    // If the second carriage in the direction of travel has a value of 3,
+    // consumers will detect an issue with the sequencing.
+    // Even if protobuf keeps sequencing, it will help with feed validation
+    optional uint32 carriage_sequence = 4;
```
skinkie, line 439: How is switch-back / zig-zag / leversal / changing ends / turnround / turning / to horse over envisioned within a single trip?

gtfs-realtime/proto/gtfs-realtime.proto
lfontolliet: Added.

gtfs-realtime/proto/gtfs-realtime.proto
lfontolliet: Added.

lfontolliet:
@barbeau I added the experimental labels. Can you add a comment with `@googlebot I consent.` as it now contains one of your commits. Thanks.

lfontolliet:
@googlebot I consent.

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: @skinkie My interpretation is that the sequence and label combinations should match whatever is shown to riders (i.e., the ground truth) for the given carriage configuration at the `VehiclePosition.timestamp`. So between two `VehiclePosition` updates, the sequence and label combinations could change, which indicates a change in configuration.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +408,58 @@ message VehiclePosition {
...
   // It is possible that the value goes over 100 if there are currently more passengers than what the vehicle was designed for.
   // The precision of occupancy_percentage should be low enough that you can't track a single person boarding and alighting for privacy reasons.
+  // If multi_carriage_status is populated with per-carriage occupancy_percentage, 
+  // then this field should describe the entire vehicle with all carriages accepting passengers considered.
   // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional uint32 occupancy_percentage = 10;
 
+  // Carriage specific details, used for vehicles composed of several carriages
+  message CarriageStatus {
+
```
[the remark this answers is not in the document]
barbeau: @jxeeno I'm certainly not against that if you and others feel like it's necessary, although the combination of vehicle ID + carriage sequence should uniquely identify a carriage.

barbeau:
@googlebot I consent

gtfs-realtime/proto/gtfs-realtime.proto
skinkie: @barbeau this is not sustainable. This means that ever time the train changes direction it either is enforced to become a new trip (which is not enforced in GTFS-static) or arbitrary need to reverse all the values in RT.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +408,58 @@ message VehiclePosition {
...
+    // the carriage. Example: "7712", "Car ABC-32", etc...
+    optional string label = 1;
+
+    // Occupancy status for this given carriage, in this vehicle
+    optional OccupancyStatus occupancy_status = 2 [default = NO_DATA_AVAILABLE];
+
+    // Occupancy percentage for this given carriage, in this vehicle
+    // Follows the same rules as "VehiclePosition.occupancy_percentage"
+    // -1 in case data is not available (as protobuf defaults to 0 otherwise)
+    optional int32 occupancy_percentage = 3 [default = -1];
```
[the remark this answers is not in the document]
barbeau: Unfortunately no - https://developers.google.com/protocol-buffers/docs/proto#optional says:

>When a message is parsed, if it does not contain an optional element, the corresponding field in the parsed object is set to the default value for that field. The default value can be specified as part of the message description...If the default value is not specified for an optional element, a type-specific default value is used instead: ...For numeric types, the default value is zero. For enums, the default value is the first value listed in the enum's type definition.

That means right now protobuf-generated client code is therefore interpreting `VehiclePosition.occupancy_percentage` as `0` if the producer hasn't populated the field, and `VehiclePosition.occupancy_percentage` is being interpreted as `EMPTY` if unpopulated.

To fix this `VehiclePosition.occupancy_percentage` needs to change from `uint32` to `int32` (which [should be](https://developers.google.com/protocol-buffers/docs/proto#updating) forwards/backwards compatible), with a default of -1 defined, and `OccupancyStatus` needs the additional `NO_DATA_AVAILABLE`, with `VehiclePosition.occupancy_percentage` using that as a default as well. Not defining these in the initial proposals for these fields was an oversight.

I plan to open a separate PR for this.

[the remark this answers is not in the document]
skinkie: @barbeau I sincerely hope uint32->int32 is also ABI compatible with end users.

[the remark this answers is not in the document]
barbeau: @skinkie That was my concern too. The link I referenced above, https://developers.google.com/protocol-buffers/docs/proto#updating, says:

>int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. If a number is parsed from the wire which doesn't fit in the corresponding type, you will get the same effect as if you had cast the number to that type in C++ (e.g. if a 64-bit number is read as an int32, it will be truncated to 32 bits).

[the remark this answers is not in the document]
skinkie: @barbeau it is wire compatible. I suppose this works for platforms with types like Python and Java. But for a C application this will be interesting. Now the occupancy percentage is something new, so I doubt there are many implementations.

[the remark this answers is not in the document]
barbeau: So thinking about this more, the main question would be what happens when a `-1` value that was serialized by a new `int32` producer is read by a legacy consumer still using the `uint32` in the .proto? I'll do some testing to see how this works.

gtfs-realtime/proto/gtfs-realtime.proto
jxeeno: Keep in mind a `label` is something that's customer facing.  If carriages have digital outward facing signage, the label may change at the time of train divide / amalgamate / switch-backs.

Could this be modelled by the `id` field?  You'll still need to reverse the carriage sequence at the time of the switch-back, but the consumer will be able to track flip through an idempotent reference to the physical carriage.  Consumers can also track carriages for divide and amalgamate operations.

It doesn't solve conveying these operations ahead of time, but at least it can be modelled and you'll know after it happens.

gtfs-realtime/spec/en/reference.md
```diff
@@ -251,12 +253,28 @@ The degree of passenger occupancy for the vehicle.
...
-| _**NOT_ACCEPTING_PASSENGERS**_ | _The vehicle can not accept passengers._ |
+| _**NOT_ACCEPTING_PASSENGERS**_ | _The vehicle or carriage is not accepting passengers. The vehicle or carriage usually accepts passengers for boarding._ |
+| _**NO_DATA_AVAILABLE**_ | _The vehicle or carriage doesn't have any occupancy data available at that time._ |
+| _**NOT_BOARDABLE**_ | _The vehicle or carriage is not boardable and never accepts passengers. Useful for special vehicles or carriages (engine, maintenance carriage, etc…)._ |
+
+
+## _message_ CarriageStatus
+
+Carriage specific details, used for vehicles composed of several carriages
+
```
barbeau, line 269:
```suggestion
Carriage specific details, used for vehicles composed of several carriages

**Caution:** this message is still **experimental**, and subject to change. It may be formally adopted in the future.

```

gtfs-realtime/spec/en/reference.md
```diff
@@ -251,12 +253,28 @@ The degree of passenger occupancy for the vehicle.
...
+#### Fields
+
+| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
+|------------------|------------|----------------|-------------------|-------------------|
+| **label** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | User visible label that may be shown to the passenger to help identify the carriage. Example: "7712", "Car ABC-32", etc... |
```
barbeau, line 274:
```suggestion
| **label** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | User visible label that may be shown to the passenger to help identify the carriage. Example: "7712", "Car ABC-32", etc... <br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future. |
```

gtfs-realtime/spec/en/reference.md
```diff
@@ -251,12 +253,28 @@ The degree of passenger occupancy for the vehicle.
...
+| **occupancy_percentage** | [int32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Optional | One | Occupancy percentage for this given carriage, in this vehicle. Follows the same rules as "VehiclePosition.occupancy_percentage". Use -1 in case data is not available for this given carriage.<br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future. |
+| **carriage_sequence** | [uint32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | Identifies the order of this carriage with respect to the other carriages in the vehicle's list of CarriageStatus. The first carriage in the direction of travel must have a value of 1. The second value corresponds to the second carriage in the direction of travel and must have a value of 2, and so forth. For example, the first carriage in the direction of travel has a value of 1. If the second carriage in the direction of travel has a value of 3, consumers will detect an issue with the sequencing. |
```
barbeau, line 277:
```suggestion
| **carriage_sequence** | [uint32](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | Identifies the order of this carriage with respect to the other carriages in the vehicle's list of CarriageStatus. The first carriage in the direction of travel must have a value of 1. The second value corresponds to the second carriage in the direction of travel and must have a value of 2, and so forth. For example, the first carriage in the direction of travel has a value of 1. If the second carriage in the direction of travel has a value of 3, consumers will detect an issue with the sequencing. <br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future. |
```

gtfs-realtime/spec/en/reference.md
barbeau: I'd added these in suggestions below.

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: @jxeeno Yes, if `id` persists across sequence changes, you should be able to detect these types of changes in real-time. So adding the `id` field seems like a reasonable approach to me.

gtfs-realtime/proto/gtfs-realtime.proto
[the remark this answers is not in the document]
barbeau: Per discussion below, I agree, let's add `id` so carriage sequence configuration changes can be detected.

barbeau:
@lfontolliet Is it possible to resolve the remaining changes discussed above soon so we can call a vote for adding as an experimental field early next week?

lfontolliet:
> @lfontolliet Is it possible to resolve the remaining changes discussed above soon so we can call a vote for adding as an experimental field early next week?
@barbeau Yes, reviewing the current comments/proposals right now and will reply before EOD.

gtfs-realtime/proto/gtfs-realtime.proto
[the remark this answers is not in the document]
lfontolliet: Added ID as part of latest commit.

gtfs-realtime/spec/en/reference.md
lfontolliet: Thanks, added.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +408,58 @@ message VehiclePosition {
...
+
+  // Details of the multiple carriages of this given vehicle.
+  // The first occurrence represents the first carriage of the vehicle, 
+  // given the current direction of travel. 
+  // The number of occurrences of the multi_carriage_details 
+  // field represents the number of carriages of the vehicle.
+  // It also includes non boardable carriages, 
+  // like engines, maintenance carriages, etc… as they provide valuable 
+  // information to passengers about where to stand on a platform
+  repeated CarriageStatus multi_carriage_status = 11;
```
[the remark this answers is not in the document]
lfontolliet: Good catch. I have renamed the field as multi_carriage_details. I'd like to make sure the field name states that it should be used only in context of multiple carriages. So we prevent the case where a mono carriage vehicle relies on this field, instead of the vehicle one.

gtfs-realtime/proto/gtfs-realtime.proto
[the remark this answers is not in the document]
lfontolliet: Added ID in my new changes.

lfontolliet:
Summary of latest changes:
- Renamed the protobuf message into `CarriageDetails`. Renamed the field into `multi_carriage_details`.
- Added an optional `id` field in the `CarriageDetails` message.

gtfs-realtime/proto/gtfs-realtime.proto
skinkie: @barbeau @jxeeno I would even suggest to use the coach registration number here.

lfontolliet:
Hello all,

Vote will be closed on Monday September 7th 2020 at 23:59:59 UTC.

Thanks,
Loïc

wafisher:
+1 from the LIRR

skinkie:
-1 I still see to many issues regarding normal operations such as reversing direction or splitting which have not been addressed. I think this producer (alone) is not a suitable candidate. I would suggest to ask Deutsche Bahn for a review here.

jxeeno:
> -1 I still see to many issues regarding normal operations such as reversing direction or splitting which have not been addressed. I think this producer (alone) is not a suitable candidate. I would suggest to ask Deutsche Bahn for a review here.

I agree it doesn't fully handle all cases of changes to carriages, but I don't think it should have to.  This is the vehicle position feed, so should be describing the current state of the vehicle.  Current vehicle state shouldn't care about a future vehicle state as a result of an operation like divide, amalgamate or reversals.

imo, what @Dave-TfNSW suggested in https://github.com/google/transit/pull/240#issuecomment-682331951 + this PR could be a way for operators to convey away divide, amalgamate and reversals by providing departure and arrival occupancy at a per stop level.

jxeeno:
+1 from AnyTrip (consumer)

skinkie:
> I agree it doesn't fully handle all cases of changes to carriages, but I don't think it should have to. This is the vehicle position feed, so should be describing the current state of the vehicle. Current vehicle state shouldn't care about a future vehicle state as a result of an operation like divide, amalgamate or reversals.

I respectfully disagree. Since this will eventually be used to track the occupancy of trains, and conveniently people are going to use the sequence which at some point does not make sense anymore. I recall that CityMapper used sequence to suggest where to board for the best place to transfer later. Also for this behavior you need proper scenarios to follow up historic state. The main problem with trains is that it is not a vehicle but a composition of formations. Pushing it back to some trivial examples that work and standardise that is not in our best interest, hence a proper review from DB, SNCF aka any serious European Rail Agency would probably benefit us all, to prevent reinventing the wheel. (Yes, this is what happens a lot lately here...)

stotala:
+1 from Google

davidlewis-ito:
+1 from Ito World

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -399,9 +407,64 @@ message VehiclePosition {
...
+    optional int32 occupancy_percentage = 4 [default = -1];
+
+    // Identifies the order of this carriage with respect to the other
+    // carriages in the vehicle's list of CarriageDetails.
+    // The first carriage in the direction of travel must have a value of 1.
+    // The second value corresponds to the second carriage in the direction
+    // of travel and must have a value of 2, and so forth.
+    // For example, the first carriage in the direction of travel has a value of 1.
+    // If the second carriage in the direction of travel has a value of 3,
+    // consumers will detect an issue with the sequencing.
```
juanborre, line 444: This behaviour could be more clear.

Does this mean that if a sequence is missing the consumer must drop the CarriageDetails, the vehiclePosition or the whole GTFS-rt?
Can it be that if a sequence is missing we should assume NO_DATA_AVAILABLE for that carriage?

barbeau: Good catch. I think an invalid sequence should result in dropping the multi-carriage data as it indicates corruption at some level. So for example:

>    // If the second carriage in the direction of travel has a value of 3,
    // consumers will discard data for all carriages (i.e., the multi_carriage_details field).

juanborre: That makes sense since we are enforcing the inclusion of all carriages, so it must mean error somewhere 👍

barbeau: And, it's probably worth adding:

>// Carriages without data must be represented with a valid carriage_sequence number and the fields without data should be omitted (alternately, those fields could also be included and set to the "no data" values).

lfontolliet:
The vote is now closed. Here is the tally.

**Yes (+1)**
- LIRR (@wafisher)
- AnyTrip (@jxeeno)
- Google (@stotala)
- Ito World (@davidlewis-ito)
- Transit (@juanborre)

**No (-1)**
- Bliksem Labs B.V. (@skinkie)

_Result: 83.3% Yes_

We need at least 80% yes for experimental GTFS-RT features [per the official process](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#experimental-fields), **so this proposal passes**.

Note: I will modify the comment describing the `carriage_sequence` field as suggested by @juanborre and @barbeau [in this GitHub comment](https://github.com/google/transit/pull/237#discussion_r481149890), unless someone wants to change their vote if this change in the comment is made.

gtfs-realtime/proto/gtfs-realtime.proto
harshverma2000: We need to consider and handle the case -
For car ordering or car position, it will be done regardless of the direction of travel -
For many trains and train operators, generally, the train Consists always remains the same, and stay in the same order, with the car adjacent to the locomotive always being termed as Car # 1 and the next car as # 2, then Car # 3 and so on. 
So, the reporting of Car # 1 may need to remain the same, particularly with the AMTRAK Capitol Corridor CCJPA. In some other transit systems, it may not be the same.
Even though the train may be going in either direction, the locomotive is many times always on one side and would be either pulling or pushing (behind the cars) the train coaches.
Car 1 is always adjacent to the locomotive, regardless of the direction of travel.

harshverma2000: > We need to consider and handle the case -
> For car ordering or car position, it will be done regardless of the direction of travel -
> For many trains and train operators, generally, the train Consists always remains the same, and stay in the same order, with the car adjacent to the locomotive always being termed as Car # 1 and the next car as # 2, then Car # 3 and so on.
> So, the reporting of Car # 1 may need to remain the same, particularly with the AMTRAK Capitol Corridor CCJPA. In some other transit systems, it may not be the same.
> Even though the train may be going in either direction, the locomotive is many times always on one side and would be either pulling or pushing (behind the cars) the train coaches.
> Car 1 is always adjacent to the locomotive, regardless of the direction of travel.

For multi-car occupancy, “order of Car numbering may not be flipped every time the train direction of travel is changed”.
Carriage Sequence is always going to be 1, 2, 3, 4, 5
The first occurrence of the repeated field will represent the first car or vehicle, given the direction of travel is not always true generally with rail operations and passenger rail industry, specially in the case of AMTRAK Capitol Corridor and this needs to be accounted for, to have an operator view of the same.
There needs to be a special case for this consideration.

barbeau: @harshverma2000 I think you're referring to how the operator numbers carriages internally? The internal numbering doesn't need to match the "external" number following the rules defined here in GTFS-realtime - the operator would just need to keep a mapping between the two internally to ensure the right numbers are published to GTFS-realtime. There are certainly cases that the current GTFS-realtime model doesn't address, but IMHO it seems adequate for the purpose of showing occupancy per-carriage to the end user. @lfontolliet and @wafisher (and others) can certainly weigh in if they see any obstacles so far in implementation.

lfontolliet: Agreed with @barbeau. The operator would need to know the sequence of carriage for a given segment of a trip, but keeping a mapping between the internal identifier and their external facing sequence seems to be easier to handle in the operator's systems, compared to surfacing a complex logic in the standard.

harshverma2000: Thanks @barbeau and @lfontolliet for your thoughts and comments.  
The idea of keeping a mapping between the internal identifier and their external facing sequence to handle the operator's systems is excellent.
I believe operators could do that.  What could be important is for the purpose of understanding occupancy-by-carriage position by various end users - they would need to know the direction-of-travel in order to understand the mapping and this point has to be documented for operator-enduser understanding. Direction-of-travel field could help end users realize this mapping...

harshverma2000:
So, I was away for a while and just checking it out.  The CarriageDetails field could include a DirectionOfTravel and this would address the requirements for operators like AMTRAK/ Capitol Corridor and some other transit agencies as well!

Issue #223, opened 2020-05-06
OccupancyStatus vs. occupancy_percentage


barbeau:
In GTFS-RT, we now have two experimental fields that can contain real-time vehicle occupancy information:
* [OccupancyStatus](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-occupancystatus) - A enum of 7 values added to the spec 6 years ago. We know of 3 producers of this data - see [this pull request](https://github.com/google/transit/pull/212) for details.
* [occupancy_percentage](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-vehicleposition) - A percentage value representing the degree of passenger occupancy of the vehicle. Just adopted as an experimental field ~30 days ago in [this pull request](https://github.com/google/transit/pull/213)

Unless there is a specific use case to have both fields in the spec, we should try to narrow down to one.

Producers and consumers - if you were to pick **one** of the two, which would you pick and why?

For producers - is one field easier to produce than the other? Would adoption be faster with one field in particular?

skinkie:
If it would be one or the other; To be compatible with SIRI I would pick the enumeration. In in addition, it is the least commercial sensitive information to be shared.

darylweinberg:
At Capital Metro we are working to produce OccupanyStatus by August. It will not have a percentage value at this time.

stevenmwhite:
`occupancy_percentage` would be the pick.
- Lends itself better to situational adjustments (for example during the COVIC crisis, 100% could represent a policy-related limit rather than a physical limit). It makes sense as "per agency regulations (whatever they may be at a given time), the vehicle is X percent of its limit. The scale of 100% can change as needed.
- It is less open to interpretation (what is many, what is few?) by different agencies and different consumers.
- It does not claim precession that may not be accurate (when status says a bus is standing room only, it may not necessarily be... depending on whether passengers have chosen to take all available seats before standing. It's not uncommon to be on a crowded transit vehicle with many people standing and still have some open seats.)
- It is easier to produce and I believe adoption would be faster as a result.
- Gives consumers more flexibility in how they display it, rather than attempting to map the enums to approximate percentage levels.

What it doesn't do is give any sort of accessibility information. While `occupancyStatus` can be unclear, the mention of "seats" can be helpful to riders if it is accurate, and a pure percentage doesn't give that.

mike-swiftly:
[...]

PR #240, opened 2020-08-10
GTFS-Occupancies

scmcca:
### Proposal
Given the recent momentum behind describing vehicle crowdedness in GTFS, we ([MobilityData IO](https://mobilitydata.org/)) are proposing GTFS-Occupancies (fka GTFS-OccupancyPredictions, fka GTFS-MeanOccuancies) as a static method for describing usual or predicted occupancies at the `trip` or `stop_times` level. 

As occupancies can currently be described in GTFS Realtime using [`occupancy_status`](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-occupancystatus) and [`occupancy_percentage`](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-vehicleposition), GTFS-Occupancies aims to complement the availability of crowdedness information by providing static predictions for future trips based on historical trends, which can help riders plan trips based on their crowdedness preferences and comfortability.

The most up-to-date proposal is contained in this PR at the "Files changed" tab. Don't hesitate to voice your needs or concerns that can improve this proposal.

Thanks!

### Edits
- 2020-12-17: Newest thinking [here](https://github.com/google/transit/pull/240#issuecomment-740756958). The PR has been changed to Option D for modeling dates.
- 2020-11-20: Option A has been abandoned for [Option B or C](https://docs.google.com/document/d/1uGpx39_ZrF9lZ8z3d-NOb4kN5_2rCs-Z4TQ3uorusnI/edit#heading=h.u9g6vw1vuv7e).


skinkie:
This is again so over engineered, that I wonder which producer stated that they could actually make it in this way? What was wrong by keeping this simple? We are talking about a prediction. Something simple like trip_id,date,stop_sequence,stop_id,occupancy_status.

stevenmwhite:
Maybe my concerns are unfounded but I'm a little worried about this _as a concept._

So far everything we have in GTFS is either **planned** or an actual **real-time status**. This adds another category of historical-based averages.

Of course, historical performance does not equal future performance, particularly as conditions change in the real world (whether that be possible pandemics, changing seasons, special events, cyclical school calendars). Is there any guidance to how far back the mean should be calculated? Should it be updated using data from the past week, the past month, all time, the same trip last year, etc? At what point is an average of many instances not representative of _any particular future one_ real-world instance?

I'm mostly concerned that if we get into standardizing a format to allow predictions of something based on past performance but not real-time information and don't standardize the calculation of that data, it will all be unreliable because quality and meaning will vary from one producer to another.

skinkie:
> Is there any guidance to how far back the mean should be calculated? Should it be updated using data from the past week, the past month, all time, the same trip last year, etc? At what point is an average of many instances not representative of any particular future one real-world instance?

From September 2020 onward we have predictions for at least the next 3-7 days available as predicted data for The Netherlands. So a journey planner might steer commuters to a different trip under COVID-19 flatting the curve concerns. I think 3 days is the least usable, and we are likely to use this in our GTFS-RT integration.

So the quality **will** be different from operator to operator. And I do hope they will take into account their tap-in-tap-out stuff, including a calendar of nearby events.

scmcca:
Hi @skinkie 
> What was wrong by keeping this simple? We are talking about a prediction. Something simple like trip_id,date,stop_sequence,stop_id,occupancy_status.

A simple approach like the one you mentioned is possible with the current PR. The main difference being a requirement to provide a `mean_occupancy_percentage` value alongside the `mean_occupancy_status` enum. This was proposed as per the lengthy discussion in #223 around disambiguating percentages and enums when only one or the other is provided. However, we welcome collaboration in changing the proposal based on community feedback!

> This is again so over engineered, that I wonder which producer stated that they could actually make it in this way?

I imagine you are referencing the per-day and per-coach features of this proposal, which I admittedly presume are demanding features in terms of primary data input for producing the spec. The feasibility of gathering such data is a question that MobilityData would love some insight on from producers/agencies! I believe Cal-ITP (@e-lo) are undertaking a related [market sounding](https://dot.ca.gov/-/media/dot-media/documents/cal-itp/cal-itp-market-sounding-real-time-transit-vehicle-occupancy-070720-a11y.pdf) that could be useful here.

Similarly, the point raised by @stevenmwhite is an important one worth discussing with producers:

> I'm mostly concerned that if we get into standardizing a format to allow predictions of something based on past performance but not real-time information and don't standardize the calculation of that data, it will all be unreliable because quality and meaning will vary from one producer to another.

It would be good to know what producer's historical trend data requirements and calculations look like (as @skinkie has done 👍), and the limits/guidelines the community could establish around producing this spec.

timMillet:
@googlebot I consent.

stevenmwhite:
LA Metro and Transit just launched this as a rider-facing feature.

@gcamp - did you guys introduce a GTFS-like extension with Metro or are you getting data from them in a more proprietary format?

gcamp:
@stevenmwhite For our data exchange with LA Metro it's a proprietary format, but we do use an internal GTFS extension for our own use. That format was shared with MobilityData and heavily influenced the current proposition.

Dave-TfNSW:
I may complicate the discussion here slightly, but at Transport for NSW we have just developed occupancy predictions for trains based off past performance, that also updates in realtime.  For example a late running vehicle is likely to accumulate more passengers. Where would this kind of occupancy belong? It’s not a real-time actual like our weight bearing sensors that we disseminate using the vehpos feed, but we currently don’t have these sensors on all trains so are looking to fill in the gaps on the network.
We would be happy to extend stop_times.txt to include mean_occupancy, up to 2 weeks in advance for the purely static predictions, but could the pseudo-realtime occupancy prediction be given out in say an extended trip-update message?

Possible example:

`    stop_time_update {
      arrival {
        delay: 155
        mean_occupancy_status: MANY_SEATS_AVAILABLE
      }
      departure {
        delay: 155
        mean_occupancy_status: MANY_SEATS_AVAILABLE
      }
      stop_id: "2508172"
      schedule_relationship: SCHEDULED
    }`

This would then be overridden by vehpos information where it's available.
What are peoples thoughts on this approach?

skinkie:
@Dave-TfNSW I like how you make _explicitly_ this is a prediction, but wouldn't it be better if we would have something like a occupancy_source enumeration, such as "historic, counter, ..."

ghynes:
At [CitySwift](http://cityswift.com/) was are predicting occupancy for a number of the bus companies that we work for (e.g. https://when2travel.co/ )
I just wanted to post to say that we would like to publish this data in GTFS and are interested in this proposal.

scmcca:
The feedback on this PR and through other channels has been very informative to the drafting process. Some updates:

- GTFS-MeanOccupancies has been rebranded to GTFS-OccupancyPredictions to better reflect practices of generating predictions and to not semantically prescribe primary data processing methods
- Option B has been added to the [proposal doc](http://bit.ly/gtfs-occupancies), taking shape of the simpler approach proposed by @skinkie

I invite everyone to review and compare Option A and Option B so we can work towards a final proposal. As always, you can comment and make suggestions directly in the proposal document, or you can continue to express your needs and concerns in this thread.

Thanks!

scmcca:
@Dave-TfNSW I think the proposal to add occupancy predictions in realtime using historical data is a valuable, but different use case from describing usual or expected occupancies at the scheduled `trip` or `stop_time` level. Providing predictions in realtime operates on the immediate future while describing usual or expected occupancies can potentially operate on the order of weeks into the future (and be integrated into trip planners). Moreover, I think there is a strength in being able to describe usual or expected occupancies that are schedule-based, and not vehicle-based for agencies who do not implement GTFS Realtime. Indeed, any realtime occupancies may be used to override schedule occupancies, or can be shown side-by-side like Google Maps [popular times](https://miro.medium.com/max/1580/1*BaMCF9GLd3v4vUBMSu3MhQ.png).

@caywood @skinkie On having information on the source data: this is important information but hard to standardize across all methodologies. I think a limited number of enums could be of value (such as 0: Historical average, 1: Modeled predictions) for consumers to rapidly know what they are dealing with. Beyond that, more in-depth clarification would seem better communicated across other side channels between parties. 

I've added **Option C** to the working [proposal document](https://docs.google.com/document/d/1uGpx39_ZrF9lZ8z3d-NOb4kN5_2rCs-Z4TQ3uorusnI/edit#heading=h.u9g6vw1vuv7e). This one takes the form of Option B, initially proposed by @skinkie, but aims to express forecasted dates more efficiently for both frequent updates, and longer-term (i.e., seasonal updates). It accomplishes this by adding fields for every day of the week similar to `calendar.txt`, Monday through Sunday, and then specifying a start and end date. This structure allows for weekly patterns of forecasted occupancies. The rationale is that data producers who are meticulous about forecasting occupancies will likely be updating `occupancies.txt` on a weekly basis, not needing to specify specific dates beyond that. Meanwhile, data producers who have more stable and recurring occupancies have the flexibility to define occupancies for a given week, and forecast that same weekly pattern out to as long as they deem valid. Overall, this method would serve the granularity of agencies with the need for frequent updates while not burdening smaller agencies with redundant occupancy descriptions for each individual date over a longer period of time, where they are stable and recurring.

Also added as per conversations above, is a `data_source` field to specify if the data that were used to populate the occupancy fields were from `0: Historical average`, or `1: Modeled prediction`.

What do we think about the proposed weekly pattern method vs the individual date method, and the description of the data source? 

I think we are close to reaching something final here! Thanks.

jxeeno:
We'll probably need to cover off handling frequency-based trips defined by `frequencies.txt` as well, potentially by adding `start_time`/`end_time` to `occupancy_patterns.txt`

caywood:
@scmcca I think the description of the data source in Option C is fine. 

Weekly patterns are a nice way to compress the predictions and I expect most models/averages will only have that level of accuracy anyway (rather than predicting every single date). 

I do wonder how holidays etc. fit into this scheme - could they be treated as exceptions? But in any case lower crowding than predicted is not as much a problem as higher crowding. 

So long as this meets the requirements of the producers, speaking as a consumer we are OK with it!

skinkie:
@scmcca so how would option C look like for single dates? The correct day of week with the same begin/end date?

scmcca:
@skinkie For describing single dates beyond one week without weekly patterns, that is certainly the way to do it!

@caywood From what skinkie suggested above, holidays can be described in Option C. For example, if you wanted to display a usual occupancy for Friday, December 25th, 2020 that is different from typical Fridays, you could specify the `occupancy_status` for `friday=1` with `start_date=20201225` and `end_date=20201226`. Of course, the data would have to be optimized such that there are no overlapping occupancy descriptions covering the `start_date` and `end_date` range for the same `trip_id` or `stop_sequence`.

This is where the `data_source` field is important to semantically distinguish what the data are. If you are providing `data_source=0` (historical averages) for a date range that covers `20201225`, it should be interpreted by the consumer that the occupancy is not typical for the individual date, but typical for Fridays on average. If you are distinguishing a different occupancy for `20201225` in particular, `data_source=1` (modeled prediction) should be used to communicate that reality.

skinkie:
> @skinkie For describing single dates beyond one week without weekly patterns, that is certainly the way to do it!

From what I heard of Dutch agencies everyone is using a moving average, hence weekly patterns would never happen. I honestly wonder how this can ever be as predictable as **option C** suggests. Especially if you take in account _events_.

caywood:
@scmcca thanks for clarifying how holidays are handled. 

Last question. Would it be clearer to allow separate files for short term predictions and for steady state day-of-week patterns? They're both important, but the short term predictions are most important right now. 

This Option C hybrid format is trying to do multiple things at once and I can see people using it incorrectly.

RTL-Gilles-Ferland:
Hello,

I represent RTL an agency from Longueuil, Quebec, Canada region. We are working on providing estimated occupancy for our clients through the different mobile applications and thought it would be interesting to share.

We’ve had a few discussions with other agencies in our region, exo and STL, as well as people from TransitApp and Chrono Mobile (Two local mobile applications). We looked at different ways of providing this data before finding this discussion here. After reading and discussing this specification, we are opting for Option C. We feel that it is the option that is most adapted to the data we have and the simplest for us to provide.

We think important that we keep control on how the estimated occupancy level for a stop_time is calculated and that we decide the occupancy_status that is communicated. We also want to be able to update these estimated occupancy on a weekly basis. (Specially with the impacts of COVID-19) Some agencies in our region will probably even update daily.

Option C gives us the ability of specifying the occupancy_status and the ability of defining a validity period for the data that we provide. (We don’t intend to use the occupancy_percentage.) We will use it to provide historical averages. We think that predictions should rater be in the GTFS-RT probably under trip_update as they would be based on what the APC is currently reporting and the type of vehicle currently making the trip, like @Dave-TfNSW suggested earlier. (But this is just a thought as we’re not able to provide predictions at this time we won’t be working on this.)

We hope to have a working version of Option C for Q1 2021 and are working with the two mobile application providers of our region TransitApp and Chrono Mobile. We will try and give feedback as we move forward with development.

Thanks

scmcca:
@skinkie 
> From what I heard of Dutch agencies everyone is using a moving average, hence weekly patterns would never happen. I honestly wonder how this can ever be as predictable as option C suggests. Especially if you take in account events.

This was echoed by CitySwift (@ghynes) through other channels of communication as well. However, the supposed strength of Option C is that it allows for _both_ individual day projections as well as a single-week projection over multiple weeks, sharply reducing the potential volume of data in a given `occupancies.txt` file for agencies with recurring and stable patterns, or where they wish to express purely historical trends rather than granular daily predictions. Moreover, Option C can be optimized for `trip_id` or `stop_sequences` where the `occupancy_status` can be classified for many days at once. 

It aims to find a happy medium between granular predictions and historical generalizations in order to be a simple and accessible spec. Most importantly, nothing is lost in Option C that is captured by Option B.

@caywood 
> Would it be clearer to allow separate files for short term predictions and for steady state day-of-week patterns? 

I'm not sure that separate files would be better for reducing ambiguity, I think it might even complicate things for all sides. 

>They're both important, but the short term predictions are most important right now.

You can do both in one `occupancies.txt` file. The shorter-term projections would have a more immediate `start_date` `end_date` range, while more general steady state day-of-week projections would have a later `start_date` `end_date` range following the more immediate one. Updates to `occupancies.txt` could specify precision as date ranges approach IRL and projections become more precise.

> This Option C hybrid format is trying to do multiple things at once and I can see people using it incorrectly.

Do you see specific pitfalls that might lead someone to use Option C incorrectly? What you can do with Option C is relatively limited. What consumers will get is: an occupancy description; whether its a pure historical average (i.e., usual occupancy) or modeled prediction (i.e., what you should expect); the `trip_id` or `stop_sequence` it is associated with; and when it applies.

skinkie:
> I'm not sure that separate files would be better for reducing ambiguity, I think it might even complicate things for all sides.

I think this should not have to be embedded within the GTFS 'network and timetable' package. This might be the first opportunity to have a standardised static publication not part of that package, still living inside that scope. As long and feedinfo would relate to the published GTFS package that would sound workable from a publisher and client perspective.

ghynes:
Option C seems to be unnecessarily combining (i) the ability to have occupancies apply for a range of dates and (ii) how those occupancies were calculated.

I think the option to have predictions apply for a range of dates sounds fine, although we would be unlikely to use it.

What problem are we trying to solve by stating how the predictions are created? It feels like the problem that's trying to be addressed is to understand the accuracy of predictions, but the result seems to be overly prescriptive guidelines about historical occupancy calculations. The one thing you can be certain of is that every route should break those guidelines at multiple points throughout the year.
I'd question whether we need that field at all.

I wouldn't agree with the assumption that predictions will be calculated once per week. If this was the case, immediately before the predictions were run you wouldn't have any predictions for the following day. We run our predictions at least once every day and make predictions two weeks into the future. Therefore it's very important to us that we can model multiple weeks of data.

caywood:
> What problem are we trying to solve by stating how the predictions are created?

@ghynes if an agency is giving us historical averages that just say "9:30 AM bus is usually full" this is nearly worthless to display to users of our real-time information apps and displays (it does have some applications in trip planning)

If an agency has a predictive model that is augmenting APC data with historical data to give us a predicted occupancy at a future stop, that is highly useful to everyone

So for us it is effectively the difference between (1) nearly useless, and (2) better than current GTFS-Realtime occupancy data

ghynes:
@caywood Thanks for outlining your use case. 
It seems to me that what we're trying to establish is the appropriateness of the occupancy values for different use cases. 

As things stand with Option C if there's a feed with date_source=1 (modeled predictions) occupancies can be 6 or 7 days old by the time the passenger is traveling. Additionally, there's no reference or requirement to use APC data. 

Instead of data_source, would we be better off with either (i) an accuracy range, or (ii) an Enum of suitable uses with guideline accuracy requirements?

scmcca:
**Dates**:
From the conversation thus far it seems that the method to describe occupancies on a weekly and recurring pattern (Option C) does not hold its own weight for its supposed advantages, with individual date descriptions as the preferred method (Option B). I've updated the PR to use a single `date` field. If there are any objections or further thoughts please feel free to voice them.

**Data source**:
It seems that the `data_source` field is somewhat of an incomplete attempt to proxy data quality and methods used to populate the occupancy fields. 

@ghynes 
> Instead of data_source, would we be better off with either (i) an accuracy range

Standardizing this range of possibilities may simply leave too much room for ambiguities, even with an enum of suitable uses.

> or (ii) an Enum of suitable uses with guideline accuracy requirements?

A `margin_of_error` field had been proposed in early iterations but was decided against because it would likely not be meaningful to riders, and producers may not know how to use it. However, it may be useful for consumers as per the reasons stated by @caywood and others.

Another option would be to remove the `data_source` field altogether and leave it up to producers and consumers to communicate quality standards and methodology in other channels of communication.

RTL-Gilles-Ferland:
**Dates :**
We don't agree with use of single date. It forces a repetition of all trip_id/stop_sequence for each and every day that we are providing data. Whereas the use of a date begin/end allows to provide the data for every day of the week in a single trip_id/stop_sequence combination. File size and data to manage is simpler we feel, and matches with the calendar file concepts.

**Data source :**
We don't see the value added of this field. The occupancies.txt file, for us, is a way to provide estimated occupancy levels on our schedule. Allowing transit users to have an idea of the usual occupancy of each trip when consulting our schedule or planning their trip. We think this field can be removed.

**Predictions :**
For us, predictions of occupancy are the action of combining the estimated occupancy and the realtime information coming from the APC system. Therefore, predictions should be integrated to the GTFS-Realtime specification in the trip update message.

gcamp:
On the dates : 
We would clearly prefer a range of dates vs single dates on the existing propositions for the same reason @RTL-Gilles-Ferland mentioned. 

However I would propose that we re-use the calendar and calendar_date system. Instead of having a `date` in occupancy_profiles, `service_id` should be provided with all the capabilities that systems offers. The service_ids used could be the same as used from the trip_id or new ones. It would allow reuse of concepts (and eventually code). 

The main downside of that solution is to have a second table with both `trip_id` and `service_id` on the same table that would mean two different thing depending on the table, which could be confusing.

RTL-Gilles-Ferland:
The re-use of the whole service concepts from calendar and calendar_dates is a good idea. Only, we will want to update the occupancy data on a weekly basis for the same schedule that is valid for 3 months. How would we proceed in such a scenario? It seems we don't have a choice to keep some form of date system in the occupancy file allowing us to define a validity period that differs from the schedule validity.

francispeixoto:
would the weekly updates vary enough to trigger a change in the occupancy enum?

ghynes:
@scmcca That all sounds good to me.

I know @RTL-Gilles-Ferland and @gcamp are looking for a date range. I don't have any objection to that as long as it's done in such a way that we can provide daily occupancy predictions for several weeks into the future.

RTL-Gilles-Ferland:
@francispeixoto Good question. From what we've seen in the last few months with COVID, definitely. If we go back before COVID, maybe not that much, except in seasonal changes, vacation periods, or when school starts/stops. Having it completly fixed for a 3-4 month period with the schedule seems restrictive though. With the date range, both scenarios are possible, agencies that want the data fix for the period of the schedule can do so, and agencies that want to update it regularly also have the possibility.

@ghynes I'm not sure I understand how the date range would prevent providing occupancy data for several weeks in the future?

Other idea : Is there a way where we can have options in the format? For example, if no dates are specified, then the service dates of the trip_id (from the calendar file) defines the validity of the data. If a date range is specified, it overwrites the service dates. It would allow to have default occupancy data and exceptions that range from one day to a few weeks. Allowing to cover exceptions in occupancy like COVID or spring break maybe?

alesk1978:
@RTL-Gilles-Ferland using the calendar files (calendar.txt and calendar_dates.txt) would allow to manage holidays or other exceptions already managed

RTL-Gilles-Ferland:
@alesk1978 That is true.

As a data provider, how would we proceed to update estimated occupancy data without forcing a complete update of the GTFS data on the data consumers? Suppose we remove all notion of time in the occupancy file, if we notice that estimated occupancies are varying at a point we need to update the data, how would we proceed?

As an agency we prefer not to provide information rather than provide information that is not accurate. Considering that the COVID impacts will be felt on transit occupancy levels for the next two years. Having a way to update this data on a regular basis is necessary. (Without changing schedules our routes.)

alesk1978:
@RTL-Gilles-Ferland wether time is included in the occupancy file or not, we are looking into having treating the service offer (route, trips, stop, stop_time) and the complementary infos such as fare and occupancy as 2 different flows, so it would be possible for data providers to upload only the complementary informations and fo data consumers to request only those informations.

RTL-Gilles-Ferland:
@alesk1978 So data consumers would have to check last modified date/time on the occupancy.txt file to know that the data in the file has been modified?

alesk1978:
@RTL-Gilles-Ferland No they would still receive specific notification as it's done right now. But that is specific behavior on our part maybe not relevant for this forum

skinkie:
I would love to see us exploiting the powers of feed info for this purpose. For example an established version could match a static occupancies delivery.
-- 
Stefan

RTL-Gilles-Ferland:
@skinkie Interesting idea.

Although, if there's no date range for data validity in the occupancies file how can we ensure that data consumers are updating the information at an acceptable frequency. If data consumers base their udpates on services, then they will miss any update in between that we may publish. Having a date range for data validity provides garantee to agencies that the data is updated at the desired frequency and that the clients are getting the latest informations.

skinkie:
> Although, if there's no date range for data validity in the occupancies file how can we ensure that data consumers are updating the information at an acceptable frequency.

This cannot be assured for normal GTFS timetables either. This is trust based.

> If data consumers base their updates on services, then they will miss any update in between that we may publish. Having a date range for data validity provides garantee to agencies that the data is updated at the desired frequency and that the clients are getting the latest informations.

I envision just two files;

GTFS-20201207.zip feed_version=0.1
GTFS-OC-20201207.zip followed up by GTFS-OC-20201212.zip followed up by GTFS-OC-20201215.zip feed_version=0.1

Next month;

GTFS-20210107.zip feed_version=0.2
GTFS-OC-20210107.zip followed up by GTFS-OC-20210112.zip followed up by GTFS-OC-20210115.zip feed_version=0.2

RTL-Gilles-Ferland:
What would be found in the : "GTFS-OC-20201207.zip followed up by GTFS-OC-20201212.zip followed up by GTFS-OC-20201215.zip feed_version=0.1" files? Just the occupancies.txt file?

francispeixoto:
@skinkie  Updating the GTFS data on a weekly basis can be taxing for the systems that consume it, depending on the amount of data they contain. I would rather see a baseline being set for a date range in the GTFS, and a mechanism where a data producer could temporarily overwrite the data for a specific subrange within the range of the published baseline.

skinkie:
> What would be found in the : "GTFS-OC-20201207.zip followed up by GTFS-OC-20201212.zip followed up by GTFS-OC-20201215.zip feed_version=0.1" files? Just the occupancies.txt file?

And obviously: feed_info.txt.

> @skinkie Updating the GTFS data on a weekly basis can be taxing for the systems that consume it, depending on the amount of data they contain. I would rather see a baseline being set for a date range in the GTFS, and a mechanism where a data producer could temporarily overwrite the data for a specific subrange within the range of the published baseline.

I don't know what kind of system you are running, but here we have daily updates. And at least biweekly updates of the national train operator. Our schedules and realtime are considered as academically valid at country scale.

RTL-Gilles-Ferland:
I'm not sure seperate files is the way to go. Having one zip to monitor and retrieve is simpler for consumers.

How about we add a new field in feed_info that would allow a producer to declare what has changed in the zip. Like a "feed_changes" that could be an ENUM of all the possible files. If it's empty then the whole zip has been updated, if it specifies a file, than only that file has been updated.

skinkie:
@RTL-Gilles-Ferland I consider **downloading** and reprocessing a few gigabytes of data just because an occupancy file has changed much more extra bload than configuring a new style of feed, with higher frequency updates, but lower than realtime. GTFS-RT is also configured independently.

alesk1978:
I was wondering ... in the end, who will decide what GTFS-Occupancy will be?

alesk1978:
@skinkie gigabytes? who do you work for?

skinkie:
> I was wondering ... in the end, who will decide what GTFS-Occupancy will be?

You, me, we as group. A democratic proces after a producer and a consumer have proved the solution works.

> @skinkie gigabytes? who do you work for?

Stichting OpenGeo / Bliksem Labs B.V. the Dutch GTFS schedules are 1.4GB uncompressed about 350MB compressed. Given the enormous amount of downloads (and sheer incompetence of many that download the file every hour without the use of file modification principles it is rather critical that open data retains manageble.

gcamp:
@skinkie couldn't you enforce that file modification checks are done in some way in the request to avoid bad actors? 

I understand that for very big GTFS, adding data that's significant can be problematic. I'm worried long term that adding ad-on data on GTFS, we will end up something fragmented. I think both are valid options, but I clearly prefer all-in-one GTFS feed.

skinkie:
> @skinkie couldn't you enforce that file modification checks are done in some way in the request to avoid bad actors?

If you have ideas I would love to hear them. Adding a login page is out of the realm...
 
> I understand that for very big GTFS, adding data that's significant can be problematic. I'm worried long term that adding ad-on data on GTFS, we will end up something fragmented. I think both are valid options, but I clearly prefer all-in-one GTFS feed.

Lets say that at 'production phase' we will obviously add the file. But given that we have daily updates of the predictions, and not daily updates of any of the schedules that we package... that gives a narrow path. In addiotion, big search Engines take up to a week to process GTFS updates...

alesk1978:
@skinkie if take @RTL-Gilles-Ferland's suggestion and identify in feed info which files have been updated, we get a GTFS that would be usable by any consumer wether they are looking for the whole package or simply want to update the info that has changed.

skinkie:
@alesk1978 you can only access feed_info.txt after you have downloaded the zipfile.

alesk1978:
@skinkie agree  but I doubt that dowloading a 380M zip file is a problem for most consumers

skinkie:
@alesk1978 it is for me as producer. Since we are paying those bills.

alesk1978:
@skinkie I'm not sure I follow ...

skinkie:
> @skinkie I'm not sure I follow ...

The part that someone has to pay for a download that generates network traffic?

scmcca:
@gcamp 
> I would propose that we re-use the calendar and calendar_date system. Instead of having a date in occupancy_profiles, service_id should be provided with all the capabilities that systems offers. The service_ids used could be the same as used from trips.txt or new ones. It would allow reuse of concepts (and eventually code).

This could be an option. Although it may be simpler to have occupancies in a single file that gets updated as needed without being necessarily tied to other files in a  GTFS dataset that may also need to be updated. Moreover, using `calendar.txt` for this purpose would change its function from describing trips that are in service, to also describing occupancy levels that are valid. This may cause some confusion but is certainly a viable option (as we are also exploring the use of `service_id` in [Fares v2](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#), issue #252).

@alesk1978 
>I was wondering ... in the end, who will decide what GTFS-Occupancy will be?

Please see [CHANGES.md](https://github.com/google/transit/blob/master/gtfs/CHANGES.md) for the complete specification amendment process.

@RTL-Gilles-Ferland and others in respect to occupancy validity: the proposal assigns specific dates or range of dates that explicit when an occupancy level is to be shown. This should be used in essence to define the validity of the foreseen occupancy data; for unspecified dates, there are no occupancy level projection. 

Indeed this alone does not address incoming updates that occur during "valid" occupancy date ranges. In which case a use of feed info as suggested by @skinkie and @RTL-Gilles-Ferland is an interesting discussion. 

To return to the topic of **dates**: There is a clear need to describe both individual dates as well as weekly patterns for reasons mentioned [here](https://github.com/google/transit/pull/240#issuecomment-731167653) and echoed by others. 

**Please see my most recent commit for Option D.** It takes the form of describing weekly patterns, but can also be used to describe individual dates using the same amount of data as either Option B or Option C. Additionally, I've proposed an `exception` field as described by @francispeixoto to "temporarily overwrite the data for a specific subrange within the range of the published baseline" for uses such as holidays when describing weekly patterns.

RTL-Gilles-Ferland:
@skinkie I'm curious. Why do you expect to update occupancy estimates on a daily basis? Do you expect on having such a variation that you will need a daily update of those estimates? Even with COVID impacts, at least on our side, the client behavior with transit doesn't seem to vary that much.

ghynes:
We also update our predictions every day.

skinkie:
> @skinkie I'm curious. Why do you expect to update occupancy estimates on a daily basis? Do you expect on having such a variation that you will need a daily update of those estimates? Even with COVID impacts, at least on our side, the client behavior with transit doesn't seem to vary that much.

Because our public transport operators are updating the data every day, for the next three days.

http://data.ndovloket.nl/bezetting/

RTL-Gilles-Ferland:
@scmcca I notice that the occupancy_status enum as described in Option D allows for int values from 0 to 6 but the occupancy_status enum defined in the GTFS-Rt specification is text values. Which type of value do we aim to support? I would recommend text to align with GTFS-Rt specification.

gcamp:
@RTL-Gilles-Ferland the enum in the protobuf have numerical value, the text value is simply used for better human comprehension : https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L430

RTL-Gilles-Ferland:
@gcamp Ok thanks. Will follow that.

RTL-Gilles-Ferland:
We are moving ahead with implementation of Option D on our side. We will use de date range and plan on updating the occupancy data once a week. We will publish a complet GTFS zip on every update but only the occupancies file will change. Development will begin in january, will try to give feedback along the way.

Until then, happy holidays to all!

scmcca:
@RTL-Gilles-Ferland Good to know! 

Please note that GTFS-Occupancies is dependent on methods for describing occupancies in GTFS Realtime that are currently experimental. Ideally, we should not adopt GTFS-Occupancies with occupancy descriptors (i.e., `OccupancyStatus` enums) that are experimental; we want the static and realtime methods to be representatively interchangeable. For this reason, the `occupancy_status` field in this proposal is subject to change.

I encourage everyone to check out #212 for the relevant conversation that aims to get some occupancy description in GTFS Realtime passed as official. This will be imperative for passing GTFS-Occupancies to a vote, and ultimately for riders to have this information.

RTL-Gilles-Ferland:
@scmcca Understood and will participate in the discussion concerning the enum.

mpaine-act:
I don't understand the PR specification reference changes for Option D.  The per day enums appear to be nondeterministic to me when its value is 0 while the date range is more than a week.  If say Monday is 0, which Monday would be valid and which one would be invalid?

skinkie:
> The per day enums appear to be nondeterministic to me when its value is 0 while the date range is more than a week. If say Monday is 0, which Monday would be valid and which one would be invalid?

It seems that you don't understand how calendar.txt works, which is quite surprising. It is **any** monday in that period.

mpaine-act:
> It is **any** monday in that period.

Ah, I misread.  Thank you.

Dave-TfNSW:
Hi all, I see this thread has gone stale, but Sydney Trains have produced a sample of GTFS-Occupancies and are planning a late October release. One of the topics that has come up is expressing confidence levels in the data, especially during timetable changes when the prediction algorithm can take a few days to catch up. Is there any ideas about an uncertainty field being added to the spec?
[SydneyTrainsGTFS_Occupancies.zip](https://github.com/google/transit/files/7183032/SydneyTrainsGTFS_Occupancies.zip)
e.

gtfs/spec/en/reference.md
```diff
@@ -366,7 +369,31 @@ Describe the different levels of a station. Is mostly useful when used in conjun
...
+Usual or expected in-vehicle occupancy levels.
+
+For describing historical average occupancies at least 4 weeks of occupancy data are suggested to be averaged across each unique day of the week, valid for up to 2 weeks thereafter in `occupancies.txt`.
+
+Methods for occupancy forecasting used to populated `occupancies.txt` are not strictly specified. Beyond the guidelines given above, it is the responsibility of data producers to ensure that the data provided are meaningful. 
+
+| Field Name | Type | Required | Description |
+| ----- | ----- | ----- | ----- |
+| `trip_id` | ID referencing `stop_times.trip_id` | **Required** |  Identifies a `trip_id` for which an occupancy level is described. |
+| `stop_sequence` | ID referencing `stop_times.stop_sequence` | **Conditionally Forbidden** | Identifies a `stop_sequence` along `occupancies.trip_id` for which an occupancy level is described.<br><br>Defined values in `occupancies.stop_sequence` will apply to subsequent `stop_times.stop_sequence` that are not defined in `occupancies.stop_sequence` for the same `trip_id`.<br><br>**Conditionally Forbidden**<br>- **Forbidden** if `occupancies.trip_id` is empty.<br>- Optional otherwise. |
```
paulswartz, line 385: question: `trip_id` is listed above as **required** so it doesn't seem possible for this to be forbidden?
suggestion: make this **optional**

gtfs/spec/en/reference.md
```diff
@@ -366,7 +369,31 @@ Describe the different levels of a station. Is mostly useful when used in conjun
...
+| `tuesday` | Enum| **Conditionally Forbidden** | Functions in the same way as `occupancies.monday` except applies to Tuesdays. |
+| `wednesday` | Enum| **Conditionally Forbidden** | Functions in the same way as `occupancies.monday` except applies to Wednesdays. |
+| `thursday` | Enum| **Conditionally Forbidden** | Functions in the same way as `occupancies.monday` except applies to Thursdays. |
+| `friday` | Enum| **Conditionally Forbidden** | Functions in the same way as `occupancies.monday` except applies to Fridays. |
+| `saturday` | Enum| **Conditionally Forbidden** | Functions in the same way as `occupancies.monday` except applies to Saturdays. |
+| `sunday` | Enum| **Conditionally Forbidden** | Functions in the same way as `occupancies.monday` except applies to Sundays. |
+| `start_date` | Date | **Required** | Start date of the date interval that the occupancy level is valid. |
+| `end_date` | Date | Optional | End date of the date interval that the occupancy level is valid.<br><br> If defined, `occupancies.end_date` must be greater than `occupancies.start_date`.<br><br>If empty, the occupancy level will apply only for the date specified in `occupancies.start_date`. |
```
paulswartz, line 395: suggestion: what about making this **required**? If the producer wants to represent a single day, they can set `start_date` == `end_date`.

gtfs/spec/en/reference.md
```diff
@@ -366,7 +369,31 @@ Describe the different levels of a station. Is mostly useful when used in conjun
...
+| `exception` | Enum | Optional | Indicates if an occupancy exception should override an existing occupancy description for the same trip or stop time for overlapping dates. Valid options are:<br><br> `0` or empty - Does not override an occupancy description.<br>`1` - Overrides an occupancy description.
```
paulswartz, line 396: question: I don't see how this works, or why it would be needed? I see in the PR description that it's used for holidays (I suppose in the same way that `calendar_dates.txt` is?) but I believe that's only necessary because `service_id` is the primary key in `calendar.txt`. In this file, representing a trip's occupancy on different days could be provided by non-overlapping records without violating any primary key constraints.

gtfs/spec/en/reference.md
scmcca: Thanks for flagging this. Indeed this should be optional.

gtfs/spec/en/reference.md
```diff
@@ -366,7 +369,31 @@ Describe the different levels of a station. Is mostly useful when used in conjun
...
+
+Usual or expected in-vehicle occupancy levels.
+
+For describing historical average occupancies at least 4 weeks of occupancy data are suggested to be averaged across each unique day of the week, valid for up to 2 weeks thereafter in `occupancies.txt`.
+
+Methods for occupancy forecasting used to populated `occupancies.txt` are not strictly specified. Beyond the guidelines given above, it is the responsibility of data producers to ensure that the data provided are meaningful. 
+
+| Field Name | Type | Required | Description |
+| ----- | ----- | ----- | ----- |
+| `trip_id` | ID referencing `stop_times.trip_id` | **Required** |  Identifies a `trip_id` for which an occupancy level is described. |
```
skinkie, line 384: I should not bring this up, but if this is a foreign key, then we must resolve the frequencies.txt issue.

gtfs/spec/en/reference.md
scmcca: I see your point. The consensus in this discussion seems to be that updates to `occupancies.txt` happen daily. So I agree that `exception` is not needed.

gtfs/spec/en/reference.md
```diff
@@ -380,6 +382,30 @@ Describes levels in a station. Useful in conjunction with `pathways.txt`, and is
...
+
+Usual or expected in-vehicle occupancy levels.
+
+For describing historical average occupancies at least 4 weeks of occupancy data should be averaged across each unique day of the week, valid for up to 2 weeks thereafter in `occupancies.txt`.
+
+Methods for occupancy forecasting used to populated `occupancies.txt` are not strictly specified. Beyond the guidelines given above, it is the responsibility of data producers to ensure that the data provided are meaningful. 
+
+| Field Name | Type | Required | Description |
+| ----- | ----- | ----- | ----- |
+| `trip_id` | ID referencing `stop_times.trip_id` | **Required** |  Identifies a `trip_id` for which an occupancy level is described. |
```
scmcca, line 397: Should this be referencing `trips.trip_id`? How does this interact with `occupancies.stop_sequence` if both are defined?

harringtonp:
Coming to this discussion late but was wondering why was adding occupancy_status directly to stop_times not considered ? In this way you are not having to replicate the whole service/calendar mechanism in occupancies and it would be far easier for existing consumers to implement. You also don't run into any frequencies issues as mentioned above.

And secondly regarding occupancy_percentage discussions it would seem to me that if it is being added here it should probably also be added alongside departure_trip_occupancy in TripUpdates. In other words there should be some synergy between these two (VehiclePosition on the other hand which has both occupancy_status and occupancy_percentage is different as occupancy is measurable and an exact percentage makes sense)

skinkie:
> Coming to this discussion late but was wondering why was adding occupancy_status directly to stop_times not considered ?

Because that assumes that the occupancy is equal for every trip regardless of the day.

harringtonp:
> > Coming to this discussion late but was wondering why was adding occupancy_status directly to stop_times not considered ?
> 
> Because that assumes that the occupancy is equal for every trip regardless of the day.

And when it's not, there is always the option of defining new services. In general would you not expect that if trips are using the same timings across multiple days that occupancies would also generally align (within the scope of how approximate occupancy estimations are likely to be)

skinkie:
> And when it's not, there is always the option of defining new services. In general would you not expect that if trips are using the same timings across multiple days that occupancies would also generally align (within the scope of how approximate occupancy estimations are likely to be)

If you maintain a feed with any significant size and provide real-time data for it, you will like normalisation at some point. And for weekday trips, it is considered pretty stable in The Netherlands.

gtfs/spec/en/reference.md
```diff
@@ -394,7 +394,7 @@ Methods for occupancy forecasting used to populated `occupancies.txt` are not st
 
 | Field Name | Type | Required | Description |
 | ----- | ----- | ----- | ----- |
-| `trip_id` | ID referencing `stop_times.trip_id` | **Required** |  Identifies a `trip_id` for which an occupancy level is described. |
+| `trip_id` | ID referencing `trips.trip_id` | **Required** |  Identifies a `trip_id` for which an occupancy level is described. |
```
skinkie, line 397: Are you sure you want to do a compound key to a different table?

scmcca: I was thinking since `stop_sequence` is optional it would make sense to reference `trips.trip_id` since it can live on its own. But we should reference whatever is less expensive for consumers. So `stop_times.trip_id` is indeed more favorable?

skinkie: I didn't think of the situation where the trip itself had a single occupancy prediction...

scmcca: Technically a trip with only 2 stops could use trip-level occupancy. But I agree, the gain of not including `stop_sequence` is pretty marginal. Are you suggesting me make `stop_sequence` required then, or should it remain optional?

skinkie: From my point of view it is "better" to have a stop_sequence prediction. Are there any producers that can only do trip level predictions but not stop level ones?

scmcca: I agree it is better. I would think trip level predictions are uncommon.

gtfs/spec/en/reference.md
scmcca: Indeed. The model for `occupancies.txt` seems to only support trips where `time_point=1`, and overlooks frequency-based trips (i.e., trips using `frequencies.txt`) entirely.

Its not clear how we would provide "stop time" level occupancies for frequency based trips since there is no single stop_time identifier. Perhaps by adding additional fields for "date" and "first vehicle journey departure time", but it gets a bit sketchy.

skinkie: What about adding start_time like we do with GTFS-RT?

scmcca: And start_date?

skinkie: Why would a start_data be relevant? It is not in gtfs-static...

gtfs/spec/en/reference.md
jeffkessler-keolis: Speaking from the US Commuter Rail producer side, trip-level predictions currently **are** the norm (e.g. Boston, Chicago, Los Angeles).

While we'd love to provide more granular stop-by-stop ridership data, most of us don't have robust APC systems, which forces us to rely on conductor counts, manual fare collection data, platform counts at downtown terminals, etc.

Since suburb-to-city ridership still dominates travel patterns of North American legacy systems, we've found that providing trip-level crowding assessed at the maximum load point has been a good proxy for customers, particularly since the concern of most customers is how crowded their train *will be*, rather than what it's like should they board at an early stop.

Yes, this overestimates the impact for customers traveling between stations on the outer edge. Yes, this doesn't discern where along the line the train will fill. Yes, there's far more detail that customers would find of value. But this still allows us to provide significant insights given the limited data we have… and I don't want to be yet another provider to release this data in a bespoke format (or worse, something that's not machine-readable like a PDF).

gtfs/spec/en/reference.md
```diff
@@ -380,6 +382,30 @@ Describes levels in a station. Useful in conjunction with `pathways.txt`, and is
...
 
+### occupancies.txt
+
+File: **Optional**
+
+Usual or expected in-vehicle occupancy levels.
+
+For describing historical average occupancies at least 4 weeks of occupancy data should be averaged across each unique day of the week, valid for up to 2 weeks thereafter in `occupancies.txt`.
+
+Methods for occupancy forecasting used to populated `occupancies.txt` are not strictly specified. Beyond the guidelines given above, it is the responsibility of data producers to ensure that the data provided are meaningful.
```
jeffkessler-keolis, line 393: Rather than phrasing this as a normative statement, could the first sentence be treated as guidance like the second?

There are many providers who have conducted analyses and determined that the most predictive values come from durations shorter than an arbitrary four-week period, or a given number of historical trips.

gtfs/spec/en/reference.md
```diff
@@ -380,6 +382,30 @@ Describes levels in a station. Useful in conjunction with `pathways.txt`, and is
...
...`](http://gtfs.org/reference/realtime/v2/#enum-occupancystatus) enums. Valid options are:<br><br> `0` - **Empty**. The vehicle is considered empty by most measures, and has few or no passengers onboard, but is still accepting passengers. <br> `1` - **Many seats available**. The vehicle has a large percentage of seats available. What percentage of free seats out of the total seats available is to be considered large enough to fall into this category is determined at the discretion of the producer. <br> `2` - **Few seats available**. The vehicle has a small percentage of seats available. What percentage of free seats out of the total seats available is to be considered small enough to fall into this category is determined at the discretion of the producer. <br> `3` - **Standing room only**. The vehicle can currently accommodate only standing passengers. <br> `4` - **Crushed standing room only**. The vehicle can currently accommodate only standing passengers and has limited space for them. <br> `5` - **Full**. The vehicle is considered full by most measures, but may still be allowing passengers to board. <br> `6` - **Not accepting passengers**. The vehicle can not accept passengers. |
```
jeffkessler-keolis, line 399: For those who wish to provide objective historical values without the subjective categorizations, could we make this conditionally required with an option to use occupancy_percentage instead?

```suggestion
| `occupancy_status` | Enum | ** Conditionally Required** | Indicates the state of in-vehicle  occupancy. This field refers to the GTFS Realtime [`OccupancyStatus`](http://gtfs.org/reference/realtime/v2/#enum-occupancystatus) enums. Valid options are:<br><br> `0` - **Empty**. The vehicle is considered empty by most measures, and has few or no passengers onboard, but is still accepting passengers. <br> `1` - **Many seats available**. The vehicle has a large percentage of seats available. What percentage of free seats out of the total seats available is to be considered large enough to fall into this category is determined at the discretion of the producer. <br> `2` - **Few seats available**. The vehicle has a small percentage of seats available. What percentage of free seats out of the total seats available is to be considered small enough to fall into this category is determined at the discretion of the producer. <br> `3` - **Standing room only**. The vehicle can currently accommodate only standing passengers. <br> `4` - **Crushed standing room only**. The vehicle can currently accommodate only standing passengers and has limited space for them. <br> `5` - **Full**. The vehicle is considered full by most measures, but may still be allowing passengers to board. <br> `6` - **Not accepting passengers**. The vehicle can not accept passengers. <br><br>Conditionally Required:<br>- **Required** if `occupancies.occupancy_percentage` is empty.<br>- Optional|
| `occupancy_percentage` | Float | **Required** | Indicates the typical occupancy percentage. The value 100 should represent total the maximum occupancy the vehicle was designed for, including both seating and standing capacity, and current operating regulations allow. It's not impossible that the value goes over 100 if there are currently more passenger than the vehicle was designed for. The precision of precision should be low enough that you can't track a single person boarding and alighting for privacy reasons. <br><br>Conditionally Required:<br>- **Required** if `occupancies.occupancy_status` is empty.<br>- Optional|
```

gtfs/spec/en/reference.md
```diff
@@ -406,6 +406,7 @@ Methods for occupancy forecasting used to populated `occupancies.txt` are not st
 | `sunday` | Enum| **Required** | Functions in the same way as `occupancies.monday` except applies to Sundays. |
 | `start_date` | Date | **Required** | Start date of the date interval that the occupancy level is valid.<br><br>To define single dates, `start_date` and `end_date` may be the same. |
 | `end_date` | Date | **Required** | End date of the date interval that the occupancy level is valid.<br><br>To define single dates, `start_date` and `end_date` may be the same. |
+| `start_time` | Time | **Conditionally Required** | First stop departure time for a given vehicle on a trip using `frequencies.txt`.<br><br>Must be some multiple (including zero) of `frequencies.headway_secs` plus `frequencies.start_time` for the corresponding time period.<br><br>Conditionally Required:<br>- **Required** for trips using `frequencies.txt`.<br>- **Forbidden** otherwise. |
```
skinkie, line 409: This will get very complex. But do we want to do something like start_time end_time so for frequency based trips a producer **could** do occupancy time bands? Self-note to my future self: I just shot myself in the foot.

scmcca: Maybe I am missing something, but isn't `start_time` enough to identify individual "vehicle journeys" on a frequency-based trip, and then define different occupancies per `stop_sequence`, per vehicle journey? This seems to accomplish "occupancy time bands".

skinkie: @scmcca in frequency based operation only exact times will have some useful meaning for start_time.

scmcca: This could be what I am missing. At the risk of sounding misinformed, why would it not work for `exact_times=0`? Can't `frequencies.start_time` be used to estimate vehicle journeys?

skinkie: I think that part is only possible heuristically, hence clamp to the nearest start_time.

gtfs/spec/en/reference.md
scmcca: `occupancies.txt` is contingent on getting either `occupancy_status` or `occupancy_percentage` officially adopted in GTFS Realtime for consistency. Whichever method of describing occupancies (possibly both) that becomes official will be represented here.

gtfs/spec/en/reference.md
scmcca: The word "[should](https://datatracker.ietf.org/doc/html/rfc2119#section-3)" denotes the first sentence as a recommendation. Do you have a proposal for a rephrase?

gtfs/spec/en/reference.md
barbeau: I agree with @skinkie that specifying a start and end time to define a window is a more natural solution for exact_times=0 trips. 

For exact_times=1, you have specific trips instances that are defined in static GTFS where you can say "A bus is scheduled to arrive in 5 minutes", so start_time unambigously maps to a single trip instance. 

For exact_times=0 trips, static GTFS is just defining an expected frequency (not trip instances at specific times), so (without real-time data) you can only say "Vehicles are expected to arrive every 15 minutes" during this time window. The exact time that these trip instances are expected to occur drifts throughout the day depending on the previous trip instances, so defining just a start_time may get ambiguous and we would need defined heuristics in the spec of how specific cases would be handled.

gtfs/spec/en/reference.md
jeffkessler-keolis: Even if one is officially adopted, I'd still request that both options be available in the static version. (Or, at the very least, we may choose to release both with one being treated as an experimental field.)

gtfs/spec/en/reference.md
jeffkessler-keolis: That's valid, although I took the "should" to apply to the average and not the amount of data.

My personal preference would be to explicitly state it as a recommendation and swap the two sentences. I think that clarifies the suggestion to producers, while reminding consumers that other mechanisms might be employed in the computation of the data. (e.g. I wouldn't want a consumer to display that as the methodology for how metrics are computed, leaving us to be an exception.)

```suggestion
Methods for occupancy forecasting used to populated `occupancies.txt` are not strictly specified. Beyond the guidelines given below, it is the responsibility of data producers to ensure that the data provided are meaningful.

For describing historical average occupancies, it is recommended that at least 4 weeks of occupancy data be averaged across each unique day of the week, and valid for up to 2 weeks thereafter in `occupancies.txt`.
```

scmcca: I applied your suggestion :) Please sign the [CLA](https://cla.developers.google.com/about/google-individual). If you can't for whatever reason let me know.

jeffkessler-keolis:
@googlebot I fixed it.

scmcca:
@googlebot I fixed it.

Issue #228, opened 2020-06-17
"Direction names" in GTFS


timMillet:
For the last update (2020-11-20) on the issue, please go to [this comment](https://github.com/google/transit/issues/228#issuecomment-731450282).
******
Hi everyone,

I’m opening an issue regarding “direction names” (also called “directions”, “route directions”, “direction destinations”, or “direction headsigns”) as some GTFS stakeholders already provide and consume these data. Before making any specification proposal, I would appreciate having your input on:
1. If you would like these data to be officially adopted in the GTFS specification.
2. On how these data should be modeled.
3. On the name you would give to these "direction names".

The needs are: 
- Providing riders with more detailed information on directions, especially for routes with complex trip patterns (e.g: routes with branches). The direction name would be a piece of information complementary to the trip_headsign and/or stop_headsign.
- Being defined per route at the trip level, at the direction level (trips.direction_id), and/or at the stop_time level. The use cases would be:
  - trip and direction levels: defining a direction that is true for all trips going to the same direction at all their stop_times (e.g with the Red Line in Boston: direction “Southbound” for trips heading either to “Ashmont” or to “Braintree”)
  - stop_time level: defining a direction that is true for all trips going to the same direction at that particular stop_time (e.g: direction “Inbound” until the first downtown stop is reached, then “Outbound”)
- Being compatible with all king of general directions (e.g: “South”, “Southbound”, “Inbound”, “Clockwise”, “Loop”, “Intérieur”, “Sud vers Paris”, “T-Centralen”, “Fruängen/Norsborg”).

The current implementations are ([Gdoc with more details](https://bit.ly/gtfs-direction-names)):
- 1 new file with 3 to 4 fields: MBTA, MTC (SF Bay Area), Trillium, TriMet
- 2 new fields in already-existing files: Transit

Thank you for all your feedback!

laurentg:
Added a small comment in the Gdoc for Mecatran implementation and support.

We would be generally in favor of specifying (even if non-formal) at least direction per route/direction level (two directions names per route), as many feeds do produce such information but many different standards co-exist at the moment.

gcamp:
Even if it's not our current implementation, I would suggest using the directions.txt file. 

For places that would change direction name during their trip, I would suggest using `block_id` to link together two splitted trips that each have their direction. Imagine for CTA's 'L' loop, the trip could be split in two in the middle of the loop to show the change of direction.

stevenmwhite:
[...]

PR #242, opened 2020-08-17
Proposal: Introduce direction names to GTFS

dudebrains01:
Hello GTFS community!

Brody here coming to you from MobilityData with a proposal for direction names in GTFS.

This PR is related to issue #228.
## Context
Currently, the GTFS specification only provides the `trips.trip_headsign` and `stop_times.stop_headsign` field to provide information on the direction of a trip. However, the value in these fields is sometimes superseded by another name like "eastbound" or "inbound", to indicate the general direction of travel.

## Proposal

This proposal would introduce the concept of direction names to GTFS, to allow more general wayfinding instructions to be added to GTFS.

### Features
- New `direction_names.txt` file;
- New `trips.direction_name_id` field;
- New `stop_times.direction_name_id` field

## Producers and consumers
We are actively working with producers and consumers and should have a producer and consumer in the next few weeks.
If you are producing or consuming the extensions in this PR, please let us know in the comments.


dudebrains01:
@googlebot I signed it!

scmcca:
@googlebot I consent.

skinkie:
This is not the result of the discussion.

stevenmwhite:
@skinkie It sounds like you're fairly adamantly against Option D, which is this essentially the one in the PR, but would be ok with Option A.
Is anyone adamantly against Option A and for this option instead? I didn't particularly see that in the previous discussion

I'm going to give this a +1 for us, but I would gladly also give Option A a positive vote.

sam-hickey-ibigroup:
For IBI:  We're happy this is moving forward. but we think this proposal has moved forward prior to consensus being reached in the previous discussion at [Issue 228](https://github.com/google/transit/issues/228).
 
We think that it is premature to open a proposal for this, since 1) the discussion of Options was still ongoing and it makes it more challenging if discussions are split across multiple issues/PRs, and 2) the proposal is based on Option D with some modifications, but the modifications were not discussed. For example: this line in the pull request "The same `trips.direction_name_id` must be provided per combination of `route_id` and `direction_id`." was not in Option D in the Google Doc linked at Issue 228. Also, more details were added in the definition of the direction_names.txt file that were not in the Google Doc linked at Issue 228.
 
That said, although we might be ok with this proposal going ahead as written, our concern is that there were still discussions ongoing at Issue 228 that have not be completed.

timMillet:
This PR seems too early regarding the fact that options were still discussed in issue #228. Sorry about that.
I'm closing this PR, and we will open a new one when a consensus on one of the options presented in issue #228 will be reached.

Thank you, Stefan, Steve, and Sam for your feedback on this PR.

Issue #127, opened 2018-12-09
Language agnostic programmatic definition of GTFS-Static


Subzidion:
I was curious if there was a language agnostic programmatic definition, similar to the proto definition for GTFS-Realtime, but for GTFS-Static, that can be used with an ORM for creating Class representation of the text files. This would simplify the coding process, and would eliminate the need to write code to represent each text file and possibly needing to update that code if/when the spec changes.

skinkie:
It is called SQL ;-) and I think it does exists via GTFSdb.

Subzidion:
GTFSdb is all Python, nothing that's just SQL. I was thinking something more along the lines of what [OneBusAway does with Hibernate](https://github.com/OneBusAway/onebusaway-gtfs-modules/blob/master/onebusaway-gtfs-hibernate/src/main/resources/org/onebusaway/gtfs/model/GtfsMapping.hibernate.xml), but usable in any language.

skinkie:
GTFSdb produces SQL tables in different flavors. You could use that in your agnostic definition.

Subzidion:
Using GTFSdb for this would still requiring updating the Python code to reflect any changes to the spec, running it to create your database schema, then using some other tool to map from the schema to classes in whatever language you want to use. Feels a bit cumbersome. I understand there's probably some class representation for most languages already created, but needing to check and hope they get updated if the spec gets changed seems annoying. If there was some definition, similar to the Hibernate one, that could be used in any language, it would make any of those language-specific GTFS-Static ORMs a lot easier.

barbeau:
@Subzidion There isn't anything official, but the closest thing I'm aware of in concept to what you're looking for is this Data Package specification:

- [GTFS Data Package Specification](https://github.com/Stephen-Gates/GTFS) - A [Data Package specification](http://specs.frictionlessdata.io/data-packages/) with validation accomplished with [Good Tables](http://goodtables.okfnlabs.org/). Includes a data package, schemas, tests, and uses South East Queensland GTFS data as an example.

I started generalizing this to any GTFS:
https://github.com/CUTR-at-USF/GTFS

I think I have some work stashed somewhere beyond what's currently in the above branch...

Subzidion:
This Data Package Specification is exactly what I was looking for. Is there any way we can make this specification a part of the main GTFS package? I would think defining GTFS in terms of the JSON schema would help clarify ambiguity instead of attempting to dfeine the JSON schema from markdown.

barbeau:
[...]

PR #244, opened 2020-09-04
Need for a JSON schema for GTFS (schedule data)?

LeoFrachet:
Hi everybody,

**Who among you would be interested into using a JSON Schema for GTFS?** (let's start with schedule data at first, since we already have the bindings for Realtime).

Having a machine-parsable spec is a request that comes from time to time. But since there is no schema format for CSV like there is the JSON Schema for JSON and XSD for XML, I've drafted JSON Schema version with a few assumptions listed below (mostly for the CSV to JSON conversion), but some extra work would be needed to automatize the tests and to add more complex validations (like conditionally required values).

If they are interest, to move forward we'll have to keep it updated down the road, so we have to do it only if it's gonna be used.

Keep us posted.

Thanks!

Editorial choices:
- JSON structure: `{"agency.txt": [{}, {}, ...], "stops.txt": [{}, {}, ...], ...}`
- Assume type conversion for integers, floats and enum.
- Assume "empty" means either no key, value is `null`, or value is empty string `""`
- Assume required is the opposite of empty. 
- Rely as much as possible on JSON-Schema formats definition (for uri, email adresses...) but still had to define regex for: phone numbers and currency code.
- JSON schema defined is `gtfs-schema-strict.json`, aka it doesn't allow extra field, we can just flip the `additionalProperties` to `true` to create a loose version.

Not validated (but could be done with more work):
- "Conditionally required" rules
- Language code not validated beyond string validation

### DIFF
--- a/gtfs/spec/en/json-schema-strict.json
+++ b/gtfs/spec/en/json-schema-strict.json
@@ -1,0 +1,1039 @@
+{
+  "$schema":"http://json-schema.org/draft/2019-09/schema#",
+  "$id":"http://ids.mobilitydata.org/schema/gtfs",
+  "title":"General Transit Feed Specification (GTFS) for scheduled data",
+
+  "definitions": {
+    "type:string:trimmed:required": {
+      "type":"string",
+      "minLength": 1,
+      "pattern":"^([^ ](.*[^ ])?)$"
+    },
+    "type:string:trimmed:optional": {
+      "anyOf": [
+        {"const": null},
+        {"const": ""},
+        {"$ref": "#/definitions/type:string:trimmed:required"}
+      ]
+    },
+    "format:uri:trimmed:required": {
+      "allOf": [
+        {"$ref": "#/definitions/type:string:trimmed:required"},
+        {"format": "uri"}
+      ]
+    },
+    "format:uri:trimmed:optional": {
+      "anyOf": [
+        {"const": null},
+        {"const": ""},
+        {"$ref": "#/definitions/format:uri:trimmed:required"}
+      ]
+    },
+    "gtfs-type:phone-number:required": {
+      "allOf": [
+        {"$ref": "#/definitions/type:string:trimmed:required"},
+        {"pattern":"^\\+[\\d,]+$"}
+      ]
+    },
+    "gtfs-type:phone-number:optional": {
+      "anyOf": [
+        {"const": null},
+        {"const": ""},
+        {"$ref": "#/definitions/gtfs-type:phone-number:required"}
+      ]
+    },
+    "gtfs-type:email:required": {
+      "allOf": [
+        {"$ref": "#/definitions/type:string:trimmed:required"},
+        {"format":"email"}
+      ]
+    },
+    "gtfs-type:email:optional": {
+      "anyOf": [
+        {"const": null},
+        {"const": ""},
+        {"$ref": "#/definitions/gtfs-type:email:required"}
+      ]
+    },
+    "gtfs-type:latitude:required": {
+      "type":"number",
+      "maximum":90,
+      "minimum":-90
+    },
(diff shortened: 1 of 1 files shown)


LeoFrachet:
Also, I'm unsure that this repo would be the best place for that, but it could be since it would be updated at the same time that the markdown spec. So this is also up to discussion.

barbeau:
Another option is to use a Table Schema that is expressed in JSON defined here - https://specs.frictionlessdata.io/table-schema/#language. Apparently this Table Schema is used by data.gov.uk and data.gouv.fr (see https://frictionlessdata.io/). 

We could use this instead of JSON-izing GTFS and implementing a schema for the JSON version. Someone that works for frictionlessdata.io (I believe) implemented a first draft of Queensland GTFS in the Table Schema at https://github.com/Stephen-Gates/GTFS if you want to see an example.

skinkie:
I think we need  to establish what the format is used for. Let me give an example. If I would like to exchange a single GTFS line in an (json-)object it is obvious that my object would look like  { "route": { "route_id": "1" } }, instead if I wanted to exchange multiple routes the choice becomes between a row store { "routes": [{ "route_id": "1" }, { "route_id": "1" }] } and a column store { "routes": { "route_id": ["1", "2"] } } The latter is likely to compress much better, especially if we would consider other serialisation formats such as PBF too.

jamespfennell:
> But since there is no schema format for CSV like there is the JSON Schema for JSON and XSD for XML

I would say yes and no. It's true there's no CSV validator as such. Ultimately, though, I'm not sure GTFS Static is best viewed as "definitions for a bunch of CSV files". Rather, GTFS Static defines (a) a relational database schema and (b) a specific mechanism for transmitting a dump of a database with that schema.

One way to define the schema in "machine terms" would just be to write out the SQL commands necessary to create the database, paying close attention to types, nullability, defaults, and foreign key constraints (`trips.route_id` foreign keys into `routes.route_id` for example). In theory, to "validate" a specific GTFS Static file you just convert the CSV dump into the appropriate `INSERT` statements and the database will tell you if it's conformant.

jamespfennell:
Honestly, I'm quite nervous about committing things to the repo that would encourage JSON or alternative transmission mechanisms. What I can see happening is that some transit agency decides that it doesn't need to distribute the data in CSV format anymore, and does it in JSON or the alternative format instead. On the consumer side, that is an absolute nightmare scenario. All of us consumers will have to write lots of new code to facilitate such a new approach, and for zero gain on our part.

e-lo:
Without a doubt, representing the canonical GTFS specification in a well-recognized structured data format would have a great deal of benefit to:
- *developers* who wouldn't need to translate markdown into data schemas
- *users* who would benefit from increased ease of transporting the specification requirements from one UX to another programmatically without 'loss or latency in translation' 

The data-based specification files can easily drive documentation ([here is a horrid example](https://e-lo.github.io/GMNSpy/spec.html) that I whipped up from frictionless data spec and sphinx in a few min...but you get the idea) 

**Format**
There isn't a perfect solution here. 

1. **JSON-Schema** Is awesome, but I agree with the comments worrying about signaling JSON support. 
2. **Frictionless** is the best alternative for CSV right now, but their tools are somewhat outdated, don't all do what they say they do.  That said, I think it might be the best solution and we can **fix** what is broken fairly easily (I believe)
3. **Protobuff** is my least favorite option. They are not self-documenting (see multitude of comments in the `.proto` files) and thus the documentation cannot be automatically rendered. They are also not designed for txt/csv and similar to json comment, we don't want to signal an intent to

e-lo:
Suggest we migrate general discussion about definition to #127 and open a PR when community has general consensus on scope/format.

LeoFrachet:
Agreed

PR #248, opened 2020-09-24
fix: Define defaults for VehiclePosition occupancy percentage and status

### LINKED ISSUE #247: VehiclePosition OccupancyStatus and occupancy_percentage return incorrect default values
This issue surfaced during discussion of the multi-carriage occupancy proposal:
https://github.com/google/transit/pull/237#discussion_r470345036

In GTFS-realtime feeds using the current [gtfs-realtime.proto](https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto) file, if the producer doesn't populate `VehiclePosition.occupancy_percentage` and `VehiclePosition.occupancy_status`, then when reading the serialized protocol buffer data consumer software may read these values as `VehiclePosition.occupancy_percentage` = `0` and `VehiclePosition.occupancy_status` = `EMPTY`.

This is due to the issue that protobufs don't have a concept of `null`. https://developers.google.com/protocol-buffers/docs/proto#optional says:

>When a message is parsed, if it does not contain an optional element, the corresponding field in the parsed object is set to the default value for that field. The default value can be specified as part of the message description...If the default value is not specified for an optional element, a type-specific default value is used instead: ...For numeric types, the default value is zero. For enums, the default value is the first value listed in the enum's type definition.

That means right now protobuf-generated consumer code is therefore interpreting `VehiclePosition.occupancy_percentage` as `0` if the producer hasn't populated the field, and `VehiclePosition.occupancy_status` is being interpreted as `EMPTY` if unpopulated.

To fix this `VehiclePosition.occupancy_percentage` needs to change from `uint32` to `int32` (which [should be](https://developers.google.com/protocol-buffers/docs/proto#updating) forwards/backwards compatible), with a default of -1 defined, and `VehiclePosition.occupancy_status` needs to use the value `NO_DATA_AVAILABLE` (added in the multi-carriage proposal) as the default. Not defining these defaults in the initial proposals for these fields in `VehiclePosition` was an oversight. 

https://developers.google.com/protocol-buffers/docs/proto#updating says:

>int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. If a number is parsed from the wire which doesn't fit in the corresponding type, you will get the same effect as if you had cast the number to that type in C++ (e.g. if a 64-bit number is read as an int32, it will be truncated to 32 bits).

My main question would be what happens when a -1 value that was serialized by a new int32 producer is read by a legacy consumer still using the uint32 in the .proto? I plan to do some testing to see how this works prior opening a PR to fix this.

This issue doesn't exist in the new multi-carriage occupancy fields because defaults have been properly defined for these fields within `CarriageDetails`:
https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L465

barbeau:
As discussed in https://github.com/google/transit/issues/247, default values aren't currently defined for `VehiclePosition.occupancy_percentage` and `VehiclePosition.occupancy_status`.

This pull request adds defaults for `VehiclePosition.occupancy_percentage` and `VehiclePosition.occupancy_status` so erroneous defaults aren't read by GTFS-realtime consumers.

TODO:
- [x] Test backwards compatibility for occupancy_status - Done for Java in https://github.com/MobilityData/gtfs-realtime-bindings/commit/a4464a4413fb7a5e71ae4781c76b51291aaa2931 in branch https://github.com/MobilityData/gtfs-realtime-bindings/tree/occupancy-defaults-status. No changes in behavior when reading post default PBs with pre-default code.
- [x] Test backwards compatibility for occupancy_percentage -  what happens when a -1 value that was serialized by a new int32 producer is read by a legacy consumer still using the uint32 in the .proto. Done for Java in https://github.com/MobilityData/gtfs-realtime-bindings/commit/48bd41addbdbe627aa189461897fdee35fec9a63 in branch https://github.com/MobilityData/gtfs-realtime-bindings/commits/occupancy-defaults-status-percentage. No changes in behavior when reading post default PBs with pre-default code, including reading -1 values for occupancy_percentage.
- [ ] Test reading -1 occupancy_percentage value on legacy C++ consumers

Fixes #247

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -437,18 +437,19 @@ message VehiclePosition {
   }
   // If multi_carriage_status is populated with per-carriage OccupancyStatus,
   // then this field should describe the entire vehicle with all carriages accepting passengers considered.
-  optional OccupancyStatus occupancy_status = 9;
+  optional OccupancyStatus occupancy_status = 9 [default = NO_DATA_AVAILABLE];
 
   // A percentage value representing the degree of passenger occupancy of the vehicle.
   // The values are represented as an integer without decimals. 0 means 0% and 100 means 100%.
   // The value 100 should represent the total maximum occupancy the vehicle was designed for,
   // including both seated and standing capacity, and current operating regulations allow.
   // It is possible that the value goes over 100 if there are currently more passengers than what the vehicle was designed for.
+  // -1 in case data is not available for this vehicle (as protobuf defaults to 0 otherwise).
   // The precision of occupancy_percentage should be low enough that you can't track a single person boarding and alighting for privacy reasons.
   // If multi_carriage_status is populated with per-carriage occupancy_percentage, 
   // then this field should describe the entire vehicle with all carriages accepting passengers considered.
   // This field is still experimental, and subject to change. It may be formally adopted in the future.
-  optional uint32 occupancy_percentage = 10;
+  optional int32 occupancy_percentage = 10 [default = -1];
 
   // Carriage specific details, used for vehicles composed of several carriages
   // This message/field is still experimental, and subject to change. It may be formally adopted in the future.


skinkie:
Could you elaborate why this optional value needs an initialisation? If I would parse this feed without the value present, it would be just "not available".

barbeau:
@skinkie Could you read https://github.com/google/transit/issues/247 and let me know if you still have questions? I explained it there.

skinkie:
> @skinkie Could you read #247 and let me know if you still have questions? I explained it there.

I have only active experience with C, Java and Python parsing, both obviously all support NULL, null and None, but that is not my point here. When I want to check for a stopTimeUpdate, I first have to ensure stopTimeUpdate.hasDeparture() for which languages are optional values in GTFS-RT not supporting this hasAttribute method?

barbeau:
>I first have to ensure stopTimeUpdate.hasDeparture(). for which languages are optional values in GTFS-RT not supporting this hasAttribute method?

Yes, that's the issue here. We can't guarantee that all protobuf consuming software include methods like `hasX()`, in which case you'd just need to read the value directly. And if you read the value directly for an optional value that doesn't have an explicit default set, you get the "default default" defined in the protobuf docs.

Here's one example for C# - https://github.com/MobilityData/gtfs-realtime-bindings/issues/52.

You can see this actually applies to other optional GTFS-realtime fields as well that we could fix, including bearing, speed, and vehicle stop status.

barbeau:
I should also mention that this change will help developers new to protobufs that don't know they need to call `hasX()` before reading the value.

skinkie:
> > I first have to ensure stopTimeUpdate.hasDeparture(). for which languages are optional values in GTFS-RT not supporting this hasAttribute method?
> 
> Yes, that's the issue here. We can't guarantee that all protobuf consuming software include methods like `hasX()`, in which case you'd just need to read the value directly. And if you read the value directly for an optional value that doesn't have an explicit default set in the .proto, you get the "default default" defined in the protobuf docs.

How is that binding handling a missing Departure or Arrival then technically (kind of the basics when using GTFS-RT)? I honestly think this should not be solved by default values, rather the binding generator for C# should be fixed.

darylweinberg:
Just a note, Capital Metro Austin will begin publishing occupancy_status in about 2 weeks, for our entire bus fleet. We are only choosing to publish CRUSHED_STANDING_ROOM_ONLY, FULL and NOT_ACCEPTING_PASSENGERS. When one of these values is not present, we do not include occupancy_status for that vehicle.

We have worked with our partners to display “Notice: Crowded Vehicle” if any of these three values are present. I will be reaching to other consumers to explain what we are doing and see if they want to provide the data that is available.

Thank you,

Daryl Weinberg
Transit Systems Architect
o: 512-369-6216
|
|

jxeeno:
Yeah, I think I'm on the same page as @skinkie on this one.  Even though it's not explicitly supported by proto2, it's implicitly a requirement for the bindings to support expressing optional values as null-like or have a `hasX()` in order for `delay`/`time` in GTFS-R to be interpreted in any meaningful way.

barbeau:
@skinkie @jxeeno Thanks for the feedback. Could you elaborate on the downside to defining defaults in the gtfs-realtime.proto file? What do you see as the main drawbacks?

gcamp:
I agree that generally there are workaround in the implementation to achieve the same behaviour. However since there's nothing in the spec that mention what is the actual default, we have a vendor that insist that they can not include the `occupancy_status` value if the occupancy is `EMPTY` (the first value of the enum which is the default if no default is specifically defined). 

Adding a default value would clear that ambiguity.

PR #249, opened 2020-09-28
Change type of headway_secs to positive integer

aababilov:
headway_secs=0 does not have a physical meaning. This must be the time between departures for the same stop for a single trip.


aababilov:
Also, Google shows a validation error for feeds that give headway_secs=0.

gcamp (approved):

timMillet (commented):
I agree with the proposition. However, "Positive Integer" is not defined in the Field Types section and should be added for consistency.

aababilov:
You are right, it is missing in the section. However, it is already used for pathways.traversal_time. I will add the definition in a separate pull request.

aababilov:
https://github.com/google/transit/pull/251 for the new types.

PR #250, opened 2020-09-28
Proposal: Update TripUpdate.timestamp definition to reflect prediction generation time

barbeau:
As discussed in https://github.com/google/transit/pull/241#issuecomment-676429064, the current definition of `TripUpdate.timestamp` doesn't seem to fit predictions well.

It's currently:

>Moment at which the vehicle's real-time progress was measured.

As @jxeeno mentioned in https://github.com/google/transit/pull/241#issuecomment-678266898:

>There are producers out (e.g. Sydney Trains) there that are following the spec -- i.e. populating the timestamp field with the last vehicle position reported rather than last prediction. As @tleboulenge alluded to, that's a problem when the AVL source is a transponder-based or track-side signalling based that only reports when a train begins to occupy a section of the track. If a train is unexpectedly held at a danger signal, in the background, delay predictions are still being updated because it's not occupying the next section of track within the timetabled period of time. However, the timestamp isn't being incremented because no new location is provided.

This proposal changes it to be:

>Moment at which the arrival and departure prediction(s), schedule_relationship, etc. in this TripUpdate were generated. This value should only change when the data within this TripUpdate is refreshed. If data is updated but the value remains the same, the timestamp should still be updated to the most recent data generation time. In POSIX time (i.e., the number of seconds since January 1st 1970 00:00:00 UTC).

This change will allow consumers to use the timestamp as a hash of the prediction values for each trip, in that they only need to apply the prediction to their internal data model if the timestamp is different from the timestamp of the predictions from the previous feed message.

I'm curious if any producers or consumers would consider this proposal a breaking change.

Proposal announced on the GTFS-realtime Google Group at https://groups.google.com/g/gtfs-realtime/c/c6ysJtZ4nXk.


skinkie:
Lets say I think it would be sensible to uncouple TripUpdates from VehiclePositions. But to be honest I find the reasoning rather vague. If a train keeps occupying a section, then the next sample that reports the occupation (in order to calculate the delay) would also have an updated timestamp. Your proposal is actually to shift the meaning of timestamp more towards VDV453/SIRI _ResponseTimestamp_ opposed to the current _RecordedAtTime_. That is a big semantic change. Has been evaluated what the downside would be if the "now correct implementations" would not migrate? I guess their data wouldn't be updated anymore right?

IvanVolosyuk:
We use the data to measure end-to-end staleness of the TUs. It would be nice to have the same meaning of the field for both VPs and TUs, so that we compare apples to apples. Hash of predicted values can be produced by... hashing the predicted values. It will be more reliable than using an optional timestamp.

skinkie:
@IvanVolosyuk I would be against mapping VP timetable towards this. In my perspective this must be the value matching the data acquisition.

barbeau:
>We use the data to measure end-to-end staleness of the TUs. It would be nice to have the same meaning of the field for both VPs and TUs, so that we compare apples to apples.

@IvanVolosyuk But if the TU.timestamp is the vehicle update timestamp (same as VP.timestamp), and not the time when the prediction was generated, then the field is just a potential upper-bound for the TU staleness - not the actual TU staleness. Given that a vehicle could move and the prediction not be updated, it's use as an upper-bound for TU staleness with the current definition is even questionable, IMHO.

Also, to me it doesn't seem to make sense for the TU.timestamp and VP.timestamp to have effectively the same definition, as the TU.timestamp field is then redundant. What additional value do you get from this?

>That is a big semantic change. Has been evaluated what the downside would be if the "now correct implementations" would not migrate? I guess their data wouldn't be updated anymore right?

@skinkie Potentially, yes - hence my question to the community if this is considered a breaking change. If so, we could either introduce this as a new field with the new definition, or bump the spec version.

paulswartz:
Does the header timestamp not do what you're looking for? In our current implementation, that's the timestamp at which we generated the predictions (they're made as a group). The timestamp in an individual TU is the timestamp of the vehicle's location. I get that makes them slightly redundant, but the TU feed does not have vehicle updates in it, and vice versa.

barbeau:
>In our current implementation, that's the timestamp at which we generated the predictions (they're made as a group). The timestamp in an individual TU is the timestamp of the vehicle's location.

@paulswartz In MBTA's interpretation of the header timestamp, that would work, but there isn't any guarantee that others interpret or use the field the same way. For example, if predictions aren't made as a group, you'd need timestamps at the individual TU level, not the header (an aside - I'd like to clarify the header timestamp definition too. I took at shot at this a while back in https://github.com/google/transit/pull/55, but there wasn't sufficient interest at that time).

So in MBTA's implementation - is it possible for the header timestamp to change but all the TU timestamps remain the same as the previous feed iteration?

paulswartz:
> So in MBTA's implementation - is it possible for the header timestamp to change but all the TU timestamps remain the same as the previous feed iteration?

@barbeau I think so, but it would likely be an edge-case where a TU which was updated is also filtered out of the output.

IvanVolosyuk:
I am fine with upper bound staleness for the TU.timestamp. Vehicle position is by far the most important signal for us.

> This change will allow consumers to use the timestamp as a hash of the prediction values for each trip, in that they only need to apply the prediction to their internal data model if the timestamp is different from the timestamp of the predictions from the previous feed message.
This looks like micro-optimization to me. If you need a hash of prediction values - hash the prediction values. This is nano-seconds operation. There is no point on relying on an optional timestamp for this purpose and changing the spec. If the consumer decides to do the stateful processing as implied by the suggestion and can track a vehicle across bundles - for the consumer a comparison of old and new predictions even without any hashing can be trivial operation.

barbeau:
Given the above discussion, and the potential for this being a breaking change, I'm going to close this discussion for now. If there is interest in the future to pursue this change we can re-open. If this is something important to you please feel free to comment on this PR.

PR #251, opened 2020-10-02
Define positive and non-null number field types

aababilov:
They are already used in the table descriptions.

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -8,9 +8,10 @@ This document defines the format and structure of the files that comprise a GTFS
 
 1.  [Term Definitions](#term-definitions)
 2.  [Field Types](#field-types)
-3.  [Dataset Files](#dataset-files)
-4.  [File Requirements](#file-requirements)
-5.  [Field Definitions](#field-definitions)
+3.  [Field Signs](#field-signs)
+4.  [Dataset Files](#dataset-files)
+5.  [File Requirements](#file-requirements)
+6.  [Field Definitions](#field-definitions)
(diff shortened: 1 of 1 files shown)


timMillet (approved):
Sounds good to me!

If you want to be exhaustive with this PR, "Float" alone should be also defined as it is used for `levels.level_index` and `pathways.max_slope`.

scmcca:
I'm wondering if we could separate **Field Types** from their factors as to avoid exhaustively defining every factor combination of integer and float (and potentially [currency amount](https://github.com/google/transit/issues/254)). Something like:

### Field Types
* ...
* **Float** - A floating point number.
* **Integer** - An integer.
* ...

### Field Factors
Factors applicable to **Float** or **Integer** field types:

* **Non-null** - Non-zero.
* **Non-negative** - Greater than or equal to zero.
* **Positive** - Greater than zero.

timMillet:
I like the idea!
However, each field type and field factor should have a precise definition (e.g: "Non-zero - Non-zero" is not a precise definition).

scmcca:
@timMillet @aababilov Please see the PR I made (https://github.com/aababilov/transit/pull/1) containing my revised suggestion as outlined in my previous comment. If it does not address the need, feedback for improvement is welcomed!

aababilov:
Updated. PTAL.

timMillet (approved):
All good for me!

scmcca:
@aababilov @timMillet I forgot a detail. As Field Signs is a new heading, I suggest to add it to the TOC. See my PR [here](https://github.com/aababilov/transit/pull/2). Thoughts?

Issue #257, opened 2020-12-22
Can stop_times reference stops with location_type = 3 or 4?


mmorang:
The current documentation for the stop_id field in stop_times.txt says "Referenced locations must be stops, not stations or station entrances."  It seems this has not been updated since location_type 3 (Generic Node) and 4 (Boarding Area) were added.  If I had to guess, I would say it should not be allowed for stop_times to reference generic nodes and boarding areas, but I think this should be clarified and made explicit in the documentation for stop_times.

I cannot make a PR in this repo without extensive bureaucratic hassle from my company, so it would be wonderful if someone else could take on this task.

skinkie:
@mmorang agreed. But I guess it should also be made clear we are mixing 'the infrastructure' with 'the transit network' here. I know (many) producers, including myself who would at least have the ability to schedule on stations, which is not allowed, since only stops are allowed. This leads to the introduction of 'virtual stops' representing the station. If we make clear what things in stop_times can be used, we might want to do so for things such as transfers as well.

timMillet:
@mmorang 
Hi Melinda,
Effectively, stop times cannot reference stops with `stops.location_type=3` or `=4`. I made the PR (https://github.com/google/transit/pull/258) you requested! You will note that I removed the mentions of the other location types -it explicitly tells what _must_ be done only- so that the definition will not need to be changed again if new location types are added in the future. Please review it and tell me if you agree!

@skinkie
Hi Stefan,
I agree there is a need for debating on the ability to schedule on stations, i.e referencing in `stop_times.stop_id` stops with `stops.location_type=1`. I did not include this ability in the PR though, as it would be an extension of the specification and not only a clarification. If you want, I'd suggest opening a dedicated issue, as the title of this current issue could be confusing for other community members.

PR #258, opened 2020-12-22
Clarification of stop_times.stop_id

timMillet:
Following demand in the issue https://github.com/google/transit/issues/257:

This pull request clarifies the definition of the fields `stop_times.stop_id`.
- Explicitly tells that the referenced locations must have `stops.location_type` values of `0` or empty.
- Future-proof phrasing: it references only what must be done.


mmorang:
Thanks, @timMillet. This looks good to me and is consistent with my understanding of the specification.

skinkie:
I am impatient to vote. Go Go Go! (OpenGeo)

timMillet:
@skinkie I'd tend to think that this change is editorial only, and would not require a vote. Do you think this change is worth a vote on your side?

skinkie:
No, I think this just makes things more clear.

timMillet:
Ok, thanks. I'll merge!

PR #259, opened 2021-01-07
Clarification of GTFS Realtime occupancy descriptions

scmcca:
Following the conversation in #212, this pull request aims to disambiguate the intended use of `OccupancyStatus` as describing the nominal status of passenger occupancy levels, and `occupancy_percentage` as describing passenger occupancy levels on a linear scale.

Summary:
- Clarification to `OccupancyStatus` as indicating the nominal status of passenger occupancy; removal of "percentage" language to avoid confusion with `occupancy_percentage`.
- Editorial clarification of occupancy percentage description.
- Added line breaks to **Caution** statements for improved legibility.

Feedback is welcomed.


gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -425,21 +425,21 @@ message VehiclePosition {
   }
   optional CongestionLevel congestion_level = 6;
 
-  // The degree of passenger occupancy of the vehicle or carriage. This field is still
+  // The nominal status of passenger occupancy for the vehicle or carriage. This field is still
```
barbeau, line 428: I'd like to see more explicit language here that individual producers may not publish all of these values, and therefore consumers must NOT assume that the enumerations are on a linear scale - for example, with `EMPTY` representing the minimum value and `NOT_ACCEPTING_PASSENGERS` or `FULL` representing the max value. Also that producers must not use enumeration values that don't correspond to actual real-world vehicle occupancy states - for example, a producer must NOT internally take a % full numeric value and evenly map it across the enumerations.

It could also recommend (via `should`) that consumers only represent this information to end users based on the enumeration label and not attempt to further interpret the enumeration. For example, a consumer shouldn't map a `STANDING_ROOM_ONLY` value to a roughly half-way full UI graphic bar. Instead, they should show the text "Standing room only" to end users.

scmcca: Please see my latest commit where I added these details.

stevenmwhite:
These changes are helpful, particularly with the added details.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -425,21 +425,28 @@ message VehiclePosition {
   }
   optional CongestionLevel congestion_level = 6;
 
-  // The degree of passenger occupancy of the vehicle or carriage. This field is still
-  // experimental, and subject to change. It may be formally adopted in the future.
+  // The nominal status of passenger occupancy for the vehicle or carriage.
+  // Individual producers may not publish all OccupancyStatus values. Therefore, consumers
+  // must not assume that the OccupancyStatus values follow a linear scale.
+  // Consumers should represent OccupancyStatus values as the nominal status indicated 
+  // and intended by the producer. Likewise, producers must use OccupancyStatus values that
```
devadvance, line 432: I assume localization is acceptable here, hence "represent" as the chosen wording?

scmcca: The use of the word "represent" is meant to convey what @barbeau described:
> a consumer shouldn't map a STANDING_ROOM_ONLY value to a roughly half-way full UI graphic bar. Instead, they should show the text "Standing room only" to end users.

Localization, such as translations, is acceptable as long as it does not change the meaning of the enum value.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -425,21 +425,28 @@ message VehiclePosition {
...
+  // correspond to actual vehicle occupancy states, and must not interally map OccupancyStatus 
+  // values to interpretted percentages. For describing passenger occupancy levels on a linear
```
barbeau, line 434:
```suggestion
  // values to interpreted percentages. For describing passenger occupancy levels on a linear 
```

gtfs-realtime/spec/en/reference.md
```diff
@@ -296,19 +298,19 @@ The degree of passenger occupancy for the vehicle or carriage.
...
+| **occupancy_status** | [OccupancyStatus](#enum-occupancystatus) | Optional | One | Occupancy status for this given carriage, in this vehicle. Default is set to `NO_DATA_AVAILABLE`.<br><br>**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future.|
```
barbeau, line 311: The more detailed description added in the .proto file should be added in the reference.md as well. Historically the more detailed description has actually lived in the reference.

scmcca: I have added the .proto description to the reference.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -425,21 +425,28 @@ message VehiclePosition {
   }
   optional CongestionLevel congestion_level = 6;
 
-  // The degree of passenger occupancy of the vehicle or carriage. This field is still
-  // experimental, and subject to change. It may be formally adopted in the future.
+  // The nominal status of passenger occupancy for the vehicle or carriage.
+  // Individual producers may not publish all OccupancyStatus values. Therefore, consumers
+  // must not assume that the OccupancyStatus values follow a linear scale.
+  // Consumers should represent OccupancyStatus values as the nominal status indicated
```
barbeau, line 431: Should "nominal status" be "nominal states" here? Not sure if that's intended or a typo. It works either way but "states" aligns better with the next sentence, IMHO.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -425,21 +425,28 @@ message VehiclePosition {
...
-  // The degree of passenger occupancy of the vehicle or carriage. This field is still
-  // experimental, and subject to change. It may be formally adopted in the future.
+  // The nominal status of passenger occupancy for the vehicle or carriage.
+  // Individual producers may not publish all OccupancyStatus values. Therefore, consumers
+  // must not assume that the OccupancyStatus values follow a linear scale.
+  // Consumers should represent OccupancyStatus values as the nominal status indicated 
+  // and intended by the producer. Likewise, producers must use OccupancyStatus values that
+  // correspond to actual vehicle occupancy states, and must not interally map OccupancyStatus 
+  // values to interpreted percentages. For describing passenger occupancy levels on a linear 
+  // scale, see `VehiclePosition.occupancy_percentage`.
```
barbeau, line 435: Given that this applies to carriage details as well, this should probably be just `occupancy_percentage` instead of `VehiclePosition.occupancy_percentage`.

skinkie:
+1  OpenGeo

abyrd:
It does seem like a good idea to replace the word "percentage" on fields that do not actually contain percentages. However, this PR replaces "percentage" with "allocation", which implies that someone allocated the seats, i.e. someone assigned them to a specific person or usage. The new word seems less clear than the one it replaces.

It also replaces the expression "degree of passenger occupancy" with "nominal state of passenger occupancy". I think I understand the intended meaning here, that this is a "name for the occupancy state" rather than a quantitative measure, but I suspect most readers will perceive it as opaque jargon. In common usage "nominal" means something exists in name only, has characteristics different than those implied by its name, or is very small. In economics it might also mean something like "unadjusted", and in engineering something like "within acceptable range".

scmcca:
@abyrd Good points. 

Perhaps we can replace "allocation" with "number" and "amount" such that the spec reads: 
> The vehicle or carriage has a large ~allocation~ number of seats available. The ~allocation~ amount of free seats out of the total seats available to be considered large enough to fall into this category is determined at the discretion of the producer.

As for "nominal", perhaps we can remove it and simply rely on the word "state" such that the spec reads:
> The ~nominal~ state of passenger occupancy for the vehicle or carriage.

Thoughts?

scmcca:
Hi everyone,

It seems like we've settled on the final proposal. As the proposal concerns an editorial clarification to a historical feature, I believe we can assume implementation from at least one GTFS-realtime producer and consumer. Also, we are well past the required 7-day discussion period as outlined in the specification amendment process. Feel free to voice any remaining feedback.

**Formally, I'm calling for a vote on the official adoption of the changes proposed in this PR. The vote will close on 2021-01-29 at 23:59:59 UTC.**

skinkie:
+1 OpenGeo

gcamp:
+1 Transit

stevenmwhite:
+1 GMV Sync

scmcca:
The vote ended on 2021-01-29 23:59:59 UTC with a unanimous consensus of 3 yes votes from OpenGeo (@skinkie), Transit (@gcamp), and GMV Sync (@stevenmwhite). As per the specification amendment process, this proposal is accepted! Thanks to everyone involved.

PR #260, opened 2021-02-09
Add departure_occupancy_status to TripUpdate

jakehoare:
The idea behind this proposal is that for the user, the occupancy of a vehicle several stops away (supplied in VehiclePosition) may be difficult to act upon, because occupancy will change by the time the vehicle arrives at the stop where the user intends to board.

This new field allows agencies to provide expected occupancy for future stops.
It is left open for agencies to decide how to calculate these occupancies, they may use any model provided it uses realtime data.

The natural place in GTFS for these occupancies is in TripUpdate.StopTimeUpdate - they are analogous to StopTimeEvents.

If #248 is adopted, that default would apply here as well.

Thoughts, comments?


skinkie:
When I read the topic I thought this ment: the crowdedness at the stop. I think it would be good to change the wording to prevent this ambiguity.

jakehoare:
I used StopOccupancy, since it's analogous to StopTimeUpdate - except there might not be a previous value so it isn't an update.
But a clarification would be fine.
e.g.. TripStopOccupancy? or OccupancyOnTrip?

mpaine-act:
How will rider know if the prediction for the occupancy at each trip stop is arrival or departure based?  Shouldn't there be a separate occupancy prediction for both arrival and departure StopTimeEvent?

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -364,6 +364,24 @@ message TripUpdate {
...
+    // Must be the same as in stop_times.txt in the corresponding GTFS feed.
+    optional uint32 stop_sequence = 1;
+    // Must be the same as in stops.txt in the corresponding GTFS feed.
+    optional string stop_id = 2;
+
+    // Expected occupancy after departure from the given stop.
+    optional VehiclePosition.OccupancyStatus occupancy_status = 3;
+  }
+
+  optional StopOccupancy stop_occupancy = 7;
```
barbeau, line 384: >However then it's necessary to specify time updates in order to specify occupancy updates - this proposal makes times and occupancies independent.

There is precedent for specifying other stop attributes within a `StopTimeUpdate` without specifying predictions. https://github.com/google/transit/pull/219 introduced dynamic stop assignments, which says:

 "To assign a stop without providing any real-time arrival or departure predictions, populate stop_time_properties.assigned_stop_id and set StopTimeUpdate.schedule_relationship = NO_DATA."

We could do something similar here if we want to include the new data within StopTimeUpdate.

I'd be more inclined to bundle the data within StopTimeUpdate and not have a separate array here. Having another array of values that have to be matched to stops and can be keyed on stop_id or stop_sequence seems like it's inviting trouble, as that's already a point of error in feeds. If this stays as a separate array, I'd strongly recommend that it be keyed only on stop_sequence and omit stop_id.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -364,6 +364,24 @@ message TripUpdate {
     extensions 9000 to 9999;
   }
   optional TripProperties trip_properties = 6;
+  
+  // Realtime occupancy immediately after departure for a given stop on a
```
barbeau, line 368: Is there a reason to allow data for past events?

I'd be inclined to have stronger language that requires this to be a predicted future value immediately after departure for the given stop. One sticking point for predictions currently, especially surrounding TripUpdate.delay, is that it's not clear if the value is observed fact or a prediction about the future. I'd prefer to have stricter guidance here to know exactly what we're seeing in feeds. It seems to me if you just want real-time or past info, you should be looking at VehiclePosition.occupancy_status. This new field should include some predictive intelligence that presumably adds value over the data in VehiclePosition.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -364,6 +364,24 @@ message TripUpdate {
...
+    // The update is linked to a specific stop either through stop_sequence or
+    // stop_id, so one of the fields below must necessarily be set.
+
+    // Must be the same as in stop_times.txt in the corresponding GTFS feed.
+    optional uint32 stop_sequence = 1;
+    // Must be the same as in stops.txt in the corresponding GTFS feed.
+    optional string stop_id = 2;
+
+    // Expected occupancy after departure from the given stop.
+    optional VehiclePosition.OccupancyStatus occupancy_status = 3;
```
barbeau, line 381: tl;dr - There is a lot of discussion if OccupancyStatus is the right way forward for occupancy. See https://github.com/google/transit/issues/223 for discussion vs occupancy_percentage and an older proposal to revise OccupancyStatus at https://github.com/google/transit/pull/212. There is also a newer draft proposal at https://github.com/TransitApp/transit/pull/3 to define a fixed linear scale for occupancy, which could deprecate OccupancyStatus.

All or some of these (OccupancyStatus, occupancy_percentage, CrowdLevel) could be used here.

barbeau (commented):
Thanks @jakehoare! Some of my thoughts are in-line below.

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: Also, this would need to change from `optional` to `repeated` if it stays as a separate array.

jakehoare:
> How will rider know if the prediction for the occupancy at each trip stop is arrival or departure based? Shouldn't there be a separate occupancy prediction for both arrival and departure StopTimeEvent (or put the occupancy prediction inside StopTimeEvent).

If occupancy on departure from stop n is equal to occupancy on arrival at stop n+1 (i.e. assume nobody get on or off between stops) then clients of the data can use occupancy on departure as occupancy on arrival at the next stop. So 2 fields would repeat the same data, except shifted by 1 stop.
It's up to clients of the data to convey to users if they are showing departure or arrival occupancy.

gtfs-realtime/proto/gtfs-realtime.proto
jakehoare: That's fine, the intention here is to use the standard unit of occupancy that applies to the rest of GTFS, whatever that may change to in the future.

gtfs-realtime/proto/gtfs-realtime.proto
jakehoare: The reason was only consistency with other existing fields of TripUpdate. But it would be fine to change to "future only", as you mention.
That would imply consistency with VehiclePosition.current_stop_sequence, or allow current_stop_sequence to be inferred if it isn't set.

gtfs-realtime/proto/gtfs-realtime.proto
jakehoare: Thanks. That makes sense to me. I suppose the negatives are having to provide empty `StopTimeUpdate`s in order to provide occupancy predictions (which is probably a niche case), and the name `StopTimeUpdate` possibly implying that it's about time and not occupancy.
They are not show-stoppers to me. Particularly if there is precedent, then a separate repeated looks good. I'll propose some changes after others have had a chance to comment on this and the other points under discussion.

jakehoare:
I have revised this in line with comments above - mainly to put the new occupancy inside StopTimeUpdate (i.e. using the same stop_sequence or stop_id), instead of a separate message.

Note that ScheduleRelationship.NO_DATA could be misleading in that times cannot be provided but occupancies could.

mike-swiftly:
Sorry I'm a bit late to the discussion as Jake has called for a vote on this PR. But there is a seeming discrepancy. The departure_occupancy_status is marked as optional in the proto file but as Required in the documentation. Is the documentation incorrect?

barbeau (commented):
@jakehoare Some comments in-line. 

For those not following the Google Group, the vote was called at https://groups.google.com/forum/#!topic/gtfs-realtime/GQFcNL8Mlds (in the future please call it here too). There are going to need to be some updates to the PR to clarify exactly what's being proposed, so I'd suggest re-starting the vote after everything is set.

Also, please clarify if this is a vote for adoption as an RT experimental field (i.e., we don't currently have any producers or consumers) or official field (requires producers and consumers). See https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#experimental-fields.

gtfs-realtime/spec/en/reference.md
```diff
@@ -186,6 +187,7 @@ The update is linked to a specific stop either through stop_sequence or stop_id,
...
 | **arrival** | [StopTimeEvent](#message-stoptimeevent) | Conditionally required | One | If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED.  If schedule_relationship is NO_DATA, arrival and departure must be empty. |
 | **departure** | [StopTimeEvent](#message-stoptimeevent) | Conditionally required | One | If schedule_relationship is empty or SCHEDULED, either arrival or departure must be provided within a StopTimeUpdate - both fields cannot be empty. arrival and departure may both be empty when schedule_relationship is SKIPPED.  If schedule_relationship is NO_DATA, arrival and departure must be empty. |
+| **departure_occupancy_status** | [OccupancyStatus](#enum-occupancystatus) | Required | One | The state of passenger occupancy for the vehicle immediately after departure from the given stop. |
```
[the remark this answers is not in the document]
mike-swiftly: I see CrowdLevel vs OccupancyStatus as a separate issue that should be resolved by #261. I think we can move ahead with predicted occupancy using OccupancyStatus at this time and then add CrowdLevel once #261 is hashed out. I don't see us removing OccupancyStatus as part of #261 so I think it would be acceptable to also have two values for predicted occupancy. At that time I would also like to see occupancy_percentage be discussed as a possible part of predicted occupancy since the enumerated types are likely too limited to be of universal use.

gtfs-realtime/spec/en/reference.md
```diff
@@ -149,10 +149,11 @@ Note that the update can describe a trip that has already completed.To this end,
...
 | **trip_properties** | [TripProperties](#message-tripproperties) | Optional | One | Provides the updated properties for the trip. <br><br>**Caution:** this message is still **experimental**, and subject to change. It may be formally adopted in the future. |
+| **stop_occupancy** | [StopOccupancy](#message-stopoccupancy) | Optional | Many | Expected OccupancyStatus for the trip immediately after departure from a stop (both future, i.e., predictions, and in some cases, past ones, i.e., those that already happened). The stop_occupancies must be sorted by stop_sequence, and apply for all the following stops of the trip up to the next specified stop_occupancy. <br><br>**Caution:** this message is still **experimental**, and subject to change. It may be formally adopted in the future.|
```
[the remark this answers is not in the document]
jakehoare: Thanks for pointing out. Done.

gtfs-realtime/spec/en/reference.md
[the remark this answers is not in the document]
jakehoare: Amended for the first 3 comments above.

For `CrowdLevel` vs `OccupancyStatus`, I'd prefer to separate that out from this and use `OccupancyStatus` here since that's currently used in `VehiclePosition`, then update this when/if VP changes.

mike-swiftly:
+1 

I think the changes to the documentation are good and that other issues. like departure_crowd_level, should be dealt with separately.

jakehoare:
Also note that this is for adoption as an experimental field.

Vote will be closed on Friday March 26th at 23:59:59 UTC.

You can vote with a +1 (in favor) or -1 (against) in the comments here.

Thanks
Jake

skinkie:
+1

RTL-Gilles-Ferland:
+1

With departure_crowd_level being added eventually.

juanborre:
+1 from Transit 👏

Bertware:
+1 from Trafiklab/Samtrafiken

sam-hickey-ibigroup:
+1 from IBI Group

And we agree with the comment above from @jakehoare that “the intention here is to use the standard unit of occupancy that applies to the rest of GTFS, whatever that may change to in the future”

gtfs-realtime/spec/en/reference.md
```diff
@@ -199,7 +200,7 @@ The relation between this StopTime and the static schedule.
...
-| **NO_DATA** | No data is given for this stop. It indicates that there is no realtime information available. When set NO_DATA is propagated through subsequent stops so this is the recommended way of specifying from which stop you do not have realtime information. When NO_DATA is set neither arrival nor departure should be supplied. |
+| **NO_DATA** | No data is given for this stop. It indicates that there is no realtime information available. When set NO_DATA is propagated through subsequent stops so this is the recommended way of specifying from which stop you do not have realtime timing information. When NO_DATA is set neither arrival nor departure should be supplied. |
```
paulswartz, line 203: Should the first "realtime" here be "realtime timing"?

jakehoare: Yes true, amended.

jakehoare:
Thank you all for participating! The vote ended on Friday March 26th at 23:59:59 UTC with a unanimous consensus of 6 yes votes from OpenGeo (@skinkie), Swiftly (@mike-swiftly), Transit (@juanborre), RTL (@
RTL-Gilles-Ferland), Trafiklab/Samtrafiken (@Bertware), and IBI Group (@sam-hickey-ibigroup).

barbeau:
@jakehoare Looks like there are conflicts with the master branch - could you please resolve those, and then I can merge?

jakehoare:
@googlebot I fixed it.

jakehoare:
@barbeau Ah yes, resolved conflicts now.

PR #261, opened 2021-02-10
Add CrowdLevel to VehiclePosition

scmcca:
Hi everyone,

This pull request contains a proposal for CrowdLevel continuing from #212 and is based off of stakeholder interest in #223 for a description of occupancy levels in GTFS Realtime that are expressed as enum values on a simple ordinal scale (i.e., `NOT_CROWDED`, `SOME_CROWDING`, `CROWDED`).

This is in contrast to OccupancyStatus which has been clarified in #259 to describe occupancies by states (i.e., `MANY_SEATS_AVAILABLE`, `STANDING_ROOM_ONLY`, etc.) that cannot necessarily be attributed to a unversal linear scale (as seen in a few practices), as well as `occupancy_percentage` which describes occupancy ratio at a fine-grained level.

Note that this pull request is for the experimental adoption of CrowdLevel. Feedback is welcomed!

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -481,7 +481,7 @@ message VehiclePosition {
   // If multi_carriage_details is populated with per-carriage CrowdLevel,
   // then this field should describe the entire vehicle with all carriages accepting passengers considered.
   enum CrowdLevel {
-    // The vehicle of carriage does not have any crowd level data available at that time.
+    // The vehicle or carriage does not have any crowd level data available at that time.
     NO_DATA_AVAILABLE = 0;
 
     // The vehicle or carriage is not considered crowded by most measures, having few or no passengers onboard.
@@ -521,15 +521,11 @@ message VehiclePosition {
     // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
     optional OccupancyStatus occupancy_status = 3 [default = NO_DATA_AVAILABLE];
 
-    // The passenger crowd level for this given carriage, in this vehicle.
-    // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
-    optional CrowdLevel crowd_level = 4 [default = NO_DATA_AVAILABLE];
-
     // Occupancy percentage for this given carriage, in this vehicle.
     // Follows the same rules as "VehiclePosition.occupancy_percentage"
     // -1 in case data is not available for this given carriage (as protobuf defaults to 0 otherwise)
     // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
-    optional int32 occupancy_percentage = 5 [default = -1];
+    optional int32 occupancy_percentage = 4 [default = -1];
 
     // Identifies the order of this carriage with respect to the other
     // carriages in the vehicle's list of CarriageDetails.
@@ -542,7 +538,11 @@ message VehiclePosition {
     // Carriages without data must be represented with a valid carriage_sequence number and the fields 
     // without data should be omitted (alternately, those fields could also be included and set to the "no data" values).
     // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
-    optional uint32 carriage_sequence = 6;
+    optional uint32 carriage_sequence = 5;
+
+    // The passenger crowd level for this given carriage, in this vehicle.
+    // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
+    optional CrowdLevel crowd_level = 6 [default = NO_DATA_AVAILABLE];
 
     // The extensions namespace allows 3rd-party developers to extend the
     // GTFS Realtime Specification in order to add and evaluate new features and
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -486,25 +486,6 @@ message VehiclePosition {
   // then this field should describe the entire vehicle with all carriages accepting passengers considered.
   // This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional uint32 occupancy_percentage = 10;
-  
-  // The passenger crowd level for the vehicle or carriage.
-  // This field is still experimental, and subject to change. It may be formally adopted in the future.
-  // If multi_carriage_details is populated with per-carriage CrowdLevel,
-  // then this field should describe the entire vehicle with all carriages accepting passengers considered.
-  enum CrowdLevel {
-    // The vehicle or carriage does not have any crowd level data available at that time.
-    NO_DATA = 0;
-
-    // The vehicle or carriage is not considered crowded by most measures, having few or no passengers onboard.
-    NOT_CROWDED = 1;
-
-    // The vehicle or carriage has some crowding by most measures, having some passengers onboard.
-    SOME_CROWDING = 2;
-
-    // The vehicle or carriage is considered crowded by most measures, having many passengers onboard and limited space.
-    CROWDED = 3;
-  }
-  optional CrowdLevel crowd_level = 11;
 
   // Carriage specific details, used for vehicles composed of several carriages
   // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
@@ -564,6 +545,25 @@ message VehiclePosition {
   // information to passengers about where to stand on a platform.
   // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
   repeated CarriageDetails multi_carriage_details = 11;
+
+  // The passenger crowd level for the vehicle or carriage.
+  // This field is still experimental, and subject to change. It may be formally adopted in the future.
+  // If multi_carriage_details is populated with per-carriage CrowdLevel,
+  // then this field should describe the entire vehicle with all carriages accepting passengers considered.
+  enum CrowdLevel {
+    // The vehicle or carriage does not have any crowd level data available at that time.
+    NO_DATA = 0;
+
+    // The vehicle or carriage is not considered crowded by most measures, having few or no passengers onboard.
+    NOT_CROWDED = 1;
+
+    // The vehicle or carriage has some crowding by most measures, having some passengers onboard.
+    SOME_CROWDING = 2;
+
+    // The vehicle or carriage is considered crowded by most measures, having many passengers onboard and limited space.
+    CROWDED = 3;
+  }
+  optional CrowdLevel crowd_level = 12;
 
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime Specification in order to add and evaluate new features and
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -464,7 +464,7 @@ message VehiclePosition {
     // The vehicle or carriage is not accepting passengers, but usually accepts passengers for boarding.
     NOT_ACCEPTING_PASSENGERS = 6;
 
-    // The vehicle or carriage doesn't have any occupancy data available at that time.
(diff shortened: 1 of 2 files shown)


gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -542,6 +546,25 @@ message VehiclePosition {
   // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
   repeated CarriageDetails multi_carriage_details = 11;
 
+  // The passenger crowd level for the vehicle or carriage.
+  // This field is still experimental, and subject to change. It may be formally adopted in the future.
+  // If multi_carriage_details is populated with per-carriage CrowdLevel,
+  // then this field should describe the entire vehicle with all carriages accepting passengers considered.
+  enum CrowdLevel {
+    // The vehicle or carriage does not have any crowd level data available at that time.
+    NO_DATA = 0;
```
barbeau, line 555: I was going to say let's use `NO_DATA_AVAILABLE` to match the other occupancy "no data" values. But we do have `ScheduleRelationship.NO_DATA`, so `OccupancyStatus` probably should have used this. So we could stick with `NO_DATA` here, especially if `OccupancyStatus` ends up being deprecated.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -542,6 +546,25 @@ message VehiclePosition {
...
+    // The vehicle or carriage is not considered crowded by most measures, having few or no passengers onboard.
+    NOT_CROWDED = 1;
+
+    // The vehicle or carriage has some crowding by most measures, having some passengers onboard.
+    SOME_CROWDING = 2;
+
+    // The vehicle or carriage is considered crowded by most measures, having many passengers onboard and limited space.
+    CROWDED = 3;
+  }
+  optional CrowdLevel crowd_level = 12;
```
barbeau, line 566: This should have the default of no data also

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -542,6 +546,25 @@ message VehiclePosition {
   // This message/field is still experimental, and subject to change. It may be formally adopted in the future.
   repeated CarriageDetails multi_carriage_details = 11;
 
+  // The passenger crowd level for the vehicle or carriage.
+  // This field is still experimental, and subject to change. It may be formally adopted in the future.
+  // If multi_carriage_details is populated with per-carriage CrowdLevel,
+  // then this field should describe the entire vehicle with all carriages accepting passengers considered.
+  enum CrowdLevel {
+    // The vehicle or carriage does not have any crowd level data available at that time.
```
barbeau, line 554: Nit - this comment is a mix between present and past tense. Looks like the others are in present tense, so this should probably be:

`// The vehicle or carriage does not have any crowd level data available at this time.`

gtfs-realtime/spec/en/reference.md
```diff
@@ -297,6 +303,22 @@ For describing passenger occupancy levels on a linear scale, see `occupancy_perc
...
+
+To describe in-vehicle occupancy states, see [OccupancyStatus](#enum-occupancystatus). 
+
+Both OccupancyStatus and CrowdLevel may be used side-by-side, but must respectively indicate true conditions. For example, a vehicle may have _**FEW_SEATS_AVAILABLE**_ and _**SOME_CROWDING**_, have _**STANDING_ROOM_ONLY**_ and be _**CROWDED**_, or is _**NOT_ACCEPTING_PASSENGERS**_ and is _**CROWDED**_.
+
+**Caution:** this field is still **experimental**, and subject to change. It may be formally adopted in the future.
+
+| _**Value**_ | _**Comment**_ |
+|-------------|---------------|
+| _**NOT_CROWDED**_ | _The vehicle or carriage is not considered crowded by most measures, having few or no passengers onboard, or as otherwise defined by current operating regulations._ |
```
barbeau, line 318: Only the header text should be italicized, not the table body, although I see now it was done that way for OccupancyStatus. Both should be normal font in the "Value" and "Comment" section to match the other enum tables.

barbeau (commented):
Thanks @scmcca! Mostly LGTM, some comments in-line

gtfs-realtime/proto/gtfs-realtime.proto
scmcca: FWIW I had originally used `NO_DATA_AVAILABLE` to match that in `OccupancyStatus` but the validator rejected it since `NO_DATA_AVAILABLE` has to be unique within VehiclePosition, not just CrowdLevel.

gtfs-realtime/proto/gtfs-realtime.proto
scmcca: I copied the style of this comment from that in `OccupancyStatus`. I agree that it does not read well, but if we change it here it should be edited in `OccupancyStatus` as well.

gtfs-realtime/spec/en/reference.md
scmcca: As above, I copied the style precedent set by `OccupanyStatus`. If we change one we should change both, which I think is a good idea! It could be worth a separate project to clean up the GTFS Realtime reference style more globally.

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: Ah, right, I forgot about that 👍

bbafford:
Clever Devices would like to understand if the intention is to expand GTFS-RT to include OccupancyStatus, OccupancyPercentage, and CrowdLevel. It is not clear how these three should co-exist, why such similar fields are being proposed, and the differences between OccupancyStatus and CrowdLevel. If CrowdLevel is intended to fully replace OccupancyStatus, than we believe the proposal in #261 is satisfactory. 

Our contributions to #223 advocating a clear, coherent, and not overly-precise measure of real-time occupancy were not meant as a request for yet another field. This was our perspective on the best way for GTFS-RT to communicate crowding to riders. We believe that an additional field complicates this. We are unlikely to become produces of CrowdLevel as we already extended support of OccupancyStatus to satisfy the pressing need to communicate passenger crowding.  We map crowding levels to three enumerations allowed by OccupancyStatus that have a clear hierarchy (Empty, Few Seats Available, Full).

stevenmwhite:
> We map crowding levels to three enumerations allowed by OccupancyStatus that have a clear hierarchy (Empty, Few Seats Available, Full).

@bbafford We do something similar, but the fact that we only use three and that we believe them to have a clear scale isn't actually clear from the spec, which specifically addresses the type of passenger spaces available (standing and seated).

My understanding is that `CrowdLevel` is being proposed as a way to communicate a clear scale and hierarchy of passenger load -- and that essentially the way you (and we) currently map crowding data from vehicles should be transitioned to use the new scale defined here.

Then, the documentation for the existing `OccupancyStatus` will be clarified to ensure that producers and consumers understand it to mean something qualitative about the types of passenger space available (seats vs standing). I expect most producers don't actually have this specific information and over time would switch to using the `CrowdLevel` instead -- but theoretically for those who do have access to seat-level data, there is still valuable information that can be communicated by saying whether seats are available or not.

RTL-Gilles-Ferland:
Hello,

We also think that 2 enums is not necessary. The `CrowdLevel` one is more interesting and we do find this proposal satisfactory.

Will this new `CrowdLevel` be integrated to the `Occupancy.txt` [(#240)](https://github.com/google/transit/pull/240) file with the objective of replacing the `OccupancyStatus` as well?

As a producer we will probably have to include both enum values for some time until all consumers have migrated from `OccupancyStatus` to `CrowdLevel`.

Thanks

scmcca:
@bbafford I'd like to second @stevenmwhite's comment. You can find a complete account of the discussions that led to this proposal at #223 and #212. 

To summarize, this proposal was initially made in #212 as a replacement of OccupancyStatus. However, it became clear that some stakeholders were still using the current enum values of OccupancyStatus to communicate qualitative vehicle occupancy states such as "seating available" or "standing only". So forcing a breaking change by deprecating some enum values, and changing the mappings of OccupancyStatus enum values onto an ordinal scale was not likely to be a successful strategy if put to a vote. Moreover, the description of OccupancyStatus has been recently clarified in #259 to discourage interpretive mappings of qualitative OccupancyStatus values onto a linear scale.

>If CrowdLevel is intended to fully replace OccupancyStatus, than we believe the proposal in #261 is satisfactory.

If the community migrates from OccupancyStatus to CrowdLevel (there is some momentum for that), then deprecating OccupancyStatus would be a serious consideration for the mid- to long-term, but not likely in the short-term as tried with #212.

@RTL-Gilles-Ferland The occupancy description that will be integrated into `occupancies.txt` (#240) will be the one that the community votes in as official. In fact, the reality that there is no official way to describe occupancies in GTFS is what is blocking #240 to move forward.  I've made this proposal for CrowdLevel experimental for now to ease adoption for the mid-term.  

> As a producer we will probably have to include both enum values for some time until all consumers have migrated from OccupancyStatus to CrowdLevel.

Depending on the producers who intake your feeds, this may be a good strategy in the interim.

@bbafford @RTL-Gilles-Ferland The three occupancy descriptions co-exist by each indicating respectively true conditions. I left a note for the interaction of OccupancyStatus and CrowdLevel in the proposal: 

> Both OccupancyStatus and CrowdLevel may be used side-by-side, but must respectively indicate true conditions. For example, a vehicle may have **FEW_SEATS_AVAILABLE** and **SOME_CROWDING**, have **STANDING_ROOM_ONLY** and be **CROWDED**, or is **NOT_ACCEPTING_PASSENGERS** and is **CROWDED**.

The idea is that since are each distinct in describing occupancies (qualitative states, ordinal level, fine-grained ratio), and results may differ by vehicle type or context, that all _can_ be provided but must ultimately hold up to their own statements. At the same time, all occupancy descriptions are optional. Producers can provide whichever one(s) they want.

RTL-Gilles-Ferland:
> The idea is that since are each distinct in describing occupancies (qualitative states, ordinal level, fine-grained ratio), and results may differ by vehicle type or context, that all can be provided but must ultimately hold up to their own statements. At the same time, all occupancy descriptions are optional. Producers can provide whichever one(s) they want.

Understood, but are we making things more complicated than they need to be? What's the initial business requirement? The requirement is to inform the client on the potential crowding of the next departure. It seems to us that one simple ordinal enum successfully answers that requirement.

We feel that by having 3 different optional values and giving freedom to producers we are complicating things for consumers. Consumers (like mobile app providers) will end up having to support multiple combinations of these 3 optional values to get the same end result for our clients. It seems like we are failing at having a "standard" by doing this.

jamespfennell:
Yes unfortunately proposals like this are always biased to the benefit of producers. Producers have the luxury of picking and choosing which of the N (in this case 3) methods they want to use to transmit data; meanwhile, consumers are forced to implement each of the N methods.

Of course we need to advance the spec, but perhaps somewhat redundant additions should be tied to deprecations of other fields?

Also, the GTFS spec (both static and realtime) does support extensions. Should high specialized features like this, which will likely see little use, just be implemented as extensions?

paulswartz:
@mbta is interested in implementing this. We would add this to our existing implementations of the larger `OccupancyStatus` and `occupancy_percentage` fields.

harshverma2000:
Thanks @scmcca - we need a way for the operator to change their occupancy indicators if they desire some time and to maintain flexibility for the designer, imho, the static text files could be the best place and simplest way to keep that info.

barbeau:
This proposal was closed due to inactivity, but I wanted to add this comment in case it resumes.

It appears that VDV (Germany) is standardizing on a 3-level occupancy enumeration - "low", "moderate", "high":
https://github.com/VDVde/TRIAS/pull/14

This is similar to what is proposed in this PR.

PR #263, opened 2021-03-08
Clarification of TripUpdate.timestamp

ericouyang:
In scenarios where StopTimeUpdates are specified for events in the past, there may be multiple "Moment[s] at which the vehicle's real-time progress was measured" as a part of a TripUpdate when there are StopTimeUpdates for events in the past. This proposal explicitly specifies that the appropriate timestamp to use is "The most recent moment at which the vehicle's real-time progress was measured to estimate StopTimes in the future."

One scenario where this ambiguity has created challenges is when a TripUpdates consumer relies on this timestamp to understand how old a particular prediction is for the purposes of better understanding the accuracy of predictions.

StopTimeUpdates in the past are important for informing passengers about recently departed trips, especially when a vehicle is running early, as described in [the TripUpdates spec](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/trip-updates.md#stop-time-updates
):

> Producers should not drop a past StopTimeUpdate if it refers to a stop with a scheduled arrival time in the future for the given trip (i.e. the vehicle has passed the stop ahead of schedule), as otherwise it will be concluded that there is no update for this stop.

As such, additional language has also been added in-line to make this behavior explicit: "When StopTimes in the past are provided, arrival/departure times may be earlier than this value."


ericouyang:
I've called for a vote on this clarification for adoption on the Google Group (https://groups.google.com/g/gtfs-realtime/c/6wo_0jC6lQ0/m/ZyjO6DwoBQAJ). 

Please vote with a +1 (in favor) or -1 (against) before Tuesday, March 30th at 23:59:59 UTC. Thanks!

skinkie:
+1

brodyFlannigan:
**+1** for Brody Flannigan Transit Data.

gcamp:
+1 Transit

paulswartz:
+1 MBTA

harringtonp:
+1 YourStop

sam-hickey-ibigroup:
+1 IBI Group

ericouyang:
Thank you all for participating! The vote ended on Tuesday, March 30th at 23:59:59 UTC with a unanimous consensus of 6 yes votes from OpenGeo (@skinkie), Brody Flannigan Transit Data (@brodyFlannigan), Transit (@gcamp), MBTA (@paulswartz), YourStop (@harringtonp), and IBI Group (@sam-hickey-ibigroup).

PR #265, opened 2021-03-23
Proposal: Add MODIFIED and pickup and drop-off types to GTFS-RT; experimentally deprecate SKIPPED

ericouyang:
### Background
To remedy operational challenges, transit providers dynamically change the level of service provided by a vehicle for a given trip. For example, for a vehicle that is running late, a dispatcher may instruct a vehicle to go to "Drop Off Only" to regain time when running late or address a gap in service.

Today, in GTFS-RT, there is only the ability to indicate a stop as entirely [SKIPPED](https://gtfs.org/reference/realtime/v2/#enum-schedulerelationship). However, there is not a way to indicate that a stop is going to be partially served, such as stopping there upon request to let a passenger disembark and not picking up additional passengers in the scenario above.

### Proposal 
This proposal adds support for a more detailed relationship of a vehicle to a scheduled stop beyond skipping it completely and brings GTFS-realtime into consistency with GTFS-static's [stop_times.txt](http://gtfs.org/reference/static#stop_timestxt). It ​builds on a recent addition (#219) to support real-time changes using `StopTimeProperties` and adds `pickup_type` and `drop_off_type` as experimental fields. GTFS static already supports specifying `pickup_type` and `drop_off_type` as a part of [stop_times.txt](https://gtfs.org/reference/static/#stop_timestxt). Those existing definitions are leveraged and referenced as a part of this change. 

To avoid an ambiguous state where the currently supported `SKIPPED` scenario could also be alternatively represented as `drop_off_type=NO_DROP_OFF` and `pickup_type=NO_PICKUP` using `StopTimeProperties`, this proposal experimentally deprecates `SKIPPED` in favor of that more explicit representation. When `StopTimeProperties` are modified in a way that impacts the rider experience, now a new enumeration, `MODIFIED`, will be used instead. 

Both the gtfs-realtime.proto file and documentation have been updated accordingly. Additionally, like for `DUPLICATED`, a migration guide has been provided to detail a plan for a smooth transition for existing producers and consumers.

This pull request is a subset of and based on the GTFS-ServiceChanges v3.1 spec:

This proposal builds on prior art from @lionel-nj and @barbeau (see MobilityData/transit#47)


skinkie:
This is a disaster waiting to happen. Either we allow for the exchange of fully replaceable routes including the network, or we disallow network changes, but this balance on what sort of could work for some "big players" does not make any sense.

barbeau:
Also note that `stop_time_properties.pickup_type = NO_PICKUP` is effectively the same as the existing `vehicle.occupancy_status = NOT_ACCEPTING_PASSENGERS`. Should we deprecate `OccupancyStatus.NOT_ACCEPTING_PASSENGERS` as part of this proposal as well? Or do people view them as valuable individually (in a VP vs. TU feed)?

ericouyang:
@skinkie - Could you elaborate a bit more in terms of the risks you're seeing with a piecemeal approach?

While I definitely agree that it would be great to implement more of the proposal in GTFS-ServiceChanges V3.1 to support complex route changes, I believe that there's still value in tackling it piece-by-piece (like in the scenario described above). This PR follows the precedent set by #219 and #221.

ericouyang:
> Should we deprecate `OccupancyStatus.NOT_ACCEPTING_PASSENGERS` as part of this proposal as well? Or do people view them as valuable individually (in a VP vs. TU feed)?

@barbeau - Good question! Since one is describing the vehicle in real-time vs. the other describing a trips' relationship with upcoming stops, I view them as distinct but related. I could see value in terms of adding an inline clarification about how they can be used together, such as:

> If NO_PICKUP is set because the vehicle is at capacity and no longer accepting additional boardings, vehicle.occupancy_status = NOT_ACCEPTING_PASSENGERS should also be set.

There's also the issue of determining how many stops to set `NO_PICKUP` for while a vehicle is `NOT_ACCEPTING_PASSENGERS`. Would it make sense to set it for all upcoming stops until the vehicle has an occupancy status that supports boarding again?

On a practical level, I also wonder how often `NOT_ACCEPTING_PASSENGERS` is used by feed creators, how it's displayed in passenger apps, and if it has any ramifications on the display of predictions, like the way that `Alert.effect = NO_SERVICE` will cause predictions to be suppressed. If it's not actively used, no one has plans to implement it and/or there's no special treatment that distinguishes it from `FULL`, that could be a reason to remove it in favor of an alternative approach.

skinkie:
> @skinkie - Could you elaborate a bit more in terms of the risks you're seeing with a piecemeal approach?

Because this is a moving target from a software architecture perspective. If the target is replacing full trips: standardize that.

It is impossible to keep up to date with partial implementations, unless _you you already support all that_, and are waiting for a full implementation that is being stopped by incremental implementations.


> While I definitely agree that it would be great to implement more of the proposal in GTFS-ServiceChanges V3.1 to support complex route changes, I believe that there's still value in tackling it piece-by-piece (like in the scenario described above). This PR follows the precedent set by #219 and #221.

Does not mean that the precedent has resulted in global support for it.

ericouyang:
> If the target is replacing full trips: standardize that.

@skinkie Thanks for your thoughtful response. At this time, our target is not replacing full trips or being able to create arbitrary ad-hoc trips on the fly (DUPLICATED, as previously adopted, is in scope). As noted in the original description, our intent at this time is to support on-the-fly changes to partially cancel service, such as to have a vehicle go drop-off only.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -54,7 +54,8 @@ message FeedMessage {
 // Metadata about a feed, included in feed messages.
 message FeedHeader {
   // Version of the feed specification.
-  // The current version is 2.0.  Valid versions are "2.0", "1.0".
+  // The current version is 2.0. Valid versions are "3.0", "2.0", "1.0".
```
barbeau, line 57: I'm not sure we really need a version bump to "v3" as part of this proposal. The proposed changes aren't breaking, and it's really an incremental improvement to TripUpdate. In the Service Changes v3.1 docs the term "TripUpdates v3" was used to differentiate it from some of the other groups of feature, but I don't think it's as relevant for the whole spec. I'd prefer to reserve v3 for breaking changes like requiring more fields in VehiclePositions.

Unless I'm missing something - let me know.

ericouyang: The main thing is that feels a bit breaking is that in an existing feed without these fields, `SKIPPED` could be appropriate for the drop-off only scenario as the behavior for a rider-facing application we might expect is to not show predictions for `NO_PICKUP`. With these feeds available, we're imagining that we would keep the relationship `SCHEDULED` since the vehicle _may_ stop, but only upon rider request. 

Agree, though, that there are other cases that feel more additive. For example, a rider application could choose to implement a net new feature to notify a rider on the vehicle upon seeing `MUST_ASK_DRIVER_DROP_OFF ` in the real-time feed.

I did see that there was a bunch of discussion in #221 about the migration path to `DUPLICATED` within the same feed (thank you for documenting that!). I'm not sure where consumers are today, but I'm primarily concerned about time-to-implementation, so I feel that it's safest to maintain two separate feeds, for the time being, and for consumers to adopt the new version once they are ready.

skinkie:
-1 (OpenGeo)
Pro implementing this as complete solution, against doing this as experimental stuff just causing fragmentation.

ericouyang:
I've called for a vote for adoption of this experimental future on the Google Group (https://groups.google.com/g/gtfs-realtime/c/YWY9IoMQF7g).

Please vote with a +1 (in favor) or -1 (against) before Thursday, April 8th at 23:59:59 UTC. Thanks!

ericouyang:
Since we didn't get much feedback last time, I've called for a revote for adoption of this experimental future on the Google Group (https://groups.google.com/g/gtfs-realtime/c/YWY9IoMQF7g?pli=1).

Please vote with a +1 (in favor) or -1 (against) before Wednesday, Aug 24th at 23:59:59 UTC. Thanks!

gcamp:
I think the version problem (can you really have experimental versions?) at the very least needs to be addressed before an other vote. 

I would personally not increase the version considering it has a bigger meaning and the proposal is only experimental. 

You can still maintain two versions side by side with other means like different URLs.

barbeau:
I agree with @gcamp. Experimental versions isn't something in the current CHANGES doc and specifying how that works is probably it's own proposal.

>The main thing is that feels a bit breaking is that in an existing feed without these fields, SKIPPED could be appropriate for the drop-off only scenario as the behavior for a rider-facing application we might expect is to not show predictions for NO_PICKUP. With these feeds available, we're imagining that we would keep the relationship SCHEDULED since the vehicle may stop, but only upon rider request.

I think this could be handled via a [migration guide like we did for DUPLICATED](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/examples/migration-duplicated.md)?

These conditions should be clarified somewhere so it's clear to producers and consumers how the data will be interpreted.

ericouyang:
Thanks, @gcamp and @barbeau! I've gone ahead and reverted the version bump change and will take a first crack at documenting a migration guide for this experimental addition.

gtfs-realtime/proto/gtfs-realtime.proto
ericouyang: Resolving this thread now that the version bump has been removed in favor of a in-place migration guide

gtfs-realtime/spec/en/examples/migration-modified.md
```diff
@@ -0,0 +1,40 @@
...
+      stop_time_properties {
+        pickup_type: NO_PICKUP
+        drop_off_type: MUST_ASK_DRIVER_DROP_OFF
+      }
+    }
+  }
+}
+~~~
+
+It is suggested that you notify existing consumers (e.g., via a developer mailing list) that the use of `SKIPPED` for partial service at a stop is being deprecated in favor of more nuanced information by a set deadline and that consumers should start consuming the `MODIFIED` enumeration as well as `pickup_type` and `drop_off_type`. You should also provide a link to this migration guide. After the deadline passes, you can stop using `SKIPPED` entities for scenarios where `pickup_type` or `drop_off_typ` are included and start using `MODIFIED` instead.
```
paulswartz, line 37: typo
```suggestion
It is suggested that you notify existing consumers (e.g., via a developer mailing list) that the use of `SKIPPED` for partial service at a stop is being deprecated in favor of more nuanced information by a set deadline and that consumers should start consuming the `MODIFIED` enumeration as well as `pickup_type` and `drop_off_type`. You should also provide a link to this migration guide. After the deadline passes, you can stop using `SKIPPED` entities for scenarios where `pickup_type` or `drop_off_type` are included and start using `MODIFIED` instead.
```

paulswartz:
+1 @mbta

paulswartz:
@googlebot I fixed it.

gtfs-realtime/spec/en/examples/migration-modified.md
```diff
@@ -0,0 +1,40 @@
...
+      stop_time_properties {
+        pickup_type: NO_PICKUP
+        drop_off_type: MUST_ASK_DRIVER_DROP_OFF
+      }
+    }
+  }
+}
+~~~
+
+It is suggested that you notify existing consumers (e.g., via a developer mailing list) that the use of `SKIPPED` for partial service at a stop is being deprecated in favor of more nuanced information by a set deadline and that consumers should start consuming the `MODIFIED` enumeration as well as `pickup_type` and `drop_off_type`. You should also provide a link to this migration guide. After the deadline passes, you can stop using `SKIPPED` entities for scenarios where `pickup_type` or `drop_off_type` are included and start using `MODIFIED` instead.
```
barbeau, line 37: It looks like you're effectively deprecating `SKIPPED`? If so, please mark it as deprecated in the .proto with `[deprecated=true]` and docs.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -238,6 +238,12 @@ message TripUpdate {
       // NOTE: This field is still experimental, and subject to change. It may be
       // formally adopted in the future.
       UNSCHEDULED = 3;
+
+        // The vehicle's relationship with the static schedule differs in a way that impacts the rider experience,
+        // such as a vehicle not picking up passengers because a vehicle is "drop-off only" to catch up back to
+        // schedule when it's running late. StopTimeProperties must included to specify how the relationship
+        // has been modified.
+      MODIFIED = 4;
```
barbeau, line 246: The inclusion of the `MODIFIED` enum deviates a bit from our original plan with [GTFS-ServiceChanges v3.1](https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit#heading=h.mm6bblup6h8d). 

We intentionally did not want to introduce any enums that could be interpreted as being able to alter the stop sequence of a trip (beyond the allowances for `assigned_stop_id`). This was the justification behind keeping `SCHEDULED` as the schedule relationship, because the stop sequence still matches the schedule data. It seemed sufficient at the time that the presence of `StopTimeProperties` was sufficient to indicate that those properties had changed from the static data, and a new enum wasn't needed. I'm open to counter-arguments for this, though, if others feel it's needed.

If we keep this `MODIFIED` enum, then it seems it should be used for any data within `StopTimeProperties` (including the already existing `assigned_stop_id`). It should also specifically stated that it does NOT apply to any of the fields outside of `StopTimeProperties` - in other words, you can't just say `MODIFIED` and throw any combination of `stop_id` and `stop_sequence` in a `StopTimeUpdate` and expect a consumer to change the trip this way.

A few other questions:
1. How should I indicate that I don't have arrival predictions for a stop that I `MODIFIED` (i.e., if we can't use `NO_DATA`)?
2. Can we used `MODIFIED` and `StopTimeProperties` for exact_times=0 trips (i.e., if we can't use `UNSCHEDULED`)?

barbeau, line 246: Needs the "experimental" label

barbeau (commented):
@ericouyang Thanks for putting this together! A few thoughts in-line.

skinkie:
-1 (OpenGeo)

ericouyang:
Hi all! I've called for a vote to adopt this change as experimental on the Google Group (https://groups.google.com/g/gtfs-realtime/c/YWY9IoMQF7g?pli=1).

Please vote with a +1 (in favor) or -1 (against) before Tuesday, September 28th at 23:59:59 UTC. Thanks!

@paulswartz, @colemccarren - Since there have been content changes since your vote, I believe we'll need a new +1, if you are still in support of the updated proposal.

brodyFlannigan:
+1 Brody Flannigan Transit Data.

ericouyang:
@skinkie - I completely understand your initial -1 due to not stopping voting when content changes were being made to address feedback (still new to this all!). 

I was surprised, however, by the additional -1 because in my [prior comment](https://github.com/google/transit/pull/265#issuecomment-907360465), I believe I addressed your (very valid!) concern about route deviations. Is there something beyond that you're looking for to gain your support for this proposal to allow for a more expressive real-time specification, so that agencies can better provide more complete real-time data that reflects operational realities, like needing to go drop-off only?

sam-hickey-ibigroup:
We (IBI Group) like the additional granularity offered by `pickup_type` and `drop_off_type` but want to make sure the transition to those fields goes smoothly for producers and consumers. Some comments/questions:

1. Marking `SKIPPED` as `[deprecated=true]` while the new `pickup_type` and `drop_off_type` are experimental seems in conflict. Doesn't seem like there is a lot of precedence for deprecating fields, but seems it would make more sense to wait to mark `SKIPPED` as `[deprecated=true]` until `pickup_type` and `drop_off_type` are formally adopted.

2. Regarding deprecating a field, the [GTFS-realtime change process](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md) says the following:

> By using [deprecated=true] (instead of RESERVED), producers and consumers that have already adopted the field do not have to remove it from use.

However, that is in the "Experimental fields" section. Is it safe to say the same will apply to fields that are formally part of the spec then deprecated (as is proposed here with `SKIPPED`)? 

3. How would a producer know when all of their feed consumers are ready to consume the data for fully skipped stops using `MODIFIED` and `pickup_type=NO_PICKUP` with `drop_off_type=NO_DROP_OFF`? And until all consumers are ready, should a producer use `SKIPPED` and `pickup_type=NO_PICKUP` with `drop_off_type=NO_DROP_OFF` (even though this technically violates the following):

> If pickup_type is specified, StopTimeUpdate.schedule_relationship must be MODIFIED for schedule-based service.
> If drop_off_type is specified, StopTimeUpdate.schedule_relationship must be MODIFIED for scheduled-based service.

ericouyang:
All great questions, @sam-hickey-ibigroup! 

1. The intent with marking `SKIPPED` as `[deprecated=true]`, as noted by @barbeau ([link](https://github.com/google/transit/pull/265#discussion_r696133584)), is to have an explicit signal about the experimental feature and therefore `SKIPPED` could removed in the future, should this proposal be officially adopted. Given that `[deprecated=true]` allows for people to continue using it, as you noted, it feels more helpful for consumers/producers, particularly any new ones, to mark it as such rather than the ambiguity of having 2 ways to represent the same scenario. Totally open to other suggestions for signaling that explicitly, but I'm not aware of another in-built mechanism with protobufs. A comment referring to the new fields feels like it doesn't convey strong enough of a message.

    Should this proposal itself be deprecated or later removed, I think the cleanest approach then would be to un-deprecate `SKIPPED` to signal that as the recommended approach again for this scenario.

2. Yes, that does apply here. In the interim period while this proposal is experimental, producers/consumers do not need to remove it, though it may be requested by a feed producer per the migration guide. I'll add some language to clarify that expectation. If the experimental feature becomes officially adopted, I think that point would be an appropriate time to change `SKIPPED` to [reserved](https://developers.google.com/protocol-buffers/docs/proto#reserved) to prevent further usage.

3. There isn't a perfect way for producers to know other than manually coordinating with each one, so this proposal follows the precedent established by #221 around explicit guidelines for communicating with consumers around timelines. But yes, as noted in the migration guide, what feels like a good interim approach is to use `SKIPPED` with `pickup_type=NO_PICKUP` with `drop_off_type=NO_DROP_OFF`. I'll make an editorial change referencing the migration guide about it being appropriate to use `SKIPPED` in a short transition period.

skinkie:
I would still prefer to keep SKIPPED (_implying_ both drop-off is false) for the situation that the stop is not passed through.

barbeau:
>I would still prefer to keep SKIPPED (implying both drop-off is false) for the situation that the stop is not passed through.

@skinkie Doesn't this approach have backwards compatibility issues though? If producers are currently using `SKIPPED` to describe the situation where a vehicle passes though a stop (but does not stop), which is allowed by the current spec, going forward how do which know which producers intend the new definition (stop is not passed through) vs. the original definition (stop is passed through but vehicle doesn't stop)?

barbeau:
>Should this proposal itself be deprecated or later removed, I think the cleanest approach then would be to un-deprecate SKIPPED to signal that as the recommended approach again for this scenario.

I agree with this - we should probably add an explicit note on that in the .proto comments on SKIPPED. There shouldn't be an issue with this in terms of protobuf compatibility of existing producers/consumers.

>If the experimental feature becomes officially adopted, I think that point would be an appropriate time to change SKIPPED to `reserved` to prevent further usage.

So I know removing an enum value from the .proto will break new consumers of existing feeds that are producing the value (this happened [with REPLACEMENT](https://groups.google.com/g/gtfs-realtime/c/wbPfMIpXsJE/m/AHwwNqruFwAJ) in the past). 

However, I *believe* using `reserved` on enum values does not have this issue as the new consumer bindings still recognize the previous existence of the enum value and therefore don't throw a `(not a valid enum value)` error. My understanding is that `reserved` just prevents you from using the enum in your code the next time you regenerate your bindings from the .proto (i.e., your code won't compile if you try to use the enum value). Conversely, for enum values marked `[deprecated=true]` you can still use the enum value in your code and it will compile, but it will give you warnings that it's deprecated.

Unfortunately the exact mechanics and backwards compatibility issues with `reserved` aren't well-documented in the Protocol Buffer documentation, so it would be good to get feedback from someone with more experience managing backwards compatibility issues in protobufs or test ourselves to confirm before using `reserved`.

skinkie:
> > I would still prefer to keep SKIPPED (implying both drop-off is false) for the situation that the stop is not passed through.
> 
> @skinkie Doesn't this approach have backwards compatibility issues though? If producers are currently using `SKIPPED` to describe the situation where a vehicle passes though a stop (but does not stop), which is allowed by the current spec, going forward how do we know which producers intend the new definition (stop is not passed through) vs. the original definition (stop is passed through but vehicle doesn't stop)?

Functionally the effect for the traveler is the same. The only difference may be that the vehicle location may follow the _expected_ path, while the Skipped-message suggest it may follow _unexpected_ path.

SteveGladstone:
Definitely have a concern as a producer (via Trapeze/Swiftly/others) over `SKIPPED` being deprecated.  Uncertainty over how the reserved keyword in the proto behaves makes me cautious.  The last thing we as an agency (Maryland Transit) want is to be required to go through a costly upgrade process.  And as a consumer of feeds for use in other applications, `SKIPPED` plus `NOT_ACCEPTING_PASSENGERS` combined already covers things well enough by letting riders waiting know their stop won't be served because the bus is not accepting passengers.  Even if it's not explicitly what's intended, it fill the gap conceptually.

That being said, the granularity of explaining what's happening with service in travel can be powerful and important.  But it's a bit unclear how this would get reflected for riders IMO.

One example - train service that will skip stops to make up time due to <reasons> would go from `SKIPPED` to `pickup_type=NO_PICKUP` and `dropoff_type=NO_DROP_OFF`.  Without a corresponding Alert with `effect=NO_SERVICE`, what should this look like on a website, mobile app, or other digital medium?  And given the `vehicle.occupancy_status=NOT_ACCEPTING_PASSENGERS` enum, should that be triggered as well?  I agree with Eric that Vehicle and Trip data is unique, though there is overlap and consistency across all feeds would be preferred.  As far as I'm aware (?), there's still a discussion around `occupancy_status`, `occupancy_percentage`, and `crowding_levels` to reflect passenger loads and associated effects.  If a vehicle has passengers but is not accepting passengers, what should this look like in such a way as to not cause confusion amongst riders?

Another example that could make this a little more difficult.  Consider bus service with stops potentially every 1-2 minutes (inefficient, but humor me).  How should this change be reflected by producers when vehicles go in and out of a `NO_PICKUP` state?  A bus may get too crowded, go into `NO_PICKUP` for 3 stops to a transfer hub, resume pickup for 3 stops, then have to go `NO_PICKUP` after stop 4 because it's another transfer hub.  The oscillation between states seems like it would be problematic- though this is as much an agency policy issue as it would be a feed consumer issue.  If the state is based on VLU logic tied to passenger load and able to be manually triggered, you could give riders a headache over whether their bus will or will not let them board.

Thus a question to the group - how should this info by communicated to riders and how should feed consumers utilize it?

ericouyang:
> Functionally the effect for the traveler is the same.

@skinkie - I respectfully disagree about this and share @barbeau's concern around backwards compatibility. Upon a re-read of the documentation for `SKIPPED`, making a semantic change _does_ impact how the values are interpreted as "Delay from a previous stop in the trip does propagate over the `SKIPPED` stop." I interpret that as meaning, by default, that a vehicle isn't performing, say a detour that would further the delay more than previously indicated.

If you think it's a valuable distinction between vehicles that pass through a stop vs. ones that do not, I would suggest that the best way to do that is introducing another enum (such as `DETOURED`) with explicit semantics, perhaps with a requirement of specifying a value for `delay` or otherwise requiring a `StopTimeEvent` at the next possible `StopTimeUpdate`. However, I don't think this particularly useful of a concept in `TripUpdate`. Maybe it could be an extension to `VehicleStopStatus` instead?

ericouyang:
@barbeau, @SteveGladstone - Regarding `reserved`, upon further reflection, I think how we choose to remove `SKIPPED`, if at all, can be a decision for the future. I've added new language into the spec to reflect interim guidance. The decision between keeping it as `deprecated` (and then removing the additional experimental language) or `reserved` to me primarily rests on how strict we want to be in terms of fully removing `SKIPPED` from the ecosystem over time vs. more of a softer encouragement.

ericouyang:
> The last thing we as an agency (Maryland Transit) want is to be required to go through a costly upgrade process. 

@SteveGladstone - We totally share the same sentiment and, as a feed producer, intend to allow for agencies to decide if and when they'd like to exercise the migration notification process documented in `migration-modified.md`. The choice to use this and other experimental features defined by GTFS-ServiceChanges v3 is fully opt-in and based on the needs of agencies.

Regarding `NOT_ACCEPTING_PASSENGERS`, while there's certainly overlaps in scenarios, I see that and the other discussion around representing occupancy as related, but separate concept. A vehicle may be `pickup_type=NO_PICKUP` and `drop_off_type=NO_DROP_OFF` purely due to running late, but have `occupancy_status=EMPTY`. Personally, I'm not sure how often `NOT_ACCEPTING_PASSENGERS` is used today and how helpful it is as a distinct concept from `FULL`, especially, as you noted, there can be frequent changes in status as a vehicle continue to load/unload passengers (particularly if there's driver discretion), so by the time a vehicle actually arrives at a stop, the status at a prior point in time may not be valid and cause rider confusion.

With respect to how this appears to the passenger, I'd love to get any insights from @gcamp with how `pickup_type` and `drop_off_type` from static GTFS are handled today within Transit. I would presume that the behavior for these setting these in real-time to be the same as what already would happen today. The key intent behind the changes proposed here is not erroneously misleading passengers that they could board an incoming vehicle, but while in a routing/navigation mode, the estimated arrival time could still leverage the predicted timing.   

> Consider bus service with stops potentially every 1-2 minutes (inefficient, but humor me). How should this change be reflected by producers when vehicles go in and out of a `NO_PICKUP` state? A bus may get too crowded, go into `NO_PICKUP` for 3 stops to a transfer hub, resume pickup for 3 stops, then have to go `NO_PICKUP` after stop 4 because it's another transfer hub. The oscillation between states seems like it would be problematic- though this is as much an agency policy issue as it would be a feed consumer issue. If the state is based on VLU logic tied to passenger load and able to be manually triggered, you could give riders a headache over whether their bus will or will not let them board.

Our intended use for these new properties is for an agency-triggered intervention to help restore service, such as when there is bunching, rather than one that's automated based on any signals from a bus (so very different than setting `occupancy_status` based on APCs). The status communicated to consumers of the feed would therefore respect decisions made by a dispatcher in operations. From our research, it appears that usually there is a point along the route (often until the end of the current trip) where a drop-off only would apply, so it wouldn't oscillate and would likely be overly taxing of a process for a dispatcher to change the status frequently. That said, one could also imagine a driver-initiated flow through some onboard interface to self-report statuses and a policy could be derived to, say, limit that propagation to X stops into the future. As a rider, I'd say that I would probably prefer the pleasant surprise of a vehicle arriving earlier than expected while waiting at a stop for a further out vehicle rather than a vehicle being predicted to arrive but not being able to board. Hence, I think real-time passenger information software may want to adopt a more conservative stance to tend to mark more stops as `NO_PICKUP` should an explicit end of a drop-off only not be specified.

sam-hickey-ibigroup:
@ericouyang Thanks for addressing our comments! The “experimentally deprecated” language added in a few places looks nice.

We still have a concern about marking `SKIPPED` as `[deprecated=true]` while the new `pickup_type` and `drop_off_type` are experimental. The [protocol buffers documentation](https://developers.google.com/protocol-buffers/docs/proto#options) says the following about `deprecated`:

> `deprecated` (field option): If set to `true`, indicates that the field is deprecated and should not be used by new code.

This is in conflict with the "experimentally deprecated" language that says, for example, "In this interim period, users may continue to use this value for compatability." Our preference is to not mark the `SKIPPED` field as `[deprecated=true]` at this point and instead keep the "experimentally deprecated" language for now. We're happy to see this all adopted experimentally, but we still have concerns about the migration steps required from a practical perspective that will need to be addressed before moving out of the experimental phase such that the change doesn't leave behind existing producers or consumers of `SKIPPED`.

Also, we share the concern noted by others of changing `SKIPPED` to mean a stop is not passed through.

SteveGladstone:
> The ability to do this, and do it in a way that mirrors the (already understood) static pickup and drop-off types is really exciting.

I concur.  Having the flexibility to handle these scenarios would be awesome.  I think the sticky point right now is `SKIPPED` with deprecation and backwards compatibility given some of us with older CAD/AVL systems might need 1-3 years for "migration" (ug).

Also curious if @gcamp can provide any insight on how Transit might reflect this.

ericouyang:
The voting period for this proposal has ended, with one +1 and one -1 in this most recent iteration. So far, a total of 3 members have voted a +1 for some version of this proposal and 1 member has voted a -1.

As a next step, I'll continue to make further refinements to this proposal and, once it appears that there's some consensus, call for another vote. 

@sam-hickey-ibigroup - That perspective totally makes sense! I've gone ahead and removed the formal deprecation tag and left the "experimentally deprecated" language for a future vote.

> Also, we share the concern noted by others of changing SKIPPED to mean a stop is not passed through.

To clarify, while there was prior discussion around changing the definition, this proposal does not propose any changes to the definitions of any existing enums. As [previously shared](https://github.com/google/transit/pull/265#issuecomment-929323256), if such a notion were helpful (I do not personally think so), I think that should be a completely separate proposal to, say, add a new enum with a new meaning. From the discussion so far, it is only @skinkie who is seeking such a change.

> would it make sense for the definition of an experimental MODIFIED value to include instructions to specifically not use SKIPPED elsewhere in the same .proto?

@colemccarren - I like this idea a lot and will add some language reflecting this guidance!

skinkie:
Keep open.

PR #271, opened 2021-05-06
Editorial fix: standardize reporting of conditional requirements in GTFS Schedule

scmcca:
This PR standardizes the reporting of "Conditionally Required:" statements in field descriptions and adds the convention where it was previously missing (i.e., `stop_times.arrival_time`, `.departure_time`). The aim is to increase the legibility of the documentation.

Contained in this PR:
- Standardized reporting of conditional requirements in the body of field descriptions.
- Standardized list styles.
- Replaced "Required" heading with "Presence".
- Structured "Document Conventions" section.

Feedback welcomed!

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -6,12 +6,10 @@ This document defines the format and structure of the files that comprise a GTFS
 
 ## Table of Contents
 
-1.  [Term Definitions](#term-definitions)
-2.  [Field Types](#field-types)
-3.  [Field Signs](#field-signs)
-4.  [Dataset Files](#dataset-files)
-5.  [File Requirements](#file-requirements)
-6.  [Field Definitions](#field-definitions)
+1.  [Document Conventions](#document-conventions)
+2.  [Dataset Files](#dataset-files)
+3.  [File Requirements](#file-requirements)
+4.  [Field Definitions](#field-definitions)
     -   [agency.txt](#agencytxt)
     -   [stops.txt](#stopstxt)
     -   [routes.txt](#routestxt)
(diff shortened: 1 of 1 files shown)


scmcca:
Follow-up: While no changes are being made to the content of the spec, we invite a general review of the changes, and whether the community thinks a vote is necessary in this case. If not, we will move forward with merging the proposed fixes in about a week. Thanks!

skinkie:
What I dislike from this patch is that it mixes aesthetic (letter cases) with content changes, this must be avoided.

brodyFlannigan:
+1 Brody Flannigan Transit Data.

gcamp:
Agree with @skinkie + if possible in the future the 4 changes listed could have been 4 PRs to make the review substantially easier. Alternatively, it could be 4 different commits that could be reviewed individually. 

That said, I'm ok with the changes. The changes don't require a vote in my opinion.

scmcca:
@skinkie The challenge is standardizing the reporting of "Conditionally Required" statements in field descriptions where they were previously embedded, this inherently changes the form of the content, but should not change the meaning or interpretation of the spec. Beyond that, I understand that restructuring a "Document Conventions" heading as well as replacing "Required" headings with "Presence" is more than an aesthetic fix. However, the end goal is to make the documentation more clear, so if there are specific areas that were made less clear in this process, please let me know and we can adjust them!

scmcca:
@gcamp Great point! Separating these changes out would have been better.

If the changes look good and the consensus is that a vote is not needed, we'll merge them later this week. Feel free to provide any remaining feedback. Thanks!

PR #276, opened 2021-06-03
Allow fare gates (pathway_mode=6) to be bidirectional

scmcca:
As flagged by @skinkie in the [GTFS Changes mailing list](https://mail.google.com/mail/u/0/#inbox/FMfcgzGkXScXfknhsbCkJclwhTDkvSNs), there is a clarification needed to `pathways.is_bidirectional` regarding the behavior of fare gates (`pathway_mode=6`).

Currently, the specification does not allow fare gates to be bidirectional (i.e., the action of providing proof of payment to cross areas separated by a "fare gate"). This is in contradiction to the common case of tap-in tap-out systems where the rider needs to provide proof of payment in either direction of the physical gate, which could be modeled in `pathways.txt` as such. Perhaps I am missing something, so please let me know!

Please see my suggestion to allow `pathway_mode=6` to be bidirectional in the "Files changed" tab.


flocsy:
LGTM

scmcca:
My sense is that data producers who have their pathways for bidirectional fare gates "duplicated" as per the original specification are not affected by the proposed patch here. However, I realize that this PR entails a breaking change for data consumers.

I am wondering if (i) a data producer can share a pathways dataset containing bidirectional fare gates as proposed in this PR, and (ii) some thoughts from data consumers and the community on whether or not this patch will break their products or other's products. The outcome of this PR will likely depend on (ii). Thanks!

skinkie:
We can share a dataset.
-- 
Stefan

abyrd (approved):

abyrd:
Not sure whether reviews are a proper way to respond to these PRs or if they're used differently by a core team. Although this is technically breaking backward compatibility, this is essentially a bugfix to a recent addition to the spec. As stated above this should not pose problems for GTFS producers; GTFS consumers should have a chance to comment on potential problems but I wouldn't anticipate major problems/costs adapting to this change.

scmcca:
@abyrd Good point. 

Seeing that this is a bugfix and that there is support from producers and consumers in this thread, I think we can call a vote. 

This vote is to allow `pathway_mode=6` (fare gates) to be bidirectional. Voting ends on 2021-09-14 at 23:59:59 UTC.

Any remaining feedback is welcomed and encouraged, particularly from consumers. Thanks!

skinkie:
Thanks @scmcca!

+1 (OpenGeo)

abyrd:
+1 Conveyal

flocsy:
+1 Moovit

scmcca:
The vote ended on 2021-09-14 at 23:59:59 UTC with 3 votes in favor and 0 opposed. As per the Specification Amendment Process, this vote passes!

Thanks to all involved.

Issue #273, opened 2021-05-20
Update GTFS to RFC 2119


scmcca:
Hi everyone. We ([MobilityData](https://mobilitydata.org/)) want to get some thoughts from the community on updating GTFS to [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).

In short, RFC 2119 is a standard of keywords to indicate requirement levels (i.e., "MUST", "SHOULD"). It is used by similar data standards such as [GBFS](https://github.com/NABSA/gbfs/blob/master/gbfs.md) and [IMDF](https://register.apple.com/resources/imdf/) to draw clear distinctions around what is required and what is recommended in the specification. These clear distinctions are not made in GTFS, leaving room for interpretation, ambiguity, and make the spec harder to grasp for both newcomers and experienced stakeholders when it comes to producing, consuming, and creating products (i.e., validators) around GTFS.

In terms of updating GTFS documentation to RFC 2119, little change would be needed in the majority of cases. In some cases, parts of the spec may need minor edits to be RFC 2119 friendly, without changing the content (such as replacing "can" with the RFC "MAY"). The goal is to have a data standard that is legible and clear. 

Pending feedback here, we would begin this work with GTFS Schedule, followed by GTFS Realtime. Looking forward to any thoughts on this. Thanks!

skinkie:
What we as producer observe is that some more critical consuming clients (organisations that pay the bills) ask for certain features to be present "because it is in the standard". For MUST this is pretty clear, but the distinction between SHOULD, MAY, optional, best-practice is not. I think this must also be "commercially" clear that implementing a certain feature because all fields are filled in is not better than some fields ignored.

e-lo:
We are thinking about this at Cal-ITP and believe that being able to point to a clearly and consistently articulated list of SHOULD statements which are articulated in the spec itself would be easier to articulate in contract language as well as enforce than the current state of the spec + various best practices.

PR #277, opened 2021-06-03
Proposal to update GTFS Schedule to RFC 2119

scmcca:
As discussed in #273, this PR contains a proposal to update GTFS Schedule to [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) in order to disambiguate interpretations and make the spec easier for producing, consuming, and creating products (i.e., validators) around GTFS. Additionally, these changes aim to be useful when articulating obligations in contract requirements. 

Note that [BCP 14](https://www.rfc-editor.org/info/bcp14) (containing RFC 2119 and RFC 8174) has not been implemented in this initial proposal. Thus, I have not CAPITALIZED the keywords as seen in [GBFS](https://github.com/NABSA/gbfs/blob/master/gbfs.md#terminology). This has the purpose of making the initial proposal easier to review with pre-existing instances of lowercase RFC 2119 keywords, but we can decide if the CAPITALIZATION is useful and if this proposal should be updated to BCP 14, or alternative styling as seen in [IMDF](https://register.apple.com/resources/imdf/) (i.e., only capitalizing "MUST"s, "SHOULD"s, and "MAY"s).

The PR contains separate commits for each applicable section of the GTFS Schedule reference to ease review (the changes are mostly light). 

See all the most up-to-date changes in the "Files changed" tab. Feel free to provide in-line comments and any actionable feedback if any sections have mistakenly altered interpretations or otherwise.

Thanks!

**Edit**
Commit fb61255 catches some instances where "can" is suitable instead of "may". Generally, "can" is used to denote optional usage by a data consumer or producer, and "may" is used to denote optional behaviour of the specification.


gtfs/spec/en/reference.md
```diff
@@ -248,16 +250,16 @@ File: **Conditionally Required**
...
+|  `exception_type` | Enum | **Required** | Indicates whether service is available on the date specified in the date field. Valid options are:<br><br> `1` - Service has been added for the specified date.<br>`2` - Service has been removed for the specified date.<hr>*Example: Suppose a route has one set of trips available on holidays and another set of trips available on all other days. One `service_id` could correspond to the regular service schedule and another `service_id` could correspond to the holiday schedule. For a particular holiday, the [calendar_dates.txt](#calendar_datestxt) file can be used to add the holiday to the holiday `service_id` and to remove the holiday from the regular `service_id` schedule.* |
```
timMillet, line 262: Shouldn't it be "may" instead of "can"?

timMillet (approved):
1 note, otherwise LGTM!

gtfs/spec/en/reference.md
```diff
@@ -140,19 +142,19 @@ File: **Required**
...
+|  `stop_code` | Text | Optional | Short text or a number that identifies the location for riders. These codes are often used in phone-based transit information systems or printed on signage to make it easier for riders to get information for a particular location. The `stop_code` may be the same as `stop_id` if it is public facing. This field should be left empty for locations without a code presented to riders. |
+|  `stop_name` | Text | **Conditionally Required** | Name of the location. Names should be meaningful to local and tourist vernacular.<br><br>When the location is a boarding area (`location_type=4`), the `stop_name` should contains the name of the boarding area as displayed by the agency. It could be just one letter (like on some European intercity railway stations), or text like “Wheelchair boarding area” (NYC’s Subway) or “Head of short trains” (Paris’ RER).<br><br>Conditionally Required:<br>- **Required** for locations which are stops (`location_type=0`), stations (`location_type=1`) or entrances/exits (`location_type=2`).<br>- Optional for locations which are generic nodes (`location_type=3`) or boarding areas (`location_type=4`).|
```
flocsy, line 146: Maybe we should (probably in a separate PR) also clarify "local and tourist". For example in places where the local alphabet is non latin the only way to provide useful information to tourists is to have a latin (probably but not necessarily English) transciption.

gtfs/spec/en/reference.md
```diff
@@ -165,14 +167,14 @@ File: **Required**
...
+|  `continuous_drop_off` | Enum | Optional | Indicates that the rider can alight from the transit vehicle at any point along the vehicle’s travel path as described by `shapes.txt`, on every trip of the route. Valid options are: <br><br>`0` - Continuous stopping drop off. <br>`1` or empty - No continuous stopping drop off. <br>`2` - Must phone agency to arrange continuous stopping drop off. <br>`3` - Must coordinate with driver to arrange continuous stopping drop off. <br><br>Values for `routes.continuous_drop_off` will be overriden by values defined in `stop_times.continuous_drop_off` for specific `stop_time`s along the route. |
```
flocsy, line 177: 1. typo: "overriden" => "overridden" (needs 2 "d"-s) (you have it in 2 places)

flocsy, line 177: 2. maybe it should be "may be overridden" instead of "will be overridden"? I understand that if there's a stop_times.continuous_drop_off then it WILL override the value in routes.continuous_drop_off, but the language everywhere else is about the possibilities of the editor, so IMHO "may be" is better.

gtfs/spec/en/reference.md
```diff
@@ -183,10 +185,10 @@ File: **Required**
...
-|  `block_id` | ID | Optional | Identifies the block to which the trip belongs. A block consists of a single trip or many sequential trips made using the same vehicle, defined by shared service days and `block_id`. A `block_id` can have trips with different service days, making distinct blocks. See the [example below](#example-blocks-and-service-day) |
+|  `trip_headsign` | Text | Optional | Text that appears on signage identifying the trip's destination to riders. Should be used to distinguish between different patterns of service on the same route.<br><br> If the headsign changes during a trip, values for `trip_headsign` will be overridden by values defined in `stop_times.stop_headsign` for specific `stop_time`s along the trip. |
```
flocsy, line 188: If the headsign changes during a trip, values for `trip_headsign` MAY be overridden by values defined in `stop_times.stop_headsign` for specific `stop_time`s along the trip.

gtfs/spec/en/reference.md
```diff
@@ -388,26 +390,26 @@ In regions that have multiple official languages, transit agencies/operators typ
...
...e used if the text needs to be translated into languages other than the default one.<br><br>The default language may be multilingual for datasets with the original text in multiple languages. In such cases, the `feed_lang` field should contain the language code `mul` defined by the norm ISO 639-2. A translation for each language used in the dataset should be provided in `translations.txt`. If all the original text in the dataset is in the same language, then `mul` should not be used.<hr>_Example: Consider a dataset from a multilingual country like Switzerland, with the original `stops.stop_name` field populated with stop names in different languages. Each stop name is written according to the dominant language in that stop’s geographic location, e.g. `Genève` for the French-speaking city of Geneva, `Zürich` for the German-speaking city of Zurich, and `Biel/Bienne` for the bilingual city of Biel/Bienne. The dataset `feed_lang` should be `mul` and translations would be provided in `translations.txt`, in German: `Genf`, `Zürich` and `Biel`; in French: `Genève`, `Zurich` and `Bienne`; in Italian: `Ginevra`, `Zurigo` and `Bienna`; and in English: `Geneva`, `Zurich` and `Biel/Bienne`._ |
```
flocsy, line 410: The file `translations.txt` MAY be used...

gtfs/spec/en/reference.md
```diff
@@ -388,26 +390,26 @@ In regions that have multiple official languages, transit agencies/operators typ
...
...cord_id` | ID | **Conditionally Required** | Defines the record that corresponds to the field to be translated. The value in `record_id` should be a main ID of the table, as defined below:<br><br>- `agency_id` for `agency.txt`<br>- `stop_id` for `stops.txt`;<br>- `route_id` for `routes.txt`;<br>- `trip_id` for `trips.txt`;<br>- `trip_id` for `stop_times.txt`;<br>- `pathway_id` for `pathways.txt`;<br>- `level_id` for `levels.txt`;<br>- `attribution_id` for `attribution.txt`.<br><br>Fields in tables not defined above should not translated. However producers sometimes add extra fields that are outside the official specification and these unofficial fields may be translated. Below is the recommended way to use `record_id` for those tables:<br><br>- `service_id` for `calendar.txt`;<br>- `service_id` for `calendar_dates.txt`;<br>- `fare_id` for `fare_attributes.txt`;<br>- `fare_id` for `fare_rules.txt`;<br>- `shape_id` for `shapes.txt`;<br>- `trip_id` for `frequencies.txt`;<br>- `from_stop_id` for `transfers.txt`.<br><br>Conditionally Required:<br>- **Forbidden** if `table_name` is `feed_info`.<br>- **Forbidden** if `field_value` is defined.<br>- **Required** if `field_value` is empty. |
```
lionel-nj, line 396: Does the following make sense?

```suggestion
|  `record_id` | ID | **Conditionally Required** | Defines the record that corresponds to the field to be translated. The value in `record_id` should be a main ID of the table, as defined below:<br><br>- `agency_id` for `agency.txt`<br>- `stop_id` for `stops.txt`;<br>- `route_id` for `routes.txt`;<br>- `trip_id` for `trips.txt`;<br>- `trip_id` for `stop_times.txt`;<br>- `pathway_id` for `pathways.txt`;<br>- `level_id` for `levels.txt`;<br>- `attribution_id` for `attribution.txt`.<br><br>Fields in tables not defined above should not translated. However producers may sometimes add extra fields that are outside the official specification and these unofficial fields may be translated. Below is the recommended way to use `record_id` for those tables:<br><br>- `service_id` for `calendar.txt`;<br>- `service_id` for `calendar_dates.txt`;<br>- `fare_id` for `fare_attributes.txt`;<br>- `fare_id` for `fare_rules.txt`;<br>- `shape_id` for `shapes.txt`;<br>- `trip_id` for `frequencies.txt`;<br>- `from_stop_id` for `transfers.txt`.<br><br>Conditionally Required:<br>- **Forbidden** if `table_name` is `feed_info`.<br>- **Forbidden** if `field_value` is defined.<br>- **Required** if `field_value` is empty. |
```

gtfs/spec/en/reference.md
```diff
@@ -233,7 +235,7 @@ File: **Conditionally Required**
 
 |  Field Name | Type | Presence | Description |
```
lionel-nj, line 236: Should this set of modifications also clarify that "the end_date date must not precede the start_date date"? Or would this happen in a separate PR?

gtfs/spec/en/reference.md
lionel-nj, line 396: Should this be as follows:

"The value in `record_id` should be a main ID of the table, as defined" -> "The value in `record_id` must be a main ID of the table, as defined..."

lionel-nj (commented):
Thanks @scmcca for working on that! I added a few comments that actually may be addressed in a separate PR according to your workflow :)

gtfs/spec/en/reference.md
```diff
@@ -128,7 +130,7 @@ File: **Required**
 |  `agency_name` | Text | **Required** | Full name of the transit agency. |
 |  `agency_url` | URL | **Required** | URL of the transit agency. |
 |  `agency_timezone` | Timezone | **Required** | Timezone where the transit agency is located. If multiple agencies are specified in the dataset, each must have the same `agency_timezone`. |
-|  `agency_lang` | Language code | Optional | Primary language used by this transit agency. This field helps GTFS consumers choose capitalization rules and other language-specific settings for the dataset. |
+|  `agency_lang` | Language code | Optional | Primary language used by this transit agency. Should be provided to help GTFS consumers choose capitalization rules and other language-specific settings for the dataset. |
 |  `agency_phone` | Phone number | Optional | A voice telephone number for the specified agency. This field is a string value that presents the telephone number as typical for the agency's service area. It can and should contain punctuation marks to group the digits of the number. Dialable text (for example, TriMet's "503-238-RIDE") is permitted, but the field must not contain any other descriptive text. |
```
lionel-nj, line 134:
```suggestion
|  `agency_phone` | Phone number | Optional | A voice telephone number for the specified agency. This field is a string value that presents the telephone number as typical for the agency's service area. It may and should contain punctuation marks to group the digits of the number. Dialable text (for example, TriMet's "503-238-RIDE") is permitted, but the field must not contain any other descriptive text. |
```

gtfs/spec/en/reference.md
scmcca: I agree. I may be wrong but I think `translations.txt` came after this description was written. An update would be useful with some guidance on dealing with cases such as the one you're describing. Would you like to open this topic as an issue?

gtfs/spec/en/reference.md
```diff
@@ -140,19 +142,19 @@ File: **Required**
...
+|  `location_type` | Enum | Optional | Location type. Valid options are:<br><br>`0` (or blank) - **Stop** (or **Platform**). A location where passengers board or disembark from a transit vehicle. Is called a platform when defined within a `parent_station`.<br>`1` - **Station**. A physical structure or area that contains one or more platform.<br>`2` - **Entrance/Exit**. A location where passengers can enter or exit a station from the street. If an entrance/exit belongs to multiple stations, it may be linked by pathways to both, but the data provider must pick one of them as parent.<br>`3` - **Generic Node**. A location within a station, not matching any other `location_type`, that may be used to link together pathways define in pathways.txt.<br>`4` - **Boarding Area**. A specific location on a platform, where passengers can board and/or alight vehicles.|
```
scmcca, line 153: Note
> that may be used to link together pathways

should probably be reverted to:

> that can be used to link together pathways

since it is describing an ability to do something and not a RFC permission/possibility.

gtfs/spec/en/reference.md
scmcca: Good catch. I will make the corrections.

scmcca: I have to agree that the language/perspective I chose is a bit confusing. I think if when we're speaking of "possibilities of the editor" we should use "can" to denote the ability of doing something. Meanwhile "may" should be used for defining RFC permissions/possibility (i.e., behavior of the spec as a computer would see it).

gtfs/spec/en/reference.md
scmcca: Similar to above, we should use "can" to denote an ability of doing something IMHO (i.e., the original description). Using "may" to describe the behavior of the spec makes this part seem like its optional for the consumer to implement this feature, when really it is a required feature of the spec.

gtfs/spec/en/reference.md
scmcca: This section is tricky because it's describing a hypothetical usage of the fields via an example. I am tending towards "could" being okay here and even reverting the final "can" back to "could" to be consistent with the tense. 

@lionel-nj Put it elegantly. As a quick reference:
- Can - Ability to do something (i.e., scenarios, usage)
     - Example: riders ability do to things -> can; e.g. _“riders can reach a customer service…”_
- May - Permission/possibility (i.e., spec)
     - Example: restrictions from the specification -> may; e.g. _“The `stop_code` may be the same as `stop_id` if it is public facing.”_

gtfs/spec/en/reference.md
scmcca: I think making the first 8 `id` fields for the official tables defined as requirements ("must") is good. 

How would the recommended `id` fields for unofficial tables be handled in a validator? A warning in all cases that an "unofficial" `id` is present and an error if none of the `id`s described in this description are present?

gtfs/spec/en/reference.md
scmcca: IMHO it should be "can" since it is describing the ability of the GTFS producer to do something (use `translations.txt`) and not a permission for optional behaviour (RFC "may").

gtfs/spec/en/reference.md
```diff
@@ -49,9 +51,9 @@ Presence conditions applicable to fields and files:
...
+- **Color** - A color encoded as a six-digit hexadecimal number. Refer to [https://htmlcolorcodes.com](https://htmlcolorcodes.com) to generate a valid value (the leading "#" must not be included). <br> *Example: `FFFFFF` for white, `000000` for black or `0039A6` for the A,C,E lines in NYMTA.*
 - **Currency code** - An ISO 4217 alphabetical currency code. For the list of current currency, refer to [https://en.wikipedia.org/wiki/ISO_4217#Active\_codes](https://en.wikipedia.org/wiki/ISO_4217#Active_codes). <br> *Example: `CAD` for Canadian dollars, `EUR` for euros or `JPY` for Japanese yen.*
-- **Date** - Service day in the YYYYMMDD format. Since time within a service day can be above 24:00:00, a service day often contains information for the subsequent day(s). <br> *Example: `20180913` for September 13th, 2018.*
+- **Date** - Service day in the YYYYMMDD format. Since time within a service day can be above 24:00:00, a service day may contain information for the subsequent day(s). <br> *Example: `20180913` for September 13th, 2018.*
```
scmcca, line 56: Should be "Since time within a service day may..."

gtfs/spec/en/reference.md
```diff
@@ -248,16 +250,16 @@ File: **Conditionally Required**
 
 File: **Conditionally Required**
 
-The [calendar_dates.txt](#calendar_datestxt) table can explicitly activate or disable service by date. It can be used in two ways.
+The [calendar_dates.txt](#calendar_datestxt) table explicitly activates or disables service by date. It can be used in two ways.
```
scmcca, line 253: Should be "It may be used in two ways". The implication is that these are the only two ways the file may be used. Using "can" implies even more uses.

gtfs/spec/en/reference.md
lionel-nj: @barbeau please correct me if I'm wrong. As of now the gtfs-validator does not support yet the validation of conditionally required fields. 
However, this could be handled as follows:
- if `table_name` is feed_info, and `record_id` is defined: that is an ERROR;
- if `field_value` is defined,  and `record_id` is defined: that is an ERROR;
- if `field_value` is defined, and `record_id` is not defined: that is an ERROR.
⬆️ these would apply to both official and unofficial tables. 

Now, when it comes to a situation when an ERROR from the previous step was not generated - in the case of an unofficial table:

>  A warning in all cases that an "unofficial" id is present

Yes - but I'd like to discuss this further during our next technical meeting. That encompasses a broader feature to be added: conditional requirements.

gtfs/spec/en/reference.md
isabelle-dr: Should we pick either "may" or "should" here? Does "should" imply "may"?

scmcca: "Should" and "may" have different purposes as defined by [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119). "Should" denotes a recommendation (_recommendation_ for the GTFS data producer to do something) and "may" denotes that the item is optional and identifiable as OK by a validator (phone numbers containing punctuation marks are _permitted_). This sentence has both, but they have different purposes. IMHO we should define the spec first by its behavior, so if we were to pick one I would keep "may". But both are ok to have.

gtfs/spec/en/reference.md
scmcca: I propose the following:
> Values for < field > can be overridden by defining values in < field >...

gtfs/spec/en/reference.md
scmcca: Noted for a separate PR.

gtfs/spec/en/reference.md
scmcca: I committed a change to make the official files "must" as described above.

gtfs/spec/en/reference.md
scmcca: Please see my suggestion applied in the most recent changes in the "Files changed" tab. Feel free to provide a second review

gtfs/spec/en/reference.md
scmcca: Please see a revised suggestion in the "Files changed" tab. Feel free to provide a second review.

gtfs/spec/en/reference.md
```diff
@@ -248,16 +250,16 @@ File: **Conditionally Required**
...
+|  `exception_type` | Enum | **Required** | Indicates whether service is available on the date specified in the date field. Valid options are:<br><br> `1` - Service has been added for the specified date.<br>`2` - Service has been removed for the specified date.<hr>*Example: Suppose a route has one set of trips available on holidays and another set of trips available on all other days. One `service_id` corresponds to the regular service schedule and another `service_id` corresponds to the holiday schedule. For a particular holiday, the [calendar_dates.txt](#calendar_datestxt) file can be used to add the holiday to the holiday `service_id` and to remove the holiday from the regular `service_id` schedule.* |
```
timMillet, line 262: "may" instead of "can"?

gtfs/spec/en/reference.md
```diff
@@ -388,26 +390,26 @@ In regions that have multiple official languages, transit agencies/operators typ
...
 
 ### feed_info.txt
 
 File: **Optional** (**Required** if `translations.txt` is provided)
 
-The file contains information about the dataset itself, rather than the services that the dataset describes. Note that, in some cases, the publisher of the dataset is a different entity than any of the agencies.
+The file contains information about the dataset itself, rather than the services that the dataset describes. Note that, in some cases, the publisher of the dataset may be a different entity than any of the agencies.
```
timMillet, line 404: "can" instead of "may"?

timMillet (approved):

gtfs/spec/en/reference.md
scmcca: Since this is referring to "usage" around who a publisher can be, I agree.

gtfs/spec/en/reference.md
scmcca: I suppose this case is defining a *permission of the spec* in the context of the example, so I agree.

gtfs/spec/en/reference.md
abyrd, line 146: "meaningful to ... vernacular" does not seem like idiomatic English, so cannot be substituted for "in the ... vernacular". The expression "tourist vernacular" is a little hard to interpret anyway in the original spec. A possible alternative is "Names should be understandable to both local residents and tourists".

gtfs/spec/en/reference.md
```diff
@@ -140,19 +142,19 @@ File: **Required**
...
 |  `tts_stop_name` | Text | Optional | Readable version of the `stop_name`. See "Text-to-speech field" in the [Term Definitions](#term-definitions) for more. |
-|  `stop_desc` | Text | Optional | Description of the location that provides useful, quality information. Do not simply duplicate the name of the location. |
+|  `stop_desc` | Text | Optional | Description of the location that provides useful, quality information. Must not contain a duplicate of `stop_name`.|
```
abyrd, line 148: "Must not contain" is different than "must not simply duplicate". This changes the meaning of the document rather than just clarifying the term "must".

gtfs/spec/en/reference.md
```diff
@@ -165,14 +167,14 @@ File: **Required**
...
-|  `route_desc` | Text | Optional | Description of a route that provides useful, quality information. Do not simply duplicate the name of the route. <hr> _Example: "A" trains operate between Inwood-207 St, Manhattan and Far Rockaway-Mott Avenue, Queens at all times. Also from about 6AM until about midnight, additional "A" trains operate between Inwood-207 St and Lefferts Boulevard (trains typically alternate between Lefferts Blvd and Far Rockaway)._ |
+|  `route_desc` | Text | Optional | Description of a route that provides useful, quality information. Must not contain a duplicate of `route_short_name` or `route_long_name`. <hr> _Example: "A" trains operate between Inwood-207 St, Manhattan and Far Rockaway-Mott Avenue, Queens at all times. Also from about 6AM until about midnight, additional "A" trains operate between Inwood-207 St and Lefferts Boulevard (trains typically alternate between Lefferts Blvd and Far Rockaway)._ |
```
abyrd, line 170: "Must not contain" is different than "must not simply duplicate". This changes the meaning of the document rather than just clarifying the term "must".

gtfs/spec/en/reference.md
```diff
@@ -165,14 +167,14 @@ File: **Required**
...
+|  `continuous_drop_off` | Enum | Optional | Indicates that the rider can alight from the transit vehicle at any point along the vehicle’s travel path as described by `shapes.txt`, on every trip of the route. Valid options are: <br><br>`0` - Continuous stopping drop off. <br>`1` or empty - No continuous stopping drop off. <br>`2` - Must phone agency to arrange continuous stopping drop off. <br>`3` - Must coordinate with driver to arrange continuous stopping drop off. <br><br>Values for `routes.continuous_drop_off` can be overridden by defining values in `stop_times.continuous_drop_off` for specific `stop_time`s along the route. |
```
abyrd, line 177: Should "can be overridden" be replaced with "may be overridden"?

gtfs/spec/en/reference.md
```diff
@@ -183,10 +185,10 @@ File: **Required**
...
-|  `block_id` | ID | Optional | Identifies the block to which the trip belongs. A block consists of a single trip or many sequential trips made using the same vehicle, defined by shared service days and `block_id`. A `block_id` can have trips with different service days, making distinct blocks. See the [example below](#example-blocks-and-service-day) |
+|  `trip_headsign` | Text | Optional | Text that appears on signage identifying the trip's destination to riders. Should be used to distinguish between different patterns of service on the same route.<br><br> If the headsign changes during a trip, values for `trip_headsign` can be overridden by defining values in `stop_times.stop_headsign` for specific `stop_time`s along the trip. |
```
abyrd, line 188: "can be overridden" should be "may be overridden" or "should be overridden"

gtfs/spec/en/reference.md
```diff
@@ -183,10 +185,10 @@ File: **Required**
...
+|  `trip_short_name` | Text | Optional | Public facing text used to identify the trip to riders, for instance, to identify train numbers for commuter rail trips. If riders do not commonly rely on trip names, `trip_short_name` should be empty. A `trip_short_name` value, if provided, should uniquely identify a trip within a service day; it should not be used for destination names or limited/express designations. |
+|  `direction_id` | Enum | Optional | Indicates the direction of travel for a trip. This field should not be used in routing; it provides a way to separate trips by direction when publishing time tables. Valid options are: <br><br>`0` - Travel in one direction (e.g. outbound travel).<br>`1` - Travel in the opposite direction (e.g. inbound travel).<hr>*Example: The `trip_headsign` and `direction_id` fields could be used together to assign a name to travel in each direction for a set of trips. A [trips.txt](#tripstxt) file could contain these records for use in time tables:* <br> `trip_id,...,trip_headsign,direction_id` <br> `1234,...,Airport,0` <br> `1505,...,Downtown,1` |
```
abyrd, line 190: "may be used together" instead of "could be used together".  Also the following sentence may be clearer if it starts with "For example trips.txt could contain"...

gtfs/spec/en/reference.md
```diff
@@ -219,12 +221,12 @@ File: **Required**
...
-|  `shape_dist_traveled` | Non-negative float | Optional | Actual distance traveled along the associated shape, from the first stop to the stop specified in this record. This field specifies how much of the shape to draw between any two stops during a trip. Must be in the same units used in [shapes.txt](#shapestxt). Values used for `shape_dist_traveled` must increase along with `stop_sequence`; they cannot be used to show reverse travel along a route.<hr>*Example: If a bus travels a distance of 5.25 kilometers from the start of the shape to the stop,`shape_dist_traveled`=`5.25`.*|
+|  `shape_dist_traveled` | Non-negative float | Optional | Actual distance traveled along the associated shape, from the first stop to the stop specified in this record. This field specifies how much of the shape to draw between any two stops during a trip. Must be in the same units used in [shapes.txt](#shapestxt). Values used for `shape_dist_traveled` must increase along with `stop_sequence`; they must not be used to show reverse travel along a route.<hr>*Example: If a bus travels a distance of 5.25 kilometers from the start of the shape to the stop,`shape_dist_traveled`=`5.25`.*|
```
abyrd, line 229: The meaning of the sentence "they must not be used to show reverse travel along a route" is not clear, but changes here would be off-topic and should perhaps be made in a separate PR.

gtfs/spec/en/reference.md
```diff
@@ -248,16 +250,16 @@ File: **Conditionally Required**
...
+|  `exception_type` | Enum | **Required** | Indicates whether service is available on the date specified in the date field. Valid options are:<br><br> `1` - Service has been added for the specified date.<br>`2` - Service has been removed for the specified date.<hr>*Example: Suppose a route has one set of trips available on holidays and another set of trips available on all other days. One `service_id` corresponds to the regular service schedule and another `service_id` corresponds to the holiday schedule. For a particular holiday, the [calendar_dates.txt](#calendar_datestxt) file may be used to add the holiday to the holiday `service_id` and to remove the holiday from the regular `service_id` schedule.* |
```
abyrd, line 262: RFC2119 is about clarifying requirements, but the conditionals here were inside an example, rather than being requirements in the specification. Removing the conditionals can give the impression that there is only one way to represent the example in GTFS. I would rephrase this whole example as "For example, suppose a route... This could be represented with one service_id for the regular schedule..." however this change might be out of scope for the current PR and should perhaps be done in a separate PR.

gtfs/spec/en/reference.md
```diff
@@ -269,9 +271,9 @@ File: **Optional**
 |  `price` | Non-negative float | **Required** | Fare price, in the unit specified by `currency_type`. |
 |  `currency_type` | Currency code | **Required** | Currency used to pay the fare. |
 |  `payment_method` | Enum | **Required** | Indicates when the fare must be paid. Valid options are:<br><br>`0` - Fare is paid on board.<br>`1` - Fare must be paid before boarding. |
-|  `transfers` | Enum | **Required** | Indicates the number of transfers permitted on this fare. The fact that this field can be left empty is an exception to the requirement that a Required field must not be empty. Valid options are:<br><br>`0` - No transfers permitted on this fare.<br>`1` - Riders may transfer once.<br>`2` - Riders may transfer twice.<br>empty - Unlimited transfers are permitted. |
+|  `transfers` | Enum | **Required** | Indicates the number of transfers permitted on this fare. Valid options are:<br><br>`0` - No transfers permitted on this fare.<br>`1` - Riders can transfer once.<br>`2` - Riders can transfer twice.<br>empty - Unlimited transfers are permitted. |
```
abyrd, line 274: Why is the word "can" substituted for "may" here? The change is the reverse of what's being done elsewhere in the document. These words are describing what the rider is allowed to do, not requirement levels in the specification so they are not really covered by RFC 2119. If the intent is to _avoid_ confusion with the meaning of "optional" given in RFC 2119, I suggest we use the words "is permitted" or "is allowed".

gtfs/spec/en/reference.md
```diff
@@ -388,26 +390,26 @@ In regions that have multiple official languages, transit agencies/operators typ
...
 
 ### feed_info.txt
 
 File: **Optional** (**Required** if `translations.txt` is provided)
 
-The file contains information about the dataset itself, rather than the services that the dataset describes. Note that, in some cases, the publisher of the dataset is a different entity than any of the agencies.
+The file contains information about the dataset itself, rather than the services that the dataset describes. Note that, in some cases, the publisher of the dataset can be a different entity than any of the agencies.
```
abyrd, line 404: I find this reads better with the "is". This is not really related to RFC 2119 but if we're editing for clarity I'd keep "is" and remove the superfluous "Note that":
```suggestion
The file contains information about the dataset itself, rather than the services that the dataset describes. In some cases, the publisher of the dataset is a different entity than any of the agencies.
```

abyrd (changes requested):

gtfs/spec/en/reference.md
```diff
@@ -8,7 +8,7 @@ This document defines the format and structure of the files that comprise a GTFS
 
 1.  [Document Conventions](#document-conventions)
 2.  [Dataset Files](#dataset-files)
-3.  [File Requirements](#file-requirements)
+3.  [File Conventions](#file-conventions)
```
abyrd, line 11: RFC 2119 provides words "used to signify the requirements in the specification" which according to the title are "Key words ... to Indicate Requirement Levels". Is there a reason we should stop calling these requirements and start calling them conventions? I think the word "conventions" is better suited for the following paragraph that lists the key words (where it is already used).

scmcca: Good point. Reverting to "File Requirements" works for me.

gtfs/spec/en/reference.md
scmcca: Good point. Since this is explicit about "the rider", perhaps we can keep "may".

gtfs/spec/en/reference.md
scmcca: I agree that the sentence about reverse travel is not clear. The description is clear in the first sentence why values must increase (i.e., "actual distance traveled"). Nonetheless, I think "they must not be used to show reverse travel along a route" is still a valid statement here?

gtfs/spec/en/reference.md
scmcca: Great insight around RFC 2119 and examples. We can revert the proposed changes and dedicate any clarifications to a separate PR, as suggested.

gtfs/spec/en/reference.md
scmcca: I chose the word "meaningful" as a substitute for "will understand" to indicate that names should represent stops according to how people ("locals" and "tourists") associate names with place. That is to say, not only should the name be legible, but it should reflect what the stop is conventionally called. 

The suggested alternative to "Names should be understandable to both local residents and tourists" might be too ambiguous IMHO, as it may only allude to legibility and localization for different audiences (but say nothing about producing "meaningful" names).

@flocsy also [raised the point](https://github.com/google/transit/pull/277#discussion_r646094801) that we should add a note about the use of `translations.txt` along the lines of "local and tourist" (perhaps a separate issue).

gtfs/spec/en/reference.md
scmcca: Good catch. Maybe we can edit this to:
> Must not be a duplicate of `stop_name`.

gtfs/spec/en/reference.md
scmcca: Same suggestion as above in `stop_desc`. Change "must not contain" to "must not be".

isabelle-dr:
I think this section below could be part of this review. 
"For times occurring after midnight, enter the time as a value greater than 24:00:00 " in Field Types under Time.
I'm not sure if "must" or "should" should be used here.

gtfs/spec/en/reference.md
```diff
@@ -140,19 +142,19 @@ File: **Required**
...
 |  `parent_station` | ID referencing `stops.stop_id` | **Conditionally Required** | Defines hierarchy between the different locations defined in `stops.txt`. It contains the ID of the parent location, as followed:<br><br>- **Stop/platform** (`location_type=0`): the `parent_station` field contains the ID of a station.<br>- **Station** (`location_type=1`): this field must be empty.<br>- **Entrance/exit** (`location_type=2`) or **generic node** (`location_type=3`): the `parent_station` field contains the ID of a station (`location_type=1`)<br>- **Boarding Area** (`location_type=4`): the `parent_station` field contains ID of a platform.<br><br>Conditionally Required:<br>- **Required** for locations which are entrances (`location_type=2`), generic nodes (`location_type=3`) or boarding areas (`location_type=4`).<br>- Optional for stops/platforms (`location_type=0`).<br>- Forbidden for stations (`location_type=1`).|
```
isabelle-dr, line 154: Should we add "must" in _Stop/platform_, _Entrance_ and _Boarding area_ sections of `stops.parent_station`?

scmcca:
@isabelle-dr 

> I think the two sections below could be part of this review as well.
"For times occurring after midnight, enter the time as a value greater than 24:00:00" in Field Types under Time.

I can think of a clarification to make this sentence less instructional and more descriptive: 

"Values greater than 24:00:00 are allowed to define times occurring after midnight for the service day on which the trip schedule begins."

What is also missing from the "Time" definition are invalid values. There is no maximum time value described. perhaps after some value like +12 or +24 hours above 24:00:00 a warning is issued in the validator? Any other suggestions?

> "Timezone names never contain the space character but may contain an underscore." in Field Types under Timezone.

I'd like to reduce the "Timezone" description more generally to:

"TZ timezone as described in https://www.iana.org/time-zones. Refer to http://en.wikipedia.org/wiki/List_of_tz_zones for a list of valid values.<br> _Example: Asia/Tokyo, America/Los_Angeles or Africa/Cairo._"

isabelle-dr:
@scmcca 

> What is also missing from the "Time" definition are invalid values. There is no maximum time value described. perhaps after some value like +12 or +24 hours above 24:00:00 a warning is issued in the validator? Any other suggestions?

Currently, the [Canonical GTFS Validator](https://github.com/MobilityData/gtfs-validator) accepts values in formats H:MM:SS, HH:MM:SS, as well as HHH:MM:SS, and emits an ERROR otherwise. 
There have been [discussions](https://github.com/MobilityData/gtfs-validator/pull/21) on having a separate WARNING if a time field contains a value above a certain threshold. Since this would be an addition to the specification, it could happen in a separate PR or added to the Best Practices if we see a need for such a threshold.

I think those two suggestions work perfectly in the context of this PR!

gtfs/spec/en/reference.md
```diff
@@ -165,14 +167,14 @@ File: **Required**
...
-|  `route_desc` | Text | Optional | Description of a route that provides useful, quality information. Do not simply duplicate the name of the route. <hr> _Example: "A" trains operate between Inwood-207 St, Manhattan and Far Rockaway-Mott Avenue, Queens at all times. Also from about 6AM until about midnight, additional "A" trains operate between Inwood-207 St and Lefferts Boulevard (trains typically alternate between Lefferts Blvd and Far Rockaway)._ |
+|  `route_desc` | Text | Optional | Description of a route that provides useful, quality information. Must not be a duplicate of `route_short_name` or `route_long_name`. <hr> _Example: "A" trains operate between Inwood-207 St, Manhattan and Far Rockaway-Mott Avenue, Queens at all times. Also from about 6AM until about midnight, additional "A" trains operate between Inwood-207 St and Lefferts Boulevard (trains typically alternate between Lefferts Blvd and Far Rockaway)._ |
```
aababilov, line 170: This looks too strict. I suggest to say "should not" here.

Numerous feeds that are consumed by Google do have this notice and they are considered valid.

A route that has the same `route_desc` and `route_short_name` or `route_long_name` may be still served to the end user, so this is not an error. In fact, many consumers may just ignore route_desc completely.

See also my patch to Validator:
https://github.com/MobilityData/gtfs-validator/pull/917

gtfs/spec/en/reference.md
scmcca: Can you give a suggestion of what this would look like, or has this been resolved?

gtfs/spec/en/reference.md
scmcca: I've applied this suggestion to stop_desc as well. Thanks!

gtfs/spec/en/reference.md
abyrd: Yes, I think the pair of sentences "must increase" / "must not be used" reads well. Clarification could be handled some other time.

abyrd:
> @isabelle-dr
> 
> > I think the two sections below could be part of this review as well.
> > "For times occurring after midnight, enter the time as a value greater than 24:00:00" in Field Types under Time.
> 
> I can think of a clarification to make this sentence less instructional and more descriptive:
> 
> "Values greater than 24:00:00 are allowed to define times occurring after midnight for the service day on which the trip schedule begins."

As I understand it, trips may begin after 24:00:00, i.e. times over 24:00:00 are not restricted to trips beginning before 24:00:00. This is because some agencies internally define an overnight service break (when no service runs, maintenance can be done, etc.) and the next service day only begins after that overnight service break. If the service break is 3-5AM for example, a final trip of the day from 2-2:30 AM could reasonably be coded on the previous day using 26:00:00-26:30:00. This pull request is described as being focused on the use of certain specific terms, so I think it would be best to avoid any potential changes to the meaning of the specification.

> What is also missing from the "Time" definition are invalid values. There is no maximum time value described. perhaps after some value like +12 or +24 hours above 24:00:00 a warning is issued in the validator? Any other suggestions?

I agree it is reasonable to issue warnings on strange values of the time field. However there are services that run for several days and will exceed these limits - notably some ferries in Norway that are known to be represented in GTFS data. The values that are considered "unusual" are somewhat subjective and might be different in different validation contexts. To me this seems like good material for the "best practices" guide, with the specification addressing only more objective limits.

> I'd like to reduce the "Timezone" description more generally to:
> 
> "TZ timezone as described in https://www.iana.org/time-zones. Refer to http://en.wikipedia.org/wiki/List_of_tz_zones for a list of valid values.
> _Example: Asia/Tokyo, America/Los_Angeles or Africa/Cairo._"

This wording seems good to me, but this change seems out of scope for the current pull request. Given that some people may decide to read, comment, and vote on the PR based on its title and initial description, it might be better to make such changes in another PR.

gtfs/spec/en/reference.md
abyrd: I can see why you might want to replace "understand" with some other word, but my sense is that the revised version of the sentence is not grammatically correct, specifically the expression "meaningful to ... vernacular". Perhaps the whole sentence should be rewritten once a clearer meaning is agreed upon, and perhaps without using the word "vernacular". As mentioned in other comments below about time fields, this change seems out of scope in a PR described as being about RFC 2119 terms. I'm not insisting on PRs being purely single-subject, but if a PR does include additional changes it seems important to mention those in the description or title.

scmcca:
@abyrd @skinkie 

Fair points. I'll revert the changes made to the 24:00:00+ and TZ Timezone descriptions. 

After that do you think we're good to go?

gtfs/spec/en/reference.md
scmcca: I agree that this needs more attention, and that it is a separate topic from RFC. I reverted it to the original for now.

scmcca:
Thanks to everyone involved in the revisions. This was a lot of work! We seem to have reached a stable consensus. 

As per the [Specification Amendment Process](https://github.com/google/transit/blob/master/gtfs/CHANGES.md), I am calling a vote to update GTFS Schedule to [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) as outlined in this PR.

Voting will end on 2021-08-26 at 23:59:59 UTC.

westontrillium:
+1 from @Trillium

abyrd:
+1 (Conveyal)

mgilligan:
+1, [TriMet](https://developer.trimet.org/)

e-lo:
+1 [Cal-ITP](calitp.org)

flocsy:
+1 Moovit

Issue #266, opened 2021-03-26
Static spec doesn't differentiate between primary and foreign IDs


botanize:
No where in the spec is it clear what the rules are for ID fields. The only "documentation" is in the validators.

For example, it seems reasonable that `trip_id` would only need to be unique within the combination of service and route IDs in the `trips.txt` file (in SQL terms, `PRIMARY KEY (service_id, route_id, trip_id)`). However, validators tell me that `trip_id` is required to be unique within `trips.txt` regardless of the combination of other ID fields (`PRIMARY KEY (trip_id)`), but the other ID fields,  `route_id` and `service_id`, are not expected to be unique. 

Relying on implied rules results in further confusion. For example, the `service_id` ID is the only ID in `calendar_dates.txt`, but is not expected to be unique (`FOREIGN KEY (service_id)`), which you might expect if you assumed that an ID needs to be unique if it's the only ID, or if the ID name is a close match to the file name or file purpose. 

Maybe adapting language from relational databases, e.g. primary ID and foreign ID, would clarify the role of each ID in each file. If there are any examples where there is a multi-column primary key (multiple ID fields used to uniquely identify a row) and that constraint is expected by validators, it should be noted in the file information. 

For example, the spec for stop_times.txt could look like this:

```
### stop_times.txt

File: **Required**

PRIMARY KEY(trip_id, stop_sequence)

|  Field Name | Type | Required | Description |
|  ------ | ------ | ------ | ------ |
|  `trip_id` | foreign ID referencing `trips.trip_id` | **Required** | Identifies a trip.  |

```

scmcca:
This looks good to me! Any thoughts on turning this into a pull request to work through the details?

botanize:
Yep, happy to, but I'm working on getting my employer to sign a contributor agreement first.

abyrd:
Agreed, this is a good idea. I believe GTFS emerged historically as a text dump of a relational database, so relational concepts like primary keys are completely relevant. Especially since GTFS is often used by importing it back into a relational database.

IvanVolosyuk:
Trip id is supposed to be globally unique within a static feed regardless
of the routes they are in. A realtime update can specify only the trip_id
and it should be sufficient to identify the trip uniquely without
additional information.

mmorang:
[...]

PR #278, opened 2021-07-14
Add Primary and Foreign ID fields

botanize:
Re #266

- Define Primary ID and Foreign ID types
- Re-type existing IDs
- Add `Primary ID  (ID, ...)` to file introduction to clarify requirements to uniquely identify rows

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -55,6 +55,8 @@ Presence conditions applicable to fields and files:
 - **Email** - An email address. <br> *Example: `example@example.com`*
 - **Enum** - An option from a set of predefined constants defined in the "Description" column. <br> *Example: The `route_type` field contains a `0` for tram, a `1` for subway...*
 - **ID** - An ID field value is an internal ID, not intended to be shown to riders, and is a sequence of any UTF-8 characters. Using only printable ASCII characters is recommended. IDs defined in one .txt file are often referenced in another .txt file. <br> *Example: The `stop_id` field in [stops.txt](#stopstxt) is a ID. The `stop_id` field in [stop_times.txt](#stop_timestxt) is an ID referencing `stops.stop_id`.*
+- **Primary ID** - An ID or set of IDs and fields that uniquely identify a row. <br> *Example: the `trip_id` and `stop_sequence` fields make the Primary ID of [stop_times.txt](#stop_timestxt).*
+- **Foreign ID** - An ID that references an ID in another .txt file. <br> *Example: The `stop_id` field in [stop_times.txt](#stop_timestxt) is a Foreign ID referencing `stops.stop_id`.*
 - **Language code** - An IETF BCP 47 language code. For an introduction to IETF BCP 47, refer to [http://www.rfc-editor.org/rfc/bcp/bcp47.txt](http://www.rfc-editor.org/rfc/bcp/bcp47.txt) and [http://www.w3.org/International/articles/language-tags/](http://www.w3.org/International/articles/language-tags/). <br> *Example: `en` for English, `en-US` for American English or `de` for German.*
 - **Latitude** - WGS84 latitude in decimal degrees. The value must be greater than or equal to -90.0 and less than or equal to 90.0. *<br> Example: `41.890169` for the Colosseum in Rome.*
 - **Longitude** - WGS84 longitude in decimal degrees. The value must be greater than or equal to -180.0 and less than or equal to 180.0. <br> *Example: `12.492269` for the Colosseum in Rome.*
@@ -122,9 +124,11 @@ The following example demonstrates how a field value would appear in a comma-del
 
 File: **Required**
 
+Primary ID (`agency_id`)
+
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ | ------ |
-|  `agency_id` | ID | **Conditionally Required** | Identifies a transit brand which is often synonymous with a transit agency. Note that in some cases, such as when a single agency operates multiple separate services, agencies and brands are distinct. This document uses the term "agency" in place of "brand". A dataset may contain data from multiple agencies. <br><br>Conditionally Required:<br>- **Required** when the dataset contains data for multiple transit agencies. <br>- Optional otherwise. |
+|  `agency_id` | Primary ID | **Conditionally Required** | Identifies a transit brand which is often synonymous with a transit agency. Note that in some cases, such as when a single agency operates multiple separate services, agencies and brands are distinct. This document uses the term "agency" in place of "brand". A dataset may contain data from multiple agencies. <br><br>Conditionally Required:<br>- **Required** when the dataset contains data for multiple transit agencies. <br>- Optional otherwise. |
 |  `agency_name` | Text | **Required** | Full name of the transit agency. |
 |  `agency_url` | URL | **Required** | URL of the transit agency. |
 |  `agency_timezone` | Timezone | **Required** | Timezone where the transit agency is located. If multiple agencies are specified in the dataset, each must have the same `agency_timezone`. |
@@ -137,9 +141,11 @@ File: **Required**
 
 File: **Required**
 
+Primary ID (`stop_id`)
+
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ | ------ |
-|  `stop_id` | ID | **Required** | Identifies a location: stop/platform, station, entrance/exit, generic node or boarding area (see `location_type`). <br><br>Multiple routes may use the same `stop_id`. |
+|  `stop_id` | Primary ID | **Required** | Identifies a location: stop/platform, station, entrance/exit, generic node or boarding area (see `location_type`). <br><br>Multiple routes may use the same `stop_id`. |
 |  `stop_code` | Text | Optional | Short text or a number that identifies the location for riders. These codes are often used in phone-based transit information systems or printed on signage to make it easier for riders to get information for a particular location. The `stop_code` can be the same as `stop_id` if it is public facing. This field should be left empty for locations without a code presented to riders. |
 |  `stop_name` | Text | **Conditionally Required** | Name of the location. Use a name that people will understand in the local and tourist vernacular.<br><br>When the location is a boarding area (`location_type=4`), the `stop_name` should contains the name of the boarding area as displayed by the agency. It could be just one letter (like on some European intercity railway stations), or text like “Wheelchair boarding area” (NYC’s Subway) or “Head of short trains” (Paris’ RER).<br><br>Conditionally Required:<br>- **Required** for locations which are stops (`location_type=0`), stations (`location_type=1`) or entrances/exits (`location_type=2`).<br>- Optional for locations which are generic nodes (`location_type=3`) or boarding areas (`location_type=4`).|
 |  `tts_stop_name` | Text | Optional | Readable version of the `stop_name`. See "Text-to-speech field" in the [Term Definitions](#term-definitions) for more. |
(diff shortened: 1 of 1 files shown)


scmcca:
@botanize Thanks for putting this together! Here are my suggestions and feedback.

**Style**
I wonder if it would be enough to simply indicate what the Primary ID is at the top of each file as suggested in this PR in the style of "Primary ID (field)". 

Currently in the proposal, the Primary ID is indicated at the top of each file and then again as a field type. This creates confusing instances where we are defining a Primary ID for the file, and then again as a related but separate concept per field. In simple cases these match, such as for routes.txt and `route_id`. But ambiguities/redundancies arise. For example, a Primary ID at the file level might contain Foreign ID type values as well as non-ID type values such as in stop_times.txt (`trip_id`, `stop_sequence`). Moreover, some ID fields do not fit in the "Primary" or "Foreign" concept such as `stops.zone_id`.

My suggestion is to only indicate what the Primary ID is at the file level, and do away with defining "Primary" or "Foreign" at the individual field levels (i.e., keep them as they were defined originally as "ID" or "ID referencing"). "Foreign" IDs are already clear IMHO with the phrase "ID referencing". 

I'm a fan of keeping the definitions of "Primary ID" and "Foreign ID" in the Field Types section of the documentation. 

**Tech**
- **The definition for "Foreign ID"** in the Field Types section should be edited to include ID referencing IDs in the same file (and not just in other files). An example of this would be `stops.parent_station`, which references `stops.stop_id` in the same stops.txt file.
- **fare_rules.txt**: No Primary ID was defined for this file. My interpretation is that the collection of all fields defined in a record in fare_rules.txt must be unique.
- **frequencies.txt**: The Primary ID should include `start_time` and `end_time` to yield Primary ID (`trip_id`, `start_time`, `end_time`). My understanding is that you can define successive time intervals for the same `trip_id` to define different frequencies throughout a trip. This would make Primary ID (`trip_id`) alone false.
- **transfers.txt**: The Primary ID should be defined as Primary ID (`from_stop_id`, `to_stop_id`).
- **pathways.txt**: The Primary ID should be defined as Primary ID ('pathway_id`, 'from_stop_id`, `to_stop_id`). If different pathways need to be defined for the same `from_stop_id` and `to_stop_id` pair to define different `pathway_mode`s, for example, a new `pathway_id` must be used. Correct me if I'm wrong!
- **translations.txt**:  I think the collection of all fields defined in a given record of translations.txt must be unique. Additionally, I think `record_id` should be of type "ID referencing".
- **feed_info.txt**: Is the implication of `feed_start_date` and `feed_end_date` that only 1 record is allowed in the feed_info.txt file? If so, this should be made clear in the file description.
- **attributions.txt**: Seems like the Primary ID here should be Primary ID (`attribution_id`, `organization_name`). In which case, `attribution_id` should have its presence changed to **Required**. This is perhaps a separate PR.

Let me know what you think and if I overlooked anything! Thanks.

botanize:
@scmcca thank you for the feedback! 

> I wonder if it would be enough to simply indicate what the Primary ID is at the top of each file as suggested in this PR in the style of "Primary ID (field)".
> 
> Currently in the proposal, the Primary ID is indicated at the top of each file and then again as a field type. This creates confusing instances where we are defining a Primary ID for the file, and then again as a related but separate concept per field. In simple cases these match, such as for routes.txt and route_id. But ambiguities/redundancies arise. For example, a Primary ID at the file level might contain Foreign ID type values as well as non-ID type values such as in stop_times.txt (trip_id, stop_sequence). Moreover, some ID fields do not fit in the "Primary" or "Foreign" concept such as stops.zone_id.
> 
> My suggestion is to only indicate what the Primary ID is at the file level, and do away with defining "Primary" or "Foreign" at the individual field levels (i.e., keep them as they were defined originally as "ID" or "ID referencing"). "Foreign" IDs are already clear IMHO with the phrase "ID referencing".

I intended the file-level and field-level Primary ID concepts to be identical. The file-level Primary ID uniquely identifies a row in the file, if there's only one field required to do that, that field is a Primary ID. Seeing that `trip_id` is of type Primary ID tells me that it is sufficient to uniquely identify a record, which is exactly the same information provided at the file-level, ```Primary ID (`trip_id`)```. When multiple fields (whether ID types or not) are required to identify a row, the file level Primary ID lists them all out, but it's possible that some of them are not IDs (`frequencies.start_time`) and others are plain IDs (`shapes.shape_id`), and some may be Foreign IDs (`frequencies.trip_id`). And I think it's helpful to know that `shapes.shape_id` isn't a Primary ID, despite it being pretty obvious, but the ID is named after the file, so one could assume otherwise.

While the existing language for foreign IDs (`ID referencing…`) is clear when the ID can reference only one field in one file, it fails when there's an ID that can reference a variety of IDs across files. For example, `translations.record_id` is a Foreign ID, but what it references isn't set in stone the way most of the foreign IDs are (like `stop_times.stop_id`). So I think it's more clear to mark that field as `Foreign ID` than `ID` or `ID referencing *`.


Thank you for the technical notes, I was unsure of the primary keys for many of these files, which was the motivation for the issue and pull request!

> * **fare_rules.txt**: No Primary ID was defined for this file. My interpretation is that the collection of all fields defined in a record in fare_rules.txt must be unique.

I think we could use ```Primary ID (`*`)``` to indicate that the combination of all provided fields must be unique, I could document that convention in the Field Type definition for Primary ID. Otherwise I can spell it out and list all of the fields as part of the primary ID.

> * **frequencies.txt**: The Primary ID should include `start_time` and `end_time` to yield Primary ID (`trip_id`, `start_time`, `end_time`). My understanding is that you can define successive time intervals for the same `trip_id` to define different frequencies throughout a trip. This would make Primary ID (`trip_id`) alone false.

That was also my understanding.

> * **pathways.txt**: The Primary ID should be defined as Primary ID ('pathway_id`, 'from_stop_id`, `to_stop_id`). If different pathways need to be defined for the same `from_stop_id` and `to_stop_id` pair to define different `pathway_mode`s, for example, a new `pathway_id` must be used. Correct me if I'm wrong!

Makes sense to me, do we know what any consumers of **pathways.txt** are doing?

> * **translations.txt**:  I think the collection of all fields defined in a given record of translations.txt must be unique. Additionally, I think `record_id` should be of type "ID referencing".

I label `record_id` as "Foreign ID", which means "ID referencing…". In this case it can reference a lot of different things, and that's explained well in the description, so I don't know what else to say here other than `Foreign ID`.

> * **feed_info.txt**: Is the implication of `feed_start_date` and `feed_end_date` that only 1 record is allowed in the feed_info.txt file? If so, this should be made clear in the file description.

I agree, my understanding is that this is a one-line file. I could add something like ```Primary ID (none)``` and replace the start of the file description with "This file contains a single row of…".

> * **attributions.txt**: Seems like the Primary ID here should be Primary ID (`attribution_id`, `organization_name`). In which case, `attribution_id` should have its presence changed to **Required**. This is perhaps a separate PR.

I don't know how consumers interpret this, OneBusAway, OpenTripPlanner and TransitClock don't seem to consume it. I used the "ID named after the file" heuristic to select the Primary ID, but now I see that it's not even a required field. I don't see anything to gain from requiring `attribution_id` to be unique only in combination with `organization_name`, I think it should become **Required** and become the Primary ID.

gtfs/spec/en/reference.md
```diff
@@ -178,16 +186,18 @@ File: **Required**
...
 
+Primary ID (`trip_id`)
+
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ | ------ |
-|  `route_id` | ID referencing `routes.route_id` | **Required** | Identifies a route. |
-|  `service_id` | ID referencing `calendar.service_id` or `calendar_dates.service_id` | **Required** | Identifies a set of dates when service is available for one or more routes. |
-|  `trip_id` | ID | **Required** | Identifies a trip. |
+|  `route_id` | Foreign ID referencing `routes.route_id` | **Required** | Identifies a route. |
+|  `service_id` | Foreign ID referencing `calendar.service_id` or `calendar_dates.service_id` | **Required** | Identifies a set of dates when service is available for one or more routes. |
```
skinkie, line 194: This should be and/or. Because the file could contain either a normalised, normalised+exceptions, or denormalised setup.

gtfs/spec/en/reference.md
```diff
@@ -231,9 +243,11 @@ File: **Required**
 
 File: **Conditionally Required**
 
+Primary ID (`service_id`)
+
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ |------ |
-|  `service_id` | ID | **Required** | Uniquely identifies a set of dates when service is available for one or more routes. Each `service_id` value can appear at most once in a [calendar.txt](#calendartxt) file. |
+|  `service_id` | Primary ID | **Required** | Uniquely identifies a set of dates when service is available for one or more routes. Each `service_id` value can appear at most once in a [calendar.txt](#calendartxt) file. |
```
skinkie, line 250: I disagree on this point. This is actually a foreign id pointing towards a not yet existing structure in GTFS: services. It is used as grouping. Technically one could argue that multiple ranges _could_ be supported.

botanize: I agree that that would be a reasonable interpretation, but the spec already states that ```Each `service_id` value can appear at most once in a [calendar.txt](#calendartxt) file.```, which is essentially the definition of "Primary ID". I think changing that part of the spec would require a separate pull-request.

skinkie: It is still not a primary key. Because it does not have to be in the file, if only calendar_dates.txt is delivered...

gtfs/spec/en/reference.md
botanize: I left the original phrasing of the field, only adding that it was a "Foreign" ID. At least in American usage "or" is taken to mean one, the other or both (like logical OR), and it seems pretty clear here that "or" isn't intended to mean "not both".

gtfs/spec/en/reference.md
botanize: Sure, calendar.txt is **Conditionally Required**, but doesn't change the fact that `service_id` is **Required** if calendar.txt is provided, and is allowed to appear at most once in calendar.txt, and that's what makes it a primary key for the file.

abyrd: Conceptually there is a set of services which is the union of all service IDs referenced in calendar or calendar_dates. But that's just a set of IDs with no other attached fields, so it can be implicit. I believe the original idea was that all services would be defined in calendar.txt and exceptions referencing those services in calendar_dates.txt, but in some systems this would yield services in calendar.txt with no active days, and an informal rule was made that any such service with no active days can be dropped. For robustness and consistency, it might have been better to just require these empty services to be defined anyway in calendar.txt but that's not the spec we have. Looking at it this way, `service_id` is the primary key of `calendar`, and is a foreign key in `calendar_dates` which is allowed to reference these implicit fully-inactive rows in `calendar`.

abyrd:
I definitely support clearly stating the primary keys of each table in GTFS. Overall the direction of this proposal seems good, but I share some of the same concerns as @scmcca, who says "This creates confusing instances where we are defining a Primary ID for the file, and then again as a related but separate concept per field."

This confusion arises because this PR is conflating the two distinct concepts of type and key. The primary key is a set of attributes, and every attribute has a type whether or not it is in the primary key. The fact that an attribute is a member of the primary key or not is independent of its type. It so happens that in GTFS we also have (retroactively) defined a type called ID. Primary keys often include fields of type ID, but when they are included in the primary key their type remains ID.

I like that the primary key is shown as a tuple at the top of each table definition. This will be immediately recognizable and understandable to anyone familiar with the relational model of data, including many people implementing GTFS consumer systems. I don't think it's contradictory or problematic to also flag which fields are members of the primary key or are foreign keys, but these should not be treated as types - doing so is likely to create subtle confusion. Either they should not be in the type column, or if they are placed in one of the existing columns for brevity, it should be clear they are just annotations and do not replace the values in that column.

I think the discussion of primary and foreign keys should also be moved out of the types list into a different section.

barbeau:
Note that primary and foreign keys have been defined in the canonical GTFS validator that MobilityData has been working on with input from the community. See the table schemas at:
https://github.com/MobilityData/gtfs-validator/tree/master/main/src/main/java/org/mobilitydata/gtfsvalidator/table

Primary keys are fields annotated with `@PrimaryKey`, and foreign keys are annotated with `@ForeignKey` and a definition of the field they are a foreign key to. For example, for trips.txt route_id is annotated with `@ForeignKey(table = "routes.txt", field = "route_id")`.

I'd suggest cross-checking this PR with that tool to see if there are any discrepancies.

scmcca:
> Based on the comments in the linked issue, the key on frequencies.txt should be trip_id + start_time, not trip_id alone, correct?

I can reiterate the suggestion I made earlier to have the primary key for frequencies.txt be (`trip_id`, `start_time`, `end_time`). Having the primary key only at (`trip_id`, `start_time`) would theoretically allow for multiple records with the same `trip_id` and `start_time` but with confounding `end_time`s. Maybe I'm missing something?

Also, it's a picky detail but to keep consistency should we be using [common nouns](https://www.grammarly.com/blog/common-noun/) for "primary key", "unique ID", and "foreign ID"? For example "Primary Key" would be "Primary key" at the top of each file and "unique ID" would be "Unique ID" in the Type column.

Otherwise LGTM.

botanize:
> I can reiterate the suggestion I made earlier to have the primary key for frequencies.txt be (`trip_id`, `start_time`, `end_time`). Having the primary key only at (`trip_id`, `start_time`) would theoretically allow for multiple records with the same `trip_id` and `start_time` but with confounding `end_time`s. Maybe I'm missing something?

I think it's actually the opposite. A primary key of (`trip_id`, `start_time`) means there can be only one row for any combination of `trip_id` and `start_time`. If you make the primary key (`trip_id`, `start_time`, `end_time`) you'd allow multiple rows with the same `trip_id` and `start_time` as long as they're disambiguated by `end_time`, which makes no sense to me.

scmcca:
@botanize That makes sense. Would that mean that the primary key would be equally correct if it were (`trip_id`, `end_time`)?

botanize:
Given that headways for the same trip must not overlap (see the description of `headway_secs`), either `start_time` or `end_time` could be used in combination with `trip_id` to make the primary key. However, it seems far more conventional to me to use the start time of the service.

Bertware:
Regarding transfers.txt: 

This PR would define (`from_stop_id`, `to_stop_id`) as the primary key, which would prevent us (Samtrafiken/Trafiklab) from extending this file with trip-to-trip transfer information as we do today. Right now we publish this data according to the [Google trip-to-trip transfer extension](https://developers.google.com/transit/gtfs/reference/gtfs-extensions#TripToTripTransfers), but this would become invalid as it would cause duplicate keys. The same functionality is included in [MBTA's transfers.txt extension](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#transferstxt). How should this be handled?

@mbta @paulswartz

paulswartz:
@Bertware internally, we use (`from_stop_id`, `to_stop_id`, `from_trip_id`, `to_trip_id`) as the unique key.

Bertware:
We use (`from_stop_id`, `to_stop_id`, `from_trip_id`, `to_trip_id`) as unique key as well, where `from_trip_id `and `to_trip_id `can be empty for normal (default) stop-to-stop transfers. I'm not sure if there are producers using the `from_route_id `and `to_route_id `fields. Should these fields be proposed in a separate issue and PR, and possibly included in the official spec, so they can be used as unique key in this PR?

skinkie:
> We use (`from_stop_id`, `to_stop_id`, `from_trip_id`, `to_trip_id`) as unique key as well, where `from_trip_id `and `to_trip_id `can be empty for normal (default) stop-to-stop transfers. I'm not sure if there are producers using the `from_route_id `and `to_route_id` fields. Should these fields (`from_trip_id`, `to_trip_id`, and possibly `from_route_id `, `to_route_id` if there are producers and consumers) be proposed in a separate issue and PR, and possibly included in the official spec, so they can be used as unique key in this PR?

I think this PR should not address it, but the trip-to-trip extension should.

Bertware:
Just checked the repository and there already is [an open PR for the trip-to-trip extension](https://github.com/google/transit/pull/32) #32 I was unaware of. Whichever gets merged in second should take the other into account, but if the trip-to-trip transfer proposal can be merged first we would prevent backwards compatibility issues (since this PR would declare (from_stop_id, to_stop_id) to be unique, while this would be undone if #32 gets merged after this one)

antrim:
Related: I closed PR #32. I think it makes sense to open a separate PR that's cleaner covering transfer rules for routes and trips, and in-seat transfers.

skinkie:
@antrim what was the reason not to resolve that in #32?

scmcca:
@skinkie @antrim Correct me if I'm wrong, but I think the decision to close #32 is because the PR would have to resolve conflicts by spec changes since 2016 and the will to continue #32 is simply no longer there by the advocate (it went stale before reopening, but no meaningful activity has continued despite momentum here). 

Since this proposal calls out the need for `transfers.from_trip_id` and `transfers.to_trip_id`, I suggest we open a new PR for trip-to-trip transfers to address the block we are currently facing here. Anyone want to advocate it? If not I can certainly do it to move the discussion along, but I wouldn't want to steal the PR glory from anyone who is more involved in the extension.

botanize:
I think the bigger issue is, what are you supposed to do if you develop a GTFS extension that violates a primary key described in this PR?

If the extension becomes part of the spec the primary key would be updated, but while it's a proprietary extension there would be a conflict between the extended feed and the spec.

We could just add something to the new Dataset Attributes subheading that says un-official extensions may change these relationships by adding new fields to the end of the table's primary key? That would prevent someone from extending `trips.txt` by adding `service_id` to the primary key, since `service_id` already exists in the spec. It would also allow adding `transfers.from_trip_id` and `transfers.to_trip_id` to the end of the primary key resulting in (`from_stop_id`, `to_stop_id`, `from_trip_id`, `to_trip_id`), which is backwards compatible with (`from_stop_id`, `to_stop_id`) in the sense that null values for `from_trip_id` and `to_trip_id` would mean that each `from_stop_id`, `to_stop_id` pair would need to be unique.

scmcca:
@botanize Do you have examples of other extensions where this problem presents itself? If so it would be useful to note them here. 

Otherwise, it seems that there is demand for `from_trip_id` and `to_trip_id` as an official part of the spec without causing problems here (if merged before this PR).

botanize:
I don't have any examples, but it seems bound to come up again at some point.

How long do we sit on this PR to try to get #32 or a substitute PR approved?

scmcca:
> but it seems bound to come up again at some point.

Agreed that this is a foreseeable issue. I can open a substitute PR tomorrow, at which point we'll have to let 7 discussion days pass followed by 7 days for voting. So we should be able to resume with #278 by early October.

scmcca:
@botanize #284 for trip-to-trip and route-to-route transfers has been merged. Feel free to move forward with defining the IDs for `transfers.txt` as discussed above.

botanize:
I ordered the fields in the transfers.txt key to keep the existing keys for producers of the extension compatible: (`from_stop_id`, `to_stop_id`, `from_trip_id`, `to_trip_id`, `from_route_id`, `to_route_id`)

@Bertware @paulswartz does this meet your needs?

Bertware:
@botanize looks good to me!

skinkie:
+1 OpenGeo

flocsy:
fare_rules.txt is a problem IMHO. Although mostly not because changes in this PR. However Primary key (`*`) would allow me to have N lines that define (fare_id, route_id, origin_id, destination_id, {contains_id_1, ...., contains_id_N}) (this is what we want to allow), but it also allows me to add another line: (fare_id, route_id, origin_id, destination_id, null), which makes no sense and I think it shouldn't be allowed. So maybe contains_id should be Conditionally Required with an explanation that is all the other fields equal between any two rows then contains_id is required.

Side note: However the bigger problem is that the current gtfs doesn't allow me to do this:
(fare_id, route_id, origin_id, destination_id, {contains_id_1, contains_id_2, contains_id_3})
(fare_id, route_id, origin_id, destination_id, {contains_id_1, contains_id_2})
Maybe we should start a separate discussion on how to solve this problem.

PR #283, opened 2021-09-08
Feature/image in alerts

gcamp:
Service alerts sometime can benefit of having an image next to the alert description to enhance the comprehension of the alert. 

MBTA is already sending [alerts with images on their twitter](https://twitter.com/MBTA/status/1435588731575185409). UTA also uses a webpage to detail their alerts and [add an image to the content](https://content.govdelivery.com/bulletins/gd/UTTA-2ee87a5). IBI service alerts too already support attaching images that way. 

This proposal adds a URL to an image for it to be displayed by apps. 

cc @ibi-group-team @paulswartz


skinkie:
If we are not even able to follow up #76 for static, I do not want to see this proposal for RT fast tracked.

gcamp:
@skinkie feel free to continue the discussion of #76 in that PR. The proposal here is not fast tracked, it just follow the normal process as any other proposal. 

I welcome your comments on this PR, if you have any!

skinkie:
It might make sense to make it a ilst, consisting of the image url and an enum with a type. Photo, Map, ..., come to mind.

brodyFlannigan:
> It might make sense to make it a ilst, consisting of the image url and an enum with a type. Photo, Map, ..., come to mind.

I like where this is going, especially with maps.
Often times, disruptions are very complicated to explain in a GTFS-RT alert where there is no formatting (i.e. headers, bullet points, etc.), and since not all consuming display/link to the URL provided, it can be very difficult to get information to riders. Being able to link to a map or image would be really helpful.

skinkie:
> > It might make sense to make it a ilst, consisting of the image url and an enum with a type. Photo, Map, ..., come to mind.
> 
> I like where this is going, especially with maps.
> Often times, disruptions are very complicated to explain in a GTFS-RT alert where there is no formatting (i.e. headers, bullet points, etc.), and since not all consuming display/link to the URL provided, it can be very difficult to get information to riders. Being able to link to a map or image would be really helpful.

Lets not forget accessibility, complementary information :)

gcamp:
Yes I almost want to explicitly discourage text in image, that's certainly not what I wanted.

brodyFlannigan:
Absolutely, there should NOT be any text in images. I'm thinking more like
having an image of a map somewhat like OC Transpo (Ottawa, Canada) does for
their detours (Like this
<https://web.archive.org/web/20210624003349/https://www.octranspo.com/images/files/maps/detours/2021/Piperville_2021_222.pdf>
).

stevenmwhite:
Since we did refer to agencies who include images with Twitter posts, when it comes to the idea about making sure there's not text in the images, we should note that plenty of agencies use images like the below on Twitter simply as a more visually appealing/branded way to share the message. 

https://twitter.com/metrolaalerts/status/1439968121230667776

https://twitter.com/SMBigBlueBus/status/1438941139449548808

https://twitter.com/lbtransit/status/1405615776271278080

I think we all agree the intent here is not looking for this kind of "branded visual" that just replicates the text of the tweet (often with even less information), but we should at least recognize that this type of content would be the natural use case for many agencies.

As suggested, specifically calling out what types of images are desired and including the enum of type to force a selection into those desired categories could certainly be helpful.

gcamp:
@stevenmwhite Having an enum seems overkill to me. I understand the goal of forcing different types of images but in practice we'll need an `OTHER` which defeats the purpose.

stevenmwhite:
@gcamp Yeah, I agree with that as well...

I do think it will be difficult to enforce any kind of image that we _don't_ want agencies to publish and we should recognize that. 

Perhaps some language around how it shouldn't be used for "marketing or branding" images, but I don't think any of those agencies consider what they're posting to Twitter to be marketing or branding necessarily.

gcamp:
@stevenmwhite let me know what you think of the updated language

stevenmwhite:
@gcamp I think that's helpful.

I should also note that I'm not _opposed_ to any of this and would be in favor of the feature generally. We'd love to include these features in our alert publishing software for GTFS-RT. As a consumer, you will certainly be more picky about what types of images you do or do not want to have showing up in your app than I will about what types of images agencies desire to include.

I just wanted to make the point that it's _likely_ agencies will naturally publish images for visual branding purposes if their alert usage on Twitter is any indication, so we should either be prepared to accept that or to actively discourage it somehow.

gcamp:
@skinkie added list of images. The enum has been discussed earlier and I don't think it's required. 

Let me know if you have other comments before I call a vote.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -634,6 +634,12 @@ message Alert {
 
   optional SeverityLevel severity_level = 14 [default = UNKNOWN_SEVERITY];
 
+  // List of URL linking to an image to be displayed along the alert text. Used to explain visually the alert effect of a detour, station closure, etc. The image should enhance the understanding of the alert and must not be the only location of essential information.
```
skinkie, line 637:
```suggestion
  // List of URLs linking to an image to be displayed along the alert text. Used to explain visually the alert effect of a detour, station closure, etc. The image should enhance the understanding of the alert and must not be the only location of essential information.
```
?

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -634,6 +634,12 @@ message Alert {
...
+  // The following types of images are discouraged : image containing mainly text, marketing or branded images that add no additional information. 
+  // The image linked need to be PNG or JPG of less than 500kB.
```
skinkie, line 639:
```suggestion
  // The images linked need to be PNG or JPG of less than 500kB. 
```
Is there a reason you want to exclude other formats such as WebP?

gtfs-realtime/spec/en/reference.md
```diff
@@ -38,38 +38,61 @@ Fields labeled as **experimental** are subject to change and not yet formally ad
...
-            *   [TimeRange](#message-timerange)
-            *   [EntitySelector](#message-entityselector)
-                *   [TripDescriptor](#message-tripdescriptor)
-                    *   [ScheduleRelationship](#enum-schedulerelationship-1)
-            *   [Cause](#enum-cause)
-            *   [Effect](#enum-effect)
-            *   [TranslatedString](#message-translatedstring)
-                *   [Translation](#message-translation)
-            *   [SeverityLevel](#enum-severitylevel)
+- [Elements](#elements)
```
barbeau, line 41: Are these changes to the table of contents intended?

gcamp: No I'll revert those, it's from my editor.

gtfs-realtime/proto/gtfs-realtime.proto
gcamp: I'm not strongly against WebP but the fact that the iOS support is [relatively recent](https://css-tricks.com/webp-image-support-coming-to-ios-14/) I wouldn't have included it. .

paulswartz:
Possible additions, which would involve a new type besides `TranslatedString`:
- should we include the MIME type of the image, so clients could provide a WebP image and others could filter it out?
- should we provide alt text?

stevenmwhite:
On the alt text... I would generally highly suggest YES.

However, I'm curious how this would work practically. As-is, the image itself is kind of like an alt image for the alert text. Would we expect producers to provide an image that gives substantially new information (as opposed to just the same information visually) than the alert text, and thus the image could be described in a manner that also gives additional information? Or do we expect that the alt text for the image would reasonably be equivalent to the text of the alert in its meaning.

I'm imaging the following scenario:

> **Alert Header:** Detour on route 5 downtown
> **Alert Text:** Buses detour onto Spring Street from 1st to 9th, missing normal stops at Main & 1st, Main & 5th, and Main & 8th. Stops will be available at Spring & 1st, Spring & 5th, and Spring & 8th. This detour will last for the remainder of the day.
> **Image:** _A map with stops crossed out on Main Street, showing the new path of travel on Spring Street with new stops highlighted._

Is the phrase in italics that I wrote above to describe the image something that would be useful as alt text? Or is it duplicative of/less useful than the actual alert text?

gcamp:
I think we're ready for a vote. This vote is for adding an experimental field in of images in alerts. 

Voting ends on 2021-11-10 at 23:59:59 UTC.

gtfs-realtime/spec/en/reference.md
```diff
@@ -502,20 +503,40 @@ An internationalized message containing per-language versions of a snippet of te
...
+| **media_type** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | IANA media type as to specify the type of image to be displayed. The type must start with "image/" |
 | **language** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | BCP-47 language code. Can be omitted if the language is unknown or if no internationalization is done at all for the feed. At most one translation is allowed to have an unspecified language tag - if there is more than one translation, the language must be provided. |
 
+
 ## _message_ Shape
 
 Describes the physical path that a vehicle takes when the shape is not part of the (CSV) GTFS, such as for an ad-hoc detour. Shapes belong to Trips and consist of an encoded polyline for more efficient transmission.  Shapes do not need to intercept the location of Stops exactly, but all Stops on a trip should lie within a small distance of the shape for that trip, i.e. close to straight line segments connecting the shape points
 
-**Caution:** this message is still **experimental**, and subject to change. It may be formally adopted in the future.<br>.
```
skinkie, line 516: Is this a change that should be added to this specific commit?

gcamp: I reverted the editorial changes, thanks for pointing it out.

gtfs-realtime/spec/en/reference.md
```diff
@@ -502,11 +503,32 @@ An internationalized message containing per-language versions of a snippet of te
 
 A localized string mapped to a language.
 
+| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
+|------------------|------------|----------------|-------------------|-------------------|
+| **text** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | A UTF-8 string containing the message. |
+| **language** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | BCP-47 language code. Can be omitted if the language is unknown or if no internationalization is done at all for the feed. At most one translation is allowed to have an unspecified language tag - if there is more than one translation, the language must be provided. |
```
skinkie, line 509: Was this a change that also appeared out of nowhere?

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -906,6 +911,56 @@ message TranslatedString {
...
+//    the first matching translation is picked.
+// 2. If a default UI language (e.g., English) matches the language code of a
+//    translation, the first matching translation is picked.
+// 3. If some translation has an unspecified language code, that translation is
+//    picked.
+// NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
+message TranslatedImage {
+  message LocalizedImage {
+    // String containing a fully qualified URL linking to an image
+    // The image linked needs to be less than 500kB.
```
flocsy, line 928: it's unclear whether the 500kb size limit is a recommendation or it is enforced?

gtfs-realtime/spec/en/reference.md
gcamp: This is just the way git is displaying the changes. `language` from Translation is "moved" to `LocalizedImage` and `text` is deleted and re-added. This doesn't affect the content itself. 

There's no way to fix that afaik.

skinkie: @gcamp I _assumed_ that might have been the case. Thanks for confirming.

flocsy:
Sorry, it looks I missed the discussion, but a few thoughts:
1. it's unclear if the language describes the language used on the image (i.e: the station names are in Japanese or English) or it describes the language of the alternate text. IMHO both combinations should be allowed. I can see that there only exists an image with Japanese text on it but there is an English alternative text for it.
2. I think it could be useful to allow multiple image sizes, so the client could chose the most appropriate size to display. Or at least the comment could encourage the use of vector images.
3. Is the 500k image size limit a recommendation or is it "enforced"?

gtfs-realtime/proto/gtfs-realtime.proto
gcamp: Yes my intent was that is enforced. I can change to SHOULD.

gcamp:
@flocsy 

1. The intent was the text and image should always be the same langage, but you're making a good point. Should we remove TranslatedImage and just keep an list of TranslatedString? @barbeau also seemed to think this could have been a good solution. 
2. I feel this should be the responsibility of the consumer to handle different sizes and not the producers. I'm not sure if others have thoughts on this. 
3. It should be enforced IMO. I'll change the text to SHOULD unless there's disagreement.

gtfs-realtime/proto/gtfs-realtime.proto
barbeau: >Yes my intent was that is enforced. I can change to SHOULD.

Do you mean it's a requirement? If so, it should be a MUST.

gtfs-realtime/spec/en/reference.md
```diff
@@ -502,11 +503,32 @@ An internationalized message containing per-language versions of a snippet of te
...
-| **text** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | A UTF-8 string containing the message. |
+| **localized_image** | [LocalizedImage](#message-localizedimage) | Required | Many | At least one localized image must be provided. |
+
+## _message_ LocalizedImage
+
+A localized image URL mapped to a language.
+
+| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
+|------------------|------------|----------------|-------------------|-------------------|
+| **url** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | String containing a fully qualified URL linking to an image. The image linked needs to be less than 500kB. |
```
barbeau, line 529: Sorry for the late comment, I didn't think of this until now - I wonder if we should strongly encourage the producer to support (and consumer to use) [If-Modified-Since HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) for this URL.

I could see producers swapping out the image at the URL for incidents as they develop and are more fully understood, and we really don't want people downloading a .5 MB file repeatedly just to see if it changed.

I suppose this applies to anything hosted at URLs (including the feeds themselves), but given the potentially large-ish file size here it seems more relevant.

flocsy:
@gcamp the consumer can maybe scale down an image but not scale up. 0.5MB is not that much, certainly not in the 21st century, even in mobile apps. So the sentence that the consumer should take care of it is kind of contradicting the strong restriction about the size.
Instead I think we could enable the producer (not enforcing him) to use different (even multiple) size of images.

@barbeau I don't think you need to write here about http headers. If they or their server doesn't know about it it's their problem, you won't be able to fix it. You might maybe do something else: say that it's not encouraged to change the image content without changing the url.

gcamp:
I'm ok being more relaxed on the image size. Honestly didn't put a lot of thought when writing that initially, do you have suggestion on a target size? Adding multiple size is something that I don't think will end up being used in practice (because producers won't produce it). At least I don't think IBI supports it right now even if they support images already. cc @ibi-group-team 

+1 on not changing content of the same url

sam-hickey-ibigroup:
> Adding multiple size is something that I don't think will end up being used in practice (because producers won't produce it). At least I don't think IBI supports it right now even if they support images already. cc @ibi-group-team

That’s correct. We (IBI) do not currently support multiple sizes for images. We also support only one image per alert. This proposal shows image with a cardinality of “Many”. We’re not completely against this, but we don’t yet see a strong need for more than one image per alert from a producer or consumer perspective, especially given the ambiguity of how a consumer would display more than one image for an alert.

gcamp:
If it wasn't clear already, the vote has been cancelled due to changes discussed. 

I pushed the following changes : 
- Size limit is now enforced (MUST)
- Size limit upped to 2MB
- Added langage about making sure the URL changes if the content of the images changes
- I moved `alternative_text` out of `TranslatedImage` and into the parent `Alert`, and renamed it `image_alternative_text`. As @flocsy pointed out, the alternative text and image can be provided with different translations and might not match 1 to 1. 

Re the ability to have multiple images seems unnecessary @sam-hickey-ibigroup. It might not happen often but I see multiple images as a nice to have. I can see for example an images with a map of a detour and a second image with an detailed map of a new stop location in a dense area.

flocsy:
Regarding multiple images: if we allow multiple images then since not all consumers will be able to support more than one image - and even for those that do - there should be a way to determine which one is the main image (or the "order of importance" of the images)

gtfs-realtime/spec/en/reference.md
```diff
@@ -502,11 +505,31 @@ An internationalized message containing per-language versions of a snippet of te
...
+|------------------|------------|----------------|-------------------|-------------------|
+| **text** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Required | One | A UTF-8 string containing the message. |
+| **language** | [string](https://developers.google.com/protocol-buffers/docs/proto#scalar) | Conditionally required | One | BCP-47 language code. Can be omitted if the language is unknown or if no internationalization is done at all for the feed. At most one translation is allowed to have an unspecified language tag - if there is more than one translation, the language must be provided. |
+
+## _message_ TranslatedImage
+
+An internationalized message containing per-language versions of an image. One of the images from a message will be picked up. The resolution proceeds as follows: If the UI language matches the language code of a translation, the first matching translation is picked. If a default UI language (e.g., English) matches the language code of a translation, the first matching translation is picked. If some translation has an unspecified language code, that translation is picked.
```
flocsy, line 515: I completely understand the intent of this sentence, but I think the wording is not correct. This sounds line a consumer's api documentation, while we mainly are talking about how to produce these files. So all the "is picked" should be changed IMHO. Maybe to something that makes it as a recommendation to the consumers. Or maybe a bit stricter than that... like "should be picked"

gcamp: This is text copied + adapted from TranslatedString. I'm for clarifying this but maybe that can be done in an other PR and fix both at the same time?

sam-hickey-ibigroup:
> It might not happen often but I see multiple images as a nice to have. I can see for example an images with a map of a detour and a second image with an detailed map of a new stop location in a dense area.

This is a good example.

If multiple images are allowed, there should be a way to define their ordering as noted by @flocsy and the  `image_alternative_text` would need to move back into the `TranslatedImage` because each image could have different `alternative_text`.

gcamp:
Right, my bad for not thinking of alternative_text relationship with the multiple stops. Even if I pushed for multiple images before, the effect is marginal. I would remove that capability to fix the alternative_text problem.

gcamp:
I've gone ahead and changed the proposal to only allow one image.

gcamp:
Let me know if there are more comments, if not I'll start an other vote by the end of this week.

gcamp:
I think we're again ready for a vote. This vote is for adding an experimental field in of images in alerts.

Voting ends on 2021-11-26 at 23:59:59 UTC.

flocsy:
What about adding the required url-encoding to the url field? IMHO it should be in the standard and not a free choice.

flocsy:
What about adding the required url-encoding to the url field? IMHO it should be in the standard and not a free choice.

flocsy:
What about adding the required url-encoding to the url field? IMHO it should be in the standard and not a free choice.

gcamp:
@flocsy I added the media_type in the TranslatedImage, is that what your question is about?

flocsy:
> @flocsy I added the media_type in the TranslatedImage, is that what your question is about?

No, I'm not talking about the media type of the file that the url points to, but the encoding in the value of the url field. i.e:
"http://example.com/file name" vs "http://example.com/file+name" vs "http://example.com/file%20name". Or more interesting example of non-ASCII character in the url: "http://example.com/Fővám tér" vs "http://example.com/F%C5%91v%C3%A1m%20t%C3%A9r" (url-encode) vs "http://example.com/RsWRdsOhbSB0w6ly" (base64), etc. IMHO the standard should explicitly define what to use to prevent confusion for both producers and consumers

gcamp:
I updated the URL description for escaping, using the same description that GTFS static uses.

sam-hickey-ibigroup:
+1 IBI Group

flocsy:
+1

paulswartz:
+1

skinkie:
+1 OpenGeo

stevenmwhite:
+1 GMV

lauramatson:
+1 Metro Transit (Minneapolis/St. Paul)

doconnoronca:
+1 TransSee

gcamp:
The voting period is finished and result are : 7 for, 0 against. The proposition passes! 🎉

For the maintainer merging, if possible squash and merge because the git history is not clean in this PR.

PR #282, opened 2021-08-26
Update stop_name Best Practices

scmcca:
As flagged in #273 and #281 by @flocsy: proposal to disambiguate `stop_name` best practices to reflect actual published naming conventions and handling translations for local/foreign riders in `translations.txt`.

PR #285, opened 2021-09-17
Add best practices for cross street `stop_name`

scmcca:
As requested on a continued discussion at #282. Add best practices for `stop_names` named after cross streets.


skinkie:
Would you support it by just introducing a new field for this?

scmcca:
I'm not sure I understand. Are you proposing an additional stop name field in stops.txt specifically for cross street stop names in place of providing a recommendation in the existing `stop_name` field?

skinkie:
Exactly. NeTEx for example has a _CrossingRoad_.

paulswartz:
@mbta has `on_street` and `at_street` as additional fields in our `stops.txt`: https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#stopstxt

scmcca:
@paulswartz Are `on_street` and `at_street` used to generate stops names after cross streets? If so, what happens to `stop_name` in these cases? If not, how are these fields used?

gcamp:
`stop name` should be itself enough to represent a stop, if adding `on_street` makes that untrue because of unwanted repetition I don't think it's a good idea. 

On the proposition itself, I'm torn. This feels too much into the "best practice" realm. Is that the job of the spec to specify that?

paulswartz:
They're separate from the `stop_name`, but generally both are present in the name of the stop. We use "@" for stops where the `at_street` is on the same side of the `on_street` as the stop, and "opp" when the `at_street` is across the `on_street`. 

Some examples from our current feed:
| stop_name | on_street | at_street |
| --- | --- | ---|
| Magazine St @ Green St | Magazine Street | Green Street |
| River St opp Blackstone St | River Street | Blackstone Street |
| 5 Charles Park Rd | Charles Park Road | Veterans Of Foreign Wars Parkway |

scmcca:
@gcamp This is part of a larger RFC 2119 conversation. 

What is the difference between RFC spec recommendations ("should") and best practice recommendations? I see your point that a line may be drawn when we start dealing with special modelling cases such as cross streets, but can keep higher level recommendations as RFC spec recommendations. 

The strength to having recommendations in the spec is that it is official and canonical with GTFS, as opposed to a separate set of potentially arbitrary best practice documentation (which might reduce the amount of data consumers actually following them). Open to hearing other strategies to manage this.

doconnoronca:
Adding something like on_street and cross_street fields was something I was going to suggest. I also had the idea of a stop_side field to indicate if the stop is near side, far side, opposite or maybe north, south, east or west side.

Redundancy is a concern, but it allows a language independent way of describing stop locations which is useful for translations for different human languages and computer data structures. 

You can see in the TransSee stop list what can be done by separating the street names: https://www.transsee.ca/stoplist?a=mbta&r=59

skinkie:
In my opinion a stop_name should be equal what is physically on the stop, the rest is a matter of attributes for positioning.

stevenmwhite:
I agree with @skinkie and @gcamp but will re-state it in a slightly different way.

This is starting to feel like dictating how an agency communicates about or brands their service, rather than defining a format to exchange that information with other computer systems.

Various agencies may have various rules (or, sometimes, no rules at all) for how they name their stops -- and it's not the job of the GTFS spec to tell them how to do it. Stop name should be formatted however the agency formats it in their literature, signage, or other communication.

e-lo:
> Stop name should be formatted however the agency formats it in their literature, signage, or other communication.

Yes but...I think it is very useful to offer a "best practice" (which I believe is what is being proposed here) when dealing with some extremely common cases, including the case of "Main Street at East Road" or "Main Street opposite East Road".  

There is also a best practice to be said about spelling things out for an annunciator unless you add a tts field.

stevenmwhite:
> Yes but...I think it is very useful to offer a "best practice" (which I believe is what is being proposed here) when dealing with some extremely common cases, including the case of "Main Street at East Road" or "Main Street opposite East Road".
> 
> There is also a best practice to be said about spelling things out for an annunciator unless you add a tts field.

I agree! But in both of these cases I don't think that the data spec is the place for a best practice like this.

If an agency doesn't spell a name out in their literature elsewhere, or uses strange local abbreviations or acronyms, I don't think the data spec is the place to suggest they change that. Best practice documentation generally sits _next to_ rather than inside the data spec and something may be completely valid from the standpoint that computer systems can transfer the data to/from each other even if the best practice (which might help make it accessible to and understandable by a wider variety of users) isn't followed.

e-lo:
> I don't think that the data spec is the place for a best practice like this.

Where would a best practice exist, then? I thought there had been a global decision to move the best practices into the spec so there was "one place to look".  I want to make it as easy as possible for everyone to know what the must/should/could for GTFS is. Having people go to multiple locations is asking for it not to be considered.

stevenmwhite:
>I thought there had been a global decision to move the best practices into the spec so there was "one place to look".

I may have missed that. Is the idea that the Best Practices repositories would be deprecated and the info would be moved into here? If I did, I apologize!

If that is the case, however, I think it's worth considering whether a best practice is one of "how to model your information" (in which case I'd say it's fair game for inclusion) or one of "what information to model." I don't think it's the place of the GTFS spec to give agencies best practices for _how to name their stops_ (what information to model) but that the spec should accommodate however agencies name their stops. Advising on how to name stops is the job of a transit planner or consultant, not a data specification.

While _I personally_ agree that it makes more sense to list the street the bus is on first, followed by the cross street, we have customers who don't do it that way for whatever reason. Sometimes it's "list the more major street first" in general and sometimes it's random and makes no sense -- but the name of their stop is what it is.

If an agency has signs, maps, schedules, or other information printed with stops named a particular way, then GTFS best practice should be limited to "match the way you name this stop on other resources, so that third party apps are consistent with the other information your riders have access to."

doconnoronca:
Perhaps adding the additional columns would be a way to seperate the streets without the need to prescribe how the stop_name is formatted.

flocsy:
There's a typo: "should contain**s** the name of the boarding area" it should be "should contain the name of the boarding area"

gtfs/spec/en/reference.md
```diff
@@ -143,7 +143,7 @@ File: **Required**
...
...r> The `stop_name` should match the agency's rider-facing name for the location as printed on a timetable, published online, or represented on signage. For translations into other languages, use `translations.txt`.<br><br>Stops named after cross streets should separate street names by a single consistent separator within `stops.txt` (e.g., "at", "and", "&", "/"). The first street name should be the street that the vehicle is travelling on at the time of the stop, and the second street name should be the cross street (i.e., < first street name > < separator > < second street name >).<br><br>When the location is a boarding area (`location_type=4`), the `stop_name` should contains the name of the boarding area as displayed by the agency. It could be just one letter (like on some European intercity railway stations), or text like “Wheelchair boarding area” (NYC’s Subway) or “Head of short trains” (Paris’ RER).<br><br>Conditionally Required:<br>- **Required** for locations which are stops (`location_type=0`), stations (`location_type=1`) or entrances/exits (`location_type=2`).<br>- Optional for locations which are generic nodes (`location_type=3`) or boarding areas (`location_type=4`).|
```
skinkie, line 146:
```suggestion
|  `stop_name` | Text | **Conditionally Required** | Name of the location.<br><br> The `stop_name` should match the agency's rider-facing name for the location as printed on a timetable, published online, or represented on signage. For translations into other languages, use `translations.txt`.<br><br>Stops named after cross streets should separate street names by a single consistent separator within `stops.txt` (e.g., "at", "and", "&", "/"). The first street name should be the street that the vehicle is travelling on at the time of the stop, and the second street name should be the cross street (i.e., < first street name > < separator > < second street name >).<br><br>When the location is a boarding area (`location_type=4`), the `stop_name` should contain the name of the boarding area as displayed by the agency. It could be just one letter (like on some European intercity railway stations), or text like “Wheelchair boarding area” (NYC’s Subway) or “Head of short trains” (Paris’ RER).<br><br>Conditionally Required:<br>- **Required** for locations which are stops (`location_type=0`), stations (`location_type=1`) or entrances/exits (`location_type=2`).<br>- Optional for locations which are generic nodes (`location_type=3`) or boarding areas (`location_type=4`).|
```

scmcca:
@flocsy @skinkie
> There's a typo: "should contains the name of the boarding area" it should be "should contain the name of the boarding area"

That was an [existing typo](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt). I agree it should be fixed but I don't think fixing typos from the spec is in scope here.

flocsy:
> @flocsy @skinkie
> 
> > There's a typo: "should contains the name of the boarding area" it should be "should contain the name of the boarding area"
> 
> That was an [existing typo](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stopstxt). I agree it should be fixed but I don't think fixing typos from the spec is in scope here.

Maybe not when it gets to fixing a random typo, but when you anyway change a sentence then I think we should fix. I haven't just stumbled uppon this typo but saw it in the diff. Also I don't see why shouldn't we fix it. Is there any disagreement in how this part of the sentence should be?

PR #290, opened 2021-10-26
GTFS-Occupancies (part 2)

scmcca:
Due to GitHub workflow and CLA issues, this pull request is a continuation of discussions at #240.

GTFS-Occupancies is an extension proposal to describe usual or expected vehicle occupancy levels in GTFS Schedule.

As occupancies can currently be described in GTFS Realtime using [`occupancy_status`](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-occupancystatus) and [`occupancy_percentage`](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-vehicleposition), GTFS-Occupancies aims to complement the availability of crowdedness information by providing static predictions for future trips based on historical trends, which can help riders plan trips based on their crowdedness preferences and comfortability.

To be resolved:
- [ ] Officializing an occupancy indicator in GTFS Realtime: Currently, both `OccupancyStatus` and `occupancy_percentage` are experimental. Ideally, we want consistency between static and realtime occupancy indicators. Whichever becomes official in GTFS Realtime will be the one used here.
- [x] Elaborate GTFS-Occupancies to support frequency-based service. Some thinking on this [here](https://github.com/google/transit/pull/240#discussion_r713393734).

TfNSW has published a dataset: https://github.com/google/transit/pull/240#issuecomment-921526701.


paulswartz:
> I do have a bigger question. Does the standard currently state anything regarding a requirement that the files must be bundled as a single zip archive? In our case we would like to add the latest predictions we receive but forcing hundreds of megabytes of unchanged data data just does not make sense.

@skinkie we were also thinking about this @mbta. We might include it in our GTFS (which updates a few times a week) as well as a standalone file (updated as frequently as necessary).

skinkie:
> @skinkie we were also thinking about this @mbta. We might include it in our GTFS (which updates a few times a week) as well as a standalone file (updated as frequently as necessary).

That is probably going to be our approach as well, but I think generically we should state something about bundling.

PR #295, opened 2021-11-24
In seat transfers

gcamp:


Marilyn-Boston (commented):
Note: I'm a contract technicaL writer at Google working on transit documentation, and I don't have a CLA, so I can't submit a PR myself. I noticed a few minor errors that should be fixed. I hope someone can submit the following changes as a PR for me:

- The word _preferred_ is misspelled as _preffered_ twice (lines 109 and 206). 
- The word _the_ is repeated in 364.  
- _Usage of `block_id` is now deprecated_ should be changed to _Usage of `block_id` to implicitly enable in-seat transfer is now deprecated_ in lines 198 and 206.  
- Change _trip to trip transfers_ to _trip-to-trip transfers_ in lines 198 and 206. 
- Change _geographically close from_ to _geographically close to_ in line 364.

skinkie:
block_id has absolutely nothing to do with what you want to achieve. Don't touch it.

Marilyn-Boston:
> block_id has absolutely nothing to do with what you want to achieve. Don't touch it.

Most of my comment was about correcting a few writing errors. The wording change I requested for the *block_id* statement wasn't intended to suggest a change to how *block_id* works but to clarify it. The original statement seemed to imply that *block_id* should no longer be used at all, when in fact it still can be used for its original purpose to indicate which trips are run by the same vehicle.

Of course, I may have miunderstood the whole issue. In any case, I think the spec should be clearer.

PR #301, opened 2022-01-06
Clarification: Trip schedule_relationship and start_time for UNSCHEDULED trips

barbeau:
Per the discussion on the Google Group at https://groups.google.com/g/gtfs-realtime/c/ujeIp2UTnCs, TripDescriptor `start_time` can refer to UNSCHEDULED trips as well, so it shouldn't reference a "schedule" in it's global definition.

This change removes the word "schedule" from the initial definition and also adds an explicit reference to the `UNSCHEDULED` enumeration to further clarify it's use.

**EDIT** Further clarifications surrounding TripDescriptor.schedule_relationship have been added as well to make it clear that SCHEDULED is the default.


barbeau:
Per additional comments in https://groups.google.com/g/gtfs-realtime/c/ujeIp2UTnCs/m/nJp8ubxiFQAJ, I've updated this PR to include more clarifications surrounding the use of TripDescriptor.schedule_relationship in the .proto and reference, including:
* Declare the default SCHEDULED enum explicitly in the .proto and mention it in the reference. 
* Remove the outdated description above the TripDescriptor.ScheduleRelationship enum in the .proto to avoid referencing "schedule" and ADDED.

To be clear this doesn't change any existing behavior, just makes it clearer to new readers of the spec and .proto what the behavior is.

barbeau:
Any additional thoughts on this proposal? If not I'd like to call for a vote in the next few days.

gcamp:
We don't support frequency-based real time, but the changes looks good to me. 

Do you know of any agency that has a frequency-based real time?

evantrillium:
@gcamp we supported GTFS-rt w/ Swiftly for washingtonparkshuttle-or-us but I believe that one is now deprecated. We also were supporting it with qline-mi-us, but they have since switched to schedule-based GTFS. @ericouyang do you know of any others that Swiftly is currently supporting?

barbeau:
We were publishing a feed for our USF Bull Runner shuttle but that's since been deprecated as well.

@stevenmwhite Is GMV Syncromatics publishing TripUpdates for any exact_times=0 trips (true frequency-based)?

barbeau:
I'd like to call for a vote on this clarification. Voting will be open until Friday 2022-03-11 23:59:59 UTC.

Please vote by commenting with a +1, -1, or abstain in comments below.

gtfs-realtime/spec/en/reference.md
```diff
@@ -439,14 +439,14 @@ TripDescriptor.route_id cannot be used within an Alert EntitySelector to specify
...
+|------------------|------------|----------------|-------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
skinkie, line 443: These changes are unrelated to the clarification. Can you prevent them in the future?

ericouyang:
Thanks for tagging me in @evantrillium, though I'm actually no longer at Swiftly.

@meremagee, @shedlickj - Can you help share any details about agencies with frequency-based predictions and tag in the appropriate folks to vote on the proposed clarification?

gtfs-realtime/spec/en/reference.md
barbeau: There are some changes here in the text related to the clarification. GitHub doesn't always do great rendering diffs of tables. Try clicking on the "Display the rich diff" button on the referenced.md file and you should see an easier to read version (see red circle in image for button, and strikethrough change starting at the bottom):

image

barbeau: Oh, maybe you're referring to the expanded table header? If so, yes, not sure why that happened. Maybe artifact of the editor I was using for markdown. Will try to avoid next time.

barbeau: I reverted the table header but I'm fighting IntelliJ to fix the other line endings. I'll work on reverting the changes to the pipe characters on the remaining lines early next week.

barbeau: Ok, I believe I've removed the unrelated formatting changes.

gtfs-realtime/spec/en/reference.md
```diff
@@ -439,14 +439,14 @@ TripDescriptor.route_id cannot be used within an Alert EntitySelector to specify
 
 #### Fields
 
-| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
```
IvanVolosyuk, line 442: Are there any changes in this section? If it is just moved, can we do that in separate change?

gtfs-realtime/spec/en/reference.md
```diff
@@ -439,7 +439,7 @@ TripDescriptor.route_id cannot be used within an Alert EntitySelector to specify
...
+| _**Field Name**_ | _**Type**_ | _**Required**_ | _**Cardinality**_ | _**Description**_ |
```
IvanVolosyuk, line 442: Can you squash all your commits? It is pretty hard to follow what is actually changed? And if the table is just moved to another place and reformatted put it in a separate commit, please.

IvanVolosyuk (changes requested):

barbeau:
>Can you squash all your commits? It is pretty hard to follow what is actually changed? And if the table is just moved to another place and reformatted put it in a separate commit, please.

@IvanVolosyuk Sure, I just squashed all the commits into https://github.com/google/transit/pull/301/commits/53a158a9ab18f0380726a02efaa4d1765e0f2257, and this is only content changes (no formatting changes or moved text). It's the exact same content as the head of the branch that this PR was previously based on.

Note that GitHub doesn't always do great rendering raw diffs of tables in markdown. Try clicking on the "Display the rich diff" button on the referenced.md file and you should see an easier to read diff version (see red circle in image for button, and strikethrough change starting at the bottom):

image

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -818,7 +817,7 @@ message TripDescriptor {
     // NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
     DUPLICATED = 6;
   }
-  optional ScheduleRelationship schedule_relationship = 4;
+  optional ScheduleRelationship schedule_relationship = 4 [default = SCHEDULED];
```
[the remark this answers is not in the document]
barbeau: I agree, it is. The reason for the change is that it hasn't been clear to newcomers in the spec, especially those that aren't familiar with the rules of protobufs, what the default value is here. Also, we should change this to be consistent, as StopTimeUpdate.schedule_relationship already has an explicit default defined in the .proto:

https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L252-L253

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -774,12 +774,11 @@ message TripDescriptor {
...
   optional string start_date = 3;
 
-  // The relation between this trip and the static schedule. If a trip is done
-  // in accordance with temporary schedule, not reflected in GTFS, then it
-  // shouldn't be marked as SCHEDULED, but likely as ADDED.
+  // The relation between this trip and the trip declared in static data.
   enum ScheduleRelationship {
     // Trip that is running in accordance with its GTFS schedule, or is close
-    // enough to the scheduled trip to be associated with it.
+    // enough to the scheduled trip to be associated with it.  This enumeration must NOT be used for any trip that is
```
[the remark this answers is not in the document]
barbeau: >The UNSCHEDULED has a comment mentioning that it is experimental and subject to change.

So there are two different schedule_relationship.UNSCHEDULED enumerations:
1. [`TripDescriptor.schedule_relationship.UNSCHEDULED`](https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L794-L796) - This field is NOT experimental. This change is related to this field.
2. [`StopTimeUpdate.schedule_relationship.UNSCHEDULED`](https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L252-L253) - This field is experimental, but only because we haven't seen anyone publish it in the wild. Conceptually it exists for the exact same reason as `TripDescriptor.schedule_relationship.UNSCHEDULED`.

>This will cause us to lose realtime for all frequency based trips, as we currently don't handle UNSCHEDULED. 

From your response it sounds like there are producers labeling updates for exact_times=0 trips with `schedule_relationship.SCHEDULED` - is that right? If so I'd argue that already goes against the spec - it's pretty clear from the existing spec that those updates should be labeled with UNSCHEDULED. 

That being said, this is the problem we're trying to solve with this PR, so if there are widespread issues with existing producers doing this wrong then we might need a migration path forward to get everyone in-line with the spec. Thoughts on this are welcome.

barbeau:
So the original voting window closed on this with no votes (just comments above).

Before I open the vote again, @IvanVolosyuk or @skinkie after reviewing my replies do you have any other questions about this proposal?

gtfs-realtime/proto/gtfs-realtime.proto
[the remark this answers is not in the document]
IvanVolosyuk: "Trips with ScheduleRelationship=UNSCHEDULED must also set all 
 stopTimeUpdates.ScheduleRelationship=UNSCHEDULED"
It looks non-experimental field suggests to use experimental feed? It is very confusing. Agency setting STU.schedule_relationship = UNSCHEDULED will be surprised that their updates are dropped because we actually don't support this experimental value and require STU.SCHEDULED for such STUs. There is actually a schedule for stop times even for exact_times=0. It just happen that the start of the schedule is shifted to the trip 'start_time' of the trip. I believe agency can express relationship to the schedule in terms of STU.{arrival,departure}.delay. That schedule will also be used if agency haven't specified any STU for a stop and the stop arrival/departure times will be deduced from it.

gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -746,14 +746,14 @@ message TripDescriptor {
   // direction of travel for trips this selector refers to.
   optional uint32 direction_id = 6;
 
-  // The initially scheduled start time of this trip instance.
```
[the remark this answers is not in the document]
barbeau: I'm intentionally trying not to use the word "scheduled" here because the `TripDescriptor` can refer to both scheduled (normal GTFS without frequencies.txt, frequencies.txt exact_times=1) and frequency-based (frequencies.txt exact_times=0) trips. I'm currently proposing that we change the text to just:

>The start time of this trip instance.

...and let the remainder of the paragraph describe the conditions under which it's scheduled or frequency-based.

IMHO it's already confusing enough that we have scheduled trips represented in frequencies.txt with exact_times=1 trips, so I'd like to be more precise in the language and not use "schedule" when referring to exact_times=0 trips.

gtfs-realtime/proto/gtfs-realtime.proto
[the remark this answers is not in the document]
barbeau: >Agency setting STU.schedule_relationship = UNSCHEDULED will be surprised that their updates are dropped because we actually don't support this experimental value and require STU.SCHEDULED for such STUs. 

How about we put some language here to make it clearer that consumers may choose to continue to process this type of data while producers transition to what the spec says? Some like what's in [TripDescriptor](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-tripdescriptor):

"If any enumeration other than UNSCHEDULED is used for exact_times=0 trips it is considered an error and the entity containing the erroneous TripDescriptor may be discarded by consumers."

The "may" here allows you to continue consuming this if you want, even if it doesn't adhere to the spec.

>There is actually a schedule for stop times even for exact_times=0. It just happen that the start of the schedule is shifted to the trip 'start_time' of the trip. I believe agency can express relationship to the schedule in terms of STU.{arrival,departure}.delay. That schedule will also be used if agency haven't specified any STU for a stop and the stop arrival/departure times will be deduced from it.

tl;dr - I've opened an issue [here](https://github.com/google/transit/issues/314) to discuss this specifically - please comment!

For this PR I'd like to stick to clarifying the existing text without getting into those details.

barbeau:
Not stale

Issue #291, opened 2021-11-09
Inconsistency documentation https://developers.google.com/transit/gtfs-realtime/guides/vehicle-positions#vehiclestopstatus


sven4all:
The documentation in https://developers.google.com/transit/gtfs-realtime/guides/vehicle-positions#vehiclestopstatus is not consistent with the documentation in the .proto files https://developers.google.com/transit/gtfs-realtime/gtfs-realtime-proto and gtfs.org 

On https://developers.google.com/transit/gtfs-realtime/guides/vehicle-positions#vehiclestopstatus IN_TRANSIT_TO is described as "the referenced stop is the next stop for the vehicle - default" while in the .proto file it's described the following:
```proto
enum VehicleStopStatus {
    // The vehicle is just about to arrive at the stop (on a stop
    // display, the vehicle symbol typically flashes).
    INCOMING_AT = 0;

    // The vehicle is standing at the stop.
    STOPPED_AT = 1;

    // The vehicle has departed and is in transit to the next stop.
    IN_TRANSIT_TO = 2;
  }
  // The exact status of the vehicle with respect to the current stop.
  // Ignored if current_stop_sequence is missing.
  optional VehicleStopStatus current_status = 4 [default = IN_TRANSIT_TO];
```
current_stop_sequence is described:
```proto
// The stop sequence index of the current stop. The meaning of
  // current_stop_sequence (i.e., the stop that it refers to) is determined by
  // current_status.
  // If current_status is missing IN_TRANSIT_TO is assumed.
  optional uint32 current_stop_sequence = 3;
```

What is the correct way to interpreted it? Should the current_stop_sequence be updated directly after the bus is departed from the previous stop (what https://developers.google.com/transit/gtfs-realtime/guides/vehicle-positions#vehiclestopstatus IN_TRANSIT_TO  suggests) or should it be updated when the vehicle arrives at the next stop what the .proto file and gtfs.org are suggesting.

barbeau:
IMHO this is another example of the confusion that results when we have multiple GTFS specs online, specifically the developers.google.com page, which for some reason doesn't always mirror the canonical spec reference on GitHub.

Here's the canonical spec and .proto reference on GitHub:
* Spec - https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-vehiclestopstatus
* .proto reference - https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L421

[...]

PR #320, opened 2022-04-20
clarification: VehicleStopStatus values

scmcca:
As discussed in #291. (cc: @sven4all)

### DIFF
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -261,7 +261,7 @@ Realtime positioning information for a given vehicle.
 |-------------|---------------|
 | **INCOMING_AT** | The vehicle is just about to arrive at the referenced stop (on a stop display, the vehicle symbol typically flashes). |
 | **STOPPED_AT** | The vehicle is standing at the referenced stop. |
-| **IN_TRANSIT_TO** | The vehicle has departed and is in transit to the referenced stop. |
+| **IN_TRANSIT_TO** | The vehicle is in transit to the referenced stop. |
 
 ## _enum_ CongestionLevel
 
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -259,9 +259,9 @@ Realtime positioning information for a given vehicle.
 
 | _**Value**_ | _**Comment**_ |
 |-------------|---------------|
-| **INCOMING_AT** | The vehicle is just about to arrive at the stop (on a stop display, the vehicle symbol typically flashes). |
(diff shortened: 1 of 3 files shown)


barbeau (approved):

barbeau:
+1 for USF

scmcca:
@sven4all Good point, there is some redundancy. We can simply take out a portion of the phrase so it reads:

> The vehicle ~~has departed and~~ is in transit to the referenced stop.

IMHO "in transit" is more intuitive than "en route". Thoughts?

sven4all:
In transit to is also fine for me.

e-lo:
LGTM

gtfs-realtime/spec/en/reference.md
```diff
@@ -259,9 +259,9 @@ Realtime positioning information for a given vehicle.
 
 | _**Value**_ | _**Comment**_ |
 |-------------|---------------|
-| **INCOMING_AT** | The vehicle is just about to arrive at the stop (on a stop display, the vehicle symbol typically flashes). |
-| **STOPPED_AT** | The vehicle is standing at the stop. |
-| **IN_TRANSIT_TO** | The vehicle has departed the previous stop and is in transit. |
+| **INCOMING_AT** | The vehicle is just about to arrive at the referenced stop (on a stop display, the vehicle symbol typically flashes). |
+| **STOPPED_AT** | The vehicle is standing at the referenced stop. |
+| **IN_TRANSIT_TO** | The vehicle is in transit to the referenced stop. |
```
barbeau, line 264: We should mark `IN_TRANSIT_TO` as the default value here too in the reference.

barbeau:
While we're here, looking back at #291, to really fix the ambiguity discussed there I think we should be more explicit about the expected behavior for `current_status` in coordination with `current_stop_sequence`.

What do you think about adding something like the below explanation for how current_stop_sequence should be updated?

>If `current_status` is provided, current_stop_sequence should be updated to the next stop after the bus departs from the previous stop.
>
>Specifically, this should be the sequence of the vehicle going from stop_sequence 2 to 3 if both fields are provided:
>
>current_stop_sequence = 2, current_status = INCOMING_AT
current_stop_sequence = 2, current_status = STOPPED_AT
current_stop_sequence = 3, current_status = IN_TRANSIT_TO // When vehicle departs stop 2
>
>If current_status is not provided, then current_stop_sequence should be updated based on the definition of IN_TRANSIT_TO, which means that the value only changes after the vehicle departs a stop.

scmcca:
Pardon my lag. The vote ended on 2022-05-03 at 23:59:59 UTC. With only one +1 and outstanding feedback, the vote is inconclusive. I will rework the proposal based on the feedback above. Thanks!

PR #327, opened 2022-05-12
translations.txt: specify which translation takes precedence over another

timMillet:
## Current issue

Translations can be defined in two ways using translations.txt:
- By referring to a particular row using record_id and record_sub_id;
- By referring to a particular value using field_value;

The specification does not forbid translating the same value twice, using both referencing methods. In which case, there is no stated logic over which translation should take preference. This leads to a situation where the data consumer needs to make a discretionary choice, behavior that is not desired. Example:

routes.txt
```
route_id,route_short_name
route_id_1,subway
```
translations.txt
```
table_name,field_name,language,translation,record_id,record_sub_id,field_value
routes,route_short_name,fr,métro,route_id_1,,
routes,route_short_name,fr,train,,,subway
```

## Solution in this PR
The translation made with the referencing method (`record_id`, `record_sub_id`) takes precedence.

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -458,6 +458,8 @@ Primary key (none)
 
 The file contains information about the dataset itself, rather than the services that the dataset describes. In some cases, the publisher of the dataset is a different entity than any of the agencies.
 
+If both referencing methods (`record_id`, `record_sub_id`) and `field_value` are used to translate the same value in 2 different rows, the translation provided with (`record_id`, `record_sub_id`) takes precedence.
+
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ | ------ |
 |  `feed_publisher_name` | Text | **Required** | Full name of the organization that publishes the dataset. This may be the same as one of the `agency.agency_name` values. |


westontrillium:
Instead of establishing a hierarchy, wouldn't it be simpler to **forbid** translating the same value in one language twice? Is there a use case for doing so?

timMillet:
@westontrillium What you suggest could be another solution and we thought about that. We preferred establishing a hierarchy because it’s a simple rule, it’s efficient because you can give a general translation that you can override in special cases, and it still meets the requirement of being unambiguous / not discretionary.

timMillet:
It looks like the proposal does not raise any other concerns. I’m opening a vote on this PR.
- translations.txt is already produced and consumed (see the original [PR#180](https://github.com/google/transit/pull/180#issuecomment-566201378)). Transit is interested in consuming it once this PR is accepted.
- Voting ends on 2022-06-02 at 23:59:59 UTC.

NB: The vote is also open on [PR#326](https://github.com/google/transit/pull/326) which fixes another issue in translations.txt.

brodyFlannigan:
+1 Brody Flannigan Transit Data

laurentg:
+1 Mecatran

botanize:
+1 Metro Transit

timMillet:
The voting period has ended. With 5 votes in favor and no votes against, this proposal is adopted.
Thanks to anyone who participated and voted!

PR #329, opened 2022-05-25
Add GTFS Schedule use case data examples

scmcca:
Add GTFS Schedule use case data examples from https://gtfs.org/schedule/data-examples/ to the google/transit repo as the source-of-truth location. Adding these examples to google/transit allows for open editing by the GTFS community and synced updates on GTFS.org. Plan to add more use case examples to this document in the near future.

Originally authored by @timMillet (correct me if I'm wrong!)


timMillet:
LGTM!

PR #330, opened 2022-05-25
Update Realtime README

scmcca:
Add instructions on where to participate in discussions in parity with https://gtfs.org/realtime/. Changes to this document are synced on GTFS.org.


barbeau (approved):
👍

PR #331, opened 2022-05-25
Add GTFS Realtime code examples

scmcca:
Add GTFS Realtime code examples in sync with https://gtfs.org/realtime/code-examples/.

### DIFF
--- a/gtfs-realtime/spec/en/examples/code-examples/dotnet.md
+++ b/gtfs-realtime/spec/en/examples/code-examples/dotnet.md
@@ -1,0 +1,23 @@
+# .NET
+To use the `gtfs-realtime-bindings` .NET classes in your own project, you need to first install the module from the [NuGet repository](https://www.nuget.org/packages/GtfsRealtimeBindings/).
+
+```
+Install-Package GtfsRealtimeBindings
+```
+
+The following code snippet demonstrates downloading a GTFS realtime data feed from a particular URL, parsing it as a `FeedMessage` (the root type of the GTFS realtime schema), and iterating over the results.
+
+```vb
+using System.Net;
+using ProtoBuf;
+using transit_realtime;
+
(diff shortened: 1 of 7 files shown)


barbeau:
@scmcca Thanks for the PR! It looks like these examples are all based on the [gtfs-realtime-bindings library](https://github.com/MobilityData/gtfs-realtime-bindings). I'm hesitant to add a copy of these here, as they will easily become outdated when the library changes. In fact, it looks like there are already some code differences between the text in this PR and the instructions on the gtfs-realtime-bindings - I'll flag a few of these I saw, but a more thorough review is required. 

Maybe in this repo we can just point to the docs maintained on the gtfs-realtime-bindings repo?

gtfs-realtime/spec/en/examples/code-examples/dotnet.md
```diff
@@ -0,0 +1,23 @@
...
+
+```vb
+using System.Net;
+using ProtoBuf;
+using transit_realtime;
+
+WebRequest req = HttpWebRequest.Create("URL OF YOUR GTFS-REALTIME SOURCE GOES HERE");
+FeedMessage feed = Serializer.Deserialize<FeedMessage>(req.GetResponse().GetResponseStream());
+
+foreach (FeedEntity entity in feed.entity) {
```
barbeau, line 18: gtfs-realtime-bindings says this should be:

```dotnet
foreach (FeedEntity entity in feed.Entities) {
```

See https://github.com/MobilityData/gtfs-realtime-bindings/tree/master/dotnet#example-code

barbeau:
>I prefer pulling in the more relevant ones from the [gtfs-realtime-bindings library](https://github.com/MobilityData/gtfs-realtime-bindings) onto GTFS.org, instead of duplicating them here on google/transit. In which case I'll close this PR and make the updates on GTFS.org. Thoughts?

👍

PR #340, opened 2022-07-06
GTFS-rt : wheelchair access update

flaktack:
This pull request add the ability to specify a `wheelchair_accessible` value in a `TripUpdate.VehicleDescriptor` that updates the value found in the static GTFS. Since `VehicleDescriptor` is shared with `VehiclePosition` the value may also be specified there.

This pull is intentionally small, it doesn't support all cases of accessibility values and instead relies on what is already accepted in the GTFS.

_This is a rebase+squash of https://github.com/google/transit/pull/87, along with moving the field to `VehicleDescriptor`._

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -843,6 +843,26 @@ message VehicleDescriptor {
   // The license plate of the vehicle.
   optional string license_plate = 3;
 
+  enum WheelchairAccessible {
+    // The trip doesn't have information about wheelchair accessibility.
+    // This is the **default** behavior. If the static GTFS contains a
+    // _wheelchair_accessible_ value, it won't be overwritten.
+    NO_VALUE = 0;
+
+    // The trip has no accessibility value present.
+    // This value will overwrite the value from the GTFS.
+    UNKNOWN = 1;
+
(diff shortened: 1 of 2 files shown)


leonardehrenfried:
Seems that we have 3 producers and 2 consumers which was both lacking last time this was proposed.

leonardehrenfried:
@flaktack I just re-read the rules about opening a vote on the mailing list. It looks like you as the advocate of the proposal are supposed to send a message to https://groups.google.com/g/gtfs-changes/ to open the vote.

Or can anyone do that?

flaktack:
per the _Specification Amendment Process_: 7 days of discussion have passed and it looks like there is consensus. I am now calling a vote.

Voting ends 2022-07-21 at 23:59:59 UTC.

The mailing list thread is at https://groups.google.com/g/gtfs-realtime/c/wDUl43e44js

BodoMinea:
If it wasn't obvious from my previous reply, +1 FlashWeb IT as a producer.

skinkie:
+1 OpenGeo

hannesj:
+1 from Entur

jfabi:
+1 from the MBTA

flaktack:
The vote has ended with 7 votes in favor and 0 opposed, with which this PR has passed. Thanks for voting!

Issue #341, opened 2022-08-05
GTFS-Fares v2 - Second implementation


omar-kabbani:
## Update on GTFS-Fares v2

Hello everyone, this issue is to present MobilityData's next steps on GTFS-Fares v2 to keep the community updated as well as to hear everyone's thoughts on this. 

We are very happy to announce that MobilityData will continue working on GTFS-Fares v2 by adding more functionalities into the specification. This comes as a successor to the [base implementation](https://github.com/google/transit/pull/286), which added fare legs, fare areas, complex transfers, and tickets. We are now working on the second implementation of GTFS-Fares v2, which covers the following features:

-  Variable pricing by time of day, or by day of week, month, or year
-  Fare containers  
-  Allowing multiple routes to behave as a single route (transfer rules ignored)
-  Zone-based fares
-  Inter-agency transfers (using fare_leg_rules.transfer_only, proposal [here](https://docs.google.com/document/d/18yWhwR89pQp48VuBNPXK0djLXOtdjWhgC2h6jCC8WPM/edit#heading=h.bhbxvegn4na))
    

The data producer and consumer for the second implementation of GTFS-Fares v2 are:
-   Consumer: [Apple](https://www.apple.com/)
-   Producer: [Interline](https://www.interline.io/)

### Timeline:

We are planning to open separate pull requests for features that do not overlap to keep focus on the discussions and facilitate the consensus building process.

**Week of August 1:**

Open pull requests for:

 - [x] Variable pricing by time of day, or by day of week, month, or year
 - [x] Fare containers

The above two features are straightforward, and MobilityData does not foresee any challenges to reaching consensus within the community. We are comfortable with opening pull requests and going through the GTFS change process leading up to a vote at this time. 

For the remaining features, more time is needed for conversations to close out open issues and come to consensus. Hence, below is the plan for moving forward with regards to these features:
 
Open GitHub issues for:

 - [x] Allowing multiple routes to behave as a single route (transfer rules ignored)
 - [x] Zone-based fares
 - [x] Inter-agency transfers (using fare_leg_rules.transfer_only, proposal [here](https://docs.google.com/document/d/18yWhwR89pQp48VuBNPXK0djLXOtdjWhgC2h6jCC8WPM/edit#heading=h.bhbxvegn4na))

**Week of August 8:**

- Discussions with the community on the three open GitHub issues.

**Week of August 15:** 

- MobilityData will share a summary of the GitHub discussions.

**Week of August 22:** 

[...]

PR #343, opened 2022-08-05
Add variable fares by time or day

omar-kabbani:
Hi everyone,   MobilityData is moving forward with the second iteration of GTFS-Fares v2, for more information about the overall plan, you can check issue https://github.com/google/transit/issues/341.

This pull request covers fares that vary by time and day, which is a section of the entire [GTFS-Fares v2 proposal](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.mhjuwqk2h5r3).

Fares can vary based on the time of day (described using `timeframe_id`) and day of week/year (described using `service_id`).

- Add new file; `timeframes.txt`, to define timeframes.
- Extend `fare_leg_rules.txt` with `from_timeframe_id`, `to_timeframe_id`, and `service_id` to describe time-dependant fares.

The time-dependant variables (timeframe and day information) are modelled into fare legs (and not fare products) since in GTFS-Fares v2, `fare_leg_rules.txt` models the location and time variables (from/to area, from/to timeframe, and service ID). All other factors are modelled in fare products. 

**_Here's a quick example_** 
- A regular fare is required to travel.
- The fare is discounted if riders travel between 6:00 AM and 7:00 AM. 
- On New Year's, transit is free (no fare).

Define service days and New Year's using `calendar.txt.` and `calendar_dates.txt`
| service_id | monday |tuesday  |wednesday |thursday |friday |saturday |sunday |start_date |end_date |
|--|--|--|--|--|--|--|--|--|--|
| regular_service | 1 |1  |1 |1 |1 |1 |1 |20230101 |20231231 |

| service_id |date  |exception_type|
|--|--|--|
|regular_service  | 20230101 |2 |
|new_year_service  | 20230101 |1 |

Define the discounted timeframe using `timeframes.txt`
| timeframe_id |start_time  |end_time|
|--|--|--|
|morning  | 06:00:00 |07:00:00 |
|regular  | 00:00:00 |05:59:59 |
|regular  | 00:07:01 |23:59:59 |
|all_day  | 00:00:00 |23:59:59 |

Define the time and date-based fares using `fare_leg_rules.txt`
| leg_group_id | from_area_id |to_area_id  |network_id |from_timeframe_id |to_timeframe_id |service_id |fare_product|
|--|--|--|--|--|--|--|--|
| leg1| zoneA |zoneB  |bus |regular |all_day|regular_service |regular_fare |
| leg1| zoneA |zoneB  |bus |morning |all_day|regular_service |discounted_fare |
| leg1| zoneA |zoneB  |bus |regular |all_day|new_year_service |free_fare |
| leg1| zoneA |zoneB  |bus |morning |all_day|new_year_service |free_fare |

Data consumer: [Apple](https://www.apple.com/)
Data producer:   (To be announced)

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -108,6 +108,7 @@ This specification defines the following files:
 |  [stop_times.txt](#stop_timestxt)  | **Required** | Times that a vehicle arrives at and departs from stops for each trip. |
 |  [calendar.txt](#calendartxt)  | **Conditionally Required** | Service dates specified using a weekly schedule with start and end dates. <br><br>Conditionally Required:<br> - **Required** unless all dates of service are defined in [calendar_dates.txt](#calendar_datestxt).<br> - Optional otherwise. |
 |  [calendar_dates.txt](#calendar_datestxt)  | **Conditionally Required** | Exceptions for the services defined in the [calendar.txt](#calendartxt). <br><br>Conditionally Required:<br> - **Required** if [calendar.txt](#calendartxt) is omitted. In which case [calendar_dates.txt](#calendar_datestxt) must contain all dates of service. <br> - Optional otherwise. |
+|  [timeframes.txt](#timeframestxt)  | Optional | Service times specified using a 24-hour format with start and end times. |
 |  [fare_attributes.txt](#fare_attributestxt)  | Optional | Fare information for a transit agency's routes. |
 |  [fare_rules.txt](#fare_rulestxt)  | **Conditionally Required** | Rules to apply fares for itineraries.<br><br>Conditionally Required:<br>- **Required** if [fare_attributes.txt](#fare_attributestxt) is defined.<br>- **Forbidden** otherwise. |
 |  [fare_products.txt](#fare_productstxt)  | Optional | To describe the different types of tickets or fares that can be purchased by riders.<br><br>File [fare_products.txt](fare_productstxt) describes fare products that are not represented in [fare_attributes.txt](#fare_attributestxt) and [fare_rules.txt](#fare_rulestxt). As such, the use of [fare_products.txt](#fare_productstxt) is entirely separate from files [fare_attributes.txt](#fare_attributestxt) and [fare_rules.txt](#fare_rulestxt). |
@@ -375,7 +376,7 @@ To describe the different types of tickets or fares that can be purchased by rid
 
 File: **Optional**
 
-Primary Key (`network_id, from_area_id, to_area_id, from_timeframe_id, to_timeframe_id, fare_product_id`)
+Primary Key (`network_id, from_area_id, to_area_id, from_timeframe_id, to_timeframe_id, service_id, fare_product_id`)
 
 Fare rules for individual legs of travel.
 
@@ -388,16 +389,18 @@ To process the cost of a leg:
     - `fare_leg_rules.from_area_id`
     - `fare_leg_rules.to_area_id`
     - `fare_leg_rules.from_timeframe_id`
-    - `fare_leg_rules.to_timeframe_id`<br/>
+    - `fare_leg_rules.to_timeframe_id`
+    - `fare_leg_rules.service_id`<br/>    
 <br/>
 
 2. If the leg exactly matches a record in `fare_leg_rules.txt` based on the characteristics of travel, that record must be processed to determine the cost of the leg.
-3. If no exact matches are found, then empty entries in `fare_leg_rules.network_id`, `fare_leg_rules.from_area_id`, `fare_leg_rules.to_area_id`, `fare_leg_rules.from_timeframe_id`, and `fare_leg_rules.to_timeframe_id` must be checked to process the cost of the leg:
+3. If no exact matches are found, then empty entries in `fare_leg_rules.network_id`, `fare_leg_rules.from_area_id`, `fare_leg_rules.to_area_id`, `fare_leg_rules.from_timeframe_id`, `fare_leg_rules.to_timeframe_id`, and `fare_leg_rules.service_id` must be checked to process the cost of the leg:
     - An empty entry in `fare_leg_rules.network_id` corresponds to all networks defined in `routes.txt` excluding the ones listed under `fare_leg_rules.network_id`
     - An empty entry in `fare_leg_rules.from_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.from_area_id`
     - An empty entry in `fare_leg_rules.to_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.to_area_id`
     - An empty entry in `fare_leg_rules.from_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.from_timeframe_id`
-    - An empty entry in `fare_leg_rules.to_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.to_timeframe_id`<br/>
+    - An empty entry in `fare_leg_rules.to_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.to_timeframe_id`
+    - An empty entry in `fare_leg_rules.service_id` corresponds to all dates defined in `calendar.service_id` excluding the ones listed under `fare_leg_rules.service_id`<br/>    
 <br/>
 
 4. If the leg does not match any of the rules described above, then the fare is unknown.
@@ -412,6 +415,7 @@ To process the cost of a leg:
 | `to_area_id` | Foreign ID referencing `areas.area_id` | Optional | Identifies an arrival area.<br><br>If there are no matching `fare_leg_rules.to_area_id` values to the `area_id` being filtered, empty `fare_leg_rules.to_area_id` will be matched by default.<br><br> An empty entry in `fare_leg_rules.to_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.to_area_id` |
 |  `from_timeframe_id` | Foreign ID referencing `timeframes.timeframe_id` | Optional |  Defines a departure timeframe for the fare leg rule.<br><br>If there are no matching `fare_leg_rules.from_timeframe_id` values to the `timeframe_id` being filtered, empty `fare_leg_rules.from_timeframe_id` will be matched by default. <br><br>An empty entry in `fare_leg_rules.from_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.from_timeframe_id`  |
 |  `to_timeframe_id` | Foreign ID referencing `timeframes.timeframe_id` | Optional |  Defines an arrival timeframe for the fare leg rule.<br><br>If there are no matching `fare_leg_rules.to_timeframe_id` values to the `timeframe_id` being filtered, empty `fare_leg_rules.to_timeframe_id` will be matched by default. <br><br>An empty entry in `fare_leg_rules.to_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.to_timeframe_id`  |
+| `service_id` | Foreign ID referencing `calendar.service_id` | Optional | Identifies a set of dates that a fare leg rule is in effect.<br><br>If there are no matching `fare_leg_rules.service_id` values to the `service_id` being filtered, empty `fare_leg_rules.service_id` will be matched by default. <br><br>An empty entry in `fare_leg_rules.service_id` corresponds to all dates defined in `calendar.service_id` excluding the ones listed under `fare_leg_rules.service_id` |
 | `fare_product_id` | Foreign ID referencing `fare_products.fare_product_id` | **Required** | The fare product required to travel the leg. |
 
 ### fare_transfer_rules.txt
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -17,6 +17,7 @@ This document defines the format and structure of the files that comprise a GTFS
     -   [stop\_times.txt](#stop_timestxt)
     -   [calendar.txt](#calendartxt)
     -   [calendar\_dates.txt](#calendar_datestxt)
+    -   [timeframes.txt](#timeframestxt)    
     -   [fare\_attributes.txt](#fare_attributestxt)
     -   [fare\_rules.txt](#fare_rulestxt)
     -   [fare\_products.txt](#fare_productstxt) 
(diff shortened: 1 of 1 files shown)


skinkie:
I suggest to make a change and don't mention a 24h period, but keep it in line with the rest of the GTFS spec (hence 24+h) formats.

flocsy:
Need to clarify the start_time/end_time formats: do we require to have them be between 00:00 and 23:59? Do we allow timeframe like: start_time: 07:00, end_time: 06:00? Or is tomorrow's 06:00 represented as 30:00? Or we don't allow this and it has to be like in the example that we have it cut in two: 00:00-06:00 and 07:00-23:00? Currently the proposal has "24-hour format" so maybe we should add "allowed values: 00:00 - 23:59"

It'll be beneficial IMHO (easier to look at the data) if instead of: [06:00-07:59], [08:00-15:59] we could use [06:00-08:00).[08:00-16:00)

IMHO we need to have a constraint that the timeframes are distinct. (You can't have both: 06:00-06:59 and 05:00-07:59, and depending what we decide about the end of the timeframe: 06:59] vs 07:00) we should not allow both: 06:00-07:00 and 07:00-18:00 because 07:00 is listed in 2 timeframes) Though it's not clear to me if we want this constraint in timeframes.txt (probably not) or somehow on the data in fare_leg_rules.txt (probably yes, but it'll be very complicated and it should also incorporate service_id)

[We might also need to have a constraint that if there are timeframes then they need to cover all the 24 hours (you can't have peak-hour: 06:00-07:59 without also having the 00:00-05:59 and 08:00-23:59). Looks like it's not needed because empty timeframe_id covers the "rest", but might not be that simple, see below]

The problem with what empty timeframe_id means: "all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.[from/to/?_]timeframe_id`" is that it might be not the case. For example in weekdays (defined in services.txt, not in timeframes.txt) we might have peek: 06:00-08:00 and 16:00-18:00, ("rest of the day" should be "off-peak") but the weekends are probably considered as off-peak: 00:00-24:00 on saturday and sunday. This very realistic example also shows why the distinction constraint is not that simple.

Maybe we could add timeframes.service_id as an optional field? That way the constraint could be that timeframes having the same service_id must be distinct.

Do we really need seconds? To me it sounds unrealistic and unnecessary.

fare_leg_rules.txt: IMHO we only need to add service_id, timeframe_id. While I can see what it could be used to if we had from/to timeframe_id, it also makes it probably too complex and error prone (or needs some sentences to constrain them to consecutive timeframes, but then that is probably something that can't be enforced either...)
Let's say we have: 00:00-06:00: off-peak, 06:00-08:00: peak, 08:00-16:00: off-peak, 16:00-18:00: peak, 18:00-24:00: off-peak. Now if we could have a rule like: from_timeframe_id: off-peak, to_timeframe_id: off-peak, then a trip that starts at 05:55 and ends at 08:05 would be.... what? considered a legal off-peak trip? I don't think so...

davidlewis-ito:
Great to see this conversation started.

We have a few comments.

1) timeframe_ids & service_ids in fare_leg_rules, fare_products or elsewhere ?
I am not sure of the logic of putting these time discriminating elements in the fare_leg_rules table. Although this was the original concept during the historic "GTFS Fares V2 Google Doc" discussion, I understood that the thinking has now moved on to use the fare_products table to combine the various discriminating factors.

I would take this a step further and propose that these time discriminating elements should go into another table ("time_rules.txt") that would yield a single key to be matched in the product table.  

As well as providing a stronger semantic modelling this also avoids the fare_products table (or fare_leg_rules) from exploding in size when one needs to model the more complicated peak definitions in a context of (say) a complex zonal system.

2) Scope of timeframes
We have found use cases where the time discrimination for a fare depends on either:
- the time the rider boards the bus
- the time the train started or ended its trip
- the time the rider enters the station

We would therefore propose further attributes: from_timeframe_type and to_timeframe_type with values of trip, leg, station (and potentially others).

3) We also agree with the comments of @flocsy  above in eradicating the use of XX:59:59 times .

omar-kabbani:
Thanks @skinkie, @flocsy, and @davidlewis-ito for your input! I have responded to your comments below - please let me know if something is outstanding. 
 
@skinkie:
> I suggest to make a change and don't mention a 24h period, but keep it in line with the rest of the GTFS spec (hence 24+h) formats.

- Correct! I will change the wording to reference the `Time` field type in the specification. 

@flocsy:

> Need to clarify the start_time/end_time formats

- `timeframes.txt` is modelled the same way as `calendar.txt`. If `start_time` is greater than `end_time`, this probably would trigger an error in the validator ([just like it currently does if the start date is greater than the end date](https://github.com/MobilityData/gtfs-validator/blob/master/RULES.md#start_and_end_range_out_of_order)). To model a timeframe that goes beyond midnight, the time should be written as 24+h, just as how `Time` is described in GTFS. 

> It'll be beneficial IMHO (easier to look at the data) if instead of: [06:00-07:59], [08:00-15:59] we could use [06:00-08:00).[08:00-16:00)

- With regards to the interval, `start_time` and `end_time` are included in the interval - just like how `start_date` and `end_date` are in `calendar.txt`. I will change the wording to clarify that. 

> IMHO we need to have a constraint that the timeframes are distinct

- I think that overlapping timeframes that are part of the same `timeframe_id` should be forbidden. I can change the wording to add that.

> We might also need to have a constraint that if there are timeframes then they need to cover all the 24 hours

- This is always going to be a constraint when using empty fields in `fare_leg_rules.txt` and `fare_transfer_rules.txt`. The exact same issue arises if an `area_id` is present in more than one `network_id`. If using empty fields complicates things, then the timeframes should be defined explicitly. 

> Do we really need seconds? To me it sounds unrealistic and unnecessary.

- I believe most data producers will not use seconds, but this is how `Time` is formatted in GTFS. Adding another data type may unnecessary complicate things. Removing seconds from `Time` will break current datasets. 

> fare_leg_rules.txt: IMHO we only need to add service_id, timeframe_id.

- Currently, the field `from_timeframe_id` indicates the window for when a trip starts to be eligible for the special pricing. I realize now that in my example above I made an error by leaving `to_timeframe_id` blank (it should be a timeframe that covers all the day). I have changed that to a timeframe that covers all day, thanks for catching that! Basically any trip that starts between 6:00 AM and 7:00 AM and ends whenever is eligible for special pricing. If an agency has special restrictions on when a trip should end to be eligible for special pricing, then they should use the field `to_timeframe_id` with that timeframe. 

@davidlewis-ito:

> timeframe_ids & service_ids in fare_leg_rules, fare_products or elsewhere ?

- I leave this to what the community wants to see - currently, `fare_leg_rules.txt` models the spatial (from area - to area) and temporal (from time window - to time window) factors of a trip to determine pricing. All other variables are modelled in `fare_products.txt`. If there is a push to model timeframes separately, then we will do as the industry sees fit. 

> Scope of timeframes

Good catch! For fare calculation purposes, the start/end of a trip is not clear. Since this is specific to fares, the start time is when a fare is validated (tap on the bus, or tap to enter the station, etc.), and the end time is when a fare is validated (tap off). I can modify the wording to reflect that.

flocsy:
@omar-kabbani
Regarding keeping both from_timeframe_id and to_timeframe_id: this poses a problem even with the most common use case: most of the day is "off-peak" but 06:00-07:59 it's "peak". If the rule would be: from_timeframe_id: off-peak, to_timeframe_id: off-peak, then probably the goal of the producer would be to say that when the whole trip is in off-peak then it's cheaper. However it can't differentiate between a trip: 05:55-08:13 and 10:12-11:34 where the 1st trip probably shouldn't be considered, since it includes the whole peak hour as well.

Regarding the constraint about overlapping timeframes: you wrote above that "overlapping timeframes that are part of the same timeframe_id should be forbidden". That for sure has to be added, but is that enough? What would you make of the following rules:

timeframe_id,start_time,end_time
off-peak,00:00,05:59
peak,06:00,07:59
off-peak,07:00,23:59

Part of the problem seems to be related to what both @davidlewis-ito and I touched that the timeframes and the service dates together define the times, so there should be something that connects between them. Either by having some explicit table for it or by having complicated constraints in the definitions. The above 3 lines look like not OK because there are 2 timeframes that include 07:00-07:59, so it's probably a bug, but on the other hand if I would change it to:

timeframe_id,start_time,end_time
off-peak,00:00,05:59
peak,06:00,07:59
off-peak,08:00,23:59
off-peak,00:00,23:59

Then it would still look strange at first sight, but when I would add that the 1st 3 lines cover the whole 24 hours for work days and the 4th line covers the whole 24 hours on weekends, then it would make sense. But then this should be somewhere explicitly defined, so the service_ids work-day and weekend should be related tho them. Of course we could also change the 4th line's timeframe_id to weekend, but that would only help humans when looking at it, but on the other hand it would probably also double the number of lines in the fare_leg_rules.

Regarding the timeframe_type: I think that modifying the wording is not enough, because it would only clarify the confusion on a local level, which anyway probably would not be confusing for the locals, but it misses the whole point of GTFS Fares V2: automated fare calculations. When you have an algorithm that calculates the fares and it knows that it takes 2 minutes to walk from the station entrance to the 23rd platform then it can and should take that into account. And then it is important for the algorithm to differentiate between these nuances: start to count at the entrance of the station or when you enter the train (reach the platform) or when the train leaves...

omar-kabbani:
Hi @flocsy -

> Now if we could have a rule like: from_timeframe_id: off-peak, to_timeframe_id: off-peak, then a trip that starts at 05:55 and ends at 08:05 would be.... what? considered a legal off-peak trip? I don't think so...

Technically, yes - this would be an off-peak trip, since the fields are based on fare validation. The examples for off-peak fares that I found online only look at when the trip starts. Hence, if the discounted pricing starts at 7:00 AM and a rider taps at 6:59 AM - they will not pay the discounted price. For example, [on TfL's fare structure](https://tfl.gov.uk/fares/find-fares/tube-and-rail-fares#on-this-page-2), it is mentioned that "*Peak and off-peak fares are charged based on the time you touch in.*". Hence, `from_timeframe_id=morning` and `to_timeframe_id=all_day` (riders are eligible for a discount if they tap in the morning - it does not matter at what time their trip ends - could be during the morning period or after)

> However it can't differentiate between a trip: 05:55-08:13 and 10:12-11:34 where the 1st trip probably shouldn't be considered, since it includes the whole peak hour as well.

Same here - if we are basing this on fare validation (*tap on/tap off, etc.*) both trips are-off peak. Unless we add a field `contains_timeframe_id` similar to how areas are handled, but this sounds a bit overkill - I do not know about all agencies, but I think most of the ones who provide variable pricing base it off of the start of the journey (tapping on). 

From your posts, I take it that you are suggesting using one field for timeframes in `fare_leg_rules.txt` - what would that represent? The start of a trip, or a window which the trip should start and end in, or something else?

For associating `service_id` to `timeframe_id`, an option would be to add a `service_id` field to `timeframes.txt` - that way, the timeframe is bound a set of days. 

As for `timeframe_type`, I leave this to data consumers as to how to handle it - but the current proposal is based on the fare validation (as in when a rider taps on or pays a fare or passes through fare gates etc.). How long it takes riders to go from the entrance to the fare gates is at the discretion of the trip planning application as well as through the use of `pathways.txt`. If there is a need to make this more specific, what enumerations should be included for timeframe type?

flocsy:
I am brainstorming rather than saying that there should be only 1 timerfame_id field PLUS timeframe_type. With those two we probably can handle many logical use cases (timeframe_type: station_entrance / vehicle_entrance would mean what you described, but there would also be a possibility to have trip_duration or whole_trip that would be able to disallow a trip: 05:55-06:13) Also by having timeframe_id + timeframe_type we can handle more real life use cases IMHO than with from_timeframe_id+to_timeframe_id (and in most cases from_timeframe_id would be equal to to_timeframe_id anyway)

Regarding adding service_id to somewhere: well, yes, of course the proposal would put them in fare_leg_rules, but there are other places we could add it, like to the timeframes. We'll need to think about the pros and cons of these (and maybe more) possibilities. As I see it we'll have to consider the most common use cases, which I believe mostly will have these in common:
the most common rules would be something like day-of-week (represented in a service) and hour-of-day (represented in timeframe) combined somehow (simplest example: M-F: 06:00-07:59 => peak, rest of the day => off-peak, but S-S: 00:00-23:59 => off-peak) For this basic case it's probably best to add service_id to timeframes. Up to this point it makes sense and also it "fixes" the overlap issue.

However considering what I think would be the next most common use case (exceptions for holidays) then if we would add just 1 service exception for X-mas, then how would it look? If we add service_id to timeframes, then we would have probably 1 more (maybe 2 if there's still some peek/off-peek thing) lines to timeframes, but then we would also need to duplicate most rules in fare_leg_rules, wouldn't we? I'm not really sure what's best. And if some lines behave in X-mas as usual, but other lines as weekend, then things get even more complicated.

omar-kabbani:
Hey everyone - just a reminder that MobilityData is hosting a roundtable discussion on Wednesday 24 August at 11:00 AM ET to discuss the outstanding items regarding time-variable fares. If you would like to attend, please react to this message or send an email to [specifications@mobilitydata.org](mailto:specifications@mobilitydata.org) and we will send you the invite. 

For a summary and more details, please check out [this post](https://github.com/google/transit/issues/341#issuecomment-1219916837).

flocsy:
I'm on a family vacation, be back in September.

halbertram:
I'm afraid I can't make the discussion. I think there's a part of @davidlewis-ito's suggestion on timeframe_type that hasn't been picked up fully: the case when the determination of the peak-ness of a leg is determined by the time that vehicle leaves its origin (or reaches its destination) - eg LIRR in NYC. It's effectively an attribute of the trip and is not connected to the rider's timings of when they board that trip at all. This feels like something that would be good to capture in timeframes. It would be hard to represent in another way, short of something like trip-level fare attributes.

omar-kabbani:
## Summary of roundtable discussion on variable fares by time or day:
### August 24 11:00 AM ET
Attended by: @Cristhian-HA, @omar-kabbani, @skinkie, Chris Erickson (Trillium), and Scott Jackson (Apple).

#### Number of fields
- It was discussed that 2 fields are required to effectively describe time-based fare calculations:
  - "from" timeframe which defines the window for when a rider has to start a journey to be eligible for the special fare.
  - "to" timeframe which defines the window for when a rider has to end a journey to be eligible for the special fare. 
- Most cases, the "from" timeframe field will be used, since many agencies handle special fares depending on when riders start their journey and not when they end them (for example: [TfL](https://tfl.gov.uk/fares/find-fares/tube-and-rail-fares) and [MRT](https://www.ptc.gov.sg/fare-regulation/bus-rail/morning-pre-peak-fares)).
- Since most agencies do not use the trip end times to process time-based fares, and only rely on when the trip starts - one option is to have the field `to_timeframe_id` default to "All day" if left blank. This will go against the default behaviour of other blank fields in `fare_leg_rules.txt`, but this will save data producers the hassle of creating a separate entry in `timeframes.txt` to describe "All day". 

#### Timeframe type 
- The need for a file or field to describe the timeframe type was highlighted since transit services have different ways of defining when a fare is activated (at station entrance, upon boarding, upon departure, etc.).
- One option is to extend `fare_leg_rules.txt` with `timeframe_type` with enumerations.

#### File structure
- `from_timeframe_id` and `to_timeframe_id` to remain in `fare_leg_rules.txt`  
- `timeframe_type` to be added to `fare_leg_rules.txt`
- Use `timeframes.txt` and `calendar.txt` to define days and times - and use the fields `from_timeframe_id`, `to_timeframe_id`, and `service_id` in `fare_leg_rules.txt` to associate times and days to fare legs.

omar-kabbani:
Hey everybody, I will close this pull request since I will be leaving MobilityData and therefore I cannot remain the advocate for this proposal.
 
In order to preserve the conversation history, I will open an issue in google/transit and reference this pull request. Please continue the discussion [there](https://github.com/google/transit/issues/350), and provide feedback on how to move forward with this proposal.

I also wanted to thank you all for your constant engagement in these GTFS extension proposals and for all the valuable feedback!

flocsy:
Hi,

I was on vacation so I couldn't attend the meeting. Here are my thoughts:

1. I like that empty timeframe_id means "all day", it is much more
"natural" to me than the definition we usually have. However doesn't this
also mean we have to "cover" the whole day with timeframes? What I mean is
that if we have the easiest example: 06:00-08:59: "peek", and the rest of
the day: "normal", then you no longer can use empty to mean "normal" so we
have to define normal as: 09:00-29:59: normal. Which is something that I
think is good, I just think we should mention this explicitly.

2. Why timeframe_type is being added to fare_leg_rules and not to
timeframes? It doesn't seem logical to me. I would like hear the arguments
why it can't be in timeframes, and maybe think about the best place to put
it.

isabelle-dr:
Opened #350 to continue the work on this PR

flocsy:
can't this be reopened?

isabelle-dr:
@flocsy Omar was the advocate of this PR and I can't continue on this behalf, unfortunately.
I know it's not ideal; I've read everything in this PR and I'm updating #350 as a result so we can resume where we left off. 
Sorry for the inconvenience -

PR #342, opened 2022-08-05
Add fare containers to reference.md

omar-kabbani:
Hi everyone,   MobilityData is moving forward with the second iteration of GTFS-Fares v2, for more information about the overall plan, you can check [this issue](https://github.com/google/transit/issues/341). 

This pull request covers Fare Containers, which is a section of the entire [GTFS-Fares v2 proposal](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.mhjuwqk2h5r3). 

Fare containers correspond to cards or mobile applications that are used to load and use fares. In many cases, fares loaded on a fare container are cheaper than fares purchased as individual tickets outside a fare container. 

The changes in this pull request are:
- Add new file; `fare_containers.txt`, to define new fare containers.
- Extend `fare_products.txt` with `fare_container_id` to assign fare products to fare containers and to define the price of a fare if a fare container is used to pay.

The behaviour of `fare_leg_rules.txt` and `fare_transfer_rules.txt` is unchanged as the cost of a fare leg or a transfer is tied to a fare product. The modelling of the cost of a one-way ticket purchased as a ticket vs. loaded on a fare container factors in `fare_products.txt`.

**_Here's a quick example:_**
- A fare container costs 5 CAD.
- A fare is 3 CAD, however, if riders load the fare on their fare container and pay with it, a fare is 2.95 CAD.
- Riders can transfer once - in a period of one hour - if they do not have a fare container, they will pay 0.05 CAD to transfer, however, if they transfer using their fare container, the transfer is free.

Define a fare container in `fare_containers.txt`
|fare_container_id  | amount | currency|
|--|--|--|
| container | 5 | CAD|

Define the fare structure (paying for a fare with and without the fare container) in `fare_products.txt`
|fare_product_id|fare_container_id  | amount | currency|
|--|--|--|--|
| single_leg |  | 3 | CAD|
| single_leg | container | 2.95 | CAD|
| transfer_cost |  | 0.05 | CAD|
| transfer_cost | container | 0 | CAD|

Define fare legs using `fare_leg_rules.txt`
|leg_group_id|from_area_id|to_area_id | network_id| fare_product_id | 
|--|--|--|--|--|
|leg1| zoneA | zoneB | bus | single_leg|

Define transfer rules using `fare_transfer_rules.txt`
[...]

PR #355, opened 2022-11-01
Add fare media

isabelle-dr:
Hi everyone! 

I am opening this Pull Request as part of the second implementation of Fares v2. I am opening it in the exact same state PR #342 was when closed, and I will update the proposal based on feedback already received so we can resume where we left off. For more information on the plan, please refer to [issue #341](https://github.com/google/transit/issues/341#issuecomment-1296266122).
This pull request covers Fare Containers, which is a section of the entire [GTFS-Fares v2 proposal](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.mhjuwqk2h5r3).
**EDIT**: this proposal has been slightly modified to represent a broader "fare media", that contains fare media and passes and can be used to travel. See the latest proposal here: https://share.mobilitydata.org/fare-container-updated-proposal

**Scope of this PR**:
- To define the fare ~~containers~~ media that an agency accepts
- To describe the fares if riders pay using their fare ~~containers~~ media (vs. tickets not loaded on a fare container)

**Not doing as part of this PR**: 
- The scope does not include any functionalities that pertain to tracking when riders recharge or use up their fare containers.

[Original message posted by @omar-kabbani in PR #342]

- Add new file; `fare_containers.txt`, to define new fare containers.
- Extend `fare_products.txt` with `fare_container_id` to assign fare products to fare containers and to define the price of a fare if a fare container is used to pay.

The behaviour of `fare_leg_rules.txt` and `fare_transfer_rules.txt` is unchanged as the cost of a fare leg or a transfer is tied to a fare product. The modelling of the cost of a one-way ticket purchased as a ticket vs. loaded on a fare container factors in `fare_products.txt`.

**_Here's a quick example:_**
- A fare container costs 5 CAD.
- A fare is 3 CAD, however, if riders load the fare on their fare container and pay with it, a fare is 2.95 CAD.
- Riders can transfer once - in a period of one hour - if they do not have a fare container, they will pay 0.05 CAD to transfer, however, if they transfer using their fare container, the transfer is free.

Define a fare container in `fare_containers.txt`
|fare_container_id  | amount | currency|
|--|--|--|
| container | 5 | CAD|

Define the fare structure (paying for a fare with and without the fare container) in `fare_products.txt`
|fare_product_id|fare_container_id  | amount | currency|
|--|--|--|--|
| single_leg |  | 3 | CAD|
| single_leg | container | 2.95 | CAD|
| transfer_cost |  | 0.05 | CAD|
| transfer_cost | container | 0 | CAD|

Define fare legs using `fare_leg_rules.txt`
|leg_group_id|from_area_id|to_area_id | network_id| fare_product_id | 
|--|--|--|--|--|
|leg1| zoneA | zoneB | bus | single_leg|

Define transfer rules using `fare_transfer_rules.txt`
|from_leg_group_id|to_leg_group_id|transfer_count | duration_limit| duration_limit_type |fare_transfer_type|fare_product_id| 
|--|--|--|--|--|--|--|
|leg1|leg1|1 | 3600| 0 |0|transfer_cost| 

Data consumer: [Apple](https://www.apple.com/maps/)
Data producer:   [Interline](https://www.interline.io/)


isabelle-dr:
Based on the discussions that already happened, as of today the open items are:
- Adding a fare container type, which would be done by extending `fare_containers.txt` with a field called `fare_container_type` that defines whether the container is a physical card, a mobile app, or both.
- review the field name `amount` in `fare_containers.txt` to better represent the cost to purchase the fare container.
- discuss if we want `travel_rules.payment_method` (part of the broader [GTFS-Travel-Rules proposal](https://docs.google.com/document/d/1w9g-CRClM0P6NbnUf8HFzK8UVNxzciPIA7WMA7KbkOg/edit#)) moved to `fare_products.txt`, to provide information on how can riders can pay for the fare products.

e-lo:
We at Cal-ITP just want to be able to efficiently provide information (e.g. adding this handy icon to the travel plan in a trip planner app) that a rider can tap to pay w/a normal EMV card.
image

irees:
I think this works well with Interline's current data and implementation. 👍

isabelle-dr:
Thank you for the feedback!
Here are the latest updates to this proposal:

1. updated the description of `amount` and `minimal_initial_purchase` to clarify their intended use. I kept the `amount` field name to stay consistent with `fare_products.txt`. Let me know if you still think the field name should be changed, or if this new description solves the confusion. See the change [here](https://github.com/google/transit/pull/355/commits/58dd6bbaf1662b8560d02068800a46af64b5f130). 
2. added an optional `fare_container_type` field to define whether the container is a physical card, a mobile app, or both.
3. Updated the primary key of `fare_products.txt` as [flagged](https://github.com/google/transit/pull/342#issuecomment-1208279858) by @npaun 

For payment methods: I see the temptation to use the container to discriminate fares that depend on the payment method. Before modifying the proposal, I'd like to know if we have a general agreement on the following design. I used Interline's MTC dataset  as an example:

- fare_container.txt

fare_container_id | fare_container_name | amount | minimum_initial_purchase | currency
-- | -- | -- | -- | --
clipper | Clipper | 3 |   | USD

- fare_products.txt (current)

fare_product_id | fare_product_name | amount | currency | fare_container_id
-- | -- | -- | -- | --
AC:local:single | AC Transit local fare | 2.5 | USD | cash
AC:local:single | AC Transit local fare | 2.25 | USD | clipper
AC:transfer:transbay-upgrade | AC Transit Transbay upgrade | 3.5 | USD | cash
AC:transfer:transbay-upgrade | AC Transit Transbay upgrade | 3.75 | USD | clipper

- fare_products.txt with added `payment_method`

fare_product_id | fare_product_name | amount | currency | fare_container_id | payment_method
-- | -- | -- | -- | -- | --
AC:local:single | AC Transit local fare | 2.5 | USD |   | [3, 4, 5]
AC:local:single | AC Transit local fare | 2.25 | USD | clipper | [3, 4, 5, 6]
AC:transfer:transbay-upgrade | AC Transit Transbay upgrade | 3.5 | USD |   | [3, 4, 5]
AC:transfer:transbay-upgrade | AC Transit Transbay upgrade | 3.75 | USD | clipper | [3, 4, 5, 6]

Assuming `payment_method` would be a list of payments accepted. What we currently have in [GTFS-TravelRules](https://docs.google.com/document/d/1w9g-CRClM0P6NbnUf8HFzK8UVNxzciPIA7WMA7KbkOg/edit#) is:
0: any
1: coins
2: exact cash (includes 1)
3: cash with change (includes 1 and 2)
4: bank card
5: contactless bank card (physical or virtual card in Apple Pay &co)
6: mobile wallet (ex: Clipper)

Note that this field was initially designed as an `Enum` (pick only one) and not a `List` (pick all that apply).

- fare_leg_rules.txt

leg_group_id | from_area_id | to_area_id | network_id | fare_product_id
-- | -- | -- | -- | --
AC | AC:local | AC:local | AC | AC:local:single
AC | AC:transbay | AC:transbay | AC | AC:local:single

- fare_transfer_rules.txt

from_leg_group_id | to_leg_group_id | transfer_count | duration_limit | duration_limit_type | fare_transfer_type | fare_product_id
-- | -- | -- | -- | -- | -- | --
AC | AC:transbay |   | 7200 | 1 | 0 | AC:transfer:transbay-upgrade

Looking forward to hear your thoughts on this.
Tagging @irees @e-lo @davidlewis-ito @flocsy who were involved in the previous PR.

bdferris-v2:
The combination of `payment_method` and `fare_container_id` seem a bit duplicative of each other, especially when treating `payment_method` as a list.  In AC:local:single with fare_container_id=clipper, what does it mean that cash is accepted as a payment method but the Clipper card is the fare container?  I'd be interested to hear more from the original authors about the interactions of these fields.

flocsy:
I agree with @bdferris-v2 regarding the issue with "clipper + cash", though I thought it's mainly a typo. The payment method in the clipper lines should be [6]

However even after fixing that he's right, that it's repeating itself (fare_container_id: clipper and [6])

I don't understand some of the payment_methods:

- I guess 1:coins means no paper, because it needs to be entered to a machine that only accepts coins?

- 6:mobile_wallet is a little bit confusing, because Google wallet, Google Pay, Apple pay, Garmin pay are all considered a wallet by many (and Google wallet even called as such). Maybe mobile_app is better? But not sure, 'cause all the above are also apps :) Maybe a way to clarify the difference is to include also "NFC" in 5:contactless_bank_card. I'd even change their names to 5:contactless_nfc and 6:mobile_app This still has some overlap because of the app, but maybe because of the NFC being part of the name it's clearer that 5 means anything that uses the NFC technology (at 1st I thought it's not a good idea to include one particular technology as the name, but then I thought about it: if there will ever be another contactless technology then we'll anyway need to add a new enum value to be differentiate, because when I go abroad with my NFC compatible phone I'll need to know if I can or can't use it) and this leaves the 6:mobile_app a little bit clearer (though it'll also need a description that tells that it's an app that doesn't use NFC but rather some other way (QA code / numeric code / bar code / camera / push notification / etc...) to validate/pay.

I'll also drop in another thought/question: Let's say there's a container called "C". It can be loaded with $5 | $25 | $50. How would we indicate this information? We could add these as 3 lines in fare_products.txt but how do we indicate in fare_container.txt that it can be only loaded using these 3 amounts and not any amount I choose?

Also maybe it's better to keep all the filenames either single or plural (currently we have fare_productS vs fare_container)

davidlewis-ito:
Great to see this discussion moving forward.

**Payment Types**
I would suggest that this column should appear both on the fare_product table AND the fare_container table.

- Within the fare_product table the field should be empty where there is a fare_container_id.  this implies that option 6 of the payment_type enum is not required
- Within the fare_container table the payment_type refers to the means by which the container stored value may be increased.

**Transfers**
In many cases a (free) transfer only applies if the same fare_container or payment type is used on both legs.  Examples include the London Bus, New York MTA.

So given we are now proposing to introduce these discriminating factors into the product table it seems we need to revisit how fare_transfer_rules interact with fare_leg_rules and/or fare_products.  We haven't thought this through in great detail but perhaps introducing a fare_group_id into fare_products would allow one to filter the available transfers appropriately.  We note also that there is more work to be done with respect to fare_transfer_rules to make them fit for the real world.

gtfs/spec/en/reference.md
```diff
@@ -343,11 +345,29 @@ For examples that demonstrate how to specify a fare structure with [fare_rules.t
...
+|  `minimum_initial_purchase` | Non-negative currency amount | Optional | The amount of the minimum fare purchase required when acquiring the container. |
+|  `currency` | Currency code | **Conditionally Required** | The currency of `fare_containers.amount` or `fare_containers.minimum_initial_purchase`. <br/> Conditionally Required: <br/> - **Required** if `fare_containers.amount` or `fare_containers.minimum_initial_purchase` are defined. <br/> - **Forbidden** if `fare_containers.amount` and `fare_containers.minimum_initial_purchase` are empty. |
+
+
 ### fare_products.txt
 
 File: **Optional**
 
-Primary Key (`fare_product_id`)
+Primary Key (`fare_product_id`, `fare_container_id`)
```
isabelle-dr, line 370: @npaun
It looks like we have two possible designs here, are we OK with both?

fare_product_id | fare_product_name | amount | currency
-- | -- | -- | --
presto_fare | PRESTO Card Fare | 3.2 | CAD
cash_fare | Cash Fare | 3.25 | CAD

fare_product_id | fare_product_name | amount | currency | fare_container_id
-- | -- | -- | -- | --
regular_fare | PRESTO Card Fare | 3.2 | CAD | PRESTO card
regular_fare | Cash Fare | 3.25 | CAD |

flocsy: I don't think so. If you have presto_fare and cash_fare then you need to duplicate all the lines in all the other tables that reference fare_products. On the other hand when you have regular_fare it also makes it produce/read/understand easier and also has the additional information that the 2 options of $3.2 and $3.25 are kind of the same, and it depends if you have got the PRESTO Card or not.

npaun: @isabelle-dr The two examples you show are both legal but they mean entirely different things. 

* In the first case, there are two separate kinds of tickets and none of them are associated with a fare container (despite the fact that one of them has "PRESTO" in its name). 
* In the second example, the `regular_fare` ticket is available both when using a PRESTO Card and when paying by cash.

If I were modelling the fare system in Toronto, I'd use the second approach, for the same reasons as @flocsy explained.

isabelle-dr: Thanks for answering!
Very clear

isabelle-dr:
Hello everyone, 

MobilityData will host a roundtable discussion this week on Thursday 1 December 11:00 AM ET to discuss the open items in this PR.
We are also hosting a discussion around higher-level GTFS Fares-v2 discriminating factors and design principles, which will happen the day before on Wednesday 30 November 11:00 AM ET and it is intended to inform the decisions for this PR.

You can access both events via [this calendar](https://share.mobilitydata.org/events-calendar), and you can reach out to [specifications@mobilitydata.org](mailto:specifications@mobilitydata.org) to get an invite on your e-mail.

npaun:
We'd love to see fare containers move forward as a simple, straightforward proposal.


## Minimum requirements

If you strip the proposal down to the bare minimum, you have this:

**`fare_products.txt`**

| `fare_product_id` | `fare_product_name` | `amount` | `currency` | `fare_container_id` (new) |
|-----------------|-------------------|--------|----------|-------------------------|
| single-ride     | Single ride       | 2.50   | USD      | clipper                 |


**`fare_containers.txt`**

| `fare_container_id` | `fare_container_name` |
|-------------------|---------------------|
| clipper           | Clipper card        | 


With just this functionality, we can already tell riders whether they need to use a fare container for their trip, and what it's called. This proposal documents the existing data produced by Interline, consumed by Transit and Apple. It gives us room to expand as needs are identified concretely later.


## Payment types

I'd propose going one step further in this PR. We should add a `fare_container_type` enum including:

- Fare card
- Mobile app
- Contactless payment
- Cash

This has a number of advantages:

* As @e-lo mentioned, just being able to show riders a contactless symbol where this is available is a massive value-add to user experience.
* We would have just one consistent abstraction covering all of the different 'ways to pay', rather than treating fare cards as as a special case.

We're already using a similar approach internally (for example, we detect Interline's `cash` and `clipper` fare containers), but it would be nice to have a standardized approach.  

If necessary, I'd suggest proceding with a vote on just the 'bare minimum' proposal for now, with payment types being a 'nice to have' extension.

isabelle-dr:
Hello, 

We have worked on a modification for this proposal. We believe this would address the in-scope issues raised on this PR & during the [roundtable discussions](https://github.com/google/transit/issues/341#issuecomment-1339947915), and it could be extended to address the other needs expressed.
**Please note that Mobilitydata would like to open a vote on this proposal if producers and consumers already working with fare containers can confirm this proposal works in practice: Interline, Apple, TransitApp, Cal-ITP, Maryland Transit Administration, San Diego MTS**

Here are the changes:
- renamed `fare_containers` to `fare_payment_types` to include cash and bank cards, as proposed by @bdferris-v2 and as a result of discussions in the original PR. We acknowledge that this concept contains payment methods (cash and bank card) and "supports" used to validate a fare product that has already been purchased (transit card, mobile app). We see payment types as "what a traveler can bring with them to pay for the ride". We explored keeping these concepts separate and it led to considerably more complexity for questionable value. 
- dropped `amount` and `mimimum_initial_purchase` as @npaun suggested.
- instead of having unique ids in `fare_payment_types.txt`, each id would form a group, a bit like how `fare_products.txt` work: in the same way one `fare_product_id` is a group of fare products that have the same behaviour in leg and transfer rules, `fare_payment_method_group_id` is a group composed of payment methods that have similar behaviour. This would avoid duplicating lines in `fare_products.txt` when we have the same price for N payment methods as @bdferris-v2 flagged.

#### Functionalities this allows
- model price differences based on the payment type
- show a rider the “tap-to-pay” icon, meaning that a contactless bank card can be used, as flagged by @e-lo 
- show riders transit cards and mobile apps that can be used for the ride

#### Further possible iterations
- allowing a simple design for cash-only fare systems: consider that empty `fare_products.fare_payment_type_group_id` means cash or add `fare_payment_type` in `fare_products.txt` with a mention that only one of `fare_payment_type` and `fare_payment_type_group_id` can be specified for a record in `fare_products.txt`
- modeling transfers discounts (where a transfer is free only if a certain payment method is used): we could imagine using a method similar to the idea behind the `filter_fare_product` field: `fare_payment_method_type` in `fare_transfer_rules.txt`, for example.
- expanding `fare_payment_types.txt` to add how much a transit card costs, how much has to be (or can be) loaded on it, more enums, etc.

This proposal is available in a google doc format with a data example at [https://share.mobilitydata.org/fare-containers-to-fare-payment-types-proposal](https://share.mobilitydata.org/fare-containers-to-fare-payment-types-proposal).

Looking forward to hearing your thoughts!

isabelle-dr:
After the feedback received on the proposal document, I have made the following modifications:
- modified the file name to `fare_payment_options.txt` so that the file and the field have different names
- updated the descriptions in the Enum field

flocsy:
Shouldn't the file be called fare_payment_option_groups.txt?
Or shouldn't the primary key only contain fare_payment_option_name and the description of fare_payment_option_group_id explain that the same group id can be used on multiple lines?

isabelle-dr:
@flocsy to your points:

> Shouldn't the file be called fare_payment_option_groups.txt?

The rationale we used is that the singular form of the filename represents a single item in the file. "routes.txt" -> every line is a route. "fare_payment_options" -> every line is a fare payment option (a group contains multiple lines).

> Or shouldn't the primary key only contain fare_payment_option_name and the description of fare_payment_option_group_id explain that the same group id can be used on multiple lines?

`fare_payment_option_name` is optional, and it can be left empty if `fare_payment_option_type` is 0 or 1, see the example below.

fare_payment_option_group_id | fare_payment_option_name | fare_payment_type
-- | -- | --
non_cash | clipper | 2
non_cash |  | 1
cash |  | 0

flocsy:
Did you make up that rationale about the single naming or is it documented anywhere? I never heard of it. But anyway it makes not much sense, because it's not that we can only have 1 group in the whole file.

BTW I hate the unnormalizedness of the GTFS files. Why multiple lines define 1 record?!???! It just becomes hard to write/read/validate/understand :(

bdferris-v2:
I'm the one who made up that guidance :)  since I think it generally applies to other files in spec.  (e.g. each row of `routes.txt` is a route, each row of `stops.txt` is a stop).  And here I think @isabelle-dr comment above still applies.  The individual rows of `fare_payment_options.txt` are fare payment options, not groups.  I could imagine a hypothetical `fare_payment_option_groups.txt` file would have one entry per fare payment option group, where `fare_payment_option_group_id` would be a unique primary key.  That said, there doesn't yet appear to be a need for this file, as there aren't any properties assigned to a group itself, other than the fare payment options associated with it.  But maybe that would change in the future?

To your point about normalization, I think I recognize your concern.  Because of the way the primary keys are specified in this PR for both `fare_products.txt` and `fare_payment_options.txt`, a provider might choose to model their system in such a way that both fare products and fare payment options are duplicated in both files for the purposes of different groupings.  I'll admit I'm not sure how much of a problem that will be in practice.

For example, is it a problem if we end up with effectively duplicate entries in `fare_payment_options.txt` (e.g. multiple entries for cash), such that a particular payment type can be re-used in different payment option groups (e.g. one group for cash-only and another group for cash+contactless)?  I don't know how strongly I feel about that.  We seemed to have erred on the side of reducing the complexity of the files that providers need to generate at the expense of some normalization.

I assume the counter proposal would be a `fare_payment_options_groupings.txt` file that has a (fare_payment_option_group_id, fare_payment_option_id) pairs allowing fare payment options to be assigned to multiple groups?

flocsy:
OK, let it keep the way it is in the PR

bdferris-v2:
So... having some time to think about comments from @flocsy and the addition of `fare_payment_option_group_id` to the primary key of  `fare_products.txt`, along with implementing validation support in [PR #1305](https://github.com/MobilityData/gtfs-validator/pull/1305), I'm going to ask some potential tricky questions about fare resolution in this proposal:

Similar to the question asked by @npaun in ["Should fare transfers require that the same fare product is used on both legs?"](https://docs.google.com/document/d/1zF5i03BFQi1kVBfN17aR7N1KRrpsHs3B9YjCzE8apJE/edit?usp=sharing), how should different combinations of fare payment options be considered across multi-leg journeys?

For example, let’s consider the following example where a given fare product has a fare-card price and a more expensive cash price:

```
fare_product_options.txt
fpo_group_id,fpo_name,fpo_type
cash,Cash,0
card,Transit Card,2

fare_products.txt
fare_product_id,fpo_group_id,amount
fp0,cash,$2
fp0,card,$1
```

For a two-leg journey where both legs use fare product `fp0`, I assume a consumer app might present: $2 Transit Card or $4 Cash.  From an implementation stand-point, should a consumer app consider the combination of `cash` for the first leg and `card` for the second leg but then reject it because the consistent combinations were “better”?  And what do I mean by "better"?  Simpler?  Not necessarily cheaper.

Or should the consumer app not consider the combination of cash+card because the `fare_group_option_ids` do not match?  Maybe it’s not possible to mix and match fare payment options across a journey, for example.  Whatever the behavior, I think we might want to be explicit about this given the existing confusion around fare products and transfers.

Also, echoing on my [comment](#issuecomment-1381226777) above, we might also want to clarify when entries in `fare_product_options.txt` are considered equivalent for the purposes of combining into a single entry.  Right now, two entries are equivalent if the (fpo_name, fpo_type) are equivalent but will that hold as we add new fields to the file?

Expanding on that idea, let’s say we’ve got one fare product that can be purchased with cash or card, and another that can be purchased only with cash.  A data provider could seemingly model that in two ways:

1) Duplicate cash entry in fare_product_options.txt:

```
fare_product_options.txt
fpo_group_id,fpo_name,fpo_type
cash,Cash,0
cash_or_card,Cash,0
cash_or_card,Transit Card,2

fare_products.txt
fare_product_id,fpo_group_id,amount
fp0,cash,$2
fp1,cash_or_card,$1
```

2) Duplicate entries in fare_products.txt

```
fare_product_options.txt
fpo_group_id,fpo_name,fpo_type
cash,Cash,0
card,Transit Card,2

fare_products.txt
fare_product_id,fpo_group_id,amount
fp0,cash,$2
fp1,cash,$1
fp1,card,$1
```

Do we have a preference on how data should (or even must) be modeled?

flocsy:
To me the duplicate entries in fare_products looks more natural (and more useful because it also allows discount if one uses the card vs cash for fp1)

davidlewis-ito:
I would suggest that both approaches are valid but that it comes down to whether the products purchased by (say)  cash or card are completely the same. 

If the purchased product affords the rider exactly the same rights and same restrictions then it seems to make sense to group the payment types to avoid complicating the fare_products table.

But often that is not the case.  Different pricing has already been mentioned, but increasingly we are seeing more generous transfer pricing available to those paying via a electronic means.  In this case the purchased product is definitely not the same. 

There is also the case for a hybrid approach; perhaps using a transit card (Oyster in London) or a NFC bank card might be grouped, but the use of cash would require a separate product due to both a higher price and less transfer concession.

isabelle-dr:
Thank you @bdferris-v2, @flocsy and @davidlewis-ito for your latest comments. I'll propose a modification, so that we don't end-up with different interpretations of the same fare designs.

isabelle-dr:
Okay, it looks like there is a general agreement for option 2 drafted [here](https://github.com/google/transit/pull/355#issuecomment-1382493289). Fare payment options should be grouped when both the following statement are true:
- no fare product price depends on the options that compose a group
- the same leg transfer rules apply (although I'd argue this is more a statement for fare products, which in turn, should also be true for fare payment options)

I propose to also explicitely mention that `fare_payment_options.txt` shouldn't contain duplicate records (same `fpo_name` and `fpo_type`).

I agree with @flocsy that it looks more natural. Additionally, as @bdferris-v2 flagged, we might want to expand  `fare_payment_options.txt` to represent additional information about each fare payment option (what's the minimal initial purchase to be loaded + how the rider can purchase the transit card in the case of type `2`, how validation is made, etc). 



Now to the questions around how to treat payment options in multi-leg journeys. 

> should a consumer app consider the combination of cash for the first leg and card for the second leg but then reject it because the consistent combinations were “better”? And what do I mean by "better"? Simpler? Not necessarily cheaper.

@bdferris-v2 I think this up to the consumer. A rider can, in a multi-leg journey where no transfer rules apply, pay cash for the first leg and then use a transit card for the second, but that might not be an option you want the trip planning app to propose. As @flocsy mentioned, you can optimize and only show the "best" options, or even ask the user to input the option they have.

> Very often the payment method must be the same across the legs for the transfer pricing to be applied. So the rider must use the transit card for both legs, or the NFC bank card for both legs, but not a mix.

@davidlewis-ito I think needing to use the same transit card or NFC bank card for both legs is equivalent to Transit's [Should fare transfers require that the same fare product is used on both legs?](https://docs.google.com/document/d/1zF5i03BFQi1kVBfN17aR7N1KRrpsHs3B9YjCzE8apJE/edit#heading=h.9a66f7frd08b). I believe the reason the user needs to use the same payment option is: the fare product validation is logged on their payment option (transit card or credit card). If they change the payment option, they buy a new fare product. If a system could link a user's credit card & transit card under an account, this requirement might no longer exist. Please tell me if you see this another way. 
That being said, going towards option 2 still gives us the flexibility to explicitely mention a payment option type group in another file if we needed to.

isabelle-dr:
I've updated the description of the `fare_payment_options.txt` to clarify its intended use.
I'm also taking note to improve the description of the fare products file, which I can commit to doing a separare PR.

Please let me know if this addresses these latest comments, thank you for your contribution to this work!

drewda:
We've created a sample of what it would look like to migrate the SF Bay Area Regional Feed from `fare_containers.txt` to `fare_payment_options.txt`: https://gist.github.com/drewda/b454c5f5c21ba41ca6d4c029cec88533

We've also discussed this with the teams at Transit app and Apple. Once this PR is adopted, we'll go through a 2-step migration to produce both files for consumers a while, and then after a couple weeks only produce the new `fare_payment_options.txt` file.

From our perspective at Interline and MTC, we're ready to move ahead with a vote on this PR.

isabelle-dr:
Thank you @drewda for sharing this sample. 
MobilityData is planning to open a vote on this proposal this week, on Wednesday, January 25.
May I ask @bdferris-v2, @davidlewis-ito and @flocsy to have a look at the changes I proposed in [0ea63c0](https://github.com/google/transit/pull/355/commits/0ea63c018e4f1f978bd287b8613d48a2fcbedb66) as a result of the latest comments?

Please note that we will have the opportunity to discuss this proposal in tomorrow's working group meeting ([reference here](https://github.com/google/transit/issues/341#issuecomment-1400719020)) and clarify any question or doubt that persists if any.

npaun:
From our side at Transit, we're also ready for the vote on this extension. As @drewda mentioned, we'll switch to consuming `fare_payment_options.txt` once it's adopted.

JloupGauducheau:
proposal for additional attributes for fare containers 
_ replacement_cost : several agencies invoice an extra cost for replacing a physical fare container lost by the user  : 
Attributes poposed charactéristics 
- type non negative real 
- applied currency (same as Fare_container_amount 
- optionnal 
- Fare container-type : list of possible physical supports for fares product  (allow a filtering of existing products for fare assistant of fare calculation engines)  the most common categories are 
               - paper ticket  
               - anonymous contactless card 
               - non transferable card
               - QR code 
               - smartphone 
   an enum of possible values should be rather selected

JloupGauducheau:
open question / suggestion on relation between agencies and fare containers
context :   the GTFS Zip archive is provided by a Public transport Authority and usually represents the globality of the PT network  covering a City ant its suburbs. This archive may gather several agencies (representing the main company and its subcontractors operating several lines of the global networks 
fares products and fare container may be associated to an agency or a group of one like this example https://www.reseau-astuce.fr/fr/carte-astuce/81  whith describes  the "carte Astuce” (a contactless card which supports fare products covering  Rouen city urban public transport)  This container can supports several fares products (up to 4)  

suggestion add Agency-id to fare product file to

isabelle-dr:
@JloupGauducheau thanks for the comments!
Did you see the latest updates to this proposal? We've modified "fare containers" to "fare payment options" to represent more broadly what travelers need to bring for their journey: https://share.mobilitydata.org/fare-containers-to-fare-payment-types-proposal.

Points 1 and 4: I believe the `fare_payment_options.txt` file could be extended to represent this information. 
Point 3: I'd point out that it is more on the operational end (compared to traveler information). We would need to discuss if such information is in scope for GTFS. 
Point 2: Is `fare_payment_option_type` solving this need?  

I will open an issue to continue the discussion around extending this proposal, in order to keep the scope of this PR to what Interline, Apple, and Transit app. are committed to producing/consuming right now.

> suggestion add Agency-id to fare product file to represent this link: possible n:n association
Good point. I suggest we also discuss this in the follow-up issue.

isabelle-dr:
Hello, 

I have updated this PR based on what was discussed in the last working group meeting (see the summary [here](https://github.com/google/transit/issues/341#issuecomment-1404472638)). The Google Doc version of this proposal has also been updated, if that makes it easier to review: https://share.mobilitydata.org/fare-containers-to-fare-payment-types-proposal.
Thank you for your patience, support, and collaboration while working through this proposal. We're almost there :)

Next step: A second 30min discussion on January 26th from 11 to 11:30 AM ET to make a decision for the `fare_payment_option_type` field concerns. Please write to [specifications@mobilitydata.org](mailto:specifications@mobilitydata.org) if you're interested in joining. The meeting notes are in the [same document](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit?usp=sharing).

isabelle-dr:
Hello all, 

Just a quick update: we were not able to reach a consensus regarding the `fare_payment_option_type` concerns.
MobilityData will explore further the implications of including payment methods (bank cards, cash) at this stage, and how it affects the [GTFS-TravelRules](https://docs.google.com/document/d/1w9g-CRClM0P6NbnUf8HFzK8UVNxzciPIA7WMA7KbkOg/edit?usp=sharing) (that builds on top of GTFS-Fares v2).
I also want to acknowledge that our producer for this PR (Interline) dataset doesn't need to group fare payment options, so the additional `fare_payment_option_groups.txt` will be removed from this PR.

isabelle-dr:
Hello, here is an update on this pull request. We have explored what the later stages would look like, and the Google Doc has been modified with examples and suggestions for further iterations: https://share.mobilitydata.org/fare-container-updated-proposal.
We would like to suggest two modifications to the proposal, and I would like to know if there is a general consensus.
- renaming the file and fields to **fare media**, and defining it as **media that can be used to load and use fares**. It is not: payment methods or validation methods, even if for some of these fare media, these additional properties won't have to be explicitly defined. I am sorry in advance for the change of names at this stage in the proposal, we believe removing "payment" from the name will reduce confusion. More about the reason we chose this name in [this section](https://docs.google.com/document/d/1xUNwjRVSUq9SqPYX1AIYINvTw2dcxeRRRJOmQ6UP_8M/edit#bookmark=id.nfb9ytve7no1).
- the categories for the `type` field. We suggest:
  - **0 or empty: None**.
  - **1: Paper Ticket**. In further iterations, this category will be linked with more information specific to paper tickets, such as where to purchase one, how to validate it, etc.
  - **2: Transit card (physical) used to store fare products and fare value.** Similarly than the paper ticket, this category will be linked with more information regarding where to purchase / load and how to validate, but also what is the cost (if any), and if an initial fare amount needs to be loaded on it or not.
  - **3: cEVM as a token container for account-based ticketing.**. This is a contactless bank card. Both physical cards and virtual cards are included in this category, because this distinction won’t be displayed to riders, and it won’t play as a requirement for other fields (validation method and payment methods won’t be affected).
  - **4: Mobile app used to store transit cards and fare products.** . This includes:
     - Virtual wallets (Apple wallet, Google wallet). Note that for a contactless credit card stored in a virtual wallet, the category is 3 (cEMV). This category is for transit cards stored in a virtual wallet.
     - Apps created around one public transit agency of transit card (PRESTO app, SNCF Connect)
     - Apps that operate across many transit agencies (Token Transit, Transit App)

Other categories could be added, such as **SMS**. More about why these categories are suggested in [this section](https://docs.google.com/document/d/1xUNwjRVSUq9SqPYX1AIYINvTw2dcxeRRRJOmQ6UP_8M/edit#bookmark=id.pmvwp0nr8kk5).

Are there any concerns with these two modifications?

isabelle-dr:
Hello everyone! 
We've received encouraging feedback on the work done to this proposal, so we will call a vote on this PR on **February 16 at 6 PM ET**.
We see more and more stakeholders producing GTFS-Fares v2 data, and we would like to move forward, acknowledging that not all questions around payment and validation methods will have a definitive answer.

We are keeping this PR as a first simple and expandable iteration. As a reminder, the scope is: 
- Define the fare media that an agency accepts
- Describe the fares if riders pay using their fare containers media (vs. tickets not loaded on a fare container)
We will coordinate other iterations based on the [proposal](https://share.mobilitydata.org/fare-container-updated-proposal) that contains the results of the working group discussions. 

We acknowledge that representing "cash is the only way to take this ride" is not covered by this iteration, in an attempt to reduce confusion around what "fare media" is. The cEMV is, on the other hand, considered "fare media" because it is used as a token for account-based ticketing, which allows fare capping and transfer rules.

Fare containers are previously defined is produced by Interline ([get the dataset here](https://www.interline.io/blog/mtc-regional-gtfs-feed-additions/#using-fares-and-transfer-discounts)), Cal-ITP (many datasets), Maryland Transit Administration ([dataset](https://mdotmta-gtfs.s3.amazonaws.com/mdotmta_gtfs_metro.zip)), San Diego MTS ([dataset](https://www.sdmts.com/google_transit_files/google_transit.zip)), and consumed by Apple and TransitApp.

Apple Maps "clipper card fare":
Screenshot 2023-02-14 at 2 09 06 PM

Interline, Apple, and TransitApp are committed to implementing the changes proposed here. Here is an [updated version of Interline's MTC data](https://drive.google.com/file/d/1Wx3EMEV3kMDmi2UKO7Z6_rrj6QypwngB/view?usp=share_link), that I am posting on their behalf.

We believe this is enough to cover the _Speculative features are discouraged_ mention in the [Specification Amendment Process](https://gtfs.org/schedule/process/).

gcamp:
The GTFS change process says 
> Before calling for a vote, at least one GTFS producer and one GTFS consumer should implement the proposed change.

I don't think "committing" to implementation fulfil that condition. 

Moreover, we said we would want to implement this functionality to add support for cash and contactless payment and 1) cash has been removed and 2) we would not support all enum in that field. 

We have more thoughts on the proposition that @npaun could share, but generally I don't think this is ready for a vote unfortunately.

isabelle-dr:
Hello @gcamp, thanks for sharing these concerns. 

>  1) cash has been removed

Acknowledging this change in scope. `cash` has been added to the scope when we've added a `type` field, and then removed when we've worked on defining what _fare media_ is and what the difference is with fare payment methods.
Do you have a concern about the modeling choice itself, or with the fact that we haven't gone far enough to include cash as part of the proposal, even if it requires a second file?

2) we would not support all enum in that field

Which ones would you support?

isabelle-dr:
About the condition to have the changes implemented before opening the vote, my understanding is that it's more a recommendation (because the term _should_ is employed) and that is there to make sure the changes work well in practice.
Other statements are requirements, such as "must be at least 7 calendar days".
@npaun seemed to imply [here](https://github.com/google/transit/pull/355#issuecomment-1401113356) that the vote could be opened before the changes are implemented in production.
I'll admit that having the changes up and running in production is ideal, and we will defer to what the community wants to do. 

I'd like to hear from others as well here because the vote is planned for today.

davidlewis-ito:
Hi,

I think that the rescoping of this PR to be focussed on "fare media"  is a sensible step and addresses some confusion that was present in earlier drafts.  The fare media now represents the means by which the passenger is able to demonstrate entitlement to ride.  

Some paper examples :  
- London Tube offers paper tickets that may be purchased at a station vending machine using cash or bankcard, and are then inserted into the faregate.  The media type is paper. 
- Boarding a bus which accepts cash, the rider would receive a ticket (perhaps also referred to as a receipt) and this is also a paper fare media.

Where the fare may be purchased (on vehicle, in station, 3rd party vendors) and how it is paid for (cash, bankcard,..) are separate questions that can be handled by further extensions.

NomeQ:
In this implementation of fare_media, is `fare_media_id` required to be unique, or can multiple `fare_media_type` be grouped together under the same ID, such as how `leg_group_id` and `fare_product_id` currently work?

isabelle-dr:
@NomeQ it would be unique, at this point. 
The [proposal](https://docs.google.com/document/d/1xUNwjRVSUq9SqPYX1AIYINvTw2dcxeRRRJOmQ6UP_8M/edit) includes an additional `fare_media_groups.txt` to create groups that can then be referenced from `fare_products.txt`.

drewda:
We've updated our data sample from the SF Bay Area Regional GTFS Feed to reflect the latest version of the proposal in this PR: https://gist.github.com/drewda/b454c5f5c21ba41ca6d4c029cec88533

Interline and MTC are fine adopting this and producing this new scheme. However, we want to make sure that Transit app and Apple (as current consumers of fares data in the SF Bay Area Regional Feed) are also fine consuming this and that it will provide sufficient information regarding cash/default journey cost.

isabelle-dr:
Hello, 
Just a heads-up that MobilityData has decided not to call a vote on this proposal quite yet because of the concern that has been raised (this addition is not in a production environment). 
In the meantime, please review the proposal sooner than later, and do not hesitate to ask any questions you might have so that our early adopters, who will need to update their implementation from _fare containers_ to _fare media_, don't need to go through this process several times.

drewda:
For those who need a full zip rather than just a subset of the operative changes, here is a full copy of the SF Bay Area Regional Feed that has been manually edited to reflect the current version of this proposal: [rg-with-fare-media-2023-02-17.zip](https://drive.google.com/file/d/1_mjZITrgoT2v9qqluJ3srAKiyXFl9woh/view?usp=share_link)

NomeQ:
PR in Mobility Database for feed published with `fare_medium_type=3` -- https://github.com/MobilityData/mobility-database-catalogs/pull/253

westontrillium:
Typo: `fare_medium_type=2`'s description: "Physical transit card that have stored tickets, passes or monetary value."

Should be "Physical transit card that **has** stored tickets, passes or monetary value."

isabelle-dr:
Hello everyone, 
The proposal has been updated to reflect the last details (using _media_ instead of _medium_ for the singular, updating the description of `fare_media_type = 0`, and resolving small typos).
Apple has updated the changes to their end-to-end test environment, and they confirm that the dataset posted by Interline works and can replace the current production dataset (with fare_containers.txt).
Please note that `fare_media_type=1` is reserved for physical paper tickets and will be adopted when we have a producer for it.

We are opening a vote for adding Fare Media, which is part of the GTFS-fares v2 proposal.

There are at least two producers:
- [Interline for SF Bay Area Regional](https://www.interline.io/blog/mtc-regional-gtfs-feed-additions/#using-fares-and-transfer-discounts) (@drewda)
- [Cal-ITP](https://dot.ca.gov/cal-itp) (@e-lo)

And at least one consumer:
- [Apple Maps](https://www.apple.com/ca/maps/)

Voting ends on 2023-03-13 at 23:59:59 UTC.

bdferris-v2:
I guess someone needs to go first:

+1 from Google

Though I'll note most of the test feeds still reference `fare_medium_*` field names, not `fare_media_*`.   I don't think that's a blocker, but wanted to call it out.

flocsy:
Added a comment about what I think is a small typo. I'll let the native English speakers decide and fix that if it's necessary. Other than that I am voting for it:
+1 Moovit

matikin9:
+1 LA Metro!  💃

drewda:
+1 from Interline (as producer of the SF Bay Area Regional Feed)

e-lo:
+1 🥳 👩‍🎤 🎸  Cal-ITP and Friends. 🙇‍♀️ to all who helped make this happen. 

We are really looking forward to getting this information to customers!

westontrillium:
+1 from Trillium!

brodyFlannigan:
+1 Transcollines!

gtfs/spec/en/reference.md
```diff
@@ -343,21 +345,37 @@ For examples that demonstrate how to specify a fare structure with [fare_rules.t
...
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+|  `fare_media_id` | Unique ID | **Required** | Identifies a fare media. |
+|  `fare_media_name` | Text | Optional | Name of the fare media.<br><br>For fare media which are transit cards (`fare_media_type =2`) or mobile apps (`fare_media_type =4`), the `fare_media_name` should be included and should match the rider-facing name used by the organizations delivering them. |
+|  `fare_media_type` | Enum | **Required** | The type of fare media. Valid options are:<br><br>`0` - None.  Used for when there is no fare media involved in purchasing or validating a fare product, such as paying cash to a driver or conductor with no physical ticket provided.<br>`2` - Physical transit card that has stored tickets, passes or monetary value.<br>`3` - cEMV (contactless Europay, Mastercard and Visa) as an open-loop token container for account-based ticketing.<br>`4` - Mobile app that have stored virtual transit cards, tickets, passes, or monetary value.|
```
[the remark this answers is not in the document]
isabelle-dr: fixed in https://github.com/google/transit/pull/355/commits/7f0860041c53e1465b8f5a47aa38081ec1065c20

isabelle-dr:
The voting period ended on 2023-03-13 at 23:59:59 UTC.

With 9 votes in favour and no votes against, the vote to extend GTFS with Fare Media passes! The votes came from:
Google (@bdferris-v2)
Moovit (@flocsy)
ITO World (@davidlewis-ito)
Apple Maps (@saraemarcus)
LA Metro (@matikin9)
Interline (@drewda)
Cal-ITP (@e-lo)
Trillium (@westontrillium)
Transcollines (@brodyFlannigan)

The [GTFS Fares-v2 Google Doc](share.mobilitydata.org/gtfs-fares-v2) has been updated to reflect this change and also contains additional Fare Media fields that were not part of this Pull Request, mainly: 
- the `paper_ticket` value for `fare_media_type`
- the `fare_media_groups.txt` file, which provides a mechanism to model groups of fare media that can each be associated with the same fare product
- a `fare_validation_method` field

Many thanks to everyone who participated in this Pull Request and in the working group meetings! Your time and dedication have truly made this possible. Thank you all so much for your outstanding contributions!

For those of you who want to stick around, the work continues in [PR#357](fare_media_validation_method) to add time-variable fares!

drewda:
Thank you, @isabelle-dr and all.

For those interested in how this will affect the SF Bay Area Regional Feed, our team will be taking the following steps to migrate:

1. add `fare_media.txt` (and continue to include `fare_containers.txt`)
2. wait a few weeks while data consumers switch to processing the new `fare_media.txt` file)
3. remove the `fare_containers.txt` file

We'll also be updating our tables at https://www.interline.io/blog/mtc-regional-gtfs-feed-fares-updates/#gtfs-fares-v2-base-implementation-and-future-additions-to-the-specification to reflect the adoption of `fare_media.txt`

(Any questions about this and other topics related to the SF Bay Area Regional Feed are welcome at https://groups.google.com/g/511sfbaydeveloperresources/ )

PR #357, opened 2022-11-02
Add variable fares by time or day

isabelle-dr:
Fares can vary based on the time of day (described using `timeframe_id`) and day of week/year (described using `service_id`).
This pull request covers fares that vary by time and day, which is a section of the entire [GTFS-Fares v2 proposal](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.mhjuwqk2h5r3).
- Add new file: `timeframes.txt`, to define moments in time where the fare applies.
- Extend `fare_leg_rules.txt` with `from_timeframe_id`, and `to_timeframe_id` to specify that a fare leg rule applies only if the beginning or end of the leg is in a specified timeframe. 

**_Here's a quick example_** 

- A regular fare is required to travel.
- The fare is discounted if riders travel between 6:00 AM and 7:00 AM. 
- On New Year's, transit is free (no fare).

Define service days and New Year's using `calendar.txt.` and `calendar_dates.txt`
| service_id | monday |tuesday  |wednesday |thursday |friday |saturday |sunday |start_date |end_date |
|--|--|--|--|--|--|--|--|--|--|
| regular_service | 1 |1  |1 |1 |1 |1 |1 |20230101 |20231231 |

| service_id |date  |exception_type|
|--|--|--|
|regular_service  | 20230101 |2 |
|new_year_service  | 20230101 |1 |

Define the discounted timeframe using `timeframes.txt`
| timeframe_id |start_time  |end_time|service_id|
|--|--|--|--|
|morning  | 06:00:00 |07:00:00 |regular_service|
|regular  | 00:00:00 |06:00:00 |regular_service|
|regular  | 07:00:00 |24:00:00 |regular_service|
|all_day  | 00:00:00 |24:00:00 |new_year_service|

Define the time and date-based fares using `fare_leg_rules.txt`
| leg_group_id | from_area_id |to_area_id  |network_id |from_timeframe_id |fare_product|
|--|--|--|--|--|--|
| leg1| zoneA |zoneB  |bus |regular |regular_fare |
| leg1| zoneA |zoneB  |bus |morning |rdiscounted_fare |
| leg1| zoneA |zoneB  |bus |all_day |free_fare |

### [Original message posted by @isabelle-dr on November 2, 2022. OUTDATED]
Hi everyone!

I am opening this Pull Request as part of the second implementation of Fares v2. I am opening it in the exact same state PR https://github.com/google/transit/pull/343 was when closed, and I will update the proposal based on feedback already received so we can resume where we left off. For more information on the plan, please refer to https://github.com/google/transit/issues/341#issuecomment-1296266122.
This pull request covers fares that vary by time and day, which is a section of the entire [GTFS-Fares v2 proposal](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.mhjuwqk2h5r3).

Fares can vary based on the time of day (described using `timeframe_id`) and day of week/year (described using `service_id`).

- Add new file; `timeframes.txt`, to define timeframes.
- Extend `fare_leg_rules.txt` with `from_timeframe_id`, `to_timeframe_id`, and `service_id` to describe time-dependant fares.

The time-dependant variables (timeframe and day information) are modelled into fare legs (and not fare products) since in GTFS-Fares v2, `fare_leg_rules.txt` models the location and time variables (from/to area, from/to timeframe, and service ID). All other factors are modelled in fare products. 

**_Here's a quick example_** 
- A regular fare is required to travel.
- The fare is discounted if riders travel between 6:00 AM and 7:00 AM. 
- On New Year's, transit is free (no fare).

Define service days and New Year's using `calendar.txt.` and `calendar_dates.txt`
| service_id | monday |tuesday  |wednesday |thursday |friday |saturday |sunday |start_date |end_date |
|--|--|--|--|--|--|--|--|--|--|
| regular_service | 1 |1  |1 |1 |1 |1 |1 |20230101 |20231231 |

| service_id |date  |exception_type|
|--|--|--|
|regular_service  | 20230101 |2 |
|new_year_service  | 20230101 |1 |

Define the discounted timeframe using `timeframes.txt`
| timeframe_id |start_time  |end_time|
|--|--|--|
|morning  | 06:00:00 |07:00:00 |
|regular  | 00:00:00 |05:59:59 |
|regular  | 00:07:01 |23:59:59 |
|all_day  | 00:00:00 |23:59:59 |

Define the time and date-based fares using `fare_leg_rules.txt`
| leg_group_id | from_area_id |to_area_id  |network_id |from_timeframe_id |to_timeframe_id |service_id |fare_product|
|--|--|--|--|--|--|--|--|
| leg1| zoneA |zoneB  |bus |regular | all_day |regular_service |regular_fare |
| leg1| zoneA |zoneB  |bus |morning | all_day |regular_service |discounted_fare |
| leg1| zoneA |zoneB  |bus |regular | all_day |new_year_service |free_fare |
| leg1| zoneA |zoneB  |bus |morning | all_day |new_year_service |free_fare |

Data consumer: [Apple](https://www.apple.com/)
Data producer:   [ITO World](https://www.itoworld.com/)


isabelle-dr:
Hello everyone! 

Here is an update on this work. 

**What was done based on the feedback received on #343**
- Updated the `timeframe.txt` file description to remove the 24h period mention, so that it stays in line with the rest of GTFS (24+h format). Result of @skinkie's [feedback](https://github.com/google/transit/pull/343#issuecomment-1206712203).
- Mentioned that start and end times are included in the interval, as a result of @flocsy's [feedback](https://github.com/google/transit/pull/343#issuecomment-1207421143).
- Made the `end_time` from **Required** to **Optional**, and added that if left empty, it is considered that the fare only depends on start time. This is because most agencies do not use the trip end times to process time-based fares, and it avoids the hassle of creating a separate entry in timeframes.txt to describe "All day". This is a result of the [August 24 roundtable discussion](https://github.com/google/transit/pull/343#issuecomment-1240127981).

**Outstanding items**
- Timeframe type: the need for a file or field to describe the timeframe type was highlighted since transit services have different ways of defining when a fare is activated (at station entrance, upon boarding, upon departure, etc.).
One option is to extend `fare_leg_rules.txt` with `timeframe_type` with enumerations, and @flocsy asked why this couldn't be added in `timeframes.txt`. This is a result of the [August 24 roundtable discussion](https://github.com/google/transit/pull/343#issuecomment-1240127981).
- Timeframe rules: There is a need to better define rules that apply in timeframe design, mainly:
  - If one timeframe is defined for an interval on a given day, we might need to explicitly mention that the whole day must be covered with timeframes
  - We might add a mention that we must not have overlapping timeframes with the same `timeframe_id` 
- We might want to have trip-level attributes to the fare calculation, for the case when the determination of the peak-ness of a leg is determined by the time that vehicle leaves its origin (or reaches its destination). This is a result of @davidlewis-ito's [feedback](https://github.com/google/transit/pull/343#issuecomment-1209571989). 

I believe these were all the items discussed, please let me know if you think otherwise.

flocsy:
There's a typo:
`end_time` | Time | **Optional** |  **Start** time for the interval.
it should be "End"

flocsy:
"Mentioned that start and end times are included in the interval"
I'm not sure if this is what you meant (not what I proposed) or you meant what I proposed but the wording is incorrect...
I meant that when we talk, we always talk about intervals like: peak: from 08:00 to 10:00 ; off-peak: 10:00-, so maybe it's better to modify the description to make this the official format. However according to the current wording we need to write: peak: 08:00 - 09:59, off-peak: 10:00 -

isabelle-dr:
@flocsy 

> end_time | Time | Optional | Start time for the interval.
> it should be "End"

Thanks, fixed. ✅

> I'm not sure if this is what you meant (not what I proposed) or you meant what I proposed but the wording is incorrect...
> I meant that when we talk, we always talk about intervals like: peak: from 08:00 to 10:00 ; off-peak: 10:00-, so maybe it's better to modify the description to make this the official format. However according to the current wording we need to write: peak: 08:00 - 09:59, off-peak: 10:00 -

Understood. The start and end times are modeled the same way `start_date` and `end_date` are in `calendar.txt`, or `start_time` and `end_time` in `frequencies.txt`: all field values are included in the intervals. We believe that changing this design for timeframes would lead to additional complexity in the spec. I have specified explicitly that `start_time` and `end_time` are included so that people design intervals as _08:00 - 09:59, off-peak: 10:00 - 16:00_, even though they talk about peaks as _08:00 to 10:00 ; off-peak: 10:00 - 16:00_. I'm happy to add an example below the table.

flocsy:
LOL, you just proved my point. You wrote: 

"so that people design intervals as _08:00 - 09:59, off-peak: 10:00 - 16:00_, even though they talk about peaks as _08:00 to 10:00 ; off-peak: 10:00 - 16:00_. "

You should've written: 10:00 - 15:59

davidlewis-ito:
Just to highlight a couple of areas we think could do with some more consdiration

### Determinant of peakness
@isabelle-dr Thanks for highlighting the use-case where the peakness might be related to the trip rather than the time at which the rider joins a trip.

The other case we have raised previously is where the peakness depends upon the time at which the rider enters the fare area of a station - for example the time the rider "touches in" at their origin station on the London Tube.

So we see three cases;

1. Peakness depends upon time the rider joins a trip - covered by timeframes in this PR
2. Peakness depends upon time the rider enters a station - would require amendment to this PR
3. Peakness depends upon time the trip starts or ends - requires either timeframes that apply to trips or (more preferably we believe) the ability to mark an individual trip as having a "fare class" that would then play into fare_leg_rules

### Far_leg_rule or Fare_product
Should these time discriminating factors be used to determine the correct fare via the fare_leg_rule or the fare_product table ? One approach may be to constrain the fare_leg_rule table to be concerned only with the physical journey taken (stops, distance , routes etc ).  The fare_products table would then combine this with the peakness, the containers, the rider_category etc.

This is a big topic I suspect that would be best discussed at the upcoming round-table

flocsy:
Isn't case 1 (joins a trip) and 2 (enters a station) basically the same thing, just with different timefarme_type_id?

davidlewis-ito:
@flocsy Yes !

flocsy:
Which is a hard question. I can imagine that in most if not all places in the world the situation is something like: 8:00-09:59 peak hour. However depending on the vehicle type (bus, tram vs subway,train for example) the point of validation might be different (bus, tram: at boarding time vs subway,train: at station entrance) So the question is what we prefer to "duplicate" in this case. If we add timeframe_type_id to timeframes, then you duplicate lines BOTH in timeframes.txt AND in fare_leg_rules.txt IMHO. If we add timeframe_type_id to fare_leg_rules.txt then we only duplicate lines there. And I think this also makes more sense because "timeframes" is about the time frames, and not where you measure it... 

However, maybe, there's an even better place where we could add timeframe_type_id. In my above example it could go to the networks.txt or something like that, and then we wouldn't need to duplicate lines anywhere...

bdferris-v2:
Hey all, trying to bring myself up to speed on this proposal.  I wanted to re-open a different potentially fraught subject: empty-entry semantics.

The current language:

    An empty entry in `from|to_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.from|to_timeframe_id`.

seems particularly fraught, if I understand it correctly?  I know there was a bit of discussion around this in original PR, but I believe it was highlighted that the second you have two different fare networks in the same feed who both want to use "empty" semantics for their "default" rule and then use different timeframes for their "specific" rules, you'll have a conflict and the fares won't compute as expected.  As discussed, this is generally a problem with "empty" matching in `fare_leg_rules.txt`.

Curious if you have specific thoughts on the timeframe proposal regarding this issue.

flocsy:
@bdferris-v2 I agree with you. This also bothers me every time I see it. It's confusing that usually empty matches anything, but not in this case. If you have a better (less confusing) idea, I'll most probably vote fore it.

isabelle-dr:
Hello everyone, 

MobilityData will host a roundtable discussion this week on Thursday 1 December 11:00 AM ET to discuss the open items in this PR.
We are also hosting a discussion around higher-level GTFS Fares-v2 discriminating factors and design principles, which will happen the day before on Wednesday 30 November 11:00 AM ET and it is intended to inform the decisions for this PR.

You can access both events via [this calendar](https://share.mobilitydata.org/events-calendar), and you can reach out to [specifications@mobilitydata.org](mailto:specifications@mobilitydata.org) to get an invite on your e-mail.

isabelle-dr:
not stale

isabelle-dr:
not stale

isabelle-dr:
still not stale 🙈

gtfs/spec/en/reference.md
```diff
@@ -373,14 +387,20 @@ To process the cost of a leg:
...
+3. If no exact matches are found, then empty entries in `fare_leg_rules.network_id`, `fare_leg_rules.from_area_id`, `fare_leg_rules.to_area_id`, `fare_leg_rules.from_timeframe_id`, `fare_leg_rules.to_timeframe_id`, and `fare_leg_rules.service_id` must be checked to process the cost of the leg:
     - An empty entry in `fare_leg_rules.network_id` corresponds to all networks defined in `routes.txt` excluding the ones listed under `fare_leg_rules.network_id`
     - An empty entry in `fare_leg_rules.from_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.from_area_id`
-    - An empty entry in `fare_leg_rules.to_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.to_area_id`<br/>
+    - An empty entry in `fare_leg_rules.to_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.to_area_id`
+    - An empty entry in `fare_leg_rules.from_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.from_timeframe_id`
```
skinkie, line 401: Can someone elaborate on the meaning "excluding the ones listed under"?

davidtlewis:
We have been talking about a couple of modifications to the scheme discussed so far;
- combining service_id and times within the timeframes.txt file
- moving the timeframe_id out of fare_leg_rules  and into fare_products.

We've illustrated this in the attached example: https://docs.google.com/document/d/1YecLge4nH1x69GLJsOCD5rYKzXoAlfiOBHd6gUzOXsA/edit?usp=sharing

lauramatson:
Per the Fares v2 roadmap, there's a goal to separate schedule information from fare information files (to facilitate more agencies being able to publish fares information when scheduling software vendors don't allow flexibility). Does this add more interdependence by using calendar.txt and calendar_dates.txt, which are already actively used to reflect schedule data?

skinkie:
> Per the Fares v2 roadmap, there's a goal to separate schedule information from fare information files (to facilitate more agencies being able to publish fares information when scheduling software vendors don't allow flexibility). Does this add more interdependence by using calendar.txt and calendar_dates.txt, which are already actively used to reflect schedule data?

This has not yet discussed in the group.

flocsy:
@davidlewis-ito I think there's a problem in the example:
in timeframes.txt, you have the 1st and the last 2 lines that are colliding IMHO:

timeframe_id | start_time | end_time | service_id
tfl_tube_offpeak | 00:00:00 | 06:30:00 | tfl_tube_peak_days
tfl_tube_offpeak | 19:00:00 | 30:00:00 | tfl_tube_peak_days
tfl_tube_offpeak | 00:00:00 | 30:00:00 | tfl_tube_off_peak_days

1. when there are 2 consecutive peak_days then the 1st day's last timeframe is: 19:00 - 30:00, but the 2nd day's 1st timeframe is: 00:00 - 06:30. IMHO sure this should be considered invalid.
2. The same when there's either the 1st day is peak and the 2nd day is off_peak or the other way around.

I'm not sure if I would've caught this if the 1st line would be: 00:00 - 06:00

Regarding the times: if you go for end_time: 09:30:00 and not for 09:29:59, then let's make it even shorter. I'm pretty sure there are no timetables that use seconds (even in Japan), so either remove the seconds or make them optional.

Regarding the place for the timeframe_type: what are the edge cases? Let's say you have paper ticket that you need to validate, but on trams and buses you validate when you enter the vehicle and on the subway you validate when you enter the stations. You use the same paper ticket. In this case it would make more sense to remove timeframe_type from the fare_products.
On the other hand if we move it to timeframes we would need to double the lines in timeframes AND still double the lines in fare_products, because we would have a line for each from_time_frame_id.
So maybe we should think about an even better place for it.

halbertram:
@flocsy Our thinking was that

- GTFS time inherently has overlaps - a trip can be specified as starting at Fri 25:00 or Sat 01:00
- Those are different in that they include information about which day's schedule the trip is considered to be part of
- We need to be able to specify fares in this overlapped time space

Prior to timeframes, when using service_ids, the timeframe rules would effectively be 00:00:00 to 999:00:00 - ie for any time that is specified on that day (assuming the fare calculators weren't adjusting the time and date to fit the 0-24h range)

Other thoughts

- We haven't yet encountered a fare system where the fares are actually different either side of midnight. 
- This possibly becomes more complex when having an 'enter_station' rather than a 'trip_departure' timeframe_type.
- We should add empty value rules for start_time as 00:00:00 and end_time as (effectively) 999:00:00.

flocsy:
@halbertram my problem is not about the 2 sides of midnight. That intuitively makes sense. But there was something we said about the timeframes need to a) cover the whole day (or maybe not) and b) be distinct. So IMHO you wouldn't be allowed to have both of the following 2 rules:
a: from:00:00, to: 06:00
b: from:00:00, to: 07:00

The even bigger problem is when you have:
c: from 22:00, to: 30:30
when c can be in conflict with both a and b if they can be on the same ACTUAL day. i.e: if a is workday and c is weekend, then even though c means on weekend night and until 06:30 on the next day (that to make things more problematic can be either Sunday  - another weekend day, or Monday -  a work day) then c: to: 30:30 can mean Monday: 06:30, so it can be in conflict with a: to:06:00

My concern is that if we allow overlaps then who decides what to chose? Does the standard state something about it? Is it the cheapest fare? Is it up to the fare calculator implementation?

I think if we use to: 30:00 it means that the few buses during the night (after midnight) are considered "the last buses of yesterday" and in these places they ONLY appear in the schedules as such. Or in other words the 00:00-06:00 would never appear.

The other thing I feel like a bit of overkill and a place to future trouble is the 999:00:00. I don't think that makes sense. Even if you would write 168:00:00 (or rather: 144:00:00 - [144=168-24]) I would argue it makes no sense to allow it. While I can kind of imagine that in some place there are things like the 2nd, 3rd, nth day of some holiday, I don't feel like we should allow these extremes, when it's possible to do the same using the dates in the calendar.

halbertram:
@flocsy Our intention with this aspect was only to handle the cases where the schedules themselves indicate that a trip was running past midnight, rather than as an extra way of interpreting other early morning trips. 

It would not have an effect when considering a regular epoch time/date, but only when considering a day plus hh:mm:ss offset (and only when hh>=24), as in the case of a specific trip. Ie Fri 26:00 is a different time from Sat 02:00 and would match different rules.

Whether a service is considered as very late from yesterday or early for today is up to the schedule producer - we wouldn't be trying to make judgements on that.  

Based on that approach, the 999:00:00 still makes sense as that time offset would be higher than any stop_times.txt time will ever be. However, it probably makes sense instead to word the empty time as considered not part of the selection ie `evening,19:00:00,` would mean 7pm onwards. In terms of what we 'allow', it feels like any restrictions are related to stop_times.txt rather than fare timeframes.

bdferris-v2:
Assuming I'm understanding the comments from @flocsy and @halbertram, I do believe we'd want to be more explicit on timeframe matching semantics.  Namely, it seems like there is a question about whether timeframes matched based on (I'm going to make up some terms here) universal time vs service-date + time-of-day time.

Universal time would involve expanding a given timeframe (composed of service days + intervals) into a set of UTC time intervals, finding the UTC start time for the leg, and seeing if any of the intervals contain that start time.  If so, you've got a hit.

Service-date + time-of-day time would involve *not* expanding the service days + intervals.  Instead, a given leg matches if its service day (in the calendar.txt sense) matches the set of service days associated with a timeframe AND if the start time of the leg (again, using the stop_times.txt timing convention) is contained within one of the time intervals of the timeframe.

I think the service-date + time-of-day matching semantics would work around the issue that @flocsy brought up.  Namely, if two timeframes existed:

timeframe_id | start_time | end_time | service_id
--------------|-----------|-----------|-----------
weekday         | 6:00          | 30:00       | weekdays
saturday         | 0:00          | 20:00       | saturday

Under universal time match, a trip at 1am on Saturday morning might match both the `weekday` and `weekend` intervals. However, under a service-date + time-of-day interpretation, if that trip was part of the `weekday` service pattern (or an equivalent), it's start-time would be something like `25:00` and it would match the `weekday` timeframe, not the `weekend` time frame, based on the combination of service dates and time.

I think the service-date + time-of-day interpretation, while maybe more complex to describe, probably better matches the existing conventions in GTFS.  Thoughts?

I've mentioned start time of leg in the examples above.  Acknowledged it could also be "start time through fare gate".  That said, if we did go with a service-date + time-of-day matching interpretation, that maybe pushes me to lobby for `from|to_timeframe_id` to be included `fare_leg_rules.txt` as opposed to `fare_products.txt`.  I was admittedly on the fence before, but it really seems like we are talking about properties of the leg itself here.  More generally, I also wonder if by keeping timeframes in `fare_leg_rules.txt`, it keeps our options open for supporting more complex time-of-day fare calculations (e.g. some of the hypothetical examples from https://bit.ly/gtfs-fares-v2-finite-state-machines or having the option to use fare_leg_rules.txt override in combination with time-of-day).  Those are admittedly hypothetical, but it's in the back of my mind.

skinkie:
@bdferris-v2 I still wan to have the discussion on denormalising this. Why put it in a separate table, what is the actual gain here?

bdferris-v2:
@skinkie am I understanding you correctly that you are advocating for inlining the fields of `timeframes.txt` into `fare_leg_rules.txt` (or `fare_products.txt`, depending on your persuasion)?

skinkie:
> @skinkie am I understanding you correctly that you are advocating for inlining the fields of `timeframes.txt` into `fare_leg_rules.txt` (or `fare_products.txt`, depending on your persuasion)?

Correct. For three fields (start_time, end_time, service_id), I don't see the added benefit. If the argument it "it would then be analogue to frequenties.txt" the cardinality would be wrong.

bdferris-v2:
@skinkie I'll say that I like the proposal from @halbertram because it allows one to construct more complex timeframes spanning multiple time intervals and service dates and then just reference it once for a fare leg rule.  If we had to normalize that same information into `fare_leg_rules.txt` itself, it'd lead to a potential blow-up of rules, as the same multi-entry timeframe information would have to be duplicated across every fare leg rule that referenced them.

skinkie:
@bdferris-v2 where can I find the proposal?

bdferris-v2:
Apologies, it was actually from @davidlewis-ito in [this comment above](https://github.com/google/transit/pull/357#issuecomment-1452249231).  @halbertram had been discussing it as well.

Critically, it proposes moving `service_id` into `timeframes.txt` and it also allows multiple entries in `timeframes.txt` with the same `timeframe_id` that are combined into a union of time intervals and service dates.

skinkie:
> Apologies, it was actually from @davidlewis-ito in [this comment above](https://github.com/google/transit/pull/357#issuecomment-1452249231). @halbertram had been discussing it as well.

That functionality _certainly_ justifies it to place it a seperate table.

> Critically, it proposes moving `service_id` into `timeframes.txt` and it also allows multiple entries in `timeframes.txt` with the same `timeframe_id` that are combined into a union of time intervals and service dates.

But if this would be allowed (timeframe_id not being unique in a table by the same name), why not allow calendar.txt to have multiple (the same) service_id's? That should get an informed decision, since it does not follow how GTFS works now.

bdferris-v2:
> But if this would be allowed (timeframe_id not being unique in a table by the same name), why not allow calendar.txt to have multiple (the same) service_id's?

Is this mostly an argument about the naming of `timeframe_id`?  Maybe you'd prefer something like `timeframe_group_id`? 
 I'm also not sure that `service_id` is the best counter-argument, as it's not unique in `calendar_dates.txt` either.

skinkie:
> > But if this would be allowed (timeframe_id not being unique in a table by the same name), why not allow calendar.txt to have multiple (the same) service_id's?
> 
> Is this mostly an argument about the naming of `timeframe_id`? 

I assumed wrongly that it was unique in this pull request, but I notice that the example also has "regular" twice.

> Maybe you'd prefer something like `timeframe_group_id`?

Would be better.

gtfs/spec/en/reference.md
isabelle-dr: @skinkie "excluding the ones listed under" means "excluding the unique list of values defined in all `fare_leg_rules.from_timeframe_id` records.
This design looks like this causes problems, as flagged by @bdferris-v2 in this doc: https://bit.ly/gtfs-fares-v2-fare-leg-rules-scope-proposal.

bdferris-v2: I'll echo my concern from my [previous comment](https://github.com/google/transit/pull/357#issuecomment-1325447530) - I think this style of empty-value semantics (aka empty value => all values excluding explicitly mentioned values) is problematic and is just going to lead to trouble later on down the road.

I think for the timeframe proposal, I'd stick to "empty value matches everything" default semantics and rely on the fact that the ITO World timeframe proposal allows more flexibility in specifying complex timeframes when an explicit match is needed.

isabelle-dr: We've discussed changing this behavior to: an empty value in `from_timeframe_id` or `to_timeframe_id` means that the start or end time doesn't affect the price. 
I think it would make more sense, but this goes against the empty semantics of the `from_area_id` and `to_area_id` fields.

isabelle-dr:
Hello, if I understand correctly what has been shared, there seems to be a consensus around:
- having a `timeframes..txt` file, because we want to be able to group multiple time ranges and service IDs under one timeframe ID
- moving `service_id` to the `timeframes.txt` table to be able to create the complex time/date frames there instead of in the file where time factors are modeled.
- renaming `timeframe_id` to `timeframe_group_id` to make the design clear.
- having a `timeframe_type` field in it to add more information about what starts/stops the clock.

Outstanding items:
- Timeframe semantics: How to model times? Do we want the whole day covered with timeframes when one is defined? Can we have overlapping time frames?
- Do we want to model time factors in `fare_products.txt` or `fare_leg_rules.txt`? (or somewhere else). Having it in `fare_leg_rules.txt` means we could potentially use the [override](https://docs.google.com/document/d/1Q7mo2k4j1-uEmaLYIQVoNqdR2a_xMj9Kc-iA9mGwnXo/edit#heading=h.4bej6s69xndk) functionality.
- Where should `timeframe_type` be defined (one of the two tables, or something at the route level?), and what should be its enum values?
- Empty semantics for what empty values mean in the from/to timeframe fields: does it mean that the price isn't affected by it, or does it mean "everything except" like the from/to areas fields?

I am adding to the conversation:
- I would like to challenge the case where the time-based fare is a trip's attribute (when the vehicle starts) and not a rider's action (pass fare gate, enter vehicle). Is it in scope here? Would it be simpler to treat these as "trip-dependant" fares, and define them using `trip_ids` directly (especially if we eventually want to define trip-level fares)? Would it simplify our time discussion?
- I find that `from_timeframe_id` and `to_timeframe_id` are confusing and I would like to propose renaming them to `timeframe_start` and `timeframe_end` because it's referencing to the time a rider boards & offboards. In natural language, one doesn't travel from [8-10] to [8-10], they start and end their travel in a [8-10] time range.

To @lauramatson's point
> Does this add more interdependence by using calendar.txt and calendar_dates.txt

The type of interdependence we talked about solving right now is when the same GTFS file has schedule fields and fare fields, such as the field `network_id` in `routes.txt`. Since here we aren't adding fields to the calendar files, it doesn't add this type of interdependence. I acknowledge that this doesn't solve all data procurement challenges, but it's a start!

To @flocsy 's point

> if you go for end_time: 09:30:00 and not for 09:29:59, then let's make it even shorter. I'm pretty sure there are no timetables that use seconds (even in Japan), so either remove the seconds or make them optional.

All times in GTFS contain seconds and it would be unnecessarily complicated to add a new data type of time without seconds. I don't think it's a pain for data producers to produce data in HH:MM:SS, I would advocate not doing this.

Please let me know if I am missing anything.

gtfs/spec/en/reference.md
isabelle-dr: @bdferris-v2 "empty value matches everything" or "empty value means start or end time doesn't affect the price?"

bdferris-v2: Assuming `from_timeframe_id` and `to_timeframe_id` (whatever the end up being called) stay in `fare_leg_rules.txt`, to me it's about how they affect whether a particular row in `fare_leg_rules.txt` matches a particular leg.  In that world, I'd prefer that an empty value aka an unspecified timeframe would match any and all legs, modulo the other matching predicates for that row.  Whether that ends up affecting the price will be a function of the other fare leg rule matching predicates, associated fare products, etc.

Does that make sense?

flocsy:
@bdferris-v2 @halbertram I'm still not convinced. When the bus leaves at Saturday 02:23 then it's not trivial to decide which timeframes the fare calculator uses if there are multiple matches. Besides the problem we're discussing is just one type of the problems introduced if we allow overlapping timeframes. Don't we want to make them invalid? i.e:

from_time: 14:00, to_time: 16:00
from_time: 12:00, to_time: 18:00

or:

from_time: 14:00, to_time: 16:00
from_time: 15:50, to_time: 18:00

halbertram:
@flocsy I'm not sure that I see overlapping fares as a problem. I can see that having overlaps within one `timeframe_id` is a bit untidy, but it shouldn't cause problems as it will just resolve to that one id.

In the case of multiple `timeframe_id`s overlapping, I think that's likely to happen anyway within certain `timeframes.txt` files - for example when the metro and bus fare rules have different timings for 'peak'. Currently we Ito also creates an 'anytime' entry that overlaps everything. These will result in a set of matching timeframes that the calculator will then resolve as it incorporates them into other selection criteria - ie the bus peak timeframe is only used for the bus `network_id` etc.

Finally, in our understanding of time (service-date + time-of-day as @bdferris-v2 mentioned above), a service at Sat 02:23 would only match the morning Sat timeframe - it would not match the after-midnight Fri timeframe because the schedules indicate that it is a Sat service.

halbertram:
@skinkie Our thinking behind `timeframe_id` was that the id already refers to technically multiple timeframes - ie it is more than one specific period of time (cf dictionary meaning). So given that, our thinking was that it doesn't make much difference when further lines add more multiples.

Eg.
|timeframe_id|start_time|end_time|service_id|
| --- | --- | --- | --- |
|peak|08:00:00|11:00:00|mon_to_sat|

didn't seem to be a different concept when written as

|timeframe_id|start_time|end_time|service_id|
| --- | --- | --- | --- |
|peak|08:00:00|11:00:00|weekdays|
|peak|08:00:00|11:00:00|saturdays|

But it's not something I feel particularly strongly about if other see if differently.

Additionally, it feels like the unique requirement on `calendar.service_id` is a bit of an anomaly. Given `calendar_dates.txt` exists, it doesn't make sense to me that a single id is completely capable of representing the months of Jan and Mar (but not Feb) - just not by the most convenient method.

skinkie:
@halbertram the unique requirement was on the topic of timeframe_id (as being the primary key) which it is not, but given the name does suggest so.

bdferris-v2:
I'll echo @halbertram 's comments on overlapping timeframes.  I can imagine situations where overlapping timeframes may be present in a feed as different networks in the same feed need slightly different (but potentially overlapping) timeframes, especially in the case of merged/aggregated feeds.

bdferris-v2:
From @isabelle-dr:

> I would like to challenge the case where the time-based fare is a trip's attribute (when the vehicle starts) and not a rider's action (pass fare gate, enter vehicle). Is it in scope here? Would it be simpler to treat these as "trip-dependant" fares, and define them using trip_ids directly (especially if we eventually want to define trip-level fares)? Would it simplify our time discussion?

I would maybe like to challenge that challenge :) I'll first say that I don't know how common "time-based fare determined by trip start time" is as a fare model and there is a world where we punt the decision on explicitly supporting this via `timeframe_type` or some other mechanism to some future discussion.  That said, I know we have a stated goal of making it potentially easier to publish fare updates separately from schedule data.  I think there is general value in being able to specify a timeframe as a general property of fares, even if it applies to trip start times, because it means you don't have to update your fare rules every time you update your trip times.  That could end up being important for systems where schedule publishing and fare publication end up being separate processes with different timeframes (in the general sense, not the spec sense).

All that said, not sure who is banging down the door for this feature at the moment :)  I think it could be safely tackled later without painting ourselves into a corner.

davidlewis-ito:
Our thoughts on trip time-based fare discrimination:

Use cases: We see this typically on regional or national rail systems, where a specific train is flagged as peak if it arrives into its major city destination in the morning commute period, or leaves the same in the afternoon commute period.  This “peakness” applies to all riders regardless of where they join or leave the train, regardless of how their personal journey timings relate to the trains origin or destination timing.  Examples include LIRR in New York and many national rail services in the UK.  

Modelling: Although we initially considered that this could be through timeframes.txt, closer inspection of the LIRR example revealed that there are a significant number of exceptions to the timing rules, such that it becomes necessary to discriminate by individual trips in order to be completely accurate.  We also felt that a more general ability to allocate trips to some trip class might be a more generally useful mechanism.  Consider special event trains, trains with only a certain level of service (sleepers).  Although perhaps a way off for GTFS, we felt that this might inform the way to go.  Whether this be an attribute on trips.txt or another “trip to class” table separated out into the fares domain is something for later consideration perhaps.  So with this in mind, I would tend to agree that we can exclude consideration from the current work on timeframes.txt.

Generalisation of timeframes.txt: Apart from distinguishing “peak” fares we can see other uses for this modelling.  Fare capping and period passes jump to mind.  Take for example the case of Dallas DART where morning,  midday and afternoon period  passes seem to be the primary ticket available.

flocsy:
@halbertram I don't see your example (service_id: mon_to_sat vs {weekdays, saturdays}) as conflicting or overlapping. For each trip you fall into exactly one of them.

@bdferris-v2 It's hard to see without an example but from the sentence you wrote it seems you are talking about different network_id-s, so I don't think that would cause an overlapping in the sense I was referring to.

My problem with overlapping is something that can be seen in the simplest example, when there's a bus at 04:13 and there are 2 "timeframes" defined as such that all their parameters match, except the start/end of the timeframe. The more obvious example would be:

start_time: 00:00, end_time: 06:00
start_time: 04:00, end_time: 05:00

but also the less obvious:
start_time: 00:00, end_time: 06:00
start_time: 22:00, end_time: 30:00

Maybe I'm missing something here, but I think it might cause "race condition", where depending on the choice you might end up with different fare amount at the end of the calculation.

Regarding the empty field's semantics: if I understand correctly what you wrote then I agree with you. I didn't like what we did in previous FaresV2 tables. It's confusing and against the instinct. I also would prefer if empty field would mean (as in GTFS v1 BTW) that a specific line matches anything (in regards to that field)

halbertram:
@flocsy Our understanding of fare calculations is that when there are multiple options that match, that set of options flow through the calculations together to produce a set of resulting fares. That set of fares then may need to be consolidated before presenting to the user based upon the use case - typically by filtering for the cheapest option.

For example, if there's a cheaper ticket for a limited time period, it doesn't necessarily invalidate the option of an all-day ticket that could have more transfer options.

Additionally, this isn't the only place in the spec where we can end up with multiple options for the same journey. Once you have complex area-to-stop mapping, things like

| from_area_id | to_area_id |
| --- | --- |
|anywhere|airport|
|central_station | airport |

will create two options for getting to the airport from the central station, which will be resolved when presenting to the user.

Finally, in our proposed understanding of “time” (service-date + time-of-day), the second example you gave would not be regarded as overlapping.

bdferris-v2:
+1 to the interpretation of @halbertram.  Considering @flocsy 's examples: for the first example (0-6,4-5), if those were assigned to the same `timeframe_id` and same `service_id`, I'd agree that doesn't seem right and we could probably mark that as a warning or error.  However, for the second example (0-6, 22-30), I don't think those are overlapping given service-date + time-of-day semantics.

flocsy:
@bdferris-v2 

service_id:weekday, start_time: 00:00, end_time: 06:00
service_id:weekday, start_time: 22:00, end_time: 30:00

I'm not sure what you mean on "time-of-day semantics". If I ask you about 2023-03-22 (Wed) 04:00, then it will match both of the above lines. So how you chose between them? Do we allow contradicting data?

gtfs/spec/en/reference.md
isabelle-dr: I think so, but just to confirm: you're suggesting that we keep the initial design, correct? Which is:

field  | Description
--  | --
`from_timeframe_id`  | Defines a departure timeframe for the fare leg rule. If there are no matching `fare_leg_rules.from_timeframe_id` values to the timeframe_id being filtered, empty `fare_leg_rules.from_timeframe_id` will be matched by default. </br></br>**An empty entry in `fare_leg_rules.from_timeframe_id` corresponds to all timeframes defined in timeframes.timeframe_id excluding the ones listed under `fare_leg_rules.from_timeframe_id`**.
`to_timeframe_id` | Defines an arrival timeframe for the fare leg rule. If there are no matching `fare_leg_rules.to_timeframe_id` values to the `timeframe_id` being filtered, empty `fare_leg_rules.to_timeframe_id` will be matched by default. </br></br>**An empty entry in `fare_leg_rules.to_timeframe_id` corresponds to all timeframes defined in `timeframes.timeframe_id` excluding the ones listed under `fare_leg_rules.to_timeframe_id`**.

isabelle-dr: And if this is the case, how to represent that the time that matters for the peak fare is only when you start the leg, not when you end the leg?

isabelle-dr: Should we differentiate these two cases?
- case 1: both `from_timeframe_id` and `to_timeframe_id` are empty -> interpreted as "corresponds to all timeframes [...] excluding [...]".
- case 2: one of them is specified and not the other -> interpreted as "the start/end time of the leg does not affect the fare"

isabelle-dr:
Hello, 

Based on the feedback received (thank you!), we've done the following changes:
- renamed `from_timeframe_id` and `to_timeframe_id` to `start_timeframe_id` and `end_timeframe_id`.
- changed the empty semantics so that empty timeframe values in `fare_leg_rules` indicates that the time doesn't affect the fare (although I am having second thoughts after going back to @bdferris-v2's comment [here](https://github.com/google/transit/pull/357#discussion_r1132925018)).
- added two constraints:
  - For each `service_id`, time intervals must be defined for the entire day (from 00:00:00 to 23:59:50). This would generate an ERROR in the canonical validator.
  - There should not be overlapping time intervals for the same `timeframe_group_id` and `service_id` values. This would generate a WARNING in the canonical validator.

We will be discussing the outstanding items in this proposal on Tuesday at 11AM EST. Please [read the agenda](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit#bookmark=kix.e3izyozhnv65) beforehand and add any items you'd like to discuss!
[Link to the event if you need](https://www.eventbrite.ca/e/specifications-discussions-gtfs-fares-v2-monthly-meetings-tickets-522966225057)

flocsy:
in the agenda there's this example:

Peak|00:00:00|09:30:00|mon_fri
OffPeak|19:00:00|30:00:00|mon_fri
OffPeak|00:00:00|30:00:00|weekends

So if something is on Tue-Sat between 00:00:00 and 06:00:00 (or i other words on Mon-Fri between 24:00:00 and 30:00:00) then it's both Peak and OffPeak? I think this contradicts the added constraint of "should not be overlapping time intervals for the same timeframe_group_id and service_id values" *)

*) It doesn't contradict it as it is written. because the timeframe_group_id is different, but IMHO this is either a typo or a logical mistake. Shouldn't the sentence be "should not be overlapping time intervals for the same service_id values" ?

Even then we would still have the conflict on Monday between 00:00:00 and 06:00:00 because it's also Peak (1st line) and OffPeak (last line: it falls to 24:00:00-30:00:00 of weekends)

I think this should not pass the validator and the probable fix would be to switch the 1st line to:

OffPeak|00:00:00|06:00:00|mon_fri
Peak|06:00:00|09:30:00|mon_fri
(which BTW would also make the 24-30 not needed in the last 2 lines)

flocsy:
Regarding the outstanding items: The example with the Peak_rail and OffPeak_bus. I don't think this is a problem, but now I understand why all my previous examples might not make sense to some people :) 

So I know we're talking about string ids without meaning, but this example hints it, so it was easier to spot: it sounds like a valid case when the rail and bus have different peak hours, I don't see any reason why forbid this. However I have a feeling that something else is missing. Can both be in the same service: mon_fri? Or should it actually be mon_fri_rail and mon_fri_bus? I don't know.

Maybe both the "overlapping" Peak, OffPeak in the previous comment and Peak_rail, OffPeak_bus should be allowed, BUT it shouldn't be allowed to use them in fare_leg_rules in such a way that would create a conflict. i.e as long as Pek_rail is only used for rail, and OffPeak_bus is only used for bus they are distinct, and don't interfere with each other. And similar in the previous comment in theory (though it's much harder to imagine that it would really work) if Peak+mon_fri is only used for trams and OffPeak+mon_fri is only used for busses, then it can work.

gtfs/spec/en/reference.md
bdferris-v2: I wasn't arguing for keeping the original design.  I was arguing for where we ended up (as of March 24).  Namely:

> An empty `start|end_timeframe_group_id` indicates that the time of the leg does not affect the fare.

As such, I'd argue that we can resolve this conversation.

bdferris-v2:
@flocy I think some of the back and forth here because we haven't been super crisp about what it means for a timeframe to match.  I'd like to propose the following language for `start_timeframe_group_id`:

> For a fare leg rule that specifies a `start_timeframe_group_id`, that rule will match a particular leg if there exists a row in `timeframes.txt` where all of the following conditions are true:
> * The row’s `timeframe_group_id` is equal to the `start_timeframe_group_id` value.
> * The set of service days identified by the row’s `service_id`contains the service day of the leg’s departure trip.  
> * The departure time of the leg, relative to the service day per GTFS Time field type conventions, is greater than or equal to the rows `start_time` value and less than the `end_time` value.

(You can imagine the equivalent text for `end_timeframe_group_id`).

This is what I was trying to get at with my discussion about about service day + time of day.  With these semantics in mind, back to your example:

```
Peak,    00:00:00, 09:30:00, mon_fri
OffPeak, 19:00:00, 30:00:00, mon_fri
OffPeak, 00:00:00, 30:00:00, weekends
```

> So if something is on Tue-Sat between 00:00:00 and 06:00:00 (or i other words on Mon-Fri between 24:00:00 and 30:00:00) then it's both Peak and OffPeak?

Per the semantics above, it depends on the service day of the vehicle.  Consider two trips, both of which operate Monday-Friday.  One departs at 02:00:00 relative to the service date and one departs at 26:00:00.  From a user perspective, you might say the both depart at 2am, but for the purposes of timeframe matching, the first matches Peak and the second matches Offpeak because of their departure times relative to their service day.

Does that make sense?

bdferris-v2:
This is the part where I acknowledge that the semantics I described get weird when timeframes aren't about vehicle departure and instead about time tapping at a fare gate, at which point I concede you argument about overlapping timeframes becomes more relevant.  But I think there is maybe an argument where we can make service day + time semantics work for fare gate taps, where using UTC time for trip departure times is more limiting.

bdferris-v2:
Regarding the proposal that for each service_id, time intervals must be defined for the entire day (from 00:00:00 to 23:59:50).

I understand where this is coming from, but I wanted to call out there there is a world where we adopt the `override` extension and folks will ask if they can just specify their peak times with an override without having to specify their non-peak times as well.  But I suppose that the constraint could be relaxed in the future if that happens?

halbertram:
@bdferris-v2 Re tap semantics: agree - we were thinking that, even when the type isn't a vehicle departure, the fare calculator will normally have an initial trip in the leg to consider, and so the service day + time semantics would still work for the tap and trip departure cases.

halbertram:
@bdferris-v2 and @isabelle-dr On complete day coverage, I feel cautious about this being an error (or even a warning). It seems to make more sense in the context of peak/offpeak timeframes, but for cases where eg there is just a single discounted period in the day, requiring to also have a non-discounted placeholder timeframe doesn't seem necessary.

Additionally there's also the issue that the requirement doesn't fully make sense technically in the case when there are multiple unrelated aspects with timeframes using the same service_id, eg two networks. If the bus timeframes have full coverage, that will mask any errors for the rail timeframes if they don't have full coverage. We don't have a mechanism to connect related timeframes and I don't think we're ready for `timeframe_group_group_id`...

flocsy:
> Per the semantics above, it depends on the service day of the vehicle. Consider two trips, both of which operate Monday-Friday. One departs at 02:00:00 relative to the service date and one departs at 26:00:00. From a user perspective, you might say the both depart at 2am, but for the purposes of timeframe matching, the first matches Peak and the second matches Offpeak because of their departure times relative to their service day.
> 

Maybe I am too strict :) Do we really want to allow the start time to be 00:00:00-999:00:00? At best it sounds like a hack, but more probably just a way to enable unintentional behaviour to occur. Though it looks strange to me to call a bus Sunday evening bus that departs at 02 on Monday, but this is something that I can still make sense of. But if we allow it to go on like this (even above 168 hours) then we'll really have overlaps that noone will be able to see. Shouldn't we still restrict the times to something like 00:00:00-35:59:59? Maybe up to 47:59:59, but that seems to be an overkill.

skinkie:
> Maybe I am too strict :) Do we really want to allow the start time to be 00:00:00-999:00:00? At best it sounds like a hack, but more probably just a way to enable unintentional behaviour to occur. Though it looks strange to me to call a bus Sunday evening bus that departs at 02 on Monday, but this is something that I can still make sense of. But if we allow it to go on like this (even above 168 hours) then we'll really have overlaps that noone will be able to see. Shouldn't we still restrict the times to something like 00:00:00-35:59:59? Maybe up to 47:59:59, but that seems to be an overkill.

I think there are trains that would actually do the three days plus operation. I know because we explicitly choose _not_ to support these in our journey planner as a continious trip. The question for this fare is obviously if you buy a ticket for a _date_ or for a _trip_ operating on an _operational date_.

gtfs/spec/en/reference.md
```diff
@@ -391,14 +408,18 @@ To process the cost of a leg:
...
+    - `fare_leg_rules.to_area_id`
+    - `fare_leg_rules.start_timeframe_group_id`
+    - `fare_leg_rules.end_timeframe_group_id`<br/>    
 <br/>
 
 2. If the leg exactly matches a record in `fare_leg_rules.txt` based on the characteristics of travel, that record must be processed to determine the cost of the leg.
-3. If no exact matches are found, then empty entries in `fare_leg_rules.network_id`, `fare_leg_rules.from_area_id`, and `fare_leg_rules.to_area_id` must be checked to process the cost of the leg:
+<br/>
+
+3. If no exact matches are found, then empty entries in `fare_leg_rules.network_id`, `fare_leg_rules.from_area_id`, `fare_leg_rules.to_area_id`, `fare_leg_rules.start_timeframe_group_id`, and `fare_leg_rules.end_timeframe_group_id` must be checked to process the cost of the leg:
```
bdferris-v2, line 419: Do `start_timeframe_group_id` and `end_timeframe_group_id` need to be checked here?

gtfs/spec/en/reference.md
```diff
@@ -411,6 +432,9 @@ To process the cost of a leg:
...
+|  `start_timeframe_group_id` | Foreign ID referencing `timeframes.timeframe_group_id` | Optional |  Defines the timeframe for the departure of the fare leg.<br><br>For a fare leg rule that specifies a `start_timeframe_group_id`, that rule will match a particular leg if there exists at least a row in `timeframes.txt` where all of the following conditions are true:<br>- The row’s `timeframe_group_id` is equal to the `start_timeframe_group_id` value.<br>- The set of service days identified by the row’s `service_id` contains the service day of the leg’s departure trip.<br>- The departure time of the leg, relative to the service day per [GTFS Time field type](#field-types) conventions, is greater than or equal to the rows `start_time` value and less than the `end_time` value.<br><br>If there are no matching `fare_leg_rules.start_timeframe_group_id` values to the `timeframe_group_id` being filtered, empty `fare_leg_rules.start_timeframe_group_id` will be matched by default. <br><br>An empty `fare_leg_rules.start_timeframe_group_id` indicates that the start time of the leg does not affect the fare. |
```
bdferris-v2, line 435: If there are no matching `fare_leg_rules.start_timeframe_group_id` values to the
    `timeframe_group_id` being filtered...

I'm not sure I understand what this line means.

bdferris-v2, line 435: if there exists at least a row in `timeframes.txt`

I noticed you added "at least" here.  Did you mean to say "if there exists at least one row"?  Or something different?

gtfs/spec/en/reference.md
```diff
@@ -436,6 +436,7 @@ To process the cost of a leg:
...
 | `timeframe_type` | Enum | **Conditionally Required** | Defines the type of fare validation to be associated with `timeframes.start_time` and `timeframes.end_time`.<br><br>Valid options are:<br>`0` - In vehicle.<br>`1` - At fare gate or on platform.<br><br>Conditionally Required:<br>- **Required**  if either `start_timeframe_group_id` or `end_timeframe_group_id` is defined.<br>- **Forbidden** otherwise.|
 | `fare_product_id` | Foreign ID referencing `fare_products.fare_product_id` | **Required** | The fare product required to travel the leg. |
+| `override` | Non-negative integer | Optional | When multiple entries in `fare_leg_rules.txt` match, the one with the highest `override` value will be selected.<br><br>An empty value for `override` is treated as zero.|
```
halbertram, line 439: I think this should be "the **ones** with the highest" - there's no guarantee that there will be only a single highest value. It might need extra words to make that clear.

halbertram:
[Apologies if this was discussed in detail at the last call that I wasn't able to attend and/or I've misunderstood the current wording]

I feel uneasy about the 'new' empty concept involving "if there are no matching, then empty values will match by default". It feels like this still involves implicit interactions between rules that may not be intended (especially when aggregating), which is certainly a problem with the original empty concept.

For example, for a system with a flat fare ($3 peak, $2 offpeak), it would be good if the third row would add a simple option of going from the central station to the airport at any time for $1. 

| from_area_id | to_area_id | start_timeframe_group_id | fare_product_id |
| --- | --- | --- | --- |
| anywhere | anywhere | offpeak | two_dollars |
| anywhere | anywhere | peak | three_dollars |
| central_station | airport | | one_dollar |

However, if I'm parsing the proposed wording correctly, the fact that either the first or second line will match the start_timeframe_group_id will stop the third row matching as its empty start_timeframe_group_id will not be considered. It behaves as if there is an implicit override aspect attached to the more specified rules, whereas I think the explicit override mechanism would be a preferable route to achieve this if required.  

In this way the empty then behaves differently from an 'anytime' timeframe, which feels odd.

Finally, I think the wording would get increasingly complex when there are multiple predicates supporting the 'if none matching'.

I think it would be much simpler (to understand and calculate) if the 'new' empty behaviour could be described in the form
```
... that rule will match a particular leg if
**either** start_timeframe_group_id is empty,
**or** there exists at least a row in `timeframes.txt` where all ...
```

bdferris-v2:
+1 to @halbertram's concerns (see my individual comments on the spec).

flocsy:
@isabelle-dr  Why is the new field called "override"? Isn't "priority" a better name?

I also like the change @halbertram recommends about the meaning of empty field.

gtfs/spec/en/reference.md
isabelle-dr: I took a stab at improving the description of override in [this commit](https://github.com/google/transit/pull/357/commits/d4afc6ede5eae96567b373a9400d6f0c2e907442). Let me know if this is easier to interpret.

gtfs/spec/en/reference.md
isabelle-dr: You're right, I added "at least", because we have had more than one row. 

I changed 

> if there exists at least a row in `timeframes.txt`

to:

> if there exists at least one row in `timeframes.txt`

Is this better?

isabelle-dr: That's a good point, I believe this is part of the legacy "empty" design. I removed this sentence and added what Hal recommended in [this comment](https://github.com/google/transit/pull/357#issuecomment-1498060622) in https://github.com/google/transit/pull/357/commits/91bda277d61e9dfab64d9def48d73ac2e71e0a74

isabelle-dr:
@flocsy the override field is initially an idea from ITO, you can read more about it [here](https://docs.google.com/document/d/1Q7mo2k4j1-uEmaLYIQVoNqdR2a_xMj9Kc-iA9mGwnXo/edit?usp=sharing).
We think "override" implies one field takes precedence over the other matches, making the other matches not considered, and "priority" would imply all matches are valid but some are preferred over others. That being said, we don't have a strong opinion on the name and would defer to the community. 

@halbertram and @bdferris-v2 thank you for pointing out the inconsistencies in the "empty" meaning. 

The last commits done to this PR change the following:
- Changed the override field description from:

> When multiple entries in `fare_leg_rules.txt` match, the one with the highest `override` value will be selected.


to:

> Allows certain rules to take precedence over others. When multiple entries in `fare_leg_rules.txt` match, the rule or set of rules with the highest value for `override` will be selected.

- Streamlined the new meaning of "empty" for `start_timeframe_group_id` and `end_timeframe_group_id`
  - Removed the mention of them in the "_If no exact matches are found, then empty entries in [...] must be checked to process the cost of a leg_" in the `fare_leg_rules.txt` semantics description.
  - Removed "_If there are no matching `fare_leg_rules.end_timeframe_group_id` values to the `timeframe_group_id` being filtered, empty `fare_leg_rules.end_timeframe_group_id` will be matched by default_" in the field descriptions.
  - Added Hal's suggestion [here](https://github.com/google/transit/pull/357#issuecomment-1498060622).

Please let me know if there is anything outstanding.

flocsy:
I'm not a native English speaker, but to me it sounds like override would fit better if this was a binary field, but since even in the proposal it talks about the highest override value, it is essentially describing what priority means IMHO.

halbertram:
@flocsy That makes sense. I think there's some value in including something to make it clear that this is a rule-overriding mechanism and not another predicate. Maybe 'override_priority' or 'rule_priority'? But also happy to go with 'priority' if that's the consensus.

flocsy:
rule_priority also sounds good

tzujenchanmbd:
Hello everyone!

In the past weeks there have been discussions about how to handle datasets that have stop time values above 24:00:00 and the timeframe matching semantics, as well as the semantics of empty values for start_time/end_time. As mentioned in [this comment](https://github.com/google/transit/pull/357#issuecomment-1460870354) from Brain, there are two options for timeframe matching semantics.

After exploring several use cases, MobilityData suggests timeframe matching based on “universal time”(real-world date and time matching) for the following reasons:

- It's simpler to understand; the times in timeframes match with the machine’s times, regardless of the trip (a “fare day” is from 3AM on the current day to 2:59AM the next day). Therefore it will be more intuitive to produce data.

- When the fare data is produced by a different organization than schedule data, the producer can create data based on the fare policy and doesn’t need to consider 24+ and max values in trips.

- Empty values in timeframes can be introduced with low-risk. If I have a Saturday fare, I don’t have to enter time values; empty values mean the fare is all day.

For more details, please refer to this [proposal](https://docs.google.com/document/d/1N3WpgAh2OxPuE5Vvjbd6mjvKK42M7mSVdxscVvzP8SU/edit#). Please note that in this proposal, we have presented both options, and MobilityData recommends **option 1** as the preferred choice.

In sum, even though this option may impose some time conversion burden on consumers, based on the spirit of GTFS of making it easier for producers to provide data, we recommend this option.

In addition, we will be discussing the outstanding items in the proposal on Tuesday at 11AM EST. Please read the [agenda](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit#) beforehand and add any items you'd like to discuss!
([Link](https://us06web.zoom.us/j/88524937228?pwd=MjhiYU9pYVZYcXdPS2x6Z2hFbXZXZz09) to the event if you need)

tzujenchanmbd:
Hello,

Based on the consensus in the working group meeting (thanks all!), we've made the following changes:
- Added **Timeframe Local Time Semantics** section after `timeframe.txt` fields table.
- Renamed `start_timeframe_group_id` and `end_timeframe_group_id` back to `from_timeframe_group_id` and `to_timeframe_group_id`.
- Updated `from_timeframe_group_id` and `to_timeframe_group_id` to align with Timeframe Local Time Semantics.
- Updated empty semantics for `timeframes.start_time` and `timeframes.end_time`.
- Removed `rule_priority` and `timeframe_type` fields from this PR.

The [Fares v2](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.98c4botj62iu) proposal (Google doc) has been updated as well. We kept both `rule_priority` (renamed from "override") and `timeframe_type`  as experimental fields in this proposal.

Please let us know if we are missing anything.

npaun:
A couple quick observations:

* I think it would be a good idea to add an explicit statement of what happens when multiple timeframes match for a leg. Presumably, all of the `fare_leg_rules` apply, and clients select the 'best' overall result, just like for stops that can be in multiple areas.
* This PR makes it clear that timeframe start and end times are in `stop.stop_timezone` but PR #378 makes clear that arrival time and departure time are in `agency.agency_timezone`. I feel this will be a source of confusion.

bdferris-v2:
Question for @npaun:

> what happens when multiple timeframes match for a leg

When you say "multiple timeframes", do you mean multiple entries in `fare_leg_rules.txt` with different `from|to_timeframe_group_id` values?  Or do you mean multiple matching rows from `timeframes.txt` for a single entry in `fare_leg_rules.txt`?

Regarding timezones, let me first acknowledge that I'm not aware of many systems (any?) that operate cross timezone feeds AND have time-of-day based fares (if you know of any, I'd love to hear about them).  But if such a system did exist, what would be the reasonable default?  The stop's timezone or the agency timezone?  My gut reaction is the stop timezone, as time-of-day fares are likely to be communicated in local time, not relative to the base timezone of an operating agency.  Agreed that this is complicated to communicate in the spec.  And to be fair, the spec defaults to agency_timezone if stop_timezones are not specified, so for most single-timezone feeds, it will sort of be an afterthought.

But all that said, I agree that it is complex and I could see arguments for both sides.

npaun:
@bdferris-v2 

> When you say "multiple timeframes", do you mean multiple entries in fare_leg_rules.txt with different from|to_timeframe_group_id values? Or do you mean multiple matching rows from timeframes.txt for a single entry in fare_leg_rules.txt?

I was thinking of the former case (multiple entries in fare_leg_rules.txt).

However, if I'm reading the spec right I could create two different service IDs (e.g. monday1 and monday2) for the same dates and create overlapping timeframes within the same `timeframe_group_id`....

> stop timezone vs agency timezone

I agree that this will rarely come up in practice; [NICTD/South Shore Line](https://mysouthshoreline.com) is probably one of the only "regional" agencies I can think of that crosses a timezone. Although I could go either way on stop timezone or agency timezone as the "most intuitive"  model, I'd prefer using agency timezone based on a different principle: reducing the number of possible "ways to handle something" in the spec. If you're a consumer and you're already processing the rider's trip, you're considering the departure or arrival time in the agency's timezone, and so it would be simpler (or at least less likely to give rise to mistakes due to conversion) to use this value in fare calculations too.

bdferris-v2:
So regarding multiple entries matching in `fare_leg_rules.txt`, I think you'd agree that this is already an issue with `fare_leg_rules.txt` (you mentioned stops with multiple area assignments allowing for multiple matches) that isn't particularly well specified in the spec?  I'm not sure that time variable fares makes this any better or worse.  I will concede that it's definitely possible to construct timeframes such that multiple legs could match with separate `timeframe_group_ids` that have overlapping times.  But I'm not sure if that should be explicitly disallowed, or even how to reliably detect it, given interaction with other fields in `fare_leg_rules.txt`.

That said, I'm curious how you are currently handling this (multiple matching rules in `fare_leg_rules.txt` and if there is consistency amongst consumers on this.  The general policy I've seen is "only present the cheapest fare if multiple match" but I think that strategy gets a bit muddy with multiple fare products and fare media.

npaun:
Yeah I agree, it's the same issue for areas; I'd settle for a clear notice in the text that multiple timeframes could match and consumers need to consider them all. 

Our strategy really is "cheapest overall acceptable match" and it's indeed hairy.

bdferris-v2:
@npaun would it be reasonable to add a more general notice to the "To process the cost of a leg:" section of `fare_leg_rules.txt`?  Something like:

> It is possible for a leg to match multiple rules.  All such matching rules must be considered, potentially producing multiple distinct fares for a journey.

I'd probably insert this between items 3 and 4 in the current list.

tzujenchanmbd:
@npaun @bdferris-v2 thanks for your thorough discussion and suggestion! @npaun I can add the [general notice](https://github.com/google/transit/pull/357#issuecomment-1583485943) to `fare_leg_rules.txt` if this looks good to you.

Regarding timezones, the fares v2 working group had an in-depth discussion, recognizing the complexity to communicate in the spec and burdens of time conversion for consumers. However, it was also acknowledged that compared to agency timezone, the local timezone is the most intuitive for producers and offers other advantages mentioned in [this comment](https://github.com/google/transit/pull/357#issuecomment-1553688728).
For more detailed discussion, please refer to the [working group meeting minutes](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit#heading=h.eu5q7xisejed).

flocsy:
Some of you discussed whether it's less confusing for consumers if the times are agency or stop timezone. I think there's another side of this question: on 363 days of the year probably either would work. But on the 2 days when there's the daylight saving time change it might be "crucial" to choose the "better" one and also because it only happens twice a year it's harder to test these edge cases. I'm not sure how much this happens in real world, but in theory there are many things could go wrong during the DST change night(mare): the 2 neighbouring countries might change DST on a different date (which would make 4 though days per year) but even if they are on the same day there is probably a 1 hour gap between when it occurs in the 2 countries. So I'm not sure which timezone is better to be used, but when discussing this we should consider the edge cases and the possibility to make implementation bugs.

isabelle-dr:
Hello everyone, the working group had a final discussion on this proposal and reached consensus ([see meeting notes](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit)).

- We have at least one producer: [ITO World](https://www.itoworld.com/). They have shared their dataset in [this file](https://docs.google.com/spreadsheets/d/1-cD-R2OH5xAQAbNWNlrXD7WOw594lVdW-bomuLo6bI8/edit#gid=1401059568), which is part of a private feed. 
- We have at least one consumer: [Apple Maps](https://www.apple.com/maps/). See the screenshots below for the same dataset.

As per the [GTFS Amendment process](https://gtfs.org/schedule/process/#specification-amendment-process), the requirements to open a vote are met. 
We are opening a vote for adding variable fares by time or day, which is part of the [GTFS-fares v2](https://docs.google.com/document/d/19j-f-wZ5C_kYXmkLBye1g42U-kvfSVgYLkkG5oyBauY/edit#heading=h.yunc1axz1fcj) proposal.

Voting ends on 2023-07-24 at 23:59:59 UTC.

_weekday peak_
PastedGraphic-3

_weekday off-peak_
PastedGraphic-1

_weekend_
PastedGraphic-2

gtfs/spec/en/reference.md
```diff
@@ -345,6 +347,26 @@ For examples that demonstrate how to specify a fare structure with [fare_rules.t
...
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+|  `timeframe_group_id` | ID | **Required** | Identifies a timeframe or set of timeframes. |
+|  `start_time` | Time | **Conditionally Required** |  Defines the beginning of a timeframe. The interval includes the start time.<br> Values greater than `24:00:00` are forbidden. An empty value in `start_time` is considered `00:00:00`. <br><br> Conditionally Required:<br> - **Required** if `timeframes.end_time` is defined.<br> - Forbidden otherwise |
+|  `end_time` | Time | **Conditionally Required** |  Defines the end of a timeframe. The interval does not include the end time.<br> Values greater than `24:00:00` are forbidden. An empty value in `end_time` is considered `24:00:00`. <br><br> Conditionally Required:<br> - **Required** if `timeframes.start_time` is defined.<br> - Forbidden otherwise |
+| `service_id` | Foreign ID referencing `calendar.service_id` or `calendar_dates.service_id` | **Required** | Identifies a set of dates that a timeframe is in effect. <br>There must not be overlapping time intervals for the same `timeframe_group_id` and `service_id` values. |
```
flocsy, line 363: Probably the sentence "There must not be overlapping time intervals for the same `timeframe_group_id` and `service_id` values." could be moved out from the description of service_id to a "common" place. Maybe just before the table? Not sure.

gtfs/spec/en/reference.md
```diff
@@ -411,6 +437,8 @@ To process the cost of a leg:
...
...the time at which the event is scheduled to occur.  For example, the time could be the scheduled departure time of a bus at the start of a fare leg where the rider boards and validates their fare. For the rule matching semantics below, the start time is computed in local time, as determined by [Local Time Semantics](#localtimesemantics) of [timeframes.txt](#timeframestxt).  The stop or station of the fare leg’s departure event should be used for timezone resolution, where appropriate.<br><br>For a fare leg rule that specifies a `from_timeframe_group_id`, that rule will match a particular leg if either `from_timeframe_group_id` is empty, or if there exists at least one row in `timeframes.txt` where all of the following conditions are true<br>- The row’s `timeframe_group_id` is equal to the `from_timeframe_group_id` value.<br>- The set of days identified by the row’s `service_id` contains the “current day” of the fare leg’s start time.<br>- The fare leg’s “time-of-day” is greater than or equal to the row’s `start_time` value and less than the `end_time` value.<br><br>An empty `fare_leg_rules.from_timeframe_group_id` indicates that the start time of the leg does not affect the fare. |
```
flocsy, line 440: This is a long, complex description. My comment applies both from_timeframe_group_id and to_timeframe_group_id: This part of the sentence is unnecessary (and even confusing): "if either `from_timeframe_group_id` is empty, or". The reason it is not needed is that before that we say: "For a fare leg rule that specifies a `from_timeframe_group_id`". My interpretation is that is it's specified then it's not empty.

jsteelz, line 440: > For example, the time could be the scheduled departure time of a bus at the start of a fare leg where the rider boards and validates their fare.

'Could be' makes little sense to me - are we just saying that data producers can invent what criteria apply to "start time" of a fare leg? This should be defined exhaustively, and I can't think of any cases other than the one mentioned here.

tzujenchanmbd: Removed the unnecessary part. Thanks!

skinkie:
+1 OpenGeo (not a producer)

saraemarcus:
+1 Apple

halbertram:
+1 Ito World

Would it be worth changing the empty wordings from
```An empty `fare_leg_rules.from_timeframe_group_id` indicates that the start time of the leg does not affect the fare```
to something like
```An empty `fare_leg_rules.from_timeframe_group_id` indicates that the start time of the leg does not affect the matching of this rule```
(or record?) as the start time may still affect the fare, just not via this rule?

westontrillium:
+1 Trillium

bdferris-v2:
+1 Google

jfabi:
+1 from the @mbta in Boston.

Excited to see consensus built on this important addition to Fares v2—the MBTA has a couple of near-term use cases where we may implement timeframes.

npaun:
+1 Transit

gtfs/spec/en/reference.md
isabelle-dr: Agreed, done in [12c3bb7](https://github.com/google/transit/pull/357/commits/12c3bb7d270cedd80524f99cd8b7f52d454497d3)

gtfs/spec/en/reference.md
isabelle-dr: Keeping as-is for now. Please note that we are working on adding data examples for time-variable fares, with the intention of showing how to represent the most common use-cases. (cc @Sergiodero)
Resolving this conversation.

isabelle-dr:
The voting period ended on 2023-07-24 at 23:59:59 UTC.

With 9 votes in favour and no votes against, the vote to extend GTFS with **Variable fares by time or day** passes! 
The votes came from:
Moovit (@flocsy)
OpenGeo (@skinkie)
Apple Maps (@saraemarcus)
ITO World (@halbertram )
Trillium (@westontrillium)
Google (@bdferris-v2)
MBTA (@jfabi)
Transit (@npaun)
Cityway (@jlgauducheau)

- The [GTFS Fares-v2 Google Doc](https://github.com/google/transit/pull/share.mobilitydata.org/gtfs-fares-v2) has been updated to reflect this change.
- The [Canonical GTFS Schedule Validator](https://github.com/MobilityData/gtfs-validator) has been updated to validate this data, it will be part of the `4.2.0` release, a contribution by @bdferris-v2 🙏.
- We are in the process of updating the [fares v2 extension page](https://gtfs.org/extensions/fares-v2/) and the [data examples](https://gtfs.org/schedule/examples/fares-v2/) on gtfs.org.

Many thanks to everyone who participated in this Pull Request and in the working group meetings, and we would like to extend a special recognition to **ITO World** and **Apple Maps** for their commitment to implementing the changes proposed in this PR, which is required for every GTFS addition.

PR #361, opened 2022-11-18
Fix: broken link in fare_rules.txt

isabelle-dr:
This PR updates the link to the FareExamples provided in `fare_rules.txt`, which was broken.


emmambd:
This PR fixes the link and should be merged so more users can see relevant examples. Step 2 would be adding these examples to gtfs.org and replacing the wayback link entirely.

isabelle-dr:
Thanks, @emmambd. 
I will go ahead and merge this PR since it has been open for 11 days with no opposition.
We don't need a vote, this is a fix that updates a pre-existing link that got broken.

PR #364, opened 2023-01-20
Update CHANGES.md with latest changes

isabelle-dr:
It seems like this doc hasn't been updated for a while 🙈.
I think we can add an item in the governance process for the PR advocate to update this file after their PR is merged?

It looks like this list is not exhaustive, it's just an overview of the biggest updates so if we add this item to the governance, we'll need to define what is "big enough" to require an update.

### DIFF
--- a/gtfs/CHANGES.md
+++ b/gtfs/CHANGES.md
@@ -53,6 +53,22 @@ Every new feature adds complexity to the creation and reading of feeds. Therefor
 
 ### Revision History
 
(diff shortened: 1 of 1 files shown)


emmambd (approved):
LGTM! I'd be interested to hear if anyone thinks having a more detailed breakdown of changes would be valuable.

isabelle-dr:
Thanks @emmambd for the review, merging this PR

Issue #322, opened 2022-04-28
Clarify implied timezone for time values in stop_times.txt


e-lo:
As a transit provider operating in multi-timezones, Amtrak needs GTFS to be more clear (and consistent) on what time zone the values in `stop_times.txt` refer to.

Current [stop_times.arrival_time](http://gtfs.org/schedule/reference/#stop_timestxt) definition:

> Arrival time at a specific stop for a specific trip on a route. If there are not separate times for arrival and departure at a stop, enter the same value for arrival_time and departure_time. For times occurring after midnight on the service day, enter the time as a value greater than 24:00:00 in HH:MM:SS *local time* for the day on which the trip schedule begins.

The above definition, with use of the word "local" implies that stop_times.txt should use the time zone of the local stop.  However, the `stop.stop_timezone` definition clearly states that if used, that `stop_times` shoudl refer to `agency.agency_timezone`. 

Excerpt of current [stops.stop_timezone](https://gtfs.org/reference/static/#stopstxt) definition:

> If stop_timezone values are provided, the times in [stop_times.txt](https://gtfs.org/reference/static/#stop_timetxt) should be entered as the time since midnight in the timezone specified by agency.agency_timezone.

## Example Problem
Amtrak's Southwest Chief Route travels between Kingman AZ (KNG) and Needles CA (NDL):

Amtrak has implemented the following: 

`agency.agency_timezone` is set to Eastern
`stops.stop_timezone` is set to Mountain for KNG and Pacific for NDL
`stop_times.departure_time` for KNG is specified as 12:28 AM (Mountain Time)
`stop_times.arrival_time` for NDL is specified as 12:26 AM (Pacific Time)

...which results in the validator giving a too-fast travel warning ([See this issue in the GTFS Validator Repository](https://github.com/MobilityData/gtfs-validator/issues/1110)

## Proposed Solution

Clarify definition of `arrival_time` and `departure_time` in  [`stop_times.txt`](http://gtfs.org/schedule/reference/#stop_timestxt) and Time term definition as follows

[Field Type: Time](https://gtfs.org/reference/static/#term-definitions)
Time in the HH:MM:SS format (H:MM:SS is also accepted). The time is measured from "noon minus 12h" of the service day (effectively midnight except for days on which daylight savings time changes occur) **in the time zone specified by agency_timezone. in agency.txt** For times occurring after midnight, enter the time as a value greater than 24:00:00 in HH:MM:SS ~local time~ for the day on which the trip schedule begins.
[...]

PR #378, opened 2023-05-17
Clarify implied timezone in stop_times.txt

tzujenchanmbd:
The purpose of this PR is:
- to clarify the usage of time zone for `stop_times.arrival_time` and `stop_times.departure_time` 
- to avoid producers providing time based on incorrect time zone
- to prevent potential confusion that may arise after introduction of time-variable fares

Current definition on `stop_times.arrival_time`:
> Arrival time at the stop (defined by stop_times.stop_id) for a specific trip (defined by stop_times.trip_id). If there are not separate times for arrival and departure at a stop, arrival_time and departure_time should be the same. For times occurring after midnight on the service day, enter the time as a value greater than 24:00:00 in HH:MM:SS **local time** for the day on which the trip schedule begins.

Problem (quoted from Elizabeth in #322):
The above definition, with use of the word "local" implies that `stop_times.txt` should use the time zone of the local stop. However, the `stop.stop_timezone` definition clearly states that if used, that `stop_times` should refer to `agency.agency_timezone`.

This problem caused Amtrak to provide incorrect data in 2022.

Proposed solution:
Clarify definition of `time`, `stop_times.arrival_time`, and `stop_times.departure_time`

Fixes issue#322 (see [issue#322](https://github.com/google/transit/issues/322) for previous discussion)


gtfs/spec/en/reference.md
```diff
@@ -76,7 +76,7 @@ Presence conditions applicable to fields and files:
 - **Float** - A floating point number.
 - **Integer** - An integer.
 - **Phone number** - A phone number.
-- **Time** - Time in the HH:MM:SS format (H:MM:SS is also accepted). The time is measured from "noon minus 12h" of the service day (effectively midnight except for days on which daylight savings time changes occur). For times occurring after midnight, enter the time as a value greater than 24:00:00 in HH:MM:SS local time for the day on which the trip schedule begins. <br> *Example: `14:30:00` for 2:30PM or `25:35:00` for 1:35AM on the next day.*
+- **Time** - Time in the HH:MM:SS format (H:MM:SS is also accepted). The time is measured from "noon minus 12h" of the service day (effectively midnight except for days on which daylight savings time changes occur). For times occurring after midnight, enter the time as a value greater than 24:00:00 in HH:MM:SS for the day on which the trip schedule begins. <br> *Example: `14:30:00` for 2:30PM or `25:35:00` for 1:35AM on the next day.*
```
derhuerst, line 79: Because we're already changing this definition: I'd argue that the "for the day on which the trip schedule begins" confuses more than it helps. I think we should either a) state (again here as before) that the time is relative to the service day, or b) don't add any alternative phrase as a GTFS "Time" is *always* relative to the service day.

leonardehrenfried:
I think this is a good change and would welcome further discussion and an eventual merge.

gtfs/spec/en/reference.md
tzujenchanmbd: @derhuerst what about the following definition?

Time - Time in the HH:MM:SS format (H:MM:SS is also accepted). The time is measured from "noon minus 12h" of the service day (effectively midnight except for days on which daylight savings time changes occur). For times occurring after midnight `on the service day`, enter the time as a value greater than 24:00:00 in HH:MM:SS. [remove "for the day on which the trip schedule begins"]

derhuerst: Sounds good.

npaun:
+1 Transit

leonardehrenfried:
+1 OpenTripPlanner

skinkie:
I would want to make it even more explicit by adding ", not the stop.timezone".

tzujenchanmbd:
@skinkie thanks for the suggestion!
Added ", not `stops.stop_timezone`" for both `arrival_time` and `departure_time`. - https://github.com/google/transit/pull/378/commits/9d4e426d627736e7773a263a4b9a0589c2a8994e

e-lo:
+1

doconnoronca:
+1 TransSee

albanpeignier:
+1 enRoute

tzujenchanmbd:
The vote passed on 2023-07-31 at 23:59:59 UTC.

6 votes in favour and no votes against.

The votes came from:
Transit (@npaun)
Elizabeth Sall (@e-lo)
TransSee (@doconnoronca)
Trillium (@westontrillium)
enRoute (@albanpeignier)

Thanks to everyone who contributed and voted! (Special thanks to @e-lo for bringing up the issue in 2022!)
We are in the process of updating the [GTFS Schedule Reference](https://gtfs.org/schedule/reference/) page on gtfs.org.

Issue #376, opened 2023-05-09
Add recommended presence: reconciling confusion between best practices and spec


emmambd:
### **Problem**

As originally mentioned in https://github.com/google/transit/issues/375, MobilityData’s heard a need to merge best practices into the spec to increase their visibility to producers and reduce confusion between the two documents. We’ve been exploring what a good starting “release” or iteration would look like to begin this work. Based on feedback in #375 and some other discussions, it looks like the one key initial challenge to solve is reconciling places in the best practices and spec where their advice conflicts. 

Some examples:

- feed_info.txt is Optional in the spec, but recommended in all cases in best practices. 
- agency_id is Optional if there is only one transit agency in the spec, but always recommended in the best practices

### **Solution**

Adding a Recommended and Conditionally Recommended presence in the spec that conforms to RFC conventions. This would make it possible to clearly state that a field or file is not required, but adding it is a best practice that should be considered. Making it immediately obvious that a file or field is recommended in the spec reduces confusion, while preserving backwards compatibility. 

This change would not affect validity or the quality evaluation of the dataset according to the Canonical Validator, since the best practices are already warnings in the validator. Any recommended field or file would generate a warning in the validator. 

### **Recommended Scope**

[MobilityData has done an initial review](https://docs.google.com/spreadsheets/d/109w3RFyERjtLYR-dSMsltl386xWJdMn8me1iUnwcdRk/edit#gid=0) of all places in the spec and best practices where they offer contradictory advice on when to add a file or field.

All conflicting best practices that don’t require adopting a new field into the spec relate to

- `feed_info.txt` file and fields
- `route_short_name` in `routes.txt`
- `agency_id` in `routes.txt`, `agency.txt`, and `fare_attributes.txt`
- `timepoint` in `stop_times.txt`
- `block_id` in `trips.txt`
- Looping or inlining routes
    - `shape_dist_traveled` in `stop_times.txt`
    - `shape_dist_traveled` in `shapes.txt`
    - `direction_id` in `trips.txt`
    
After the associated PR is adopted, MobilityData will open a PR on https://github.com/MobilityData/GTFS_Schedule_Best-Practices so that all best practices that have been reconciled + adopted into the spec are removed from the best practices repo.

It would be helpful to get feedback on: 

- The scope itself: is there any best practice here that there isn’t consensus on and ought to be clarified further before it’s added to the spec?
- If there are particular best practices in this list that are high value to trip planners and regulators to reconcile
[...]

Issue #375, opened 2023-04-05
Move Dataset Publishing and General Practices from Best Practices to the spec


emmambd:
### **Problem** 

MobilityData’s heard a number of pains from the community about the GTFS best practices and the spec’s SHOULD statements living in two different places:

- Producers do not always refer to the best practices, and so moving these into the official spec would give the best practices greater visibility and improve data quality for everyone 

- Merging the best practices in the spec would make it easier for regulators to point producers to one place to get the information they need to create their GTFS feeds

### **Solution**

Based on this feedback, MobilityData is interested in working to merge the best practices into the official spec. [We started this work by removing best practices already covered by the spec in 2021.](https://github.com/MobilityData/GTFS_Schedule_Best-Practices/pull/44) We recognize this is a big effort that will require much community discussion, so we want to start with a small section as a pilot to see if we can gain momentum on this and make some incremental changes that will be high value. 

We’ve seen some initial public interest in merging the [Dataset Publishing and General Practices section](https://github.com/MobilityData/GTFS_Schedule_Best-Practices/blob/master/en/best-practices.md#dataset-publishing--general-practices) of the Best Practices into the spec ([Slack conversation here](https://mobilitydata-io.slack.com/archives/C3NH9A9TQ/p1674487393867409?thread_ts=1673977488.805899&cid=C3NH9A9TQ) - you can [join the MobilityData slack](https://docs.google.com/forms/d/e/1FAIpQLSczZbZB9ql_Xl-1uBtmvYmA0fwfm1UX92SyWAdkuMEDfxac5w/viewform?_ga=2.262869135.2013075714.1680701629-892322330.1649683510&_gl=1%2A6d1cbr%2A_ga%2AODkyMzIyMzMwLjE2NDk2ODM1MTA.%2A_ga_55GPMF0W9Z%2AMTY4MDcxODY3OS4xMzcuMC4xNjgwNzE4Njg4LjAuMC4w) to see the discussion). Based on this, we’re seeking more feedback to assess interest and get insight. 

### **Recommended Scope**

MobilityData suggests keeping the scope to the minimum updates required for the Dataset Publishing and General Practices section to be merged into the official spec. This would include 

- Updating the link to the [Google transitfeed tool merge function](https://github.com/google/transitfeed/wiki/Merge) so it references a list of merge tools instead
- Making all SHOULD language explicit

There would be no changes to the severity of any statements - they would all remain SHOULD statements. 

[...]

Issue #227, opened 2020-06-06
frequencies.txt exact_times=1 trip_id semantics


derhuerst:
I have a question about the semantics of `frequencies.txt` with `exact_times=1`.

---

From my experience with GTFS and my observations of GTFS-based UIs, a lot of tools seem to make the following assumption:

A single trip (as defined by a unique `trip_id` in the GTFS dataset) is *one* vehicle (or group of vehicles) that I can use *without significant interruptions* (such as waiting for another vehicle or chasing to a different line). Often, a single trip is considered to be a vehicle which *I can travel with for the whole duration of the `trip`*.

The `frequencies.txt` documentation seems to undermine that assumption however:

> `frequencies.txt` represents trips that operate on regular headways (time between trips). This file can be used to represent two different types of service.
> 
> - Frequency-based service (`exact_times=0`) in which service does not follow a fixed schedule throughout the day. Instead, operators attempt to strictly maintain predetermined headways for trips.
> - A compressed representation of schedule-based service (`exact_times=1`) that has the exact same headway for trips over specified time period(s). In schedule-based service operators try to strictly adhere to a schedule.
>
> Field Name | Type | Required | Description
> -- | -- | -- | --
> `trip_id` | ID referencing `trips.trip_id` | Required | Identifies a trip to which the specified headway of service applies.

I think this is especially important for routing engines: **Now, they can't assume anymore that every GTFS data point referring to the same `trip_id` is tied to one vehicle allowing continuous travel**. There are >=1 "runs" of a vehicle, all under the same `trip_id`, but each of them ends at the last stop specified in `stop_times.txt`.

---

Is my understanding of the semantics correct? If it is, I'd argue that this is quite unintuitive and therefore easy to implement in a wrong way. If I misunderstood how `frequencies.txt` works, let's improve the documentation.

skinkie:
On Saturday, June 6, 2020 8:05:27 PM CEST, Jannis Redmann wrote:
> Is my understanding of the semantics correct? If it is, I'd 
> argue that this is quite unintuitive and therefore easy to 
> implement in a wrong way. If I misunderstood how 
> `frequencies.txt` works, let's improve the documentation.

Could you give an example how it could be interpreted in a wrong way?

derhuerst:
Let's consider an excerpt from the [example feed linked in the spec](https://github.com/google/transit/blob/102179ff7b0c5e2c6025dc6b8cef825d8999aed3/gtfs/spec/en/examples/README.md):

[...]

PR #386, opened 2023-07-05
Best practices: add recommended presence

emmambd:
### **Problem and Suggested Scope**

Referenced from [#375](https://github.com/google/transit/issues/375) and [#376](https://github.com/google/transit/issues/376):

MobilityData’s heard a number of pains from the community about the GTFS best practices living in two different places: the spec's SHOULD statements, and the GTFS Best Practices. 

- Stakeholders are not always aware of the existence of the best practices, and so moving these into the official spec would give the best practices greater visibility and improve data quality for everyone

- Merging the best practices in the spec would make it easier for regulators to point producers to one place to get the information they need to create their GTFS feeds

We’ve been exploring what a good starting “release” or iteration would look like to begin this work. Based on feedback in https://github.com/google/transit/issues/375 and some other discussions, it looks like the one key starting point is providing clear guidelines in the specification for when a field or file is recommended for use.

[Before creating this PR, we did an initial review](https://docs.google.com/spreadsheets/d/109w3RFyERjtLYR-dSMsltl386xWJdMn8me1iUnwcdRk/edit#gid=0) of all places in the spec and best practices where they offer advice on when to add a file or field. This spreadsheet also includes the corresponding [Canonical Validator](https://github.com/MobilityData/gtfs-validator) rule for each best practice for cross-referencing.

### **Proposed solution**

Adding a Recommended presence in the spec that conforms to RFC conventions. This would make it possible to clearly state that a field or file is not required, but adding it is a best practice that should be considered. Making it immediately obvious that a file or field is recommended in the spec reduces confusion, while preserving backwards compatibility.

This change would not affect validity or the quality evaluation of the dataset according to the Canonical Validator, since the best practices are already translated to warnings in the validator. Any missing recommended field or file would generate a warning in the validator.

If the associated PR is adopted, MobilityData will open a PR on https://github.com/MobilityData/GTFS_Schedule_Best-Practices so that all best practices that have been reconciled + adopted into the spec are removed from the best practices repo.

**When is a file/field Recommended instead of Optional?**

The Recommended presence is used for any file/field that should be used universally, e.g `feed_info.txt` or `agency_id`. This is information producers should provide in all circumstances, but is not required in order to preserve backwards compatibility. 

Any file/field that is recommended only under specific circumstances (such as `route_short_name` when there is a brief service designation) keeps the Optional presence and the words "SHOULD" and/or "recommended" are used in its description. 

### **Alternatives considered**

**Adding no additional presences and keeping all SHOULDs in the description**
The concern with this approach was that it remains confusing for producers for a file/field to be optional but recommended in all cases. [Conversation here.](https://github.com/google/transit/issues/375#issuecomment-1520564752)

**Adding a Conditionally Recommended presence**
This was originally drafted, this seemed to overly complicate the spec and create confusion about the difference between Recommended presences and the Optional presence. [Conversation here.](https://github.com/google/transit/issues/376#issuecomment-1546340051)


gtfs/spec/en/reference.md
```diff
@@ -223,7 +224,7 @@ Primary key (`trip_id`)
...
-|  `block_id` | ID | Optional | Identifies the block to which the trip belongs. A block consists of a single trip or many sequential trips made using the same vehicle, defined by shared service days and `block_id`. A `block_id` may have trips with different service days, making distinct blocks. See the [example below](#example-blocks-and-service-day). To provide in-seat transfers information, [transfers](#transferstxt) of `transfer_type` `4` should be provided instead. |
+|  `block_id` | ID | Optional | Identifies the block to which the trip belongs. A block consists of a single trip or many sequential trips made using the same vehicle, defined by shared service days and `block_id`. A `block_id` may have trips with different service days, making distinct blocks. See the [example below](#example-blocks-and-service-day). To provide in-seat transfers information, [transfers](#transferstxt) of `transfer_type` `4` should be provided instead. May be provided for frequency-based trips. |
```
gcamp, line 227: What does this change add?

gtfs/spec/en/reference.md
```diff
@@ -196,8 +197,8 @@ Primary key (`route_id`)
...
-|  `route_short_name` | Text | **Conditionally Required** | Short name of a route. Often a short, abstract identifier (e.g., "32", "100X", "Green") that riders use to identify a route. Both `route_short_name` and `route_long_name` may be defined.<br><br>Conditionally Required:<br>- **Required** if `routes.route_long_name` is empty.<br>- Optional otherwise. |
+|  `agency_id` | Foreign ID referencing `agency.agency_id` | **Conditionally Required** | Agency for the specified route.<br><br>Conditionally Required:<br>- **Required** if multiple agencies are defined in [agency.txt](#agency). <br>- Recommended otherwise. |
+|  `route_short_name` | Text | **Conditionally Required** | Short name of a route. Often a short, abstract identifier (e.g., "32", "100X", "Green") that riders use to identify a route. Both `route_short_name` and `route_long_name` may be defined.<br><br>Conditionally Required:<br>- **Required** if `routes.route_long_name` is empty.<br>- Recommended if there is a brief service designation. This should be the commonly-known passenger name of the service, no longer than 12 characters. |
```
gcamp, line 201: I would emphasize the should since the num of char is not a requirement. 

```suggestion
|  `route_short_name` | Text | **Conditionally Required** | Short name of a route. Often a short, abstract identifier (e.g., "32", "100X", "Green") that riders use to identify a route. Both `route_short_name` and `route_long_name` may be defined.<br><br>Conditionally Required:<br>- **Required** if `routes.route_long_name` is empty.<br>- Recommended if there is a brief service designation. This should be the commonly-known passenger name of the service, and should be no longer than 12 characters. |
```

gtfs/spec/en/reference.md
```diff
@@ -647,11 +649,11 @@ If both referencing methods (`record_id`, `record_sub_id`) and `field_value` are
...
+|  `feed_end_date` | Date | Recommended | (see above) |
+|  `feed_version` | Text | Recommended | String that indicates the current version of their GTFS dataset. GTFS-consuming applications can display this value to help dataset publishers determine whether the latest dataset has been incorporated. |
+|  `feed_contact_email` | Email | Optional | Email address for communication regarding the GTFS dataset and data publishing practices. `feed_contact_email` is a technical contact for GTFS-consuming applications. Provide customer service contact information through [agency.txt](#agencytxt). Should be used when `feed_contact_url` is empty. |
```
gcamp, line 655: I find this clearer than "should if empty". Should be copied on the other field too. 

```suggestion
|  `feed_contact_email` | Email | Optional | Email address for communication regarding the GTFS dataset and data publishing practices. `feed_contact_email` is a technical contact for GTFS-consuming applications. Provide customer service contact information through [agency.txt](#agencytxt). It's recommended that at least one of `feed_contact_url` or `feed_contact_url` are provided. |
```

gcamp (commented):
👏 left a couple comments

gtfs/spec/en/reference.md
Sergiodero: The `May be provided for frequency-based trips` line has been removed in the latest commit 29d268fe1a569baea3ebff3cf8fb44e11ef9095d .

Sergiodero:
Hey @gcamp, Sergio from MobilityData here. We've added a new commit that removes the `May be provided for frequency-based trips`. We’re proposing to remove this and instead incorporate a use case in the GTFS.org example section to illustrate the use of block_id with frequency based trips, this example could be similar to the one @antrim shared in issue #227.

Currently we’re exploring what could be a good example or use case, but we think that this could be a separate discussion as it would fall on the GTFS.org repo, allowing us to move forward with the discussion in this PR, while simultaneously working on defining this example.

emmambd:
Since all PR comments have been resolved and there's been engagement and interest in #376 and #376, I'm opening the vote. This PR adds a recommended presence to the spec and best practices for when specific files and fields should be added. These guidelines are taken from [GTFS Best Practices](https://gtfs.org/schedule/best-practices/).

Please vote with a +1 (for) or -1 (against) in the comments. Voting ends on 2023-07-31 at 23:59:59 UTC.

gcamp:
+1 Transit

westontrillium:
+1 Trillium

derhuerst:
+1 [bbnavi](https://bbnavi.de)

doconnoronca:
+1 TransSee

NomeQ:
+1 Garnet

evansiroky:
+1 Cal-ITP.

philip-cline:
+1 Arcadis IBI

bdferris-v2:
+1 Google

One formatting nit: for the `shapes.txt - shape_dist_traveled` addition, was the intention that the "Example" be rendered inside the table cell for the `shape_dist_traveled` field?  If so, I think you need to get rid of the line break before the `<hr>*Example` line because it causes the content to break out of the table cell.

Screenshot 2023-07-31 at 3 10 39 PM

emmambd:
@bdferris-v2 Thanks for catching that! Corrected.

emmambd:
👋 Here's an update to go vote to remove these same best practices from the Best Practices repo so we don't have duplicate guidance: https://github.com/MobilityData/GTFS_Schedule_Best-Practices/pull/59. 

Voting ends on 2023-08-29 at 23:59:59 UTC.

PR #412, opened 2023-12-07
Updated description in `translations` and `feed_info`

ponlawat-w:
Should this paragraph which appeared in `feed_info` description belong to `translations` instead?

> If both referencing methods (`record_id`, `record_sub_id`) and `field_value` are used to translate the same value in 2 different rows, the translation provided with (`record_id`, `record_sub_id`) takes precedence.


leonardehrenfried (approved):
I agree with Tzu-Chen.

tzujenchanmbd (approved):

PR #414, opened 2023-12-10
Editorial change: update text in networks.txt and routes_networks.txt

isabelle-dr:
This PR does the following:
- removed the file requirement at the file level for `networks.txt` and `route_networks.txt`, to stay consistent with the rest of GTFS where the file requirements are defined in the [Dataset files](https://gtfs.org/schedule/reference/#dataset-files) section and not re-iterated in the individual files.
- removed the file requirement for `levels.txt` for consistency (we may want to add file requirements at the file level for every file in GTFS in a future PR).
- modified some descriptive text to add a link to the `routes.txt` file in the [Dataset files](https://gtfs.org/schedule/reference/#dataset-files) section.
- Added a reference to the meaning of "a field exists" in the **Term Definition** section.

This is an editorial change and does not need a vote. 
Tagging @tzujenchanmbd for a review!

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -121,8 +121,8 @@ This specification defines the following files:
 |  [fare_transfer_rules.txt](#fare_transfer_rulestxt)  | Optional | Fare rules for transfers between legs of travel.<br><br>Along with [fare_leg_rules.txt](#fare_leg_rulestxt), file [fare_transfer_rules.txt](#fare_transfer_rulestxt) provides a more detailed method for modeling fare structures. As such, the use of [fare_transfer_rules.txt](#fare_transfer_rulestxt) is entirely separate from files [fare_attributes.txt](#fare_attributestxt) and [fare_rules.txt](#fare_rulestxt). |
 |  [areas.txt](#areastxt) | Optional | Area grouping of locations. |
 |  [stop_areas.txt](#stop_areastxt) | Optional | Rules to assign stops to areas. |
-|  [networks.txt](#networkstxt) | **Conditionally Forbidden** | Network grouping of routes.<br><br>Conditionally Forbidden:<br>- **Forbidden** if `network_id` is defined in [routes.txt](#routes.txt).<br>- Optional otherwise. |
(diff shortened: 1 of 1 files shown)


tzujenchanmbd:
I have considered both these suggestions, no strong opinion but here is my original thoughts: 

> removes the file requirement at the file level, to stay consistent with the rest of GTFS where the file requirements are defined in the Dataset files section and not re-iterated in the individual files.

It seems there wasn't consistency initially. There was repetition in `levels.txt`. My initial thought was this could make reading more convenient (without having to scroll up and down the page).

> modified some descriptive text for consistency: using defined instead of exists to describe a field, adding a link to the routes.txt file in the Dataset files section.

I am not sure if we want the "consistency" here. 

When GTFS mentions "is defined," the scope is within a single record. For instance, in the statement "`timeframes.start_time` is required if `timeframes.end_time` is defined," it refers to conditions within the same record. When the `end_time` of other records is defined, it does not necessarily mean that the `start_time` of that specific record needs to be defined. However, the condition we want to articulate here is: "if any value of this field is defined," which seems to be the first time for GTFS.

Therefore, to completely eliminate the following confusion, I've used the new phrase "field exist." -

If "is defined" refers to specific records, can `routes.network_id `and `networks.network_id` be separately defined for different routes, as long as they don’t overlap?

In addition, there could be a difference for validator in this situation:
A feed uses networks files to define networks, and there is only field header `network_id`(no values at all) in `routes.txt`.
- "field exists" seems to flag an error 
- "is defined" seems no error

westontrillium:
I've always interpreted "defined"=data exists in X column, "field exists"=X column exists, but there isn't *necessarily* data in that column, i.e., it may be empty.

As such, I assumed that requiring the use of one method or the other for the whole dataset was the intention of the original wording. Is that not the case?

isabelle-dr:
After a discussion with @tzujenchanmbd:
- we've kept file requirements details in the **Dataset Files** section only
- added a mention of what "field" exists in the **Term Definitions** section to clarify what it is

I have modified the descriptive text back to using "exists" and not "defined" in [Change defined -> exists](https://github.com/google/transit/pull/414/commits/e013c7c9bf0a3821e21ddc8d1d22d70604f44222).

tzujenchanmbd (approved):
LGTM!

Issue #413, opened 2023-12-08
Modifications to the GTFS Governance: Phasing Plan


eliasmbd:
# 🗒️ Context

After conducting numerous interviews and workshops throughout 2023, MobilityData is suggesting several refinements to address common problems in both the formal amendment process and the informal processes related to GTFS governance.

In this issue, you will find a summary of the most common problems as highlighted by the community.

# 🤔 Problems

### **High barrier to entry**

* Scattered communication leads to confusion regarding where to access key information.
* Confusion on where the source of truth for documentation lies. 
* Lack of overview on what is actively being discussed.
* The governance process is designed for major changes, creating an intimidating barrier for individuals seeking to propose minor improvements.

### **Insufficient engagement in the proposals in the early stages of development.**

* Vague process before the Pull Request stage.
* The level of consensus is frequently unclear until a Pull Request is initiated or a vote is called.
* Reviews are vague, it’s hard to know if a proposal has been thoroughly reviewed before testing and adoption.

### **First adopters are frequently impacted by last-minute changes to their implementation, leading to an increase in committed resources.**

* First Adopters need to implement changes before opening a vote. Certain problems identified too late demand multiple rounds of testing and require supporting various versions of a proposal.

# 🔮 Phasing Plan

Based on these issues, MobilityData has worked on a phasing plan, which will be updated to represent the current status.

<table>
  <tr>
   <td><strong>Phase</strong>
   </td>
   <td><strong>What’s included</strong>
   </td>
   <td><strong>Status</strong>
   </td>
  </tr>
  <tr>
   <td>1 - Effective communication and GitHub management

   </td>
   <td>
<ul>

<li>Publication of a monthly “GTFS Digest” that gives an overview of the active conversations in the community and latest additions.

<li>Using GitHub issue templates and labels to guide contributors and give a better overview of the status of each issue.

https://github.com/google/transit/pull/417
https://github.com/google/transit/issues/419
**[Modified Labels Categorization](https://github.com/google/transit/labels)**
</li>
</ul>
   </td>
   <td>✅ Done
   </td>
  </tr>
  <tr>
   <td>2 - Enhancing voting and reviews
   </td>
   <td>
<ul>

<li>Adding an earlier vote to increase the visibility and potential issues of a proposal early-on.  
<ul>
 
<li>The veto would be moved to this earlier vote to incentivise contributors to review the proposal early and share any major concerns before early adopters implement the changes proposed
 
<li>The final vote would follow a majority vote with a 80% threshold
[...]

PR #417, opened 2023-12-15
[Governance] Phase 1: Issue Templates

emmambd:
## 📓 Context

As part of Phase 1 of [Epic Issue #413](https://github.com/google/transit/issues/413), MobilityData proposes to add Issue Templates as one of the solutions to help alleviate the high barrier to entry that was rendered evident to us during multiple outreach efforts in 2023. 

## 💡 Proposed Solution 

We propose to implement the following templates and links: 

* Specification Change: Guide the contributor to describe the problem and showcase the proposed solution. 
* Ask a Question: Guide the contributor to provide sufficient information for better answers. 
* Links: Redirect contributors to valuable resources like Slack, GTFS.org, and the Google Groups
* The template page also provides the option to publish a blank issue at the bottom of the page. 

## ℹ️ Additional Information

As the governance changes in the next months, we can expect to change or add templates accordingly. 

* When editorial and "smaller" changes are defined, we can expect to create templates for those cases. 

We welcome the community to review the proposal and provide some feedback. We hope to adopt this change as an editorial change and fast track its adoption if there is a clear consensus in the comments.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,11 +1,11 @@
 contact_links:
-  - name: GTFS.org
+  - name: GTFS Documentation
     url: https://gtfs.org/
     about: The official reference for GTFS documentation. 
-  - name: GTFS Changes Google Group
+  - name: GTFS Changes Mailing List
     url: https://groups.google.com/g/gtfs-changes 
     about: Follow this group to get announcement on votes on the GTFS Schedule specification. 
-  - name: GTFS Realtime Google Group 
+  - name: GTFS Realtime Mailing List 
     url: https://groups.google.com/g/gtfs-realtime
     about: This group is the forum for discussing the GTFS Realtime specification, asking questions, and proposing changes.
   - name: GTFS Slack Channel
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,13 @@
+contact_links:
+  - name: GTFS.org
+  - url: https://gtfs.org/
+  - about: The official reference for GTFS documentation. 
+  - name: GTFS Changes Google Group
+  - url: https://groups.google.com/g/gtfs-changes 
+  - about: Follow this group to get announcement on votes on the GTFS Schedule specification. 
+  - name: GTFS Realtime Google Group 
+  - url: https://groups.google.com/g/gtfs-realtime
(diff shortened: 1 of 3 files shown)


eliasmbd (commented):
Looks good to me!

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,13 @@
+contact_links:
+  - name: GTFS.org
```
isabelle-dr, line 2: I'd call this Documentation or GTFS Documentation instead of having the URL as a title. 
I think it's a bit more aesthetic, and also more clear: this URL could also be a database of GTFS datasets.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,13 @@
...
+    url: https://gtfs.org/
+    about: The official reference for GTFS documentation. 
+  - name: GTFS Changes Google Group
```
isabelle-dr, line 5: I'd call this GTFS Changes Mailing List (this is what is used in the [Spec Amendment Process](https://gtfs.org/schedule/process/)) and more self-explanatory.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,13 @@
...
+    url: https://groups.google.com/g/gtfs-changes 
+    about: Follow this group to get announcement on votes on the GTFS Schedule specification. 
+  - name: GTFS Realtime Google Group
```
isabelle-dr, line 8: Similarly, I'd call this GTFS Realtime Mailing List

isabelle-dr (commented):
I have made a small fix to `config.yml`, you can see what this PR would look like for this repo [here](https://github.com/isabelle-dr/transit/issues/new/choose).

I've left a few minor comments inline. Thank you!

.github/ISSUE_TEMPLATE/config.yml
eliasmbd: I agree. Let's keep it close to what is used in Amendment Process.

.github/ISSUE_TEMPLATE/config.yml
eliasmbd: Yes this makes sense. I wonder if the main purpose of redirecting to GTFS.org is documentation. I would say it's likely a high percentage of the visit are for things like references pages. I would say this works.

PR #423, opened 2024-01-08
[GTFS-Fares v2] Fare product/media transfer behavior

### LINKED ISSUE #420: [GTFS-Fares v2] Multi-leg Transfer: Same product/media transfer behavior
### **Use Case**

1. When riders transfer between multiple legs during their journey using fare products, there are typically two scenarios:

- In the case of a "ticket-based system" (i.e., a product similar to a pass), special transfer rules between two legs (such as free transfers) must be applied using the same fare product.
- In a "stored-value system" (i.e., a general pay-as-you-go product), special transfer rules between two legs do not require the use of the same fare product.

2. Most transfer rules between two legs necessitate the use of the same fare media to apply, but there are exceptions. For instance, within a region, two networks might exist where one has introduced an app media while the other only uses paper tickets. In such cases, transfer rules might apply between two distinct fare media (for example, showing a paper ticket to the driver to get a free transfer).

### **Problem to solve**
The current Fares v2 spec does not have mechanism for differentiating the above use cases or provide explanation within its description, leading to potential ambiguity. This is especially relevant for certain transit systems, such as STM, that operate 100% with a ticket-based system. Transit firstly posed the question, "[Should fare transfers require that the same fare product is used on both legs?](https://docs.google.com/document/d/1zF5i03BFQi1kVBfN17aR7N1KRrpsHs3B9YjCzE8apJE/edit#heading=h.9a66f7frd08b)" Subsequently, Google also raised the following questions in this [transfer semantics document](https://docs.google.com/document/d/1Lqh3eI2YoxAaLSqUePCkcQ7WmyD-d-eXeDYymPnUfLQ/edit#heading=h.swddidrlwbq1) [_this issue only discusses the first three questions in this document_]:
- For a leg-to-leg transfer to match a row in `fare_transfer_rules.txt`, is it required that both legs have the same fare media type?
- Is it intended that multiple rows in `fare_transfer_rules.txt` can match a single leg-to-leg transfer?

### **Proposed solution**
On October 24th, the working group reached a consensus to clarify these questions, leaning towards a default behavior mechanism to minimize any potential breaking changes. Based on this consensus, we drafted [this proposal](https://docs.google.com/document/d/18e1XGgp5vMO2R-wIIk3nnZea-Ol5e9UaWkeebi3ijI8/edit#heading=h.hh2l9k6b2on9) that aims to resolve the above 3 questions.

In the November discussion, the working group is currently leaning towards option 1 proposed, which involves adding `fare_product_behavior`, `fare_media_behavior`, and `filter_fare_product_id` to `fare_transfer_rules`. This option is favored as this mechanism appears to be more like the attributes of `fare_transfer_rules `(with option 2 being an alternative).

We'd love to see any feedback on this proposal or any related use cases. Please feel free to share any thoughts on this issue.

tzujenchanmbd:
This PR covers the following questions regarding multi-leg transfer fares: 
- Should fare transfers require that the same fare product is used on both legs?
- For a leg-to-leg transfer to match a row in `fare_transfer_rules.txt`, is it required that both legs have the same fare media type?
- Is it intended that multiple rows in `fare_transfer_rules.txt` can match a single leg-to-leg transfer?

In the November 2023 meeting, the working group leaned towards option one proposed in the [spec modeling proposal](https://docs.google.com/document/d/18e1XGgp5vMO2R-wIIk3nnZea-Ol5e9UaWkeebi3ijI8/edit#heading=h.hh2l9k6b2on9), which allows producers to indicate in the `fare_transfer_rules.txt` whether specific transfer rules require the same fare product/media between two legs. 

(For more detailed context, please see [issue#420](https://github.com/google/transit/issues/420))

### Changes in this PR
- Add `fare_product_behavior`, `fare_media_behavior`, and `filter_fare_product_id` fields to `fare_transfer_rules.txt`.

There are some data examples in the [spec modeling proposal](https://docs.google.com/document/d/18e1XGgp5vMO2R-wIIk3nnZea-Ol5e9UaWkeebi3ijI8/edit#heading=h.hh2l9k6b2on9). For previous discussions, please see [working group meeting minutes](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit#heading=h.qxaj5nixu2ns). 

Please go through the changes and feel free to share your thoughts/questions here.


jfabi:
The @MBTA plans to begin producing `fare_product_behavior`, `fare_media_behavior`, and `filter_fare_product_id`, tentatively by 1 July. We'll check back here once these fields are up and running in production.

jfabi:
> The @mbta plans to begin producing `fare_product_behavior`, `fare_media_behavior`, and `filter_fare_product_id`, tentatively by 1 July. We'll check back here once these fields are up and running in production.

As of last week, the @mbta is now producing all three of these new fields in its GTFS-static feed, [downloadable here](https://cdn.mbta.com/MBTA_GTFS.zip).

tzujenchanmbd:
Not stale

tzujenchanmbd:
@felixguendling would Motis be interested in implementing this?

felixguendling:
Yes, I think this makes sense and should be easy to implement in case it gets standardized.

leonardehrenfried:
For me this is no longer a theoretical question as Trimet's (Portland) feed would lead to a large number of possible combinations for each transfer.

For now I can add a frontend filter option that stipulates that rider category and medium must match but it would be nice if the spec would be more explicit.

tzujenchanmbd:
Update:
The Fares v2 Working Group revisited this proposal during the May 27 meeting and confirmed the proposed changes.

Based on the discussion, to facilitate testing and adoption, we agreed to narrow the scope of this PR to focus on fare product/media behavior. The `filter_fare_product_id` field will be removed for now and can be added later if needed.

The behavior related to rider categories can be addressed in a separate change as well.

m-mcqueen:
TriMet is in support of this PR as it streamlines implementation for our Regional Trip Planner at trimet.org.

leonardehrenfried:
@m-mcqueen Are you enough in support to produce data for it? This matters mostly for SMART.

If so, then I can implement it in OTP and we can get it into the spec.

m-mcqueen:
@leonardehrenfried we are not producing a public SMART GTFS feed, however we are augmenting the existing public feed locally to add fares information for our regional trip planner. This is a temporary measure until SMART onboards a new GTFS vendor later this year. My understanding is that producing GTFS-Fares v2 compliant data was part of the scope.

I will pursue adding these fields in the TriMet GTFS feed should the spec be updated.

leonardehrenfried:
Well, we have a chicken and egg problem here: if there isn't an early adopter producing the data, it will not be added to the spec.

jfabi:
As of today, the @MBTA's [GTFS feed](https://cdn.mbta.com/MBTA_GTFS.zip) includes stored value card and contactless fares, and as a consequence, we are publishing transfers with both possible `fare_product_behavior` values.

We also publish `fare_media_behavior`; the way our rules/transfers/products are set up, we only publish the `0` value.

m-mcqueen:
@leonardehrenfried is the MBTA data sufficient? I can also offer a mockup feed of SMART data.

tzujenchanmbd:
@m-mcqueen If there are both `fare_media_behavior` `0` and `1` value, it would be nice to have the mockup feed.🙏

jll01:
We (Transit app) have implemented `fare_product_behavior` and tested it against the MBTA data, where it works as intended. However, we haven't deployed it, as the default value is a breaking change.

The PR defines empty (or 0) as "the transfer rule applies regardless of which products are used in both legs". We believe consumers have historically applied transfer rules only when the same fare product is used on both legs, so this default changes the meaning of every existing dataset that omits the column: rules would suddenly grant cross-product transfers that producers never intended.

**Proposal**: make the default restrictive. Empty should mean "the transfer rule applies only when both legs use the same fare product," with explicit 0/1 keeping their current meanings. This preserves backward compatibility and is consistent with `fare_media_behavior`, whose default is already restrictive. The MBTA publishes explicit values, so their GTFS would need no changes.

Two additional points:
* Semantics for 3+ leg journeys should be specified. Once a transfer rule moves a rider from product A to product B, which product does a subsequent "same product" rule match against? We propose that the acquired product (B) is used for all subsequent matching, with duration limits still counted from the first validation of the journey (with product A).
* The PR description still mentions `filter_fare_product_id`, which was removed in May.

Issue #367, opened 2023-01-31
Improve definition of Fare Products


isabelle-dr:
### Problem
There is confusion around what a fare product is in GTFS. Two sources of confusion were brought-up to our attention:

**First**, the definition of `fare_products.txt` is:
> To describe the different types of tickets or fares that can be purchased by riders.

When in reality, some fare products represent virtual products that can't be purchased (by themselves) by riders, such as a transfer cost.

**Second**: although `fare_poduct_id`'s type is `ID` and not `Unique ID`, it's unclear in the spec that several Fare Products can be grouped under the same `fare_product_id`, and under which conditions this should be done to ensure an efficient fare design.

### Solution
The file definition should be modified to include virtual products that can't be directly bought by riders.
The file definition should give insights on when fare products should be grouped under the same ID.
The definition of `fare_poduct_id` should be modified to include something like: 

> Multiple records in `fare_products.txt` may have the same `fare_product_id`, in which case all fare products with that ID will be applied when referenced from another file.

PR #426, opened 2024-01-10
[GTFS-Fares v2] Clarification: Fare product definition

tzujenchanmbd:
### Potential confusions to solve:

1. The current file description of `fare_products.txt` is:
> To describe the different types of tickets or fares that `can be purchased by riders`.

However, some fare products represent virtual products that can't be purchased (by themselves) by riders, such as a transfer cost.

2. Although the current primary key [`fare_product_id`, `fare_media_id`] already implies multiple records can have the same `fare_product_id`, it's unclear in the description that several Fare Products can be grouped under the same `fare_product_id`. (As mentioned/suggested in [issue#367](https://github.com/google/transit/issues/367))

3. There might be confusion on whether records with the same `fare_product_id` can have different prices.
(At the November meeting, the working group agreed that the same `fare_product_id` can have different prices)

### Changes in this PR:
Add clarification text to `fare_products.txt`

Since this is just a clarification, I propose we don't have to go through the voting process.


gtfs/spec/en/reference.md
```diff
@@ -407,14 +407,14 @@ File: **Optional**
 
 Primary Key (`fare_product_id`, `fare_media_id`)
 
-To describe the different types of tickets or fares that can be purchased by riders.
+To describe the different types of tickets or fares, including those directly purchasable by riders and integral virtual entities like transfer costs.
 
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ | ------ |
-| `fare_product_id` | ID | **Required** | Identifies a fare product. |
+| `fare_product_id` | ID | **Required** | Identifies a fare product.<br><br>Multiple records in [fare_products.txt](#fare_productstxt) may have the same `fare_product_id`, in which case all fare products with that ID will be applied when referenced from another file. |
```
bdferris-v2, line 414: Is it worth being more explicit about the interaction with `fare_media_id` here?  Which is to say, multiple records maybe the same `fare_product_id` with different fare media, indicating different methods for purchasing and validating the same fare product.

gtfs/spec/en/reference.md
```diff
@@ -407,14 +407,14 @@ File: **Optional**
...
 | `fare_product_name` | Text | Optional | The name of the fare product as displayed to riders. |
 |  `fare_media_id` | Foreign ID referencing `fare_media.fare_media_id` | Optional |  Identifies a fare media that can be employed to use the fare product during the trip. When `fare_media_id` is empty, it is considered that the fare media is unknown.|
-| `amount` | Currency amount | **Required** | The cost of the fare product. May be negative to represent transfer discounts. May be zero to represent a fare product that is free.|
+| `amount` | Currency amount | **Required** | The cost of the fare product. May be negative to represent transfer discounts. May be zero to represent a fare product that is free.<br><br>Different records with the same `fare_product_id` may have different amount values, indicating variations in pricing.|
```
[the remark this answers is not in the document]
tzujenchanmbd: Yes, it seems it would be better to explain within the description of the `fare_product_id` field rather than individually in all non-primary key fields. I will remove the description from the amount field.

gtfs/spec/en/reference.md
tzujenchanmbd: Yes, thanks for this suggestion @bdferris-v2!

Change made - https://github.com/google/transit/pull/426/commits/f0a12b8deb528ee804e2120c4b9be1fb2ddfa07b

Does it look clearer now?

gtfs/spec/en/reference.md
```diff
@@ -407,14 +407,14 @@ File: **Optional**
 
 Primary Key (`fare_product_id`, `fare_media_id`)
 
-To describe the different types of tickets or fares that can be purchased by riders.
+To describe the different types of tickets or fares, including those directly purchasable by riders and integral virtual entities like transfer costs.
```
isabelle-dr, line 410:
```suggestion
Used to describe the range of fares available for purchase by riders or taken into account when computing the total fare for journeys with multiple segments, such as transfer costs.

File: gtfs/spec/en/reference.md [hunk 4]
```diff
@@ -407,14 +407,14 @@ File: **Optional**
...
 
 |  Field Name | Type | Presence | Description |
 |  ------ | ------ | ------ | ------ |
-| `fare_product_id` | ID | **Required** | Identifies a fare product. |
+| `fare_product_id` | ID | **Required** | Identifies a fare product.<br><br>Multiple records in [fare_products.txt](#fare_productstxt) may share the same `fare_product_id`, in which case all records with that ID will be applied when referenced from another file. These multiple records may possess the same `fare_product_id` but with different `fare_media_id`, indicating different methods for purchasing and validating the same fare product. Based on these different methods, there may also be varying values in all non-primary-key fields, such as representing different fare product names and costs. |
```
isabelle-dr, line 414:
```suggestion
| `fare_product_id` | ID | **Required** | Identifies a fare product or set of fare products.<br><br>Multiple records in [fare_products.txt](#fare_productstxt) may share the same `fare_product_id`, in which case all records with that ID will be retrieved when referenced from another file. 
Multiple records may share the same `fare_product_id` but with different `fare_media_id`s, indicating various methods available for employing the fare product, potentially at different prices. 
|

gtfs/spec/en/reference.md
isabelle-dr: A small suggestion to make the cases where we want to define "virtual" fare products more clear.

gtfs/spec/en/reference.md
isabelle-dr: Included in this suggestion
- consistency with the definition of Fare Media (not exactly a method for purchasing, but a method for employing).
- adding "or set of" for consistency with other non-unique ID timeframe_group_id.

isabelle-dr (approved):
Thank you for working on this @tzujenchanmbd :)

I think @bdferris-v2 comments are addressed on your last commits. 
I've added suggestions, feel free to use them or portions of them, and then we're good to merge this PR 🚀.

PR #428, opened 2024-01-24
Add ignore .DS_Store

gcamp:
Ignore .DS_Store, metadata files by macOS that have no utility in the git repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ __pycache__/
 
 # C extensions
 *.so
+
+# macOS metadata files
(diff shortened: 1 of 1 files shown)


gcamp:
@jspetrak which make sense in an organisation where you can have some standard setup, I don't think we can assume that in our current context

skinkie:
Would it be a bold statement to try it upstream? Hence git for OSX by default ignore .DS_store?

eliasmbd:
@gcamp Will you continue pursuing this PR?

PR #431, opened 2024-02-12
fix(gtfs/spec/reference): remove extra backtick in`record_id` description

firestack:
This removes an extra backtick in the `record_id` row of the `translations` table. This backtick was causing all subsequent backticks and inline elements to not be applied correctly.

Before:
Before image of `record_id` row


tzujenchanmbd (approved):
Hi @firestack,

Thank you so much for catching this error. We will merge this PR soon.

eliasmbd:
@firestack Thank you for contributing to the GTFS community! We would like to highlight that you are the 40th unique PR author in google/transit, this is a great milestone! Congrats! 🎉

Issue #429, opened 2024-01-30
stops.zone_id conditional requirement with presence of route-based fare_rules?


westontrillium:
### Describe the problem

We have received multiple messages from agencies confused by an error (`stop_without_zone_id`) found in validation reports forwarded to them by the National Transit Database. I've investigated the error and am wondering if the way the spec handles `zone_id` requirements should be clarified, and if so, whether such a clarification would break backwards compatibility. I'm not sure, because it is a *conditional* requirement, and adding the exception I propose below would be easing a restriction, not adding a new one.

The current requirement is essentially: If fare_rules.txt contains records, then presence of `zone_id` is required for **all** stops in the dataset.

However, there are cases where you could define a fare_rule without the use of `zone_id`s, for example, with route-based fares. In my opinion, a feed should still be valid with partial `zone_id` presence IF those stops without a `zone_id` are present in stop_times records with a `trip_id` whose `route_id` is present in a fare_rules.txt record with only `fare_id` and `route_id` defined.

### Use cases

Consider the following example
<br>

**fare_rules.txt**

fare_id | route_id | origin_id | destination_id
-- | -- | -- | --
1855 | | A | B
7549 | 50 |   |  

**stops.txt**

stop_id | ... | zone_id
-- | -- | --
111 | ... | 
222 | ... | A
333 | ... | A
444 | ... | A
555 | ... | B
666 | ... | B

Given that `stop_id=111` is present only in `route_id=50` trips, you can determine that any itinerary that includes the stop will have a fare of `fare_id=7549` since it belongs to route_id=50. Giving the stop a `zone_id` would be superfluous because there is no fare rule that zone would belong to.

### Proposed solution

Is it possible to modify the conditional requirement, adding the following exception to the rule?

"**Required** if providing fare information using fare_rules.txt, *unless stop is present only in stop_times.txt records with a `trip_id` whose `route_id` is present in a fare_rules.txt record with only `fare_id` and `route_id` defined.*"

Ideally written in a less confusing way 😅

### Additional information

I have attached an example dataset you can run through the validator to see the `stop_without_zone_id` error in question.

[stop_without_zone_id_feedTEST.zip](https://github.com/google/transit/files/14091464/stop_without_zone_id_feedTEST.zip)

You will find several stops without `zone_id`s accounted for by route-based fares that trigger the error.

To my knowledge, feeds containing the `stop_without_zone_id` due to this reason have passed validation of consumers like Google.

Sergiodero:
[...]

PR #432, opened 2024-02-13
Clarify stops.zone_id conditional requirement for route-based fare_rules

westontrillium:
Related issue: https://github.com/google/transit/issues/429

This pull request modifies the conditional requirement for `stops.zone_id` to allow for the scenario in which the stop is only assigned to `trip_id`s whose `route_id` is present in a fare_rules.txt record with only `fare_id` and `route_id` defined (route-based fares).


Sergiodero:
Hi @westontrillium ,

Thanks for working on a proposal and opening this PR. I’m bringing over @bdferris-v2's [comment](https://github.com/google/transit/issues/429#issuecomment-1977278208) made on Issue #429, considering this point, perhaps it would good to could consider the option of making this field a recommended field instead of required? It could be recommended if zone-based info is provided (i.e. keeping the text you are proposing but replacing `required` with `recommended`). Alternatively it could just be optional all together, I wonder if anyone would have a strong preference between recommended and completely optional. 

In any case, I think in both options solve the problem of having stops without an assigned `zone_id` triggering an error by bringing more flexibility to the spec, making it a backwards compatible change for both cases.

Any thoughts?

westontrillium:
Good points from @bdferris-v2. I also can't think of a scenario where this field would be strictly *required*, especially universally.

In light of this, I think `optional` would actually make the most sense. As @bdferris-v2 stated, there are cases where an agency justifiably avoids defining `zone_id`, even if the `recommended` condition of "providing fare information using fare_rules.txt" is met. Adjusting the condition for the recommendation to just be when using fare zones would only be creating a truism: "This field is used for assigning stops to a fare zone... Recommended if assigning stops to a fare zone."

I'll wait to give others a chance to share their opinion, but I'm happy to make the eventual change to the PR to make this `optional`.

bdferris-v2:
+1 Google

westontrillium:
@e-lo Not universally. The zone that `origin_id/destination_id` refers to does not necessarily refer to all stops, so all stops do not need a `zone_id` if they are not part of a zone. And conditional requirements on the basis of an id being referenced somewhere is somewhat redundant. It is *always* the case that a defined foreign id field needs to reference a primary key that actually exists. If we based id requirements on this, we'd need to add "required if a [foreign id field] references this id" for every id in the spec. I think such a requirement is a validation rule more than anything.

e-lo:
I guess we need to differentiate between a field being _optional_ (existing as a column) vs a field being _nullable_ (have blank values).  If you have any `origin_id` / `destination_id` the **field** is required...but individiual records can always be _nullable_....right?

westontrillium:
> If you have any origin_id / destination_id the field is required

I don't think so. If you have any `origin_id/destination_id` **values**, there must be at least one stop with a `zone_id` **value** that matches. Again, this to me is a data validation rule more than a conditional requirement that should be codified into the spec. It's a given that this would be a requirement because foreign ids as a rule reference primary keys.

I don't know how the presence of the column—empty or otherwise—is relevant. Dependencies are always based on the values *within* the fields.

e-lo:
> It's a given that this would be a requirement because foreign ids as a rule reference primary keys

I don't dispute that, but the term **conditionally required** has been used throughout the spec to note this in many places and I'd just want to make sure we are being consistent.

westontrillium:
Hmm... I just went through all instances of "conditionally required" in the spec, and I didn't find a single case where "conditionally required" is invoked for the presence of a primary key if a foreign id field references it.

jfabi:
The @MBTA for a time published an internal-only version of the v1 _fare_rules.txt_ for our OpenTripPlanner instance. It appears we were leaving `zone_id` null for some stops where the fare rule could be determined entirely by `route_id`, so it seems fitting that we should support this clarification.

+1 from @MBTA

halbertram:
+1 from Ito

LuisLenta:
+1 from Ualabee

eliasmbd:
@westontrillium It looks like this PR has passed and is now adopted to GTFS. Congratulations on passing your first PR. 

+1 Google (@bdferris-v2)
+1 Spare Labs (@LeoFrachet) 
+1 MBTA (@jfabi)
+1 Ito (@halbertram)
+1 (@LuisLenta)

We have 5 votes for the change, 0 votes against the change, and 0 votes in abstention. 
The vote has passed and the change can be merged.

Issue #362, opened 2022-12-12
Static file version information is missing in the real-time data feed


Gerriko:
I was developing an application that analyses GTFS-R data feed and last week I encountered an error where my results differed from the operator's real-time app. At first I thought it was my code, but after many hours of testing I discovered that the public transit operator had just updated the GTFS static files (e.g. trips.txt, stop_times.txt, routes.txt etc.) for the winter schedule.

I went back to the GTFS-R specification and I could not find any information about how static file versions are provided within the real-time data feed. I would have assumed that this vital piece of information would be recorded in the header section at least to allow apps an automated means to check that it is using the correct file versions.

Is this correct?

If so, maybe this needs changing, or have others implemented other automated steps to let your app know that the static files it is using need updating. How does one know how to handle the transition or swap over day. Does one just have to assume that the version you've just downloaded is the same one the GTFS-R is using etc.

doconnoronca:
I have also had problems with some transit agencies where their published GTFS data does not match what is being used by their GTFS real time system. The LA Metro rail is in that state right now. Maybe adding a url and last update time to the GTFS real-time header would be useful.

evantrillium:
might be worth revisiting #255 
Additionally, Trillium produces linked_datasets.txt as a means of translating GTFS-RT feed info from the static side as well. Having a codified, complementary language between the datasets/feeds would be ideal.

skinkie:
This should be prioritized.

gcamp:
Simplest way I see would be to add `feed_version` from `feed_info.txt` in the GTFS-rt header

doconnoronca:
Adding this would help inform consumers when a new GTFS file is available and help coordinate between changes included in GTFS real-time and changes included in GTFS.

A GTFS update time (in POSIX time format) and a url to the feed being used would also be useful. Sometimes agencies have multiple GTFSs and it's not clear which one is to be used. San Francisco MUNI has this problem.

stevenmwhite:
I agree that having a reference (link) to the schedule feed in the RT feed would be incredible helpful.

We come across this issue all the time, when our customer agencies have a RT and Schedule feed from us, but also have other schedule feeds floating around out there (from their planning software, most commonly, but sometimes from other sources or old sources as well).

One of the biggest problems we see is that a consumer will look at one of those other static feeds, our RT feed, and then things don't match,

doconnoronca:
Here is a proposed addition to the FeedHeader message

```
[...]

PR #434, opened 2024-02-15
Add GTFS Feed Version to the GTFS real time FeedHeader

doconnoronca:
Based on the Issue [Static file version information is missing in the real-time data feed](https://github.com/google/transit/issues/362) opened by @Gerriko, I am proposing adding two new fields to the feed header to provide information on the GTFS file to be used with the GTFS real time data.

The feed_version would match the feed_version in the feed_info.txt file of the GTFS. This indicates which version of the GTFS file is currently being used by the real time data so the consumer knows exactly when to switch to a new file. It can also be used to identify when to check for a new GTFS file.

The GTFS_url would point to the url of the GTFS file. Sometimes agencies have multiple GTFSs and it's not clear which one is to be used. It provides the flexibility to allow the url to point to the upcoming GTFS file, but it is recommended to point to the one being used.

The next step would be for a producer to add support for this feature. This might be well suited for the @MBTA to implement it as they update their GTFS every few days, sometimes with last minute changes.

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -83,7 +83,7 @@ message FeedHeader {
   // it can be still provided.
   // The URL should be a fully qualified URL that includes http:// or https://, and any
   // special characters in the URL must be correctly escaped.
-  optional string GTFS_url = 5;
+  optional string gtfs_url = 5;
 
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime specification in order to add and evaluate new features and
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -73,6 +73,18 @@ message FeedHeader {
   // January 1st 1970 00:00:00 UTC).
   optional uint64 timestamp = 3;
 
+  // String that matches the feed_info.feed_version from the GTFS feed that the real
+  // time data is based on. Consumers can use this to identify which GTFS feed is
+  // currently active or when a new one is available to download.
+  optional string feed_version = 4;
+
+  // String containing an URL linking to the GTFS file. It should be the feed that is
+  // currently active. If a link to the upcoming GTFS file is the only one available
+  // it can be still provided.
+  // The URL should be a fully qualified URL that includes http:// or https://, and any
+  // special characters in the URL must be correctly escaped.
+  optional string GTFS_url = 5;
+
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime specification in order to add and evaluate new features and
   // modifications to the spec.
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -78,13 +78,6 @@ message FeedHeader {
   // currently active or when a new one is available to download.
   optional string feed_version = 4;
 
-  // String containing an URL linking to the GTFS file. It should be the feed that is
-  // currently active. If a link to the upcoming GTFS file is the only one available
-  // it can be still provided.
-  // The URL should be a fully qualified URL that includes http:// or https://, and any
-  // special characters in the URL must be correctly escaped.
-  optional string gtfs_url = 5;
-
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime specification in order to add and evaluate new features and
   // modifications to the spec.
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -97,7 +97,7 @@ Metadata about a feed, included in feed messages.
 | **gtfs_realtime_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Required | One | Version of the feed specification. The current version is 2.0. |
 | **incrementality** | [Incrementality](#enum-incrementality) | Required | One |
 | **timestamp** | [uint64](https://protobuf.dev/programming-guides/proto2/#scalar) | Required | One | This timestamp identifies the moment when the content of this feed has been created (in server time). In POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC). To avoid time skew between systems producing and consuming realtime information it is strongly advised to derive timestamp from a time server. It is completely acceptable to use Stratum 3 or even lower strata servers since time differences up to a couple of seconds are tolerable. |
-| **feed_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Optional | One | String that matches the feed_info.feed_version from the GTFS feed that the realtime data is based on. Consumers can use this to identify which GTFS feed is currently active or when a new one is available to download. |
+| **feed_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Optional | One | String that matches the `feed_info.feed_version` from the GTFS feed that the realtime data is based on. Consumers can use this to identify which GTFS feed is currently active or when a new one is available to download. |
 
 ## _enum_ Incrementality
 
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -97,6 +97,7 @@ Metadata about a feed, included in feed messages.
 | **gtfs_realtime_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Required | One | Version of the feed specification. The current version is 2.0. |
 | **incrementality** | [Incrementality](#enum-incrementality) | Required | One |
 | **timestamp** | [uint64](https://protobuf.dev/programming-guides/proto2/#scalar) | Required | One | This timestamp identifies the moment when the content of this feed has been created (in server time). In POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC). To avoid time skew between systems producing and consuming realtime information it is strongly advised to derive timestamp from a time server. It is completely acceptable to use Stratum 3 or even lower strata servers since time differences up to a couple of seconds are tolerable. |
+| **feed_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Optional | One | String that matches the feed_info.feed_version from the GTFS feed that the realtime data is based on. Consumers can use this to identify which GTFS feed is currently active or when a new one is available to download. |
 
 ## _enum_ Incrementality
 


gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -73,6 +73,18 @@ message FeedHeader {
...
+  // time data is based on. Consumers can use this to identify which GTFS feed is
+  // currently active or when a new one is available to download.
+  optional string feed_version = 4;
+
+  // String containing an URL linking to the GTFS file. It should be the feed that is
+  // currently active. If a link to the upcoming GTFS file is the only one available
+  // it can be still provided.
+  // The URL should be a fully qualified URL that includes http:// or https://, and any
+  // special characters in the URL must be correctly escaped.
+  optional string GTFS_url = 5;
```
AntoineAugusti, line 86:
```suggestion
  optional string gtfs_url = 5;
```

attributes should be lowercase I think

davidr1234:
+1 on this from SKI+ (SBB) in Switzerland our GTFS is updated frequently and such a linkage is highly needed.

skinkie:
One comment about the gtfs_url. Once you have a stable GTFS URL, so clients can automatically download (and use their ETag headers to see if something has changed!), having an absolute URL to that stable location would not really make sense. Sure, if you have different variants it would work, but you can not differentiate between them, unless feed_version is provided. I don't mind having a URL, but let's keep it as a reference field.

doconnoronca:
> +1 on this from SKI+ (SBB) in Switzerland our GTFS is updated frequently and such a linkage is highly needed.

Thanks for the vote, but the voting process hasn't started. What would really help is if you can implement this feature in your feed.

davidr1234:
@skinkie > One comment about the gtfs_url. Once you have a stable GTFS URL, so clients can automatically download (and use their ETag headers to see if something has changed!), having an absolute URL to that stable location would not really make sense. Sure, if you have different variants it would work, but you can not differentiate between them, unless feed_version is provided. I don't mind having a URL, but let's keep it as a reference field.

I am not sure if I correctly understand your comment. Do you suggest making gtfs_url "optional" as a field?

Generally, I agree that the feed_version is the more decisive field, while I do understand the potential benefit of having the URL.

skinkie:
> I am not sure if I correctly understand your comment. Do you suggest making gtfs_url "optional" as a field?

It is already optional ;-) But I mean this:

https://gtfs.ovapi.nl/nl/gtfs-nl.zip which points to https://gtfs.ovapi.nl/nl/NL-20240308.gtfs.zip

Historic variants here:
https://gtfs.ovapi.nl/nl/archive/NL-20240308.gtfs.zip
https://gtfs.ovapi.nl/nl/archive/NL-20240305.gtfs.zip

It is obvious that when gtfs_url would be https://gtfs.ovapi.nl/nl/NL-20240308.gtfs.zip it is pretty "unique" (under the constraint data is exported once per day). But as consumer you rather want to have https://gtfs.ovapi.nl/nl/gtfs-nl.zip as a stable url. Having a URL in GTFS-RT makes sense, but I would not want it to be the _unstable url_. I don't want to throw the baby out with the bathwater with this _significant_ improvement.

davidr1234:
> > I am not sure if I correctly understand your comment. Do you suggest making gtfs_url "optional" as a field?
> 
> It is already optional ;-) But I mean this:
> 
> https://gtfs.ovapi.nl/nl/gtfs-nl.zip which points to https://gtfs.ovapi.nl/nl/NL-20240308.gtfs.zip
> 
> Historic variants here: https://gtfs.ovapi.nl/nl/archive/NL-20240308.gtfs.zip https://gtfs.ovapi.nl/nl/archive/NL-20240305.gtfs.zip
> 
> It is obvious that when gtfs_url would be https://gtfs.ovapi.nl/nl/NL-20240308.gtfs.zip it is pretty "unique" (under the constraint data is exported once per day). But as consumer you rather want to have https://gtfs.ovapi.nl/nl/gtfs-nl.zip as a stable url. Having a URL in GTFS-RT makes sense, but I would not want it to be the _unstable url_. I don't want to throw the baby out with the bathwater with this _significant_ improvement.

True :-).

OK, this makes sense. I agree with this. I actually read it the other way around and that's why I couldn't follow your line of thought.

doconnoronca:
> > I am not sure if I correctly understand your comment. Do you suggest making gtfs_url "optional" as a field?
> 
> It is already optional ;-) But I mean this:
> 
> https://gtfs.ovapi.nl/nl/gtfs-nl.zip which points to https://gtfs.ovapi.nl/nl/NL-20240308.gtfs.zip
> 
> Historic variants here: https://gtfs.ovapi.nl/nl/archive/NL-20240308.gtfs.zip https://gtfs.ovapi.nl/nl/archive/NL-20240305.gtfs.zip
> 
> It is obvious that when gtfs_url would be https://gtfs.ovapi.nl/nl/NL-20240308.gtfs.zip it is pretty "unique" (under the constraint data is exported once per day). But as consumer you rather want to have https://gtfs.ovapi.nl/nl/gtfs-nl.zip as a stable url. Having a URL in GTFS-RT makes sense, but I would not want it to be the _unstable url_. I don't want to throw the baby out with the bathwater with this _significant_ improvement.

As long as the stable url switches over at the same time the it is activated in the real time data, I think either option would fully meet the criteria for gtfs_url.

Perhaps having a field for the current url and the upcoming url is an option, but that might be making a simple feature more complex then necessary.

skinkie:
> As long as the stable url switches over at the same time the it is activated in the real time data, I think either option would fully meet the criteria for gtfs_url.

Which doesn't happen. The stable URL is updated earlier. But it also does not mean that the next version is instantly incompatible.

doconnoronca:
I guess the question comes down to should the url point to the active GTFS or the upcoming GTFS.

The upcoming GTFS would work better for consumers who are capable of preloading the GTFS before goes into effect.

The current GTFS would be good for consumers who aren't capable of preloading the GTFS or are activating the agency for the first time.

davidr1234:
We will likely realize the feed_version in the next months. 

After some discussions we decided against an URL. The reason is a bit complicated to explain, but basically updating the RT-feed and the static file happen at different points in time, which is also related to the fact that the preparation and provision of the static file requires a varying amount of time and only after the switch has surely happened the GTFS-RT feed(version) is(would be) updated.

doconnoronca:
> We will likely realize the feed_version in the next months.
> 
> After some discussions we decided against an URL. The reason is a bit complicated to explain, but basically updating the RT-feed and the static file happen at different points in time, which is also related to the fact that the preparation and provision of the static file requires a varying amount of time and only after the switch has surely happened the GTFS-RT feed(version) is(would be) updated.

Including the URL in the GTFS-RT seems like it would be particularly useful for your agency. It took me quite a lot of searching to find your current [GTFS Data](https://opentransportdata.swiss/en/dataset/timetable-2024-gtfs2020) and it appears the url is different every version.

Could you update the url in the feed to point to the new GTFS data at the same time that the feed version switches over? That is the process I recommend in the documentation.

davidr1234:
> > We will likely realize the feed_version in the next months.
> > After some discussions we decided against an URL. The reason is a bit complicated to explain, but basically updating the RT-feed and the static file happen at different points in time, which is also related to the fact that the preparation and provision of the static file requires a varying amount of time and only after the switch has surely happened the GTFS-RT feed(version) is(would be) updated.
> 
> Including the URL in the GTFS-RT seems like it would be particularly useful for your agency. It took me quite a lot of searching to find your current [GTFS Data](https://opentransportdata.swiss/en/dataset/timetable-2024-gtfs2020) and it appears the url is different every version.
> 
> Could you update the url in the feed to point to the new GTFS data at the same time that the feed version switches over? That is the process I recommend in the documentation.

Hi @doconnoronca thank you for the feedback! If you don't mind, and not to spam this PR, may I kindly ask you to send us a small description (using https://opentransportdata.swiss/en/contact-2/) on how you searched for the dataset and what issue you had exactly with finding it (e.g., did you use the search function?, etc.)? Thanks!

doconnoronca:
The feed_version has been implemented in the Open data platform mobility Switzerland [trip updates feed](https://opentransportdata.swiss/en/cookbook/gtfs-rt/).

I am working on supporting some of the Swiss real time data in TransSee and using this feature to apply the GTFS files when the go into effect.

skinkie:
Thanks for the effort! We will certainly be a producer for this. I do see some issues with daily updated GTFS-data, but lets not forget how powerful this is to prevent false positives. 

+1 OpenGeo

davidr1234:
+1 SKI+ (Federal Office of Transportation in Switzerland)

gcamp:
+1 Transit 

Thank you for the work on this

leonardehrenfried:
+1 OpenTripPlanner

This a good way to know that you're definitely using the correct version.

laurentg:
+1 Mecatran (consumer & producer)

LuisLenta:
+1 from Ualabee! Good work!

sam-hickey-arcadis:
[reference.md](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-feedheader) needs to be updated still, but otherwise +1 for Arcadis.

skalexch (approved):

eliasmbd:
The vote has passed and will be merged into the specification. Thank you @doconnoronca @davidr1234 for your work on this. 

Here are the votes: 
+1 OpenGeo @skinkie 
+1 SKI+ (Federal Office of Transportation in Switzerland) @davidr1234 
+1 Transit @gcamp 
+1 OpenTripPlanner @leonardehrenfried 
+1 GMV @stevenmwhite 
+1 Mecatran @laurentg 
+1 Ualabee @LuisLenta 
+1 Arcadis @sam-hickey-arcadis 

8 total votes for the proposal!

gtfs-realtime/spec/en/reference.md
```diff
@@ -97,6 +97,7 @@ Metadata about a feed, included in feed messages.
 | **gtfs_realtime_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Required | One | Version of the feed specification. The current version is 2.0. |
 | **incrementality** | [Incrementality](#enum-incrementality) | Required | One |
 | **timestamp** | [uint64](https://protobuf.dev/programming-guides/proto2/#scalar) | Required | One | This timestamp identifies the moment when the content of this feed has been created (in server time). In POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC). To avoid time skew between systems producing and consuming realtime information it is strongly advised to derive timestamp from a time server. It is completely acceptable to use Stratum 3 or even lower strata servers since time differences up to a couple of seconds are tolerable. |
+| **feed_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Optional | One | String that matches the feed_info.feed_version from the GTFS feed that the realtime data is based on. Consumers can use this to identify which GTFS feed is currently active or when a new one is available to download. |
```
tzujenchanmbd, line 100:
```suggestion
| **feed_version** | [string](https://protobuf.dev/programming-guides/proto2/#scalar) | Optional | One | String that matches the `feed_info.feed_version` from the GTFS feed that the realtime data is based on. Consumers can use this to identify which GTFS feed is currently active or when a new one is available to download. |
```

tzujenchanmbd:
Before the vote was concluded, there was a last-minute change made to the reference.md - [2e286bd](https://github.com/google/transit/pull/434/commits/2e286bdf0cb22eee65c9f5edffba87beb10a215a). However, since the description remains the same, I propose that it is unnecessary to initiate another vote.

Tagging voters if they want to have another look at the change [2e286bd](https://github.com/google/transit/pull/434/commits/2e286bdf0cb22eee65c9f5edffba87beb10a215a) in reference.md  (take about 1 minute) - @skinkie @davidr1234 @gcamp @leonardehrenfried @stevenmwhite @laurentg @LuisLenta @sam-hickey-arcadis  

If there are no objections, we can proceed with merging next week.

davidr1234:
> Before the vote was concluded, there was a last-minute change made to the reference.md - [2e286bd](https://github.com/google/transit/pull/434/commits/2e286bdf0cb22eee65c9f5edffba87beb10a215a). However, since the description remains the same as proto, I propose that it is unnecessary to initiate another vote.
> 
> Tagging voters if they want to have another look at the change [2e286bd](https://github.com/google/transit/pull/434/commits/2e286bdf0cb22eee65c9f5edffba87beb10a215a) in reference.md (take about 1 minute) - @skinkie @davidr1234 @gcamp @leonardehrenfried @stevenmwhite @laurentg @LuisLenta @sam-hickey-arcadis
> 
> If there are no objections, we can proceed with merging next week.

No objections. I was not sure if you wanted an explicit response or if a "thumbs up" sufficed. Just in case, here explicitly :)

PR #449, opened 2024-05-02
Cleanup reference

polettif:
While trying to parse `reference.md` I noticed some inconsistencies in how names are used. This is just a "typo" PR which uses the same naming pattern for all fields. Please let me know if I'm missing some reason why "Required" was used as a column name instead of "Presence" for location_group and location_group_stops.

Also, the revision date (Nov 16, 2023) at the top is also not up-to-date but I refrained from changing that as well.

PR #453, opened 2024-05-14
Cleanup reference from PR#449

tzujenchanmbd:
Thanks @polettif for catching these inconsistencies! 

The changes are totally the same with [PR#449](https://github.com/google/transit/pull/449). We open this PR due to CLA problem. 

@polettif would you like to review if this covers all your suggested changes?

No voting needed.


polettif:
> @polettif would you like to review if this covers all your suggested changes?

Looks good to me, thank you. The date on top is still **Revised Nov 16, 2023**, just to let you know.

tzujenchanmbd:
Thanks! @polettif @Sergiodero 
I will update the date on the top and changes.md in another PR

jcpitre:
@tzujenchanmbd as long as you are modifying the page for revision history, there is a super minor point that you might want to correct also. 
The `Revision History` link currently points to `https://gtfs.org/schedule/revision-history` and for some reason this goes to the top of `https://gtfs.org/schedule/process/`

You should consider changing the link to `https://gtfs.org/schedule/process/#revision-history` so it jumps directly to the actual revision history.

Issue #288, opened 2021-10-19
Backwards compatibility and versioning


scmcca:
With new fields being added to the spec, we wanted to revisit the concept of backwards compatibility in GTFS. Currently, [CHANGES.md](https://github.com/google/transit/blob/master/gtfs/CHANGES.md) reads:
> When adding features to the specification, we want to avoid making changes that will make existing feeds invalid. We don't want to create more work for existing feed publishers until they want to add capabilities to their feeds. Also, whenever possible, we want existing parsers to be able to continue to read the older parts of newer feeds.

As implied, there are 2 parties to consider: data producers and data consumers.

For data producers, backwards-compatibility means that any changes to the spec do not invalidate or change the meaning of old datasets. That is to say, data producers won't need to change their datasets to make room for a new feature. This is seems to have been respected so far, and is favorable to maintain for reasons mentioned above.

For data consumers, backwards-compatibility means that any changes to the spec will only be built "on top" of current code. Meaning that if a data consumer choses not to implement a new spec feature, all other historical implementations remain valid with old and new feeds containing the addition (the consumer will simply be ignoring the feature, with no consequence on other rider information).

The problem comes in when adding features that are non-backwards compatible for data consumers. 

A recent example of this is the voting-in of #284. With the addition of more specific trip-to-trip and route-to-route transfers, consumers now have to update their code to look for trip-to-trip and route-to-route on top of the existing stop-to-stop transfers. If they don't, data consumers may be misrepresenting the specificity of transfers leading to errors in rider information.

More non-backwards compatible changes for data consumers are foreseeable. Otherwise, the development and improvement of GTFS would be severely limited.

Much of these issues stem from the fact that GTFS has no versioning (discussed previously in #215). While there is expressed interest from the community to version GTFS, there is also concern from other parties that this will complexify GTFS by encouraging breaking changes for all parties, leading to lower usage because of an additional barrier to entry/comprehension.

With the context that backwards compatibility means different things for different parties, I'm wondering if versioning is possible while maintaining that changes should be backwards-compatible for data producers. This would keep GTFS easy to use for data producers, while providing a mechanism for data consumers to track added features that break implementation.

Looking forward to thoughts on this from all parties. Thanks!

flocsy:
[...]

PR #460, opened 2024-05-22
Update requirement for feed_info.txt

isabelle-dr:
## Problem
When translations.txt was added into GTFS ([source](https://github.com/google/transit/pull/180)), the feed_info.txt file was changed from : 

> File: **Optional**

To:

> File: **Optional** (**Required** if `translations.txt` is provided)

Then, when the Recommended presence was added into GTFS ([source](https://github.com/google/transit/pull/386)), the feed_info.txt file was changed to:

> File: **Recommended** (**Required** if `translations.txt` is provided)

The issue is: the [Dataset Files](https://gtfs.org/schedule/reference/#dataset-files) table at the top of the spec was not updated with these changes; feed_info.txt is still described as **Optional**.

## Solution
In this PR, I modified the dataset file table with what I believe is the right requirement for feed_info.txt: 

> **Conditionally Required**:
> - **Required** if `translations.txt` is provided
> - **Recommended** otherwise

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -138,7 +138,7 @@ This specification defines the following files:
 |  [locations.geojson](#locationsgeojson)  | Optional | Zones for rider pickup or drop-off requests by on-demand services, represented as GeoJSON polygons. |
 |  [booking_rules.txt](#booking_rulestxt)  | Optional | Booking information for rider-requested services. |
 |  [translations.txt](#translationstxt)  | Optional | Translations of customer-facing dataset values. |
-|  [feed_info.txt](#feed_infotxt)  | Optional | Dataset metadata, including publisher, version, and expiration information. |
+|  [feed_info.txt](#feed_infotxt)  | **Conditionally Required** | Dataset metadata, including publisher, version, and expiration information.<br><br>Conditionally Required:<br>- **Required** if [translations.txt](#translationstxt) is provided.<br>- **Recommended** otherwise.|
 |  [attributions.txt](#attributionstxt)  | Optional | Dataset attributions. |
 
 ## File Requirements
@@ -805,7 +805,7 @@ If both referencing methods (`record_id`, `record_sub_id`) and `field_value` are
 
 ### feed_info.txt
 
-File: **Recommended** (**Required** if [translations.txt](#translations) is provided)
+File: **Conditionally Required**
 
 Primary key (none)
 
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -138,7 +138,7 @@ This specification defines the following files:
 |  [locations.geojson](#locationsgeojson)  | Optional | Zones for rider pickup or drop-off requests by on-demand services, represented as GeoJSON polygons. |
 |  [booking_rules.txt](#booking_rulestxt)  | Optional | Booking information for rider-requested services. |
 |  [translations.txt](#translationstxt)  | Optional | Translations of customer-facing dataset values. |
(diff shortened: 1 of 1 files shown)


skinkie:
I would rather have this upgraded to required, without the constraint.

isabelle-dr:
> I would rather have this upgraded to required, without the constraint.

I think you're not the only one, but making feed_info.txt _Required_ would make existing data invalid & be a breaking change, which goes against our [guiding principles](https://gtfs.org/schedule/process/#guiding-principles).
_Recommended_ is one solution for encouraging the industry to produce feed_info.txt while following these principles.

skinkie:
Upgrading anything from optional as required would be a breaking change. Hence, I would like to have a vote on the right solution. This still does not help us forward.

gcamp:
Unsure what you are arguing here @skinkie, that we break backward compatibility without a versioning scheme or other means to have a transition? Should consumer start rejecting dataset the minute the vote has passed?

skinkie:
> Unsure what you are arguing here @skinkie, that we break backward compatibility without a versioning scheme or other means to have a transition? Should consumer start rejecting dataset the minute the vote has passed?

As we both know nobody is actively rejecting data. Not even when their data is **not** in UTF-8, a requirement which is also part of this _great_ standard. Making an extra file mandatory as part of the specification does not break the _existing_ implementations only new implementations.

Now, the suggestion is that we could make feed_info.txt mandatory once when translations.txt was provided. Was that then a breaking change? Or can we introduce any new feature (without versioning) and then just say: "If you use this feature, then you must provide feed_info.txt."

doconnoronca:
I assume consumers would only reject files unsuitable for their particular system, which could be different from what that standards require.

isabelle-dr:
This PR doesn't _change_ the requirement level for `feed_info.txt`, it updates the table [at the top](https://gtfs.org/schedule/reference/#dataset-files) to match the requirement we have [at the file level](https://gtfs.org/schedule/reference/#feed_infotxt). 

The change to make feed_info.txt go from _Optional_ to _Recommended_ was voted on in https://github.com/google/transit/pull/386.

Discussing changing it from _Recommended_ to _Required_ would have to happen in another issue, and [we have an issue open](https://github.com/google/transit/issues/288) for discussing backward compatibility and versioning.

isabelle-dr:
Given that this PR is not changing the spec but rather adding an amendment that was missed in #386, I was thinking a review would be enough to merge this PR, and that a vote wouldn't be necessary.

That being said, I am happy to call a vote if others think it's needed.
Thoughts? Do we need a vote before merging this PR?

gcamp:
Even if it was a mistake from a previous PR, this is not from an editorial change. I would do a vote.

tzujenchanmbd (approved):

isabelle-dr:
This PR has been open for at least 7 calendar days. As per the [Spec Amendment Process](https://github.com/google/transit/blob/master/gtfs/CHANGES.md), I am opening a vote for updating the requirement of `feed_info.txt` in the top section, so that it matches what is described below.

Voting ends on 2024-07-17 at 23:59:59 UTC.

e-lo:
+1 UrbanLabs

skinkie:
+1 OpenGeo

gcamp:
+1 Transit

drewda:
+1 from @interline-io

evansiroky:
+1 Caltrans

isabelle-dr:
The voting period ended on 2024-07-17 at 23:59:59 UTC.

With 5 votes in favor and no votes against, the vote passes.
The votes came from:
- @e-lo (UrbanLabs)
- @skinkie (OpenGeo)
- @gcamp (Transit)
- @drewda (Interline)
- @evansiroky (Caltrans)

tzujenchanmbd:
This recommendation has been incorporated into the Canonical GTFS Validator V6.0.0. Please refer to the release page: https://github.com/MobilityData/gtfs-validator/releases/tag/v6.0.0.

PR #473, opened 2024-06-06
Fixing Stalebot 2.0

eliasmbd:
The stalebot was still closing issues and PRs automatically again. We hope this fix will remove automatic closure. We also updated the stale label to fit with our current label formatting type.


Sergiodero (approved):
LGTM!

PR #479, opened 2024-06-26
fix(gtfs/schedule/spec): add missing `Foreign` specifiers

firestack:
It seems like these new fields were added but were not consistent with the rest of the document, this fixes those type fields to match with other foreign ID fields.


eliasmbd:
@firestack Thanks for bringing this up! Contributions like these help us improve the overall quality of the spec. The community will get back to you and review your proposed changes soon.

tzujenchanmbd (approved):
Thanks again @firestack for catching this!

(Simple editorial changes - no need to vote)

PR #483, opened 2024-07-15
[GTFS Fares v2] Area Set matching predicates.

bdferris-v2:
Add support for matching "sets" of areas in `fare_leg_rules.txt` per [this proposal](https://docs.google.com/document/d/1aF7huR6f_Dw-6Cxnu6QFBKfjEoDmPJcwOLk2nv2KVHM/edit?usp=sharing).


gtfs/spec/en/reference.md
```diff
@@ -477,12 +479,16 @@ To process the cost of a leg:
 
 <br/>
 
+For area set predicates in `fare_leg_rules.txt` specified below, a leg “travels through an area” if any of the stops or parent stations (if the area is not defined at the stop-level) of the leg; including departure, arrival, and intermediate stops; belongs to the specified area as defined by [stop_areas.txt](#stopareastxt).
```
tzujenchanmbd, line 482: For the intermediate stops, it seems they must be stops defined in stop_times rather than "geographical pass-through" stops. Could the current description cause misunderstandings? Would it be clearer to state this explicitly?

gtfs/spec/en/reference.md
```diff
@@ -27,6 +27,7 @@ This document defines the format and structure of the files that comprise a GTFS
     -   [fare\_transfer\_rules.txt](#fare_transfer_rulestxt)
     -   [areas.txt](#areastxt)
     -   [stop_areas.txt](#stop_areastxt)
+    -   [area_sets.txt](#areasetstxt)
```
tzujenchanmbd, line 30:
```suggestion
    -   [area_sets.txt](#area_setstxt)
```

gtfs/spec/en/reference.md
```diff
@@ -126,6 +127,7 @@ This specification defines the following files:
 |  [fare_transfer_rules.txt](#fare_transfer_rulestxt)  | Optional | Fare rules for transfers between legs of travel.<br><br>Along with [fare_leg_rules.txt](#fare_leg_rulestxt), file [fare_transfer_rules.txt](#fare_transfer_rulestxt) provides a more detailed method for modeling fare structures. As such, the use of [fare_transfer_rules.txt](#fare_transfer_rulestxt) is entirely separate from files [fare_attributes.txt](#fare_attributestxt) and [fare_rules.txt](#fare_rulestxt). |
 |  [areas.txt](#areastxt) | Optional | Area grouping of locations. |
 |  [stop_areas.txt](#stop_areastxt) | Optional | Rules to assign stops to areas. |
+|  [area_sets.txt](#areasetstxt) | Optional | Collections of areas. |
```
tzujenchanmbd, line 130:
```suggestion
|  [area_sets.txt](#area_setstxt) | Optional | Collections of areas. |
```

gtfs/spec/en/reference.md
```diff
@@ -550,6 +556,19 @@ Assigns stops from [stops.txt](#stopstxt) to areas.
...
+
+File: **Optional**
+
+Primary key (`*`)
+
+Groups collections of areas into sets.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+| `area_set_id` | Unique ID | **Required** | Identifies an area set.  Must be unique in [area_sets.txt](#areasetstxt). |
```
tzujenchanmbd, line 569:
```suggestion
| `area_set_id` | Unique ID | **Required** | Identifies an area set.  Must be unique in [area_sets.txt](#area_setstxt). |
```

gtfs/spec/en/reference.md
bdferris-v2: I attempted to add explicit language.  Let me know what you think.

tzujenchanmbd: Looks good to me!

Just to confirm:

> and intermediate stops as referenced by the trip's stop-time entries; belongs to the specified area as defined by [stop_areas.txt]

This means that entries will be included regardless of their `pickup_type` and `drop_off_type` values (e.g., value=1 - No pickup/drop_off available). Is this correct?

bdferris-v2: Yes, I suppose that's correct.

I think we discussed this briefly in the last meeting, yes?  I think it's reasonable to make the default the simple case (passes through a stop).

tzujenchanmbd: Yes, got it. Just want to confirm if I correctly understand the semantics here (i.e. regardless of the pickup_type and drop_off_type values).

izzuddinzulkifli (approved):

tzujenchanmbd:
Here is a [dataset](https://docs.google.com/spreadsheets/d/1vnXMG8Tiizz1u5hm9VF_PQrpajHwEHqF4XtvLyxZYAM/edit?gid=2066352222#gid=2066352222) that has already been implemented by Ito, which is part of a private feed.

halbertram:
Whilst we are far along in this proposal, I recently had the following thought. It may be overly pedantic (and doesn't come with a proposed solution), but it might be worth discussing.

There seems to be a bit of a variation between how eg areas are formed from stops, and how area sets are formed from areas.

#### Stops and areas / Routes and networks

* *\<element\>s.txt* – element data
* *\<collection\>s.txt* – collection data such as name
* *\<element>_\<collection>s.txt* – insertion into collection

#### Areas and area sets

*area_sets.txt* defines the collections and the insertion. It's not clear where any fields that an area set might gain down the line would go.

If we view the collection as 'set', then it does partially follow the above scheme, but that will cause issues once we have network sets, and *sets.txt* would be ambiguous.

However having *area_sets.txt* and *area_area_sets.txt* doesn't seem sensible. We could find a name for an area set (in the way that we call a 'stop set' an area), possibly 'region' – but may be hard to find a word that would be appropriate in all uses of area sets. (Also would eventually have to find one for the corresponding network and trip_group cases). Or further area set level fields could go in *area_set_details.txt*.

bdferris-v2:
Did we talk about this briefly in the Fares v2 meeting last week?  Or did I just imagine talking about this? :)

Agreed that the naming is awkward here.  A couple options:

1) We just keep `area_sets.txt` as-is.  It's not clear what properties of an area set that we'd actually add (beyond a name?) and if we did need to add something after the fact, then I agree that `area_set_properties.txt` or `area_set_details.txt` could possibly work.
2) That said, I do acknowledge that this does deviate slightly from the pattern.  If we really wanted to stick to the pattern, then I guess I'd actually suggest `area_set_areas.txt` for the collection file to map an area to an area set.  A little clearer on what it means than `area_area_sets.txt` I think?  But if we went with this option, would we want to then make `area_sets.txt` the element file now (and require an entry for each area set), even if we don't have meaningful fields to associate with it?

gtfs/spec/en/reference.md
```diff
@@ -550,6 +556,19 @@ Assigns stops from [stops.txt](#stopstxt) to areas.
...
+
+File: **Optional**
+
+Primary key (`*`)
+
+Groups collections of areas into sets.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+| `area_set_id` | Unique ID | **Required** | Identifies an area set.  Must be unique in [area_sets.txt](#area_setstxt). |
```
tzujenchanmbd, line 569: I assume an area_set can be a collection that contains multiple areas. A unique ID doesn’t seem suitable for mapping area to area_set?

tzujenchanmbd:
Last week I mentioned that Hal raised this question, but it seems we didn’t dive into the discussion. :)

Currently, all similar files (e.g., `areas.txt`, `networks.txt`) only include a unique ID field and a name field. We could probably follow this pattern for now? and if additional properties are needed in the future, they can be added directly to the file.

Regarding naming, `area_set_areas.txt` does seem clearer than `area_area_sets.txt` to me. 

How does this look?

`area_set_areas.txt`: This file maps area to area set, with the following fields:
- `area_set_id` (foreign ID referencing area_sets.area_set_id, required)
- `area_id` (foreign ID referencing areas.area_id, required)

`area_sets.txt`: The file defining area sets, with the following fields:
- `area_set_id` (unique ID, required)
- `area_set_name` (optional)
Additional properties can be added to `area_sets.txt` if needed in the future.

halbertram:
I think that makes sense given the various options.

felixguendling:
I've implemented support for this feature in MOTIS (released in v2.0.71).
https://github.com/motis-project/nigiri/commit/d32d816575d24aaf88d4e1e58b5a0733588eeb6e

Maybe it's obvious for others but it could help to explicitly state how to deal with stops that are assigned to more than one area (stop_areas.txt has primary key `*` which means a stop can be part of several areas).

I implemented it like this:
- check if all areas from an area set are visited: go through all areas and search for at least one stop that is assigned to this area (including all areas a stop is assigned to)
- check if no extra areas are covered: go through each stop of the effective fare leg and check that all areas the stop is assigned to are part of the area set

tzujenchanmbd (changes requested):
Suggestions to add area_set_elements.txt, to align with the conclusions from comments above and previous working group discussions.

gtfs/spec/en/reference.md
```diff
@@ -477,12 +479,16 @@ To process the cost of a leg:
...
 | `to_area_id` | Foreign ID referencing `areas.area_id` | Optional | Identifies an arrival area.<br><br>If the `rule_priority` field does not exist AND there are no matching `fare_leg_rules.to_area_id` values to the `area_id` being filtered, empty `fare_leg_rules.to_area_id` will be matched by default.<br><br> An empty entry in `fare_leg_rules.to_area_id` corresponds to all areas defined in `areas.area_id` excluding the ones listed under `fare_leg_rules.to_area_id`<br><br>If the `rule_priority` field exists in the file, an empty `fare_leg_rules.to_area_id` indicates that the arrival area of the leg does not affect the matching of this rule. |
+| `contains_area_set_id` | Foreign ID referencing `area_sets.area_set_id` | Optional | Identifies an area set containing a collection of areas.<br><br> For this rule to match, a leg must travel through each and every area of the area set.  The leg may travel through additional areas not contained by the area set or through stops that are not assigned to an area at all.  If not specified, this predicate does not affect the matching of this rule. |
```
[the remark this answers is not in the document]
tzujenchanmbd: > should there also be rules of how contains_exactly_area_set_id and contains_area_set_id interact with from_area_id and to_area_id?

@skalexch I think the `from_area_id` / `to_area_id` case is different from `network_id`?
The `network_id` predicate matches a single network, which is logically at odds with a `network_set` that represents multiple networks within the same leg rule. In contrast, `from_area_id` and `to_area_id` describe the origin and destination areas of a leg. Producers may want to define different fares for different origin–destination pairs within the same area_set (e.g., A→D vs. A→C within area set ABCD), especially when `fare_leg_join_rule` is used. Allowing them to be used together should provide more flexibility?

> Also, if contains_area_set_id and contains_exactly_area_set_id both exist,

Agree that the interaction between `contains_area_set_id` and `contains_exactly_area_set_id` is similar to the network set case and that they should be mutually exclusive. I’ll update the suggestions accordingly.

gtfs/spec/en/reference.md
tzujenchanmbd:
```suggestion
Primary key (`network_id, from_area_id, to_area_id, from_timeframe_group_id, to_timeframe_group_id, contains_exactly_area_set_id, contains_area_set_id, fare_product_id`)
```

PR #485, opened 2024-07-23
Add that headsigns are recommended

### LINKED ISSUE #480: Make headsigns a recommended GTFS field
This issue picks up on the conversation in #461.

## Context
Certain Optional files/fields are dependent on the service (e. g. `timeframes.txt` is for modeling fares based on time of day), whereas other files/fields can and should always be added regardless of the type of service being modeled (e. g. `feed_info.txt`, ). We think there is value in calling out the latter explicitly using the terms associated with the Best Practices (_Recommended_ or _Should_), to promote higher-quality GTFS.

We looked at what are the most common optional features on the [Mobility Database](https://mobilitydatabase.org/): ~1500 GTFS Static datasets across 79 countries - note that the US accounts for approx half of this, so we also looked at non-US based data. These are the 6 top represented optional features, in both cases.

|feature                  |all data|non-US data|
|------------------------|--------|-----------|
|Shapes                  |84%  |72%     |
|Headsigns               |82%  |79%     |
|Route Colors            |73%  |57%     |
|Wheelchair Accessibility|62%  |41%     |
|Feed Information        |62%  |43%     |
|Location Types          |60%  |46%     |

There is currently [a PR open](https://github.com/google/transit/pull/470) for recommending Shapes in the spec.  

## Issue
This issue is for making [`trips.trip_headsign`](https://gtfs.org/schedule/reference/#tripstxt) recommended in the spec, given its use in production data across the globe.
Amongst the feeds that don't have headsigns, I didn't find obvious cases where it wouldn't be needed. Do we know of any?

## Solution proposed
Adding the _Recommended_ presence requirement for `trip_headsign` (see all presence requirements [here](https://gtfs.org/schedule/reference/#presence)).
This is considered a backwards-compatible change because it wouldn't change the validity of feeds that don't have headsigns. They would trigger a WARNING in the [canonical validator](https://gtfs-validator.mobilitydata.org/).

isabelle-dr:
This PR closes #480.

## Context and issue
Headsigns is one of the top-used optional field in GTFS. 82% of the datasets form the [Mobility Database](https://mobilitydatabase.org/) have it, which is more than route colors or feed_info.txt. 
It's a field that often can and should be included and we believe there is value in calling it out specifically using the Best Practices (_Recommended_ or _Should_), to promote higher-quality GTFS.

Discussions in the [GH issue](https://github.com/google/transit/issues/480) surfaced certain cases where there are no headsigns:
1. Loop routes
2. Flex datasets

## Solution proposed
Adding a statement that headsigns is recommended for all services with headsign text displayed on the vehicle which may be used to distinguish amongst trips in a route.

I acknowledge this isn't something we can easily validate automatically and trigger a WARNING for it in the [Canonical GTFS Schedule Validator](https://gtfs-validator.mobilitydata.org/). 
That being said, the spec contains instances of must/required and recommended/should that can't easily be checked by an automatic validator and would have to be checked by other means, and we think this is definitely better than nothing. For example:

- In [stop_times.stop_id](https://gtfs.org/schedule/reference/#stop_timestxt): _All stops serviced during a trip must have a record in [stop_times.txt](https://gtfs.org/schedule/reference/#stop_timestxt)._
- in [pathways.min_width](https://gtfs.org/schedule/reference/#pathwaystxt): _This field is recommended if the minimum width is less than 1 meter._
- In [agency.agency_email](https://gtfs.org/schedule/reference/#agencytxt): _This email address should be a direct contact point where transit riders can reach a customer service representative at the agency._

Recommendations don't invalidate a GTFS dataset, this change is considered backwards-compatible.

## Alternatives considered
In the GitHub issue, we've discussed adding a new `trips.has_headsigns` field that would have the _Recommended_ status and make `trips.trip_headsign` _Conditionally Required_ if `has_headsigns==1`, but then concluded the simple statement would lead to a good enough outcome with less complexity added to the spec.

Tagging folks involved in the GH issue @westontrillium @skinkie @antrim @evansiroky


evansiroky:
Sorry for not commenting on the issue, but I'm voting +0 on this. I would have preferred to have some language in the spec to enable a way to code the data in a way that explicitly says there are no headsigns on the trips.

isabelle-dr:
@evansiroky how about opening a separate GitHub issue for adding the functionality of indicating explicitly if a service has headings or not? 
This PR won’t conflict with adding that feature later on.

isabelle-dr:
I imagine the thumbs-up from @westontrillium @antrim @derhuerst means a general OK.
@evansiroky, I understand that you'd also want the spec to be able to explicitly indicate whether a service has headsigns 'in real life.' However, I believe this could be addressed in a separate GitHub issue and does not affect the relevance of the changes in this PR.

This PR has been open for at least 7 days, as per the [Spec Amendment Process](https://github.com/google/transit/blob/master/gtfs/CHANGES.md), I am opening a vote to make headsigns a recommended field in GTFS.
Voting ends on 2024-08-15 at 23:59:59 UTC.

gcamp:
+1 Transit

westontrillium:
+1 Trillium

kylerchin:
+1 Catenary Maps. Some agencies don't provide a head sign and it's confusing.

isabelle-dr:
The voting period ended on 2024-08-15 at 23:59:59 UTC.

With 3 votes in favor and no votes against, the vote passes.
The votes came from:
@gcamp (Transit App)
@westontrillium (Trillium)
@kylerchin (Catenary Maps) - 👋 welcome!

PR #142, opened 2019-02-15
Remove some trailing whitespace from files

paulswartz:
Expanded from #141.

PR #489, opened 2024-08-08
fix whitespace in .proto file

chadselph:
There was a couple of places the `}` indentation was obviously wrong on the site, I first noticed it on `SelectedTrips` (see screenshot). I found this to be confusing.

When I applied the protobuf formatter to the file it also fixed some misc. issues like tabs, spaces around ` = ` for the `deprecated` annotation, and trailing whitespace. I'm not as opinionated about these things, but I did find [precedent](https://github.com/google/transit/pull/142) for cleaning them up.

Screenshot 2024-08-08 at 10 48 49


tzujenchanmbd (approved):
Thanks @chadselph!

PR #491, opened 2024-08-16
Update revision history

isabelle-dr:
Update revision history with the latest changes


Sergiodero (approved):
Thanks @isabelle-dr ! the only change not being included from the last batch of changes in the past couple of months is an editorial one (PR #479 ), so I think we're good! LGTM!

PR #494, opened 2024-09-04
Adjust trip modifications

gcamp:
- Addition of `start_time` and `start_date` to `ModifiedTripSelector` for frequency service support
- Adding missing extensions fields for ModifiedTripSelector
- Explicitly allow use of `schedule_relationship` in `TripDescriptor` when using a `ModifiedTripSelector`.
- Addition of `dwell_time` to define time stopped at temporary stop. At the same time, explicitly support replacing existing stop by the same stop in `ReplacementStop` if some value are changing.
- Add missing documentation in reference.md (it was only specified in the .proto)


gcamp:
I did it again, sorry this was a draft and not ready for review.

PR #502, opened 2024-09-21
CHange to recommended to keep past events.

kylerchin:
Maybe this should be recommended? This is extremely useful because it's helpful to have updates on when the train entered or left the station. Maybe we should state that past events should be the actual time in which the vehicle entered or left the stop.

### DIFF
--- a/gtfs-realtime/spec/en/trip-updates.md
+++ b/gtfs-realtime/spec/en/trip-updates.md
@@ -10,7 +10,7 @@ If a vehicle is serving multiple trips within the same block (for more informati
 
 ## StopTimeUpdate
 
-A trip update consists of one or more updates to vehicle stop times, which are referred to as [StopTimeUpdates](reference.md#message-stoptimeupdate). These can be supplied for past and future stop times. You are allowed, but not required, to drop past stop times.  Producers should not drop a past `StopTimeUpdate` if it refers to a stop with a scheduled arrival time in the future for the given trip (i.e. the vehicle has passed the stop ahead of schedule), as otherwise it will be concluded that there is no update for this stop.
+A trip update consists of one or more updates to vehicle stop times, which are referred to as [StopTimeUpdates](reference.md#message-stoptimeupdate). These can be supplied for past and future stop times. Producers are recommended but not required to keep past events. Producers should not drop a past `StopTimeUpdate` if it refers to a stop with a scheduled arrival time in the future for the given trip (i.e. the vehicle has passed the stop ahead of schedule), as otherwise it will be concluded that there is no update for this stop.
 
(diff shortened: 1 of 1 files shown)


kylerchin:
CLA issue should be fixed now! : 👍

skinkie:
-1 OpenGeo
It uses the wrong wording for what you want. Your intention is to receive information, thats fine. Your implementation an client is to fetch GTFS-RT at a regular interval. It may be that your update interval does not match the producers interval. What you want to describe is some sort of time to live for StopTimeEvents. Recommending to keep **all** historic StopTimeEvents bloats the GTFS tripUpdates distribution for no good reason, therefore it should not be recommended in the way you do.

kylerchin:
@skinkie There is good reason IMO to keep previous times, such as if a consumer server cold starts, or if a fault causes a significant loss in the cache data.

skinkie:
That is not a price to pay for the producer or all other consumers.

All these kinds of things could be fixed by a final implementation for Incrementality=DIFFERENTIAL, that would allow the majority of information be provided in tiny parts and your use case in the FULL_DATASET. Personally I think a _HISTORY_ would be more appropriate.

miklcct:
I would say that past events should be kept as long as the trip hasn't been completed yet, otherwise consumers may have trouble working on delay propagation.

gcamp:
Can you explain how delay propagation is hard to do without the full trip?

miklcct:
If a bus is running early, well ahead of its scheduled time, lacking previous information may well cause the system into impossible state like time running backwards.

For example, OpenTripPlanner allows you to [configure an option when such things do happen](https://docs.opentripplanner.org/en/dev-2.x/UpdaterConfig/#u__5__backwardsDelayPropagationType).

gcamp:
That's already in the spec : 

> Producers should not drop a past `StopTimeUpdate` if it refers to a stop with a scheduled arrival time in the future for the given trip (i.e. the vehicle has passed the stop ahead of schedule), as otherwise it will be concluded that there is no update for this stop.

What is proposed here is that producers should keep all past items regardless of if they are running early or late.

evansiroky:
I agree that GTFS-RT should be reserved for realtime events only and not need to include historical data. A better way to communicate historical data would be via another standard, such as [TIDES](https://tides-transit.org/main/).

miklcct:
I want to follow up this proposal. We are now planning to have a clean up of OpenTripPlanner real-time data, and missing data are now causing us problems how to fill in the remaining times to satisfy invariants (times must be increasing along the trip).

We should write the best practice of always producing all arrivals / departures for all stops for a trip into the recommendation document, in order to ensure that consumers won't add guesswork.

PR #230, opened 2020-06-26
Proposal: Label GTFS-RT ADDED trips as not fully specified

barbeau:
Given existing discussions around how to add or modify trips in GTFS-RT [(1)](https://github.com/google/transit/issues/106) [(2)](https://github.com/google/transit/pull/221) [(3)](https://github.com/google/transit/pull/219), I hope one thing we can all agree upon is that we don't currently agree upon how `ADDED` trips are produced or consumed.

This proposal puts a label on the `ADDED` enumeration to make this lack of agreement clear to newcomers to the spec while we discuss how these concepts should be represented, similar to how the [`DIFFERENTIAL` feature](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-incrementality) is currently labeled in the GTFS-RT spec.

cc @paulswartz @gcamp @skinkie @Bertware @tleboulenge @darylweinberg @juanborre @hbruch @nathan-reynolds @lauramatson

Posted to the Google Group at https://groups.google.com/forum/#!topic/gtfs-realtime/p8pXNzGOafQ.

PR #503, opened 2024-09-25
Specify the behaviour of StopTimeEvent on different schedule relationships.

### LINKED ISSUE #490: Allow specifying both the scheduled time and delay in StopTimeEvent for added and duplicated trips.
### Describe the problem

The GTFS-RT specification for `StopTimeEvent` specifies that:

> Timing information for a single predicted event (either arrival or departure). Timing consists of delay and/or estimated time, and uncertainty.
> 
> * `delay` should be used when the prediction is given relative to some existing schedule in GTFS.
> * `time` should be given whether there is a predicted schedule or not. If both time and delay are specified, time will take precedence (although normally, time, if given for a scheduled trip, should be equal to scheduled time in GTFS + delay).

However, for an `ADDED` or `DUPLICATED` trip, there is currently no means to specify the scheduled time of the added or duplicate trip (for example, if the duplicated trip is scheduled to run slower than the original).

### Use cases

* A train company wants to add a new, ad hoc trip, not on any existing trip patterns, which adheres to a schedule, for example, an event special.
* A bus company wants to run a duplicate service timed 5 minutes before the original, but because of the slower boarding / alighting speed of the relief vehicle, it is planned to run slower than the original and eventually arrives the destination at the same time as the original.

### Proposed solution

Change the definition of the message `StopTimeEvent` in GTFS-RT to be:

Timing information for a single predicted event (either arrival or departure). Timing consists of delay and/or estimated time, and uncertainty.

* delay should be used when the prediction is given relative to some existing schedule in GTFS.
* time should be given whether there is a predicted schedule or not. If both time and delay are specified
** For scheduled trips, time will take precedence (although normally, time, if given for a scheduled trip, should be equal to scheduled time in GTFS + delay).
** For added and duplicated trips, if a delay is also specified, it represents the actual time for the new trip for which the scheduled time can be calculated
* Uncertainty applies equally to both time and delay. The uncertainty roughly specifies the expected error in true delay (but note, we don't yet define its precise statistical meaning). It's possible for the uncertainty to be 0, for example for trains that are driven under computer timing control.

|Field Name	|Type	|Required	|Cardinality|	Description|
|-----|----|----|-----|----|
|delay	|[int32](https://protobuf.dev/programming-guides/proto2/#scalar)|	Conditionally required	|One	|Delay (in seconds) can be positive (meaning that the vehicle is late) or negative (meaning that the vehicle is ahead of schedule). Delay of 0 means that the vehicle is exactly on time. <br>If `TripDescriptor.schedule_relationship` is `SCHEDULED`, the delay is measured against the GTFS static schedule.<br> If `TripDescriptor.schedule_relationship` is either `ADDED` or `DUPLICATED`, the delay is included in the actual `time` below, which must be provided. <br>If `TripDescriptor.schedule_relationship` is `UNSCHEDULED`, this field must be empty.<br>Either delay or time must be provided within a StopTimeEvent - both fields cannot be empty.|
|time	|[int64](https://protobuf.dev/programming-guides/proto2/#scalar)|	Conditionally required	|One	|Event as absolute time. In POSIX time (i.e., number of seconds since January 1st 1970 00:00:00 UTC). <br> If `TripDescriptor.schedule_relationship` is either `ADDED` or `DUPLICATED`, and both delay and time are specified, this is the actual time where the scheduled time can be calculated by subtracting the `delay`. <br>If `TripDescriptor.schedule_relationship` is `SCHEDULED`, this value overrides the time calculated by `delay`.<br>Either delay or time must be provided within a StopTimeEvent - both fields cannot be empty.|
|uncertainty	|[int32](https://protobuf.dev/programming-guides/proto2/#scalar)	|Optional	|One	|If uncertainty is omitted, it is interpreted as unknown. To specify a completely certain prediction, set its uncertainty to 0.|

### Additional information

_No response_

miklcct:
Fixes #490

### DIFF
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -170,8 +170,19 @@ Note that the update can describe a trip that has already completed.To this end,
 
 Timing information for a single predicted event (either arrival or departure). Timing consists of delay and/or estimated time, and uncertainty.
 
-*   delay should be used when the prediction is given relative to some existing schedule in GTFS.
-*   time should be given whether there is a predicted schedule or not. If both time and delay are specified, time will take precedence (although normally, time, if given for a scheduled trip, should be equal to scheduled time in GTFS + delay).
+*   `delay` should be used when the prediction is given relative to some existing schedule in GTFS.
+*   `time` should be given whether there is a predicted schedule or not.
+
+The behaviour of `delay` and `time` depends on the schedule relationship:
+
+| `TripUpdate.schedule_relationship` | `StopTimeUpdate.schedule_relationship` | Behaviour                                                                                                                                                                  |
+|------------------------------------|----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `SCHEDULED`                        | `SCHEDULED`                            | `time` will take precedence (although normally, `time`, if given for a scheduled trip, should be equal to scheduled time in GTFS + `delay`)                                |
+| `SCHEDULED`                        | `SKIPPED`                              | the vehicle was scheduled to stop, but will not do so now. `time` and `delay` may be specified but not compulsory.                                                         |
+| `ADDED` or `DUPLICATED`            | `SCHEDULED`                            | `time` must be specified for the estimated time. If `delay` is also specified, as the trip is not in the static GTFS, the scheduled time is calculated by `time` - `delay` |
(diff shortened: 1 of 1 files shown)


gcamp:
I'm not a fan of fields changing meaning depending on the context, if we would add a way to specify schedule time I would prefer having a separate field. 

In addition, I'm sure you've seen that the usage of ADDED [has been discouraged](https://github.com/google/transit/pull/230). I would think defining clearly ADDED would be useful before adding more capabilities to it.

miklcct:
I am going to work on another PR for specifying ADDED (and REPLACEMENT) types.

Issue #526, opened 2024-12-23
Add a provision for event-based trips (to support modeling of trips departing "x minutes following an event")


jeffkessler-keolis:
### Describe the problem

Many carriers operate special event service to stadiums and other event venues.

This special event service is sometimes the ONLY service available to the stadium, as is most common in the North American Rail Market during off-peak / weekend hours.

While these special-event trips typically operate on a posted schedule en-route _to_ the event, **_return_ trips typically operate a less-precise service, departing "x minutes following the conclusion of the event."**

This phenomenon is not something that can be modeled natively in GTFS, leading to three possible approaches:

- Omit the trips entirely in the GTFS Schedule, hoping that customers assume there will be some return service based on the special-event service to the event, but risking confusion around the return.

- Schedule the trips in the GTFS Schedule based on an approximate event end time, risking confusion that customers could need to depart prior to the conclusion of the event, or implying the operator should be held to an inflexible schedule.

- Schedule the trips in the GTFS Schedule based on the latest possible end time, dissuading some customers from using the service based upon an artificially-lengthy journey time, or leaving customers to believe they could stay later than possible in the event of extended event durations.

In all cases, the partial lack of data leads to confusion for riders.

### Use cases

- Show customers in static information that return service is available from an event location, even when the specific departure time is variable.
- Provide a mechanism for customers to know the latest possible time at which they must leave a special event, where applicable.
- Avoid simply displaying the trip as a static scheduled interval, without the producer also providing detail surrounding the trip's variability.

### Proposed solution

1. Trips should be scheduled in GTFS based upon the most-likely **departure** time, and standardized `stop_time` entries would be based upon the most-likely departure time.

2. A new file `event_based_trips.txt` would serve to indicate to consumers and customers that applicable trips are post-event extras, operating based on an expected **event end** time.

This approach ensures backwards-compatibility in that consumers could still display an approximate trip time based upon the most likely time of operation without modification, while more robust consumers could contextualize the trip as a post-event service, detailing the service span, post-event departure duration, onward connections, etc.

This framework also supports a future basis on which an operator could define a specific event end time in real-time, and have predictions for each trip based upon offsets adjusted from the actual event end time.

*N.B. Most North American consumers would NOT publish [1] without also being able to produce [2] to avoid indicating a precise departure time for the event without ALSO having a structured means to represent the special-event structure.*

### Additional information

This scheduling practice is employed by:

[...]

PR #527, opened 2024-12-23
Addition of event_based_trips.txt to specification

jeffkessler-keolis:
Draft PR to address #526

## Example

Transit Agency X is running a special event train service to a football game starting at 7:00 PM with an estimated end time of 10:00 PM. There are two trains scheduled to leave the football stadium, both 30 minutes and 60 minutes after the game ends.

If the event ends early, the agency can accommodate passengers immediately, meaning that the first train can leave as early as 7:30 PM.

Due to federal rest restrictions, the final train must leave by midnight, and the prior train must leave by 11:45 PM. The latest the game can end while supporting the scheduled service pattern would be 11:00 PM for the second train, and 11:15 PM for the first train.

The first scheduled departure trip has trip_id `trip_approx_1` and the second trip has ID `trip_approx_2`. Only passengers with a special event ticket are able to ride.

**stop_times.txt**

```csv
trip_id,arrival_time,departure_time,stop_id,stop_sequence,pickup_type,drop_off_type,timepoint
trip_approx_1,22:30:00,22:30:00,stadium,10,0,1,0
trip_approx_1,23:00:00,23:00:00,suburb,20,1,0,0
trip_approx_1,23:15:00,23:15:00,downtown,30,1,0,0
trip_approx_2,23:00:00,23:00:00,stadium,10,0,1,0
trip_approx_2,23:30:00,23:30:00,suburb,20,1,0,0
trip_approx_2,23:45:00,23:45:00,downtown,30,1,0,0
```

**event_based_trips.txt**

```csv
trip_id,event_end_approximate,event_end_earliest,event_end_latest,event_fares
trip_approx_1,22:00:00,19:00:00,23:00:00,1
trip_approx_2,22:00:00,19:00:00,23:00:00,1
```

## Notes

- `stop_times.txt` shows the **approximate** stop times for each trip, assuming that the event ends at its expected time.
	- By using `stop_times.txt` and the `event_end_earliest` and `event_end_latest` fields, one can determine the earliest and latest stop times for each trip.
	- Using the `event_end_apprximate` field, one can determine how long after the event each trip departs.
- The `event_end_latest` field may not be the same between trips. For example, NJ Transit's River Line must do a short trip during late-night service due to track restrictions. Thus, a hypothetical event must have an earlier end time to support trips that cover the entire line.

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -32,6 +32,7 @@ This document defines the format and structure of the files that comprise a GTFS
     -   [route_networks.txt](#route_networkstxt)
     -   [shapes.txt](#shapestxt)
     -   [frequencies.txt](#frequenciestxt)
+    -   [event_based_trips.txt](#event_based_tripstxt)
     -   [transfers.txt](#transferstxt)
     -   [pathways.txt](#pathwaystxt)
     -   [levels.txt](#levelstxt)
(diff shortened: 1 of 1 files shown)


miklcct:
How about `runs as required` trips? For example, after an event trip ends, there are two planned buses to depart, but the second will only run when the first is fully loaded.

Issue #462, opened 2024-05-22
Global trip id


davidr1234:
### Describe the problem

Hi,

**Context:** In Switzerland we've introduced the Swiss Journey ID (SJYID - documentaiton only in DE/FR/IT, sorry: https://www.xn--v-info-vxa.ch/de/datenmanagement/sid4pt-swiss-id-public-transport/swiss-journey-identification-sjyid).
It can be considered a "global" (**day-long valid**) trip identifier. One of the applications of this identifier is to allow mapping travel information among different mobility-data standards.

**What we do:** We will include the SJYID in our GTFS Schedule exports and intend to do the same for GTFS realtime. 

**What we realized:** Discussions with other European public transport providers have shown that we may not be the only ones who need such a construct. Particularly, to ensure compatibility with other exchange formats.

**What we suggest:** Although this concept has apparently been explicitly excluded (see https://gtfs.org/realtime/reference/#message-tripdescriptor & screenshot), we suggest to re-visit this and consider including a new field "trip_global_id" that allows to combine/aggregate multiple trips.

![image](https://github.com/google/transit/assets/32004282/c2b883e5-551b-402f-b8cc-63841d0f0338)

Thank you for your input.

### Use cases

Combining GTFS realtime data with NeTEx, HRDF, etc.

### Proposed solution

Amend GTFS Schedule's trips.txt with a field "trip_global_id"
Amend GTFS realtime's TripDescriptor to include "trip_global_id" referencing the scheduled "trip_global_id"

skinkie:
Why not just fill in trip_id with the contents of trip_global_id?

eliasmbd:
@davidr1234 Thanks for bringing this up here. I am sure many contributors might be facing the same problem and could provide some insight. Also, thank you for posting your first issue on google/transit! Your contributions are always appreciated!  :rocket:

davidr1234:
> @davidr1234 Thanks for bringing this up here. I am sure many contributors might be facing the same problem and could provide some insight. Also, thank you for posting your first issue on google/transit! Your contributions are always appreciated! 🚀

@eliasmbd thank you, too.

> Why not just fill in trip_id with the contents of trip_global_id?

@skinkie that wouldn't work with the trip_global_id being valid throughout a day (for us at least, maybe in other countries the time-frame varies).

skinkie:
@davidr1234 do you mean that global id is something like XX:20240524:1 or that you can have multiple trips on 20240524, that have XX:1?

[...]

PR #504, opened 2024-09-25
Deprecate TripUpdate.schedule_relationship = ADDED, add TripUpdate.schedule_relationship = NEW / REPLACEMENT to specify new / replaced trips which do not run on a schedule from the GTFS static.

miklcct:
The use of `TripUpdate.schedule_relationship = ADDED` [was unspecified and different producers / consumers used it in different ways](https://github.com/google/transit/issues/106). For example, it is sometimes used to specify additional departures on an existing route, but it is also used to specify departures which can't be matched to any existing trips.

This PR attempts to fully deprecate ADDED, and introduce NEW and REPLACEMENT, based on the implementation of OpenTripPlanner which specifies the whole journey to be added or replaced. Additional fields, such as headsigns, and pickup / drop-off types, are introduced as required to support the full specification of completely new trips.

## `NEW`
In this proposal, `TripUpdate.schedule_relationship = NEW` should be used to add trips which do not duplicate an existing trip. Such trips are considered to be unrelated to any existing trips in the GTFS Static and can serve an arbitrary pattern, including completely new patterns not found in the GTFS Static.

A typical use case is for relief trips for extra demand, typically after big events.

NEW trips are intended as a migration target for existing feeds which use ADDED to specify new trips unrelated to the static GTFS, including OpenTripPlanner. `trip_id` in the `TripDescriptor` for new trips must be completely new (not found in GTFS static) and unique, and a `start_date` should also be specified as well (I am not using the word "must" here because it is permitted not to specify start_date to match scheduled trips, in this case the trip is assumed to run today). 

The whole journey of the added trip must be specified, in stop order, as `StopTimeUpdate`s inside the `TripUpdate` without any omission. Fields are added to `TripProperties` and `StopTimeProperties` for esssential information such as names, headsigns, pickup / drop off types.

## `REPLACEMENT`
I propose to un-deprecate `TripUpdate.schedule_relationship = REPLACEMENT` as well. It works in the same way as `NEW`, apart from that the `TripDescriptor` must match one instance of a scheduled trip (like other values of `TripUpdate.schedule_relationship`), and that instance is replaced with the complete replacement trip specified in form of `StopTimeUpdate`s like an added trip. The original stop times in the GTFS static are not considered by the replacement trip in any form to avoid confusion. The replacement trip can serve an arbitrary pattern with an arbitrary schedule, the only expectation is that the passenger should associate the replacement trip to actually be a replacement of the original trip.

[...]

PR #534, opened 2025-01-30
Enhancing GTFS Schedule and Realtime with original_trip_id

davidr1234:
This pull request is related to issue https://github.com/google/transit/issues/462

**Context:** In Switzerland we've introduced the Swiss Journey ID (documentation only in DE/FR/IT: https://www.oev-info.ch/de/datenmanagement/sid4pt-swiss-id-public-transport/swiss-journey-identification-sjyid).
This ID is valid for one operating day and across different days of a scheduled year. It therefore maps to one or more trip_ids.

**Proposal:** Based on the suggestion by @miklcct (in the referenced issue) we propose to use the original_trip_id (as defined in https://developers.google.com/transit/gtfs/reference?hl=en) in GTFS Schedule and GTFS Realtime to represent constructs such as our SJYID. With this it is possible to combine trips from GTFS Schedule and GTFS Realtime with other standards such as SIRI or NeTEx, which have a similar concept. 

**Implementation:** Since the 12.12.2024 we offer the original_trip_id (filled with our SJYID) as part of GTFS Schedule (doc: https://opentransportdata.swiss/en/cookbook/gtfs/#tripstxt) and GTFS Realtime (doc: https://opentransportdata.swiss/en/cookbook/gtfs-rt/#Trip_updates). In one case our consumers use the original_trip_id in GTFS Realtime to match with timetable data in the (proprietary) HRDF format.

**Generalizability:** Based on early discussions with other public transport providers, we think this enhancement can benefit many other producers and consumers and increase the inter-operability of GTFS with other standards.


miklcct:
You need to update the .proto file for the real time field. Please use a larger number and avoid the field numbers I am proposing in #504 , as I intend to produce it as soon as I can for integration with other systems such as Darwin (my static GTFS has this field already).

davidr1234:
> You need to update the .proto file for the real time field. Please use a larger number and avoid the field numbers I am proposing in #504 , as I intend to produce it as soon as I can for integration with other systems such as Darwin (my static GTFS has this field already).

Thank you @miklcct, I missed that. I looked at https://github.com/google/transit/pull/504 and it seems that 8 is available as field number for the original_trip_id. This is also the number we currently use in our implementation. Would that interfere with your work?

miklcct:
That's great, so I can continue to use 7 for trip_headsign

Are you using 7 for something else?

davidr1234:
> That's great, so I can continue to use 5 and 6 for trip_headsign and trip_short_name respectively.
> 
> Are you using 5 or 6 for something else?

No, we only add `optional string original_trip_id = 8;`

I'll push that now.

skinkie:
Critical question: why is this important for _passenger information_. I read the interoperability argument, but not how that is used.

miklcct:
> Critical question: why is this important for _passenger information_. I read the interoperability argument, but not how that is used.

It allows consumers to match the GTFS data with external data from other sources.

miklcct:
> Concrete examples please. The GTFS ecosystem is that GTFS can be matched with GTFS-RT. In what situation is it valuable to have **other** trip identifiers. I can think of some, but those should be written down.

I am using the field to match upstream data from systems of Network Rail, where their IDs are only unique on a single day.

eliasmbd:
> Critical question: why is this important for _passenger information_. I read the interoperability argument, but not how that is used.

Are you asking why this matters to the rider?

skinkie:
> Are you asking why this matters to the rider?

Exactly.

davidr1234:
> Concrete examples please. The GTFS ecosystem is that GTFS can be matched with GTFS-RT. In what situation is it valuable to have **other** trip identifiers. I can think of some, but those should be written down.

We give an example in the introductory text under "Implementation": "In one case our consumers use the original_trip_id in GTFS Realtime to match with timetable data in the (proprietary) HRDF format."

Our consumers use both GTFS and HRDF. However, HRDF is able to better reflect certain services in Switzerland due to its more comprehensive and complex data structure, such as for linked trips. 

This approach allows to maintain the efficient structure of GTFS (which is one of the reasons for its popularity with our consumers), while providing the full bandwidth of our available customer information by combining it with our other formats.

> > Concrete examples please. The GTFS ecosystem is that GTFS can be matched with GTFS-RT. In what situation is it valuable to have **other** trip identifiers. I can think of some, but those should be written down.
> 
> I am using the field to match upstream data from systems of Network Rail, where their IDs are only unique on a single day.

I would also like to point out this statement by @miklcct, which would be another motivation for this field. This is also true for our SJYID.

skinkie:
> To summarise: I'm a bit sceptical of changing the GTFS specification to accommodate non-GTFS workflows.

I have the same skeptism. But there is fundamental thing both GTFS and NeTEx are overlooking. For every time in GTFS or NeTEx a property changes, a new identifier must be introduced. Now you could argue "this makes a lot of sense" and for some organisations (and even implementers) it does not. They are instead manageding these **validities** of properties at different levels. That is why virtually _everything_ is in conflict with each other once HRDF is mentioned. This is the absolute root cause.

miklcct:
The very reason why this field is needed is that a consumer with local knowledge can use it to reference other passenger-facing systems outside the GTFS world using the `original_trip_id` provided, when other passenger-facing systems use an ID which is not the public code.

ue71603:
It is not just HRDF. As Stefan mentions there is a core problem in NeTEx and GTFS: uniqueness of the trip in the file. Traditional public transport has uniqueness of the ServiceJourney per operating day.  Even when the trip is slightly different e.g. for Wednesday, it still is the trip that starts at 08:01 to Zürich. This can be expressed by the global id.  We have to split the trip into different ones for GTFS, but for many other use cases (and systems). It is still useful/crucial to know that this is indeed the one.

ue71603:
So what the PR really does is to accomodate both "worlds".

leonardehrenfried:
Can you not do the same "split" when converting to GTFS-RT?

skinkie:
From a GTFS standpoint it can also be interesting. For example aggregating all the "_truly_ unique trips". It is a very specific use case, therefore I hope some more examples can be provided.

ue71603:
Non sequitur @leonardehrenfried: With your argumentation we could say: Why should we produce GTFS at all. HRDF and VDV 454 contain all necessary information. It can't be the responsibility of Switzerland to facilitite work for others? 


We believe this PR is a simple way to simplify interactions between the different formats. In the ideal world one can use on realtime stream and a time table stream of your choice.

leonardehrenfried:
I'll grant you that: it's not a complicated proposal.

miklcct:
> It is not just HRDF. As Stefan mentions there is a core problem in NeTEx and GTFS: uniqueness of the trip in the file. Traditional public transport has uniqueness of the ServiceJourney per operating day. Even when the trip is slightly different e.g. for Wednesday, it still is the trip that starts at 08:01 to Zürich. This can be expressed by the global id. We have to split the trip into different ones for GTFS, but for many other use cases (and systems). It is still useful/crucial to know that this is indeed the one.

Also there is currently no facility in GTFS (unlike NeTEx) to specify that the modified Wednesday 08:01 is the same trip as the Wednesday 08:01, that if the timetable has been modified from the base timetable it is not currently possible for a client with saved pre-planned journey to know that the timetable has been changed. They will just fail to find the trip in the updated timetable.

skinkie:
> Also there is currently no facility in GTFS (unlike NeTEx) to specify that the modified Wednesday 08:01 is the same trip as the Wednesday 08:01, that if the timetable has been modified from the base timetable it is not currently possible for a client with saved pre-planned journey to know that the timetable has been changed. They will just fail to find the trip in the updated timetable.

I honestly think NeTEx did not standardise "global trip id" either. And yes, there is PrivateCode but that is not the "concept" that we mean here?

davidr1234:
> Can you not do the same "split" when converting to GTFS-RT?
> 
> BTW, I don't doubt that it would be useful to your consumers but I doubt that it's GTFS's responsibility to deal with other representations of public transport.

Indeed, but this is not a one-way street. Would you argument that the other standards (such as NeTEx/SIRI) should not facilitate inter-operability with standards, such as GTFS, as well? 

Also, and maybe we need to amend the description to clarify this:
> The very reason why this field is needed is that a consumer with local knowledge can use it to reference other passenger-facing systems outside the GTFS world using the original_trip_id provided, when other passenger-facing systems use an ID which is not the public code. 

Such systems can be as much as (in Switzerland) the national identification of journeys accross systems, standards, and operators.

leonardehrenfried:
Please don't get me wrong. I have doubts but I'm not the Guardian of GTFS Purity. :)

I would probably not vote at all.

skinkie:
OpenGeo has been producing this information as `realtime_trip_id` for over 10 years now. So you can add us towards the list of producers. We will modify our output once the vote has been done.

ue71603:
In NeTex some still think that the id attribute will suffice. They are wrong.  We will in the profiles have to make clear that the content  of the original_trip_id must be somewhere. (either in a KeyList with a given key or in the privateCodes with a given key.

tzujenchanmbd:
Given that this field may be useful for systems in different regions(?), but the current comments seems to be primarily from Europe, I’d like to invite perspectives from other regions as well :)

@jfabi @westontrillium @evansiroky @drewda @m-mcqueen @sberkley @takohei @iniad-bessho @gcamp @jll01 @bdferris-v2 Any thoughts on this field?

gcamp:
I share @leonardehrenfried and @skinkie's reservations. 

While I support GTFS interoperability with other standards I believe the current proposal for a multi-purpose, context-dependent field is problematic.  Without clearly defined contexts and specifications, this field could be used inconsistently across different systems, even those claiming interoperability with the same standard. 

A more appropriate approach in my opinion, would be to use the established extension mechanism. GTFS-rt and protobuf have explicit mechanisms to extend the schema to include private-use fields with [extensions](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md#extensions). I think this would make sense for here at least specifically for the HRDF usecase. 

I support adding to GTFS something like `global_trip_id` or `realtime_trip_id` where we could link trips to locally link them in a way that's not possible in GTFS right now. However that needs to be different to what is proposed here.

skinkie:
I think that one of the more valuable comments considering this extension is that this 'id' lives gives a (stable) aggregation during the entire calendar. I think this part must not be ignored.

t2gran:
From doc:
> It allows to compare and match GTFS-trips with similar constructs from other standards (e.g. NeTEx/SIRI) or with schedule.

 - What is the types/fields in NeTEx this mappes to?

skinkie:
> * What is the types/fields in NeTEx this mappes to?

In best case `<PrivateCode type="something">That global id</PrivateCode>`

miklcct:
> I'd also have the reservations noted by @gcamp.
> 
> One case we come across often is needing to match trips with other systems such as scheduling software for import into our CAD/AVL, while maintaining a user (dispatch and driver) facing "name" for the trip. We've done some extensions with scheduling providers where we use their trip IDs to match and a field called `trip_label` to label the trip for users. Leaving the `trip_name` empty because that's already used in GTFS to represent rider-facing names for trips (like a train number or a long-distance or commuter bus).
> 
> While that's not the use case explored here, it's an example of how we've successfully used an extension that doesn't need to be in the actual spec and I think that makes the most sense for this use case too.
> 
> For anything that is meant to reference _outside of GTFS_ I think the best place to go is to go to an extension. There was recent discussion with the Federal Transportation Administration in the USA where they wanted `agency_id` to match their National Transit Database ID. Public comment from GTFS producers suggested using an `ntd_id` extension or an `ntd.txt` file to do this. GTFS is only self-referential and I believe it should, officially in the core spec, stay that way.
> 
> I also agree with @gcamp that matching IDs between static and realtime files has been a problem where agencies use different producers for each. For that, a field in either static, or realtime, or both to reference the ID in the other could be useful, but that is still within the GTFS ecosystem and not the issue discussed here.

In the rail network in Great Britain, there are at least 3 different kinds of IDs in use:
* UID (Unique ID): uniquely identify a service in the information system, applicable for all trains (passenger, freight, empty), unique within the whole network but only on a single service day (defined as 00:00 - 23:59 local time). It is not shown to the public nor used in operation at all, only people dealing with IT will use it.
* Signalling ID: the code used internally for operational purposes, including signallers and train operators, applicable for all trains, unique only at a specific instant and within a signalling region. Multiple trains with the same signalling ID can appear in different parts of the network, and it can be reused at the same place within a single day.
* RSID (Retail Service ID): the train number, which is intended to be public facing, applicable to passenger trains only. It is supposed to be unique (but I have seen clashes before), but only on a single service day.

My current converter to GTFS maps the UID to `original_trip_id` and RSID to `trip_short_name`, and generate the `trip_id` using a way that it will be consistent across the validity of the timetable. It does not contain the signalling ID (the dispatch and driver facing "name") in the feed as it is considered operational details, not relevant to a passenger-facing operation.

abyrd:
This PR was discussed in today's OpenTripPlanner developer meeting. Given the number of comments on this PR some of us had a hard time getting a simple overview of the issue. Is the following summary correct?

This PR would add a field that can associate trips with any ID defined outside GTFS. The main use case described by its proponents is allowing consumers to recognize that a trip is a replacement for another trip the consumer already knows about (as these would have two different IDs). While this use case is widely acknowledged, many people would prefer extension fields that target specific use cases like this one, rather than a field referencing any external ID without a specified reason.

---

As there has been a lot of discussion, it would be useful to revisit the requests from @skinkie to provide a short list of very concrete use cases from the passenger perspective. Several OpenTripPlanner developers perceived such a list as quite important for allowing newcomers to understand this PR.

skinkie:
> This PR was discussed in today's OpenTripPlanner developer meeting. Given the number of comments on this PR some of us had a hard time getting a simple overview of the issue. Is the following summary correct?

I think that does not capture that the field does not have to be as unique as trip_id, and thus may suggest that some trips-in-gtfs form a group. That would be more something like _trip_group_id_, which is a completely different feature than your summary describes.

ue71603:
It is really important to note, that it is NOT only an additional id, but also fulfills an important requirement of public transport (espeically) that there are "trips" that exist on multiple days, but are not totally the same. In GTFS this results in multiple trip_id. In the rest of the communication it is the 7:02 from Bern to Zürich (so the same on every operating day). The proposed filed stores this information. Therefore there is a 1:n relation  to the trip_id, but on a given operating day only one of the trip_id can be valid.

leonardehrenfried:
I quite like @gcamp's take on it: it being framed as an interoperability concept is less useful. Like him I also support it being introduced as a new concept into the spec that just happens to line up with other standards and can therefore also be used for real time matching.

miles-grant-ibigroup:
As a GTFS consumer, this field will add more confusion than it seeks to resolve. It seems like this field is very vulnerable to being used very differently across different contexts and it feels like it’s going to be a dumping ground for various hacks. This makes it very difficult to work with as a consumer. There are better approaches for resolving the issues this PR seeks to resolve.

stevenmwhite:
> ...there are "trips" that exist on multiple days, but are not totally the same. In GTFS this results in multiple trip_id. In the rest of the communication it is the 7:02 from Bern to Zürich (so the same on every operating day).

This feels to me like it _should_ result in multiple trip IDs and that is not a problem. These are different from a _technical_ perspective, and they need to have different IDs in order to represent those differences.

Despite the differences, a GTFS consumer giving rider information will still show them each as the 7:02 trip from Bern to Zürich for the relevant day. And if a rider needs some specific identifier for that trip other than the start time, origin, destination, etc. then that's what the `trip_short_name` (which should be unique within a service day, but can be repeated across different days) is for.

Looking at the variety of comments I'm a little confused about whether we're trying to solve an issue with rider communication within the context of GTFS, or whether we're trying to link an entity in GTFS to some entity in a different file/format. If the latter, that's where an extension is helpful. If the former, I think that's a different issue that needs to be more clearly defined.

miklcct:
trip_short_name is clearly the user-facing code, but it can't be used by apps to know that the trip is a modified version of the original of the timetable when it is updated. 

The use case is if I plan a trip 90 days ahead, the timetable is adjusted, and I recheck the journey details before the day of travel.

ue71603:
In many public transport operators the id for a trip has two parts "operating_day + id (at that day)". This is the way the operating part is organised and planned. One can ignore that, in GTFS and simply do the uniqueness  with random trip_id (we do that). We build one whole year of Swiss data into one GTFS file. One looses the grouping and the understanding what happens, when this is not added. The operating day is also technical (and not always the same as the calender date). To have the information in the information that is displayed is a bad idea too. It is a technical grouping and the second part of two part technical key.

I don't see the confusion @miles-grant-ibigroup: If you don't need it, then you don't consume it. It is the same with other fields that are in the spec and in the real data. If you don't know what to do with transfers.txt or flex then don't consume it. It is not breaking the rest, but if you need it, you really need it.

We have currently trips that happen most days, that have 200 variants in the year due to miniscule and less miniscule changes in shape, additional stops, slight delays somewhere, restaurant open or not, changes information. It is virtually impossible to regroup them when the information is lost. We have a long history of mapping stuff together with id and references lost. It is never 100% and in the end it results in worse passenger information.

abyrd:
I appreciate everyone's efforts at explaining the concept, and have read the comments attentively. However, these descriptions are again veering into technical descriptions of possible concepts in external datasets that might be referenced, the cardinality of these relationships, and how such references enable general "interoperability". I was mainly seeking one or more very concise and concrete examples of how this field would be used *by end users*. We need to highlight those use cases so newcomers can understand why this PR exists and whether they support it. There are lots of references here to "grouping" trips or trips being "the same" even though they are separate. Why concretely does a passenger want to "group the trips"?

Thanks @miklcct for giving what I think is the canonical use case: "I plan a trip 90 days ahead, the timetable is adjusted, and I recheck the journey details before the day of travel." I tend to agree with him that `trip_short_name` is user-facing and does not seem intended for use as machine-readable identifier. (However, the spec reference also says `trip_short_name` "uniquely identif[ies] a trip within a service day" which is somewhat contradictory and needs to be clarified.)

For this use case, it is possible to narrow the field's semantics ("ID for the purposes of joining with realtime changes") and eliminate any reference to datasets outside GTFS. The data producer will obviously use some externally generated ID for this purpose, but where that comes from is not relevant to the GTFS spec. The meaning and use cases of the field are relevant. I think this aligns with the `realtime_trip_id` several people are mentioning.

Can anyone name any other concrete use cases, or is realtime matching the only known use case (from a passenger/end user perspective)?

@ue71603 Can you please clarify:

- Why from the passenger perspective is it an "important requirement of public transit" that trips have persistent existence across multiple days?
- Granted "It is virtually impossible to regroup [trips] when the information is lost" but why would a passenger want to "regroup" trips?
- Why would you prefer a semantic-free identifier for arbitrary grouping, instead of a field with a specific purpose?

Thanks!

abyrd:
Looking at https://www.tp-info.ch/fr/gestion-des-donnees/sid4pt-swiss-id-public-transport/swiss-journey-identification-sjyid it seems like the only concrete use case cited is matching realtime information to scheduled services.

I know as software or IT professionals a lot of us have very abstract thought processes about entities, identity, and relations etc. But GTFS is a very pragmatic and passenger-focused specification, so I think it's a good idea to reframe proposals like this for concrete use cases.

miklcct:
> I appreciate everyone's efforts at explaining the concept, and have read the comments attentively. However, these descriptions are again veering into technical descriptions of possible concepts in external datasets that might be referenced, the cardinality of these relationships, and how such references enable general "interoperability". I was mainly seeking one or more very concise and concrete examples of how this field would be used *by end users*. We need to highlight those use cases so newcomers can understand why this PR exists and whether they support it. There are lots of references here to "grouping" trips or trips being "the same" even though they are separate. Why concretely does a passenger want to "group the trips"?
> 
> Thanks @miklcct for giving what I think is the canonical use case: "I plan a trip 90 days ahead, the timetable is adjusted, and I recheck the journey details before the day of travel." I tend to agree with him that `trip_short_name` is user-facing and does not seem intended for use as machine-readable identifier. (However, the spec reference also says `trip_short_name` "uniquely identif[ies] a trip within a service day" which is somewhat contradictory and needs to be clarified.)
> 
> For this use case, it is possible to narrow the field's semantics ("ID for the purposes of joining with realtime changes") and eliminate any reference to datasets outside GTFS. The data producer will obviously use some externally generated ID for this purpose, but where that comes from is not relevant to the GTFS spec. The meaning and use cases of the field are relevant. I think this aligns with the `realtime_trip_id` several people are mentioning.
> 
> Can anyone name any other concrete use cases, or is realtime matching the only known use case (from a passenger/end user perspective)?
> 
> @ue71603 Can you please clarify:
> 
> - Why from the passenger perspective is it an "important requirement of public transit" that trips have persistent existence across multiple days?
> - Granted "It is virtually impossible to regroup [trips] when the information is lost" but why would a passenger want to "regroup" trips?
> - Why would you prefer a semantic-free identifier for arbitrary grouping, instead of a field with a specific purpose?
> 
> Thanks!

Hi Andrew, 

We plan in the future to have the functionality to save a "regular commute" in our app, which will allow you to pick a certain trip, show that the service days the trip will run, and it will check for update daily before you leave. Therefore we need a stable ID in this case even if the timetable is changed.

In the current GTFS standard, I would use the trip_id and the service_id to tell the user the calendar of the trip. If there is a planned alteration to the timetable, the applicable calendar will be changed so the trip_id will not run. We need to have a way to associate another trip to be the semantically equivalent to notify the user he should take that trip.

miklcct:
The flexibility of standards in interpretation means that it is not possible for implementations to guarantee correct desired behaviour across different feeds in different places

ue71603:
@abyrd 

>Why from the passenger perspective is it an "important requirement of public transit" that trips have persistent existence across multiple days?
It is  in my view a prerequisite to good passenger information. You seem to think it is technical and aloof. Naturally it will not be presented to the passenger in this way. You don't do this with trip_id either. Having only the local and artifically created trip_id means that it does not mesh with other items in the data space. Perhaps in your world this is something people don't do and if it can't be done in GTFS then it should not be done. Here it is a major concern of people using our data (especially additional fields that are not available in GTFS, but are in NeTEx and HRDF). We didn't do this, because we like architecture, but because we were asked to do something about the problem. We wanted for people to either stay in the GTFS bubble if they want and still get the additional data in an easy way.

>Granted "It is virtually impossible to regroup [trips] when the information is lost" but why would a passenger want to "regroup" trips?
Again, they won't see it directly usual.  If you don't want to do start/stop matching, then one needs references. And most of Europes public transport is organisaed operating+key as unique id and not just id as GTFS is.
We are working slowly towards a master data mgmt. Meaning that things like stops, lines, journeys will have a master identification. And for the journey it will be something of the type operating day + id.

>Why would you prefer a semantic-free identifier for arbitrary grouping, instead of a field with a specific purpose?
I don't understand whay you are saying.

derhuerst:
> We plan in the future to have the functionality to save a "regular commute" in our app, which will allow you to pick a certain trip, show that the service days the trip will run, and it will check for update daily before you leave. Therefore we need a stable ID in this case even if the timetable is changed.

I'd argue that it depends a lot on the passenger what is considered *their regular commute*. Some examples:
- Some passengers will have to know if their commute is wheelchair-accessible. If, one day, a different trip a differtent `wheelchair_accessible` value (due to a different physical vehicle being used) runs, is this "new" trip (with the same Swiss Journey ID) still truly *their* commute?
- For most people that I have talked to, their commute is inherently intermodal, as they *always* use some form of mobility (e.g. walking) before boarding anything that's modelled in GTFS. Therefore, an ID identifying their commute wouldn't refer to the PT trips only! IMHO the use case "identifying my commute" is something both more complex and more abstract than the use cases GTFS is designed for, and needs to be handled in a layer/spec on top of GTFS.

TLDR: The planning perspective's (of large European rail-focused organisations with fixed schedules) notion of a "commute" doesn't match the passengers' notions.

---

Often, GTFS trips are either a) referenced by multiple (relevant) companies or even b) *operated by* multiple companies. Most of them will likely have their own "original" trip ID for their planning purposes for the same GTFS trip, so effectively for one GTFS trip ID we end up with >1 "original" trip IDs.

Once we consider a one-to-many mapping, I see chaos coming. While it's *technically feasible* to just include all known "original" trip IDs and hope that consumers will recognise any >1 among them, I think this approach needs more thinking and coordination. This is why, in addition being sceptic about the use case as explained above, I think such a field should stay in an extension for now. (An extension can still be widely agreed-upon and developed in an open and backwards-compatible manner!)

miklcct:
> I'd argue that it depends a lot on the passenger what is considered *their regular commute*. Some examples:
> - Some passengers will have to know if their commute is wheelchair-accessible. If, one day, a different trip a differtent `wheelchair_accessible` value (due to a different physical vehicle being used) runs, is this "new" trip (with the same Swiss Journey ID) still truly *their* commute?

Of course it is still his commute. My app would tell the user that the wheelchair value is changed from the saved journey and asks the user how to proceed (e.g. picking up another departure).

> - For most people that I have talked to, their commute is inherently intermodal, as they *always* use some form of mobility (e.g. walking) before boarding anything that's modelled in GTFS. Therefore, an ID identifying their commute wouldn't refer to the PT trips only! IMHO the use case "identifying my commute" is something both more complex and more abstract than the use cases GTFS is designed for, and needs to be handled in a layer/spec on top of GTFS.
> 
What I have saved is the complete itinerary, with all the access / transfer / egress in addition to all the PT leg as well. All the step by step details from how to cycle from my home to a parking, walk to the platform, take the train, take to the office are saved as part of the regular commute, and can be shown to the user if the refetched PT leg is different.

doconnoronca:
It seems to me it was a mistake to require the trip_id be unique. In retrospect it should have been unique for the schedule day. It would have solved quite a few problems I seen people complain about over the years.

Adding a new way to link trips access days and schedule changes could be useful in unexpected ways.

abyrd:
> It seems to me it was a mistake to require the trip_id be unique. In retrospect it should have been unique for the schedule day. It would have solved quite a few problems I seen people complain about over the years.
> Adding a new way to link trips access days and schedule changes could be useful in unexpected ways.

I don't clearly see how linking trips across schedule changes is related to linking trips across days. How does knowing that an operator considers a certain trip on Wednesdays to be "the same" as a certain trip on Tuesdays, help me know that a new replacement trip on Wednesday 19 February corresponds to a specific original trip on Wednesday 19 February?

The idea of a dataset being built around a relation called `trips`, but the identifier of those trips (`trip_id`) not being dataset-unique is quite odd to me. I feel like we're mixing together a lot of different topics here. I suppose I can imagine the trips relation having a compound key of `(service_id, trip_id)` but it's not clear to me what this would improve.

It may be worth noting that a single trip (i.e. with a single `trip_id`) can already be instantiated across many days. This is the purpose of calendars.txt and calendar_dates.txt. The system is geared toward simpler schedules recurring regularly on certain days of the week, but is relatively flexible. The same trip_id can appear on many different service days.

miklcct:
> > It seems to me it was a mistake to require the trip_id be unique. In retrospect it should have been unique for the schedule day. It would have solved quite a few problems I seen people complain about over the years.
> > Adding a new way to link trips access days and schedule changes could be useful in unexpected ways.
> 
> I don't clearly see how linking trips across schedule changes is related to linking trips across days. How does knowing that an operator considers a certain trip on Wednesdays to be "the same" as a certain trip on Tuesdays, help me know that a new replacement trip on Wednesday 19 February corresponds to a specific original trip on Wednesday 19 February?
> 
> The idea of a dataset being built around a relation called `trips`, but the identifier of those trips (`trip_id`) not being dataset-unique is quite odd to me. I feel like we're mixing together a lot of different topics here. I suppose I can imagine the trips relation having a compound key of `(service_id, trip_id)` but it's not clear to me what this would improve.
> 
> It may be worth noting that a single trip (i.e. with a single `trip_id`) can already be instantiated across many days. This is the purpose of calendars.txt and calendar_dates.txt. The system is geared toward simpler schedules recurring regularly on certain days of the week, but is relatively flexible. The same trip_id can appear on many different service days.

You are now conflating two issues here.

1. Within the same dataset, I want to know that a trip on Wednesday runs on the same timetable as a trip on Tuesday. This is already supported.
2. Across different versions of the dataset, I want to know that a trip on Wednesday in the new timetable corresponds to a certain trip on Wednesday in the old timetable. This is not possible due to the constraint that a `trip_id` must be unique in a dataset, so if in the old timetable, the Wednesday trip is the same trip as the Tuesday trip, but in the new timetable, the Wednesday trip runs a minute different from the Tuesday trip, it is not possible to associate them.

abyrd:
> It is  in my view a prerequisite to good passenger information. You seem to think it is technical and aloof. Naturally it will not be presented to the passenger in this way...
> Perhaps in your world this is something people don't do and if it can't be done in GTFS then it should not be done. Here it is a major concern of people using our data...

I am not sure what you mean by "my world", and I am making no statements that things "should not be done". I think I should clarify here that I am not strongly invested in a particular outcome for this PR. I was just trying to encourage communication between commenters who seemed to have different perspectives.

I think most or all commenters are sympathetic to your situation and want to help transform your proposal into something that fits well into GTFS and will be well-received and well-understood by the community.

My understanding is that some commenters would prefer a field defined by its concrete use cases, focusing on relationships within GTFS, and remaining silent on any relationship to data sets outside GTFS.

My intent was to encourage communication about whether the only known use case is rediscovering trips (for example, that someone has bookmarked or purchased a ticket for) when they have been altered by realtime data or other small schedule modifications. That is: are we functionally just talking about a `realtime_trip_id`?

>> Granted "It is virtually impossible to regroup [trips] when the information is lost" but why would a passenger want to "regroup" trips?

> Again, they won't see it directly usual.

I understand that passengers will not look directly at ID strings. I was trying to encourage clear description of concrete situations encountered by the passenger where this ID would be used by software. The example provided by @miklcct would require an ID that maintains identity across small modifications to a trip on a particular day. I was asking: do we have any analogous examples of concrete use cases requiring an ID that groups trips across multiple days?

>> Why would you prefer a semantic-free identifier for arbitrary grouping, instead of a field with a specific purpose?

> I don't understand whay you are saying.

This PR seems to be about adding a field that can contain any "original" identifier for the trip, without attaching any meaning or usage context to that identifier. My understanding of comments from other contributors is that they prefer to see fields that have a specific meaning and use case.

To rephrase my question: Why would you prefer a field that can contain any arbitrary external identifier with no stated purpose, instead of a field or fields that contain identifiers with specific purposes, such as matching realtime messages to trips that have been altered?

miklcct:
Unfortunately although it was my original idea, I am now skeptical about this PR.

It is impossible to define a precise semantic about what "upstream" data the `original_trip_id` refers to, especially if a feed combines multiple sources or if the same trip appears in multiple upstream systems. We don't want to add any fields that have unclear meaning into the standard.

I think we should disregard the Google extension and define fields which cater for the use cases I mentioned above.

abyrd:
> You are now conflating two issues here.

I also consider these two different cases; I agree with your perspective. It seemed to me that they were being conflated in previous comments about the importance of `original_trip_id`. My intent was to acknowledge the validity of your example use case, and try to encourage discussion on whether the proposed `original_trip_id` was needed to solve or communicate any inter-trip relationships other than a sort of `realtime_trip_id` (your use case number 2).

miklcct:
I think that we can define a field, called `semantic_id`, in the trip model to group trips which are "semantically" considered the same by operators and passengers.

Then we can attach a specific requirement of that ID:

Two trips with the same `semantic_id` on different days should be considered the approximate equivalent from the passenger's view, for example, if there is a planned diversion on a certain Sunday, the diverted trip should share the same `semantic_id` with the original trip on different Sundays.

`semantic_id` must be consistent across different versions of the dataset. If a new version of the dataset has a trip which has the same `semantic_id` on a service day, it is considered the same trip as the previous version which may have different timetables. For example, if there is a planned diversion, consumers can notify the passenger upon receiving the new version of the feed.

`semantic_id` should be unique within a service day. Exception can be made if a logical trip is truncated into multiple parts, for example, a train - replacement bus - train forms a through journey as a replacement from the original train journey.

abyrd:
> * GTFS has some information about the accessibility and none about the formation of the train.
> * Think I want that.  I can't fetch it with the trip_id. What I need is the operating day 2025-04-03 and the journey id to hand it to the formation service or to look it up in the HRDF data that you obtained as well.
> How do you suggest this should be done in the current situation?

I can think of a few possibilities. I am not promoting these or saying they're desirable, just brainstorming:

- Extension field (proprietary or shared) for referencing very specific external data sources or formats (`hafas_journey_id`)
- Extension field (clearly and intentionally proprietary) (`sbb_journey_id`)
- An extension table to GTFS to capture train composition (specific aspects relevant to passenger information)

Keeping in mind that GTFS consumers should be tolerant of unofficial extensions. If something is clearly specific to your dataset, you could add `sbb_field_a` or `sbb_field_b` to any table, or `sbb_table_a.txt` to the archive, and it should not break anything.

stevenmwhite:
These most recent comments from @abyrd and @miklcct are getting much closer to something that seems acceptable to me. The ideas have specific, definable purposes. 

Would `semantic_id` be the first field in GTFS where consistency across datasets is required? I know it's a best practice for things like `stop_id` and `route_id` to remain consistent but I don't think that's actually a requirement for any fields currently. As long as they're internally consistent within the dataset, they are technically allowed to change with each dataset update because they're not meant to be referenced outside of that dataset.

westontrillium:
I had hoped to contribute to this discussion with something useful from a (North American) producer perspective but honestly am finding it challenging to gain a technical understanding of what’s being discussed. It sounds like the conversation may be moving beyond what has been originally proposed anyway, but I wanted to call out that even after reading through the original PR and the entirety of this comment thread, I didn’t find any comprehensive examples using actual data or tables (not just isolated trips.txt snippets) so we can clearly see the proposed file/field relationships.

As such, I’m still unclear on whether there are any restrictions for an `original_trip_id` (or equivalent) referring to multiple `trip_id`’s: Can *any* trips be grouped together? If not, how identical do the trips have to be? Same stops? Do `arrival/departure_time`s have to differ? What about `pickup/drop_off_type`? I would really love for someone to share a few simple tables showing the flow from route/calendar to `original_trip_id`+`trip_id`s to stop_times (for *all* trips in the example) so I can see what sort of “grouping” is possible/intended.

I’m also having trouble wrapping my head around some of the language used to describe the temporal restrictions for `original_trip_id` without concrete examples:

> “… This ID is valid for one operating day and across different days of a scheduled year…” “… only unique on a single day…” “… Unique per operating day…” “… Can reoccur across different days of a scheduled year…”

These statements all seem to be describing the same thing though slightly differently and potentially contradict one another. What exactly are we saying? a single `original_trip_id` only occurs for literally one calendar day (e.g., 2024-01-01), so you must structure calendar.txt/calendar_dates.txt accordingly? Or one day **of the week** across multiple dates of a scheduled year, e.g., Wednesdays on a `service_id` where `wednesday=1`? Does that mean it can’t be attached to multiple trips occurring in a single day? Again, examples would be very useful!

davidr1234:
Here's a real-world example from our implementation.

GTFS Schedule [URL](https://data.opentransportdata.swiss/dataset/6cca1dfb-e53d-4da8-8d49-4797b3e768e3/resource/82ab58b1-b303-4195-afbb-3251fa694ea7/download/gtfs_fp2025_2025-02-27.zip).

We have the following situation. The train S6 goes from Basel Bad Bf to Basel SBB.
image

This is a train that many commuters and other regulars use and for which they know its stops. So they keep in mind the "pattern", especially, because in Switzerland we have a ["Taktfahrplan"](https://de.wikipedia.org/wiki/Taktfahrplan). So the trains are part of that pattern.

On occasions "their S6" stops at a different quay. This is what we need to be able to represent to the commuters. And we do as described below using the "original_trip_id" (**Note**, we named the field based on GTFS Transit. It is used for the described use-case. At least based on what we understood from an exchange with data providers for Google).

Swiss Journey ID (SJYID) aka **original_trip_id**: ch:1:sjyid:100308:87865-001
**Trip IDs**: ['**105.TA.91-6-E-j25-1.14.H**', '44.TA.91-6-E-j25-1.8.H', '**53.TA.91-6-E-j25-1.10.H**', '74.TA.91-6-E-j25-1.13.H']
- **routes.txt**
-- route_id,agency_id,route_short_name,route_long_name,route_desc,route_type
-- "91-6-E-j25-1","351","S6","","S","109"

- **trips.txt**
-- route_id,service_id,trip_id,trip_headsign,trip_short_name,direction_id,block_id,original_trip_id,hints
-- "91-6-E-j25-1","TA+ep500","105.TA.91-6-E-j25-1.14.H","Basel SBB","87865","0","","ch:1:sjyid:100308:87865-001",""
//
-- "91-6-E-j25-1","TA+h8000","53.TA.91-6-E-j25-1.10.H","Basel SBB","87865","0","","ch:1:sjyid:100308:87865-001",""

- **stop_times.txt**
--trip_id,arrival_time,departure_time,stop_id,stop_sequence,pickup_type,drop_off_type
-- "105.TA.91-6-E-j25-1.14.H","21:45:00","21:45:00","8500090","1","0","0"
-- "105.TA.91-6-E-j25-1.14.H","21:50:00","21:50:00","8500010:0:5","2","0","0"
//
-- "53.TA.91-6-E-j25-1.10.H","21:45:00","21:45:00","8500090","1","0","0"
-- "53.TA.91-6-E-j25-1.10.H","21:50:00","21:50:00","8500010:0:3","2","0","0"

- **calendar.txt**
-- service_id,monday,tuesday,wednesday,thursday,friday,saturday,sunday,start_date,end_date
-- "TA+ep500","1","1","1","1","0","0","0","20241215","20251213"
-- "TA+h8000","0","0","0","1","0","0","0","20241215","20251213"

- **calendar_dates.txt**
-- service_id,date,exception_type
-- "TA+ep500","20241216","2" ...(dates in range with exception_type=2)... "TA+ep500","20251211","2"
//
-- "TA+h8000","20241219","2" ...(dates in range with exception_type=2)... "TA+h8000","20251211","2"

davidr1234:
@westontrillium @stevenmwhite @abyrd @miklcct - any thoughts/feedback concerning our example?

miklcct:
Whether I am going to support the proposal will depend on how it is defined.

If the definition is "an internal identifier which is used to group related trips across different dates or across different version of feeds", I'll support it.

If the definition is "an ID which is used to reference external information systems", I'll oppose it.

sebastianknopf:
One additional note from my perspective: I noted the `original_trip_id` field in the SWISS GTFS feeds. First, I was a little angry because the SJYIDs (Swiss Journey IDs) are not filled into `trip_id`, but seen from the perspective of a common traffic planning tool, there're indeed cases, where trips considered as 'one trip' by the passengers must be split into two different, technical trips in the planning system. As `trip_id` must be unique this would be collide with those two technical trips with the same SJYID.

Maybe, it would be a way to call this field simply `global_id`, which would be generic enough and also specific for the global IDs typically used in public transport (SJYID, DFID, ...). This would also be an identifier for 'a trip which is considered as **one trip** by the passenger', however it is not a human-readable ID, but a technical ID, which makes matching data between different data sources more easy, if they're used properly.

ue71603:
@tuxalp and I will try to update the definition to take namenly the comment @miklcct into account.  Concerning the renaming proposed by @sebastianknopf . We use the namein all other elements as well for some timenow. So we intend to keep it. 

The core problem is that passenger information trips and technical trips are not the same. As trip_id is a primary key it can't be used for this. We feel that if something has not really the properties of a primary key is better served as a normal element. Separation of concerns.

ue71603:
@sebastianknopf is refering to this announcement (for English text scroll down): https://in.mailing.cff.ch/html_mail.jsp?params=ya9fojwzTOqV1oC2lLo5iksvHkLMvmA2TyXrGmW6WaPQJW8p0CWBrbh8A8xGe%2FAecnDnlisnU%2FOsDFEp8blhkDqEs4N906Oi%2BJkv7cKSPCo%3D
and this example file: https://in.mailing.cff.ch/d/d.zip?l000qjri0nulnwbid0000lz0000000000m2er7dsdj7sxl3w05eu7oli000d5y00b000000dvmd7wm

Issue #530, opened 2025-01-28
GTFS Realtime Versioning


tzujenchanmbd:
### Ask a question

In recent discussions regarding Realtime spec, the community seems to have encountered difficulties in reaching consensus, leading to the topic of versioning being raised in the recent [PR#504](https://github.com/google/transit/pull/504).

To address this, we are opening this issue dedicated to discussing GTFS Realtime versioning.

The following questions may help us begin to think about this significant topic:
- If we implement versioning (i.e., a mechanism for incrementing version numbers), should this mechanism allow for breaking changes for producers by permitting new versions to be incompatible with older versions? For example, should it allow for the complete removal of certain elements or the redefinition (modification) of content from previous versions in the new version spec?
- Compared to the existing mechanism of using the [deprecated = true] tag, what are the advantages and potential risks of implementing versioning?

skinkie:
I think _removing_ elements is not what we want, because that would make the protobuf file unparsible by a newer version by default. The more complex situation is if the _structure_ would change in a higher version, effectively not being backwards compatible by the wire-protocol. In #504 the difficulty is an interpretation change, or better put, _any interpretation, being this one_. Such "clarification-change" might break existing implementations, were the standard has diverted because of the lack of clarity or an actor being unconsciously incompetent.

Now what if we would break [private extensions](https://developers.google.com/transit/gtfs/reference/extended-route-types)? If private extensions were never put here on the table, breaking something that would break the private extension: is not our problem. There are parties that have implemented feature X**'**, misreading it for feature X. To help in such transition, and to be able to make a distinction we could bump a version. This is different from [deprecated=True] because the deprecated status suggests the feature is not available anymore, while in this case the feature is merely wrongfully implemented by a group of users.

In high sight what was the reason for bumping to version 2.0? Can we learn anything from that?

I would also suggest a clear development path that would basically state which issues or features we want to resolve in a certain time frame, so it is clear that if there are "breaking changes" the timeline is clear to aggregate all changes for a new version.

PR #546, opened 2025-03-25
Add communication_period and impact_period

### LINKED ISSUE #469: Proposed Best Practice: clarify intended use for CANCELED/SKIPPED TripUpdates VS NO_SERVICE Alerts.

### LINKED ISSUE #521: Add communication_period and impact_period to GTFS Service Alerts
### Describe the problem

In the GTFS Realtime Alert spec, `active_period` is defined as: _Time when the alert should be shown to the user. If missing, the alert will be shown as long as it appears in the feed. If multiple ranges are given, the alert will be shown during all of them_.

However, there still is an ambiguity of interpretation of the `active_period` field; whether it is intended for the period of showing the alert or for the period of the related service disruption itself.

### Use cases

We need to distinguish:

1. The period when the alert is shown -> Communication period of the alert
2. The period when the alert is actually in effect -> Impact period of the alert

### Proposed solution

We propose the addition of two fields, `communication_period` and `impact_period`, to disambiguate `active_period`.

The detailed proposal is found here: [communication_period and impact_period proposal
](https://docs.google.com/document/d/1t5hMwqC0WXfgKzhkJFT_h0QwbtvqWzB4vFN1j7-ciXI/edit?usp=sharing)

### Additional information

More information regarding the extent of this issue can be found in [PR#482](https://github.com/google/transit/pull/482)

skalexch:
### **BACKGROUND**

In the GTFS Realtime Alert spec, active_period is defined as: Time when the alert should be shown to the user. If missing, the alert will be shown as long as it appears in the feed. If multiple ranges are given, the alert will be shown during all of them.

However, there still is an ambiguity of interpretation of the active_period field; whether it is intended for the period of showing the alert or for the period of the related service disruption itself.

### **PROPOSAL**

We propose the addition of two fields, communication_period and impact_period, to disambiguate active_period.

communication_period: Time when the alert should be shown to the user strictly for informative reasons.
impact_period: Time when the services are affected by the disruption mentioned in the alert.

We also define rules of interaction with active_period to ensure that the new fields are mutually exclusive with active_period.

More details and potential combinations can be found in [Issue#521](https://github.com/google/transit/issues/521) and the [proposal doc](https://docs.google.com/document/d/1t5hMwqC0WXfgKzhkJFT_h0QwbtvqWzB4vFN1j7-ciXI/edit?usp=sharing).


doconnoronca:
Perhaps the impact period should be allowed even if there is no communication period so the producer doesn't have to support a communications period.

Maybe also indicate the active period is being deprecated.

gtfs-realtime/spec/en/reference.md
```diff
@@ -340,7 +340,9 @@ An alert, indicating some sort of incident in the public transit network.
...
 |------------------|------------|----------------|-------------------|-------------------|
-| **active_period** | [TimeRange](#message-timerange) | Optional | Many | Time when the alert should be shown to the user. If missing, the alert will be shown as long as it appears in the feed. If multiple ranges are given, the alert will be shown during all of them. |
+| **active_period** | [TimeRange](#message-timerange) | Conditionally Forbidden | Many | Time when the alert should be shown to the user. If missing, the alert will be shown as long as it appears in the feed. If multiple ranges are given, the alert will be shown during all of them. <br><br> - **Forbidden** if communication_period and impact_period exist |
+| **communication_period** | [TimeRange](#message-timerange) | Conditionally Required | Many | Time when the alert should be shown to the user strictly for informative reasons. If missing, the alert will be shown as long as it appears in the feed. If multiple ranges are given, the alert will be shown during all of them. <br><br> - **Required** If impact_period is specified.<br> - **Forbidden** otherwise.|
```
gcamp, line 344:
```suggestion
| **communication_period** | [TimeRange](#message-timerange) | Conditionally Forbidden | Many | Time when the alert should be shown to the user strictly for informative reasons. If missing, the consuming application can decide when it's appropriate to be shown. If multiple ranges are given, the alert will be shown during all of them. <br><br> - **Optional** If impact_period is specified.<br> - **Forbidden** otherwise.|
```

gcamp (commented):
Some comments, I agree with @doconnoronca that we should also deprecate `active_period`

gtfs-realtime/spec/en/reference.md
```diff
@@ -340,7 +340,9 @@ An alert, indicating some sort of incident in the public transit network.
...
+| **impact_period** | [TimeRange](#message-timerange) | Conditionally Required | Many | Time when the services are affected by the disruption mentioned in the alert. <br><br> - **Required** If communication_period is specified.<br> - **Forbidden** otherwise.|
```
stevenmwhite, line 345: I'd remove the word "disruption" since alerts come in a lot of different flavors...

>Time when the services are affected by this alert.

stevenmwhite:
> What is the use case of `communication_period`? Why do you want to have an alert in a feed when it is outside the communication period?

Agencies often want to alert riders to _upcoming_ situations, and create alerts with a very broad `active_period` in order to do so. This comes out of the discussion here: https://github.com/google/transit/pull/482#issuecomment-2305391658


I'm in favor of this generally, with a few notes:
- I don't think `communication_period` should be required. If there's no communication period I think consumers should just show it when impactful.
- On the flip side side, we see agencies try to use this for general informational messages that don't have a true impact on service. Should we conditionally require _either_ the communication _or_ the impact period (or both).
- I also agree that we should deprecate `active_period`, but I'm not sure how that will work with backwards compatibility.

miklcct:
> > What is the use case of `communication_period`? Why do you want to have an alert in a feed when it is outside the communication period?
> 
> Agencies often want to alert riders to _upcoming_ situations, and create alerts with a very broad `active_period` in order to do so. This comes out of the discussion here: [#482 (comment)](https://github.com/google/transit/pull/482#issuecomment-2305391658)
> 
> I'm in favor of this generally, with a few notes:
> 
> * I don't think `communication_period` should be required. If there's no communication period I think consumers should just show it when impactful.
> * On the flip side side, we see agencies try to use this for general informational messages that don't have a true impact on service. Should we conditionally require _either_ the communication _or_ the impact period (or both).
> * I also agree that we should deprecate `active_period`, but I'm not sure how that will work with backwards compatibility.

I still don't know why there is a need for `communication_period`. If the alert is in the feed, we should show it. If the agency doesn't want to show the alert, just leave it out in the feed.

Also, is it required that the impact period is totally within the communication period. Is it allowed to have impact period outside communication period?

Also, should the alert be shown in the following situations, assuming that the impact period is within the communication period?

1. If the current time is outside the communication period, and we are checking journeys within the impact period.
2. If the current time is outside the communication period, and we are checking journeys outside the impact period, but within the communication period.
3. If the current time is inside the communication period, and we are checking journeys outside the impact period, but within the communication period.
4. If the current time is inside the communication period, and we are checking journeys outside the communication period.

stevenmwhite:
>If the alert is in the feed, we should show it. If the agency doesn't want to show the alert, just leave it out in the feed.

I wouldn't be opposed to a spec update that says the dates of the active period are the dates it has an impact, and that an alert appearing in the feed should be considered an affirmative intent from the producer that it is currently within the desired communication timeframe. That would essentially achieve the same thing, but it's not quite how it's interpreted now.

As for your four specific questions/situations... I think those are things that are pretty dependent on the specific consumer and the specific place in their app where the user is at the time. Different consumers may handle it differently, because their UX is different. If you are on a screen, for example, where you are looking at a route _without_ planning a trip. I would suggest that the alert should be shown during the communication period. But if you're on a screen in the app that is specific to a future-planned trip that is outside of the impact period, then I probably wouldn't show the alert on that screen.

In general, I do believe that the details of the user experience should still be the responsibility of the consuming app developer, and not something that's completely explicit in the spec. The spec should allow for communication of the agency's intent.

skinkie:
Some people want to change the journey planner results based on the alerts. This alone is a reason to know about communication versus actual period.

miklcct:
> >If the alert is in the feed, we should show it. If the agency doesn't want to show the alert, just leave it out in the feed.
> 
> I wouldn't be opposed to a spec update that says the dates of the active period are the dates it has an impact, and that an alert appearing in the feed should be considered an affirmative intent from the producer that it is currently within the desired communication timeframe. That would essentially achieve the same thing, but it's not quite how it's interpreted now.
> 
> As for your four specific questions/situations... I think those are things that are pretty dependent on the specific consumer and the specific place in their app where the user is at the time. Different consumers may handle it differently, because their UX is different. If you are on a screen, for example, where you are looking at a route _without_ planning a trip. I would suggest that the alert should be shown during the communication period. But if you're on a screen in the app that is specific to a future-planned trip that is outside of the impact period, then I probably wouldn't show the alert on that screen.
> 
> In general, I do believe that the details of the user experience should still be the responsibility of the consuming app developer, and not something that's completely explicit in the spec. The spec should allow for communication of the agency's intent.

So are you specifically leaving out the answers to my questions above in the spec?

In my use case, there is a journey planner, and there is also a screen where you can look for routes / stations and show their timetables without planning for a journey, and you are allowed to change the date / time in that screen.

Is the spec leaving out the behaviour on which alerts should be shown here?

westontrillium:
@doconnoronca
> Perhaps the impact period should be allowed even if there is no communication period so the producer doesn't have to support a communications period.

+1 to this. In the event that only Impact Period is defined, the alert should still be displayed, but only impact routing during the Impact Period.

But I think agencies will also want the option to have an alert with a Communication Period without the need to define the hard boundaries of an Impact Period, especially if that alert is for something more nebulous, like an expected severe weather event (pun likely intended). For other alerts, there may not really *be* an "impact." So I would actually be in support of:
- Active Period: forbidden if communication OR impact period is defined, optional otherwise
- Communication Period: Optional
- Impact Period: Optional

@stevenmwhite

> I also agree that we should deprecate active_period, but I'm not sure how that will work with backwards compatibility.

I don't think it would work, strictly speaking. Unless deprecated means we all just don't talk about it. :sweat_smile: I think it would have to be here to stay until the versioning question is resolved, yes?

@miklcct

> I still don't know why there is a need for communication_period. If the alert is in the feed, we should show it. If the agency doesn't want to show the alert, just leave it out in the feed.

Something else to consider regarding the usefulness of Communication Period is that an agency might not have staff available 24/7 to add an alert at any given moment. Such agencies may want the ability to "queue" up an alert by adding it to their feed ahead of time with a Communication Period for when they actually want it to be displayed in apps.

> Also, is it required that the impact period is totally within the communication period. Is it allowed to have impact period outside communication period?

I think there are too many possible scenarios to establish explicit rules around how Communication and Impact Periods' respective time ranges must relate to one another. I'd rather allow agencies more flexibility on how they communicate with riders and keep the responsibility of conveying information in a logical way on them (anyway, you could still publish some very whacky alerts even with limitations on how impact/communication have to relate to one another).

> In my use case, there is a journey planner, and there is also a screen where you can look for routes / stations and show their timetables without planning for a journey, and you are allowed to change the date / time in that screen.

I think that if the user is using the screen to select the date/time, then if there is a Communication Period defined and the selected date/time falls within the Communication Period, the tool should show it. If the selected date/time falls outside the Communication Period, do not show the alert. If there is no Communication Period defined but the alert is in the feed, show it.

miklcct:
> @miklcct
> 
> > I still don't know why there is a need for communication_period. If the alert is in the feed, we should show it. If the agency doesn't want to show the alert, just leave it out in the feed.
> 
> Something else to consider regarding the usefulness of Communication Period is that an agency might not have staff available 24/7 to add an alert at any given moment. Such agencies may want the ability to "queue" up an alert by adding it to their feed ahead of time with a Communication Period for when they actually want it to be displayed in apps.

We are talking about GTFS-RT feeds here. Isn't GTFS-RT supposed to be generated real-time?

skinkie:
> We are talking about GTFS-RT feeds here. Isn't GTFS-RT supposed to be generated real-time?

It is, exactly at the moment that someone enters an notification, which is known from now on, but in effect at a later time.

stevenmwhite:
A couple points to make... Let's remember that GTFS is the standard for communicating information between two systems. It is not the feature specification for the features those systems can provide.

With that in mind...

@westontrillium 
> Something else to consider regarding the usefulness of Communication Period is that an agency might not have staff available 24/7 to add an alert at any given moment. Such agencies may want the ability to "queue" up an alert by adding it to their feed ahead of time with a Communication Period for when they actually want it to be displayed in apps.
People don't edit their GTFS-RT feeds directly. They use software to do so. It's totally reasonable for that software to have the "communication period" feature and use it to determine when to include the alert in the feed it produces. So I don't think that staff wanting to set up alerts ahead of when they want them communicated is necessarily a reason we have to have communication period in the app.

@miklcct 
>In my use case, there is a journey planner, and there is also a screen where you can look for routes / stations and show their timetables without planning for a journey, and you are allowed to change the date / time in that screen.
Sorry I've been slow to respond, but I agree with what @westontrillium said here about your use case. This is how _I_ would design it, but again, if this is your product, you get to decide.
> I think that if the user is using the screen to select the date/time, then if there is a Communication Period defined and the selected date/time falls within the Communication Period, the tool should show it. If the selected date/time falls outside the Communication Period, do not show the alert. If there is no Communication Period defined but the alert is in the feed, show it.

In general, my support for this is around clarifying for everyone what the _intent_ of the people publishing alerts is. In practice, these people who use software to publish alerts have shown that they intend to inform their riders about things at times that are in advance of those things having an actual impact, if they know about them in advance. They do this already by publishing alerts with `active_period` being greater than the actual impact period, and that confuses everyone.

I'm in support of @westontrillium's suggestion here:
- Active Period: Forbidden if Communication OR Impact Period is defined, optional otherwise
- Communication Period: Optional
- Impact Period: Optional

tzujenchanmbd:
Since this PR and #504, it may be worth checking the meaning of "deprecate" in the context of GTFS Realtime. Here are some of my findings:

In 2014, `TripDescriptor.ScheduleRelationship = REPLACEMENT` was completely removed from the spec.

In 2018, during discussions on experimental field process, the practice of adding the `[deprecated=true]` tag in .proto was proposed. (https://groups.google.com/g/gtfs-realtime/c/77c3WZrGBnI)

In 2019, the previously "deprecated" TripDescriptor.ScheduleRelationship = REPLACEMENT was added back in .proto with the `[deprecated=true]` tag. The main reason for this was for backwards compatibility. (https://github.com/google/transit/commit/cabaec43e039f3425c50933ae6b227a720c74851#diff-a1a14af3b4aed7d10a3b167231b52fdb483cf71b88f1655ed2e28bccd18800ceR579)

Since this approach allows any existing consumers or producers that have implemented support to leave their code as-is, this type of "deprecate" does not seem to be considered a breaking change but rather a strong recommendation? We can also observe this tendency in the discussion of #504 (e.g., [this comment](https://github.com/google/transit/pull/504#issuecomment-2588045842)).

If there is a consensus on this, we can follow the approach in #504 and include "deprecate guidance" in both the reference and .proto.

skalexch:
Thank you all for your interest in this change and for your valuable feedback.

@gcamp , @doconnoronca , @stevenmwhite  and @westontrillium  – on deprecating active_period, please check @tzujenchanmbd 's comment above.

@stevenmwhite , @doconnoronca , and @westontrillium  – your points regarding impact_period and communication_period are valid. I will make both fields optional while keeping restrictions on active_period to ensure it doesn’t interact with any new fields.

@miklcct - Restricting impact_period within communication_period (if present) seems like a good approach, so we can add it to the spec change.

Regarding the four cases you outlined, below is how they might play out. I say “might” because, as discussed, consumers determine where to display alerts in their app and how to handle them. The intent of this PR is to clarify when an alert is purely informative vs when it indicates an active impact.

Case 1: If the current time is outside the communication_period, but the trips fall within the impact_period, then they are effectively inside the communication_period as well. Consumers may choose to adjust routing suggestions accordingly and display the alert in results or details pages since it is in effect. However, if the consumer has a “current service alerts” page, the alert might not be shown there.

Case 2: If the current time is outside the communication_period, consumers may choose not to show the alert for informational purposes. Since the journeys are also outside the impact_period, they may not display an alert at all, as those journeys remain unaffected.

Case 3: If the current time is within the communication_period, but the trips are outside the impact_period, the alert can be displayed for informational purposes, though it won’t impact current trips.

Case 4: If the current time is within the communication_period, but the user searches for trips outside of it, the alert may appear on a general “current service alerts” page, but not necessarily in the journey results, this is up to the consumer’s implementation.

ckraatz:
Sorry to be late to the conversation, which I hope I'm sufficiently caught up on to comment. As a long-time publisher of alerts within a transit agency (Santa Clara VTA) and now with a tool ([SimplifyTransit Alerts](https://www.simplifytransit.com/alerts)) used by agencies of all sizes to publish GTFS-realtime Alerts:

- I do see agencies being confused about what the time period refers to. I always emphasize for them that it corresponds to the time range that the Effect is impacting riders. The simplest solution would be to add more clarity to the reference docs to emphasize that to feed publishers.
- I think there should only be one field - rename it impact_period if you want. No need for communication_period. It's unreasonable to expect agency staff (even at large agencies) to go through the complexity of adding multiple time ranges. As the creator of a UI to make publishing GTFS-realtime Alerts easy, that also creates unnecessary design complexity. We sometimes have dispatchers creating alerts, people who are moving very quickly, and we aim to simplify rather than complicate.
- Apps and other consumers should be responsible and free to decide when and how to display alerts, and assume that the time range (especially if there's only one, as I think is appropriate) provided corresponds to when the Effect is/will be affecting riders.

That said, SimplifyTransit also has a product that consumes GTFS-realtime Service Alerts, our [Subscriptions](https://www.simplifytransit.com/subscriptions) module that sends highly personalized SMS/email alerts to riders. 

- I could see Subscriptions using a communication_period, if it were provided reliably, to give more sophisticated and staffed-up agencies control over when to send SMS/email alerts or reminders. However, that use-case is even more complex. It would be great to be able to send subscribers periodic reminders as a service change approaches, for example, or when an Alert ends. But I don't see the communication_period supporting those uses cases, as proposed.

I hope that's helpful. Happy to clarify anything or fill in gaps I missed.

tzujenchanmbd:
There has been considerable debate in #482 regarding whether the current `active_period` should be interpreted as "the period during which riders are actually impacted" or "the period during which the producer intends to notify riders." @ckraatz you can find more information there.

In theory, the simplest and most ideal approach for ambiguous fields in the Realtime spec would be to directly clarify the intended definition. However, based on recent discussions about ambiguous fields in Realtime, we can observe that this is extremely difficult in practice. Any attempt to clarify the spec by adopting one interpretation inherently results in a breaking change for those who have already implemented it differently.

As a result, the approach of "deprecating the existing ambiguous fields in the protobuf sense and introducing new, unambiguous alternatives" seems to be the compromise that the community can currently accept. Admittedly, this approach is "imperfect" for everyone. (For more background on this deprecation approach, see this [comment](https://github.com/google/transit/pull/546#issuecomment-2776803681).)

From the rider’s perspective, receiving alerts before the actual impact period begins (based on `communication_period`) can be valuable in certain situations, as it allows riders to make decisions in advance. In the current change proposal, both `communication_period` and `impact_period` are `optional`. Producers can choose to provide only one of them, or ideally, both; and consumers can decide how best to present the information to riders (such as showing alerts and/or trip planning adjustments) based on the data received. We have also added a requirement that `impact_period` must be within (or equal to) `communication_period` to ensure that consumers do not have to handle cases where there is "no alert notification during the impact period."

skalexch:
Based on the conversation above, I added [deprecated=true] to `active_period`. I also added the new enums to the .proto file.

laurentg:
@tzujenchanmbd 
> Since this approach allows any existing consumers or producers that have implemented support to leave their code as-is, this type of "deprecate" does not seem to be considered a breaking change but rather a strong recommendation?

Strongly agree with this point. At Mecatran, as a transit alert data producer for many different consumers, we have to ensure full backward compatibility for all, regardless of their support for this change. To support this new version, I see two ways of handling this:
1. produce a new "end-point" with the new version only, keeping the existing version. This would make both old and new available. With this we have to ask the consumers handling this new format to switch to a new URL.
2. produce a single data feed with the three fields present (active, communication, impact) at the same time (at least for a certain amount of time, until every consumer support the new version). This would be way easier. This feed would not be strictly following the standard however, as currently stated.

To be honest, I'd strongly prefer to follow option 2: so far the evolution of the GTFS-RT specs never forced us to version our APIs, and option 1 would require a new GTFS-RT version (this being not talked about here?). Option 2 would require to relax indeed a bit the possibility of mixing the 3 fields.

Maybe the standard could express this, to allow mixing active with impact for backward-compatibility purpose? Stating that when the two version are present at the same time, the new version taking precedence over the old one.

ckraatz:
This proposal has grown on me! The addition of communication_period will give our customers more control over when alerts are sent via SMS/email or published on their website, which they've appreciated hearing about.

SimplifyTransit is ready to join in the testing phase if that's next, and implement if there are no major obstacles.

> Maybe the standard could express this, to allow mixing active with impact for backward-compatibility purpose? Stating that when the two version are present at the same time, the new version taking precedence over the old one.

How long would we continue to have all three fields in Alerts feeds? One year? I don't think including "deprecated" fields in active feeds should go on indefinitely.

If we need to publish all three fields for backwards compatibility, SimplifyTransit would only give our customers impact_period and active_period values would be identical to impact_period. Would that be acceptable to Consumers? Any problems with that approach?

We've always trained our customers that active_period is when the Effect is impacting riders, so they'll enter impact_period the same way they enter active_period today. They won't really know the difference.

> a migration guide added into the specification.

@miklcct could a "migration guide" specify a sunset date for active_period for all Producers, giving Consumers a reasonable amount of time to adapt? One year?

ckraatz:
@miklcct that’s perfect! You would do that regardless of when your user looks up that future journey, so you’re interpreting active_period as I understand it. That’s when the Effect is happening, but there’s no reason to hide it from riders before then and many reasons to show it to them.

I honestly think the main benefit of adding communication_period will be helping agencies control website displays and our automated SMS/email. It will likely have minimal impact on apps (in large part because many agencies won’t practically spend the time creating distinct communication_periods).

miklcct:
The migration guide should say that new clients should ignore `active_period` if at least one of the new field exists.

felixguendling:
Is there any reason why this problem is not handled with semantic versioning like in any other versioned protocol or API? Version X deprecates the old field(s) and introduces the new field(s). Then version Y removes those previously deprecated fields after a grace period of M months/years/versions.

The producer clearly indicates which version is used which makes it easy for the consumer to know which fields to expect. If those fields are not present, the message is not valid and will therefore be discarded. Some protocols use two versions (compatibility version + target version, names can differ) to indicate "this message targets *`target_version`* is compatible down to *`compatibility_version`*". Therefore the consumer software can directly discard messages that are too old or too new.

GTFS-RT seems to have a version already, but doesn't seem to be used:
https://github.com/google/transit/blob/1540c87c52b4f10ed88861d5b7e154afbfa97f2a/gtfs-realtime/proto/gtfs-realtime.proto#L56-L58

ckraatz:
@skalexch I'm neutral as to whether there are three fields active (impact_period and communication_period prioritized, active_period deprecated and discouraged) or a redefinition of active_period and addition of communication_period. I see your points about backwards compatibility and clarity and would vote +1 for either option.

> Is there any reason why this problem is not handled with semantic versioning like in any other versioned protocol or API?

I strongly agree with @felixguendling that the deprecation and backwards compatibility dilemma's here would be resolved simply by a versioning protocol. It's not on @skalexch or any of us to solve in this PR, and requires a bigger conversation. As far as I can tell, it gets brought up but no one does anything about it.

Another angle: are there examples of technology standards/protocols/languages/frameworks used widely around the world that **do not** have any sort of versioning protocol? Are GTFS and GTFS-realtime alone in that? Yikes!

skalexch:
@miklcct @phil-swiftly @laurentg @felixguendling @ckraatz Thank you for the feedback!

Regarding versioning, this topic has been previously discussed and remains a point of debate within GTFS (see [PR#504](https://github.com/google/transit/pull/504) for an example). To continue the conversation, we’ve created a dedicated [issue](https://github.com/google/transit/issues/530) to focus specifically on versioning.

For the current proposal, I removed the “forbidden” restriction for `active_period` and added a migration guide explaining how to handle the three fields. Please review the migration guide and let me know if anything is missing or unclear.

knr2000:
In Switzerland, we use SIRI-SX for the national exchange of event data.

SIRI-SX has 1.n ValidityPeriods and 0.n PublicationWindow.

In addition, there are 3 perspectives:
1. general (timetable information/routing)
2. stopPoint (passenger at the stop)
3. vehicleJourney (passenger in the vehicle)


All SIRI-SX events are then converted into GTFS SA and published centrally for the entire Swiss public transport system.

According to the proposal, a planned SIRI-SX event (e.g. a construction site) with two perspectives could then be converted or interpreted by GTFS SA as follows:

- general: The event should only be displayed in the timetable during the event duration => ValidityPeriod is converted to impact_period and only published then

- stopPoint: The event should also be displayed as advance information 3 days in advance at the stop => PublicationWindow is converted to communication_period and only published then

Unfortunately, the solution discussed here only partially solves our problem. For us, it is imperative that it be possible to map the perspectives. Otherwise, we have to provide 3 separate APIs, i.e. one for each of the 3 perspectives.


This is because the following case cannot be displayed correctly:

In practice, SIRI-SX often publishes an event with customer information for the timetable (Perspective=general / entire line / e.g. ‘Construction work on line 1’) and more specific customer information per stop (Perspective=stopPoint / e.g. ‘Due to construction work on line 6, stop XY has been moved’).
Even with the CR, it would still not be clear to recipients in which perspective an event has to be published.

knr2000:
@skalexch

Thank you for your feedback.

I just realised that I made a small mistake in my example. Both events concern the same line 1.

With your suggestion, both events would be displayed to the passenger at stop_id_XY on line route_id_line_1.

And that is exactly what we want to avoid!

Only one event should be displayed in the routing system and only one event should be displayed on the display at the bus stop.

Passengers have different information needs depending on their perspective.

I hope my requirement is now clear :-)

Do you have any idea how we can solve this?

johnhelmuth-st:
> @skalexch
> 
> Thank you for your feedback.
> 
> I just realised that I made a small mistake in my example. Both events concern the same line 1.
> 
> With your suggestion, both events would be displayed to the passenger at stop_id_XY on line route_id_line_1.
> 
> And that is exactly what we want to avoid!
> 
> Only one event should be displayed in the routing system and only one event should be displayed on the display at the bus stop.
> 
> Passengers have different information needs depending on their perspective.
> 
> I hope my requirement is now clear :-)
> 
> Do you have any idea how we can solve this?

From your description and example, SIRI-SX has a single data structure (the event) with 1 to 3(?) sub-data-structures (the perspectives.) The perspectives have separate human readable messages and time periods, and (probably) share other metadata about the event, e.g. to what type of real world entity it should be applied (route/schedule signage/pages, stop signage/pages, etc.)

GTFS-RT does not have that hierarchy.  Perhaps mapping the perspectives directly to the GTFS-RT service alert on a 1 to 1 basis, might solve that.

Given that... having one GTFS-RT SA with an informed_entity with the route_id only, and the other with an informed_entity with the stop_id only (and having slightly different messages about the event,) could end up with the stop contexts showing both alerts... depending on how the consuming app controlling the signage determines which alerts to display.

It might help if there was a way to group GTFS-RT alerts together to show they are applicable to the same event, adding back in the SIRI-SX hierarchy of event -> perspective.

skalexch:
@knr2000 regarding the interpretation of different perspectives, @johnhelmuth-st's response sums it up. There is no hierarchy in GTFS-RT Service Alerts that would result in different perspective being shown at different times. Someone can create:
- Alert 1: Route perspective (`route_id` specified in `informed_entities`)
- Alert 2: Stop perspective (`route_id, stop_id` specified in `informed_entities`)
That would result in two alerts being shown in a trip planner. "Alert 1" would be shown when the route is looked up separately, while both alerts would be shown in the journey details page, unless the consumer has an internal mechanism that filters redundant alerts.

It definitely is worth discussing ways to group alerts together. In any case, the current proposal will not address this discrepancy. The proposal might however provide a direct mapping of time periods between GTFS-RT and SIRI-SX: `communication_period` -> `PublicationWindow`, `impact_period` -> `ValidityPeriod`.

ckraatz:
Must a GTFS-realtime data change account for how it interoperates with a completely separate and distinct data standard? I understand it _could_, and perhaps some would argue it _should_. Is this a requirement?

We haven’t had capacity yet, but we’re planning to test this because it has advantages in the GTFS-realtime paradigm. I wouldn’t support changes that make it align with another data standard AND interfere with the benefits in GTFS-realtime.

I’m more interested in making this parallel and consistent with ideas inside GTFS-realtime, like disambiguating informed_entity with communication_entity and impacted_entity fields.

To be clear, I’m all for walk-across bridges and translations between the data standards. I think there will be limits to that based on the paradigms they operate in.

ckraatz:
I’m excited to test and adopt this change in SimplifyTransit Alerts. We need to focus on some important backlog tasks right now, and I can’t commit to a timeline till I know those are complete.

skinkie:
> Switzerland would also produce/consume it, if it shows up. @skinkie : What for the Netherlands?

When SIRI-SX is available we can, until then it is not part of the current travel information.

sebastianknopf:
As long as VPE (Verkehrsverbund Pforzheim-Enzkreis) in Southern Germany exists, we'd also be able to adapt our implementation for testing. 

Actually, we're also converting SIRI-SX to GTFS-RT ServiceAlerts. As the effect modifies the results in Google Maps (e.g. with `NO_SERVICE` for a certain stop leading to supress all departures for that stop), we decided to publish the message in the feed, but set active_period to the value of SIRI-SX ValidityPeriod. That also means, that we could not inform passengers in Google Maps prior to the entire event, but our other consuming systems show the information as it is published. This PR would clarify this issue. 

As I am working on a complete opensource re-implementation for using it in future (even without VPE), I'd add the new proposed fields and let active_period in there for backwards compatibility.

skalexch:
@sebastianknopf glad to hear this proposal can support you! When do you think you'll have support for those two fields?

sebastianknopf:
I'd carefully say in around two weeks. Maybe a little earlier in my staging system with fictive alerts.

sebastianknopf:
My staging system has the first draft installed now and exposes currently one test alert. 

The matching GTFS static feed is here: https://gtfs.skc-hub.app/data/internal/static/gtfs.zip

The GTFS RT feed is here: https://echogtfs.skc-hub.app/api/realtime/service-alerts.pbf

Simply add `?json` to see the output as JSON. Feel free to ask for more examples, if you need. I will internally test some things and update the prod system to the draft to see whether Google Transit also works with the extended period support with backwards compatibility in `active_period`.

felixguendling:
MOTIS supports it now (part of next release).

skalexch:
Hi everyone, any implementation updates to share?

@phil-swiftly do you have a clearer timeline for when the fields will be implemented?
@felixguendling once the feature is released in MOTIS, could you please include a link to the release notes or any examples of the feature in action?

@sebastianknopf, a couple of questions about your implementation:
- As a producer, what organization do you represent? And are you the official producer for Verkehrsverbund Pforzheim-Enzkreis?
- Is the feed you shared the complete GTFS-RT Service Alerts feed for the agency? Or just a sample of it?

sebastianknopf:
@skalexch I'm working currently as external employee / consultant for Verkehrsverbund Pforzheim-Enzkreis. The agency is producing the GTFS data themselves, but I am responsible for the data pipeline. The agency is going to be dissolved with the end of this year, but as for now, we're still able to produce data.

What I just shared was my staging system for tests, you also can access the official OpenData offers from VPE at https://opendata.vpe.de - Here you'll see all service alerts which are also published to Google Maps.

felixguendling:
> @felixguendling once the feature is released in MOTIS, could you please include a link to the release notes or any examples of the feature in action?

It's available in MOTIS [v2.10](https://github.com/motis-project/motis/releases/tag/v2.10.0). There is no mention of it in the release notes because MOTIS always distinguished between communication period and impact period, so the only change is now that those fields are filled properly when working with GTFS and not just set to the same value. v2.10 is live on [Transitous](https://transitous.org).

skalexch:
@sebastianknopf @felixguendling Thank you both for clarifying! I set up a MOTIS instance and tried VPE's feed. The result looks great!
Screenshot 2026-06-02 at 5 15 11 PM
Screenshot 2026-06-02 at 5 14 23 PM

skalexch:
This proposal has been implemented by:
- MOTIS (consumer) @felixguendling 
- Verkehrsverbund Pforzheim-Enzkreis (producer) @sebastianknopf 

We have enough to proceed. We are **starting a vote** to add these two fields ~~as _experimental_~~ **formally to the spec**.

The voting period will span 2 weeks and will end at  **2026-06-25 at 23:59:59 UTC.**

felixguendling:
I am against making them experimental.

Because at the same time you kept this for the old field:

> Should not be used - for backwards-compatibility only.

So now we have
- one deprecated ("should not be used")
- and two experimental

That's not a good state.

Protobuf is in this case
- upward compatible: a old client will still understand the new way as long as `active_period` is still provided
- downward compatible: a new client will still understand the old data as long as it still parses `active_period`

So the right way to proceed would be to recommend
- producers to produce all three fields
- and consumers to prioritize `communication_period` and `impact_period` but keep support for `active_period`
- and then remove the experimental remark (because this will just hinder adoption)

skalexch:
@felixguendling I removed the experimental tags.

We will try to adopt this directly formally and see what the community thinks in the vote. In all cases, the migration guide makes it possible to continue producing the deprecated field.

We are also extending the vote by another week. So the voting now ends on **2026-06-25 at 23:59:59 UTC**.

felixguendling:
+1 MOTIS

gcamp:
+1 Transit

ue71603:
+1

westontrillium:
+1 Optibus

BodoMinea:
+1

doconnoronca:
TransSee (consumer) +1

stevenmwhite:
+1 GMV (producer)

knr2000:
+1 SKI/SBB (producer)

haeckerbaer:
+1

ckraatz:
+1 SimplifyTransit (producer)

phil-swiftly:
+1 Swiftly

miklcct:
+1

skalexch:
The vote passed with 12 votes in support and no votes against.

Thank you everyone for your continuous feedback on the proposal! We will merge this soon.

PR #556, opened 2025-04-22
[GTFS-Fares v2] Add Distance-Based Fares

skalexch:
This PR adds distance-based fares, it covers two use cases:

- Leg-by-leg distance fare.
- Total journey distance fare (eg: TfNSW)

The PR has a few adjustments to the spec:
1. Add `fare_distance_units_traveled` in `stop_times.txt`
2. Add fields for matching `min_distance_units` abd `max_distance_units` in `fare_leg_rules.txt`
3. Add `min_distance_units` and `max_distance_units` to the primary key of `fare_leg_rules.txt`
4. Adjust `fare_leg_join_rules.txt` to cover distance-based legs ~~and include `duration_limit` and `duration_limit_type` (similar to `fare_transfer_rules.txt`) to limit the fare leg within a time window.~~
5. Restrict joining legs so that distance-based legs cannot be joined with non-distance-based legs.

**Update (2026-06-03): Adding `duration_limit` and `duration_limit_type` to `fare_leg_join_rules.txt` is now in its own proposal https://github.com/google/transit/pull/639**

For more details and examples, please refer to the distance-based fares [proposal](https://docs.google.com/document/d/1HkJ0pGqywC18k4YnNrHZ71KDmibwrp09SIfzF_uvnyk/edit?usp=sharing)

The proposal doc includes an [example](https://docs.google.com/document/d/1HkJ0pGqywC18k4YnNrHZ71KDmibwrp09SIfzF_uvnyk/edit?tab=t.0#heading=h.gfjwvbr6w9i7) showcasing how distance-based fares for TfNSW can be implemented in Fares v2.

**Note**: A previous version of this PR separated the leg join mechanism for distance-based legs in a file called `fare_leg_distance_rules.txt` to cover distance-based fares of the Netherlands ([Miro](https://miro.com/app/board/uXjVLhdrOJk=/?share_link_id=656457479800)). However, the mechanism was still insufficient to fully model the fare structure. Therefore, we simplified the PR to be able to model most fares around the world using `fare_leg_rules.txt` and `fare_leg_join_rules.txt`.


skalexch:
@miklcct thank you for your remark. After reviewing the modeling for the Netherlands' fares, you are absolutely correct. It turns out that the current proposal misses a key piece in the fare calculation.

As referenced in the [notes](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit?tab=t.0#heading=h.6pydxswsohyw) for the 26/11/2024 Fares WGM, @skinkie shared an example with an equation explaining how the fare works.

E.g. in the case of a GVB (4 km) + Arriva (2 km) trip, the fare should be calculated as follows:
GVB_fare(4km) + Arriva_fare(6km) - Arriva_fare(4km).

The current proposal overlooked ` - Arriva_fare(4km)`.

We'll write something up to address this and try to figure out where this complex case fits in the proposal, or if other alternatives can be considered to address it. We'll share it soon. @skinkie can also probably provide more context into how NL fares are calculated.

skalexch:
Hello @miklcct . As promised, we have created a document exploring the possibilities of modelling the Netherlands’ case and the limitations of Fares v2 in this context.

The main outcomes are as follows:
- It might be possible to model the fare structure, but this approach would lead to the same bloated `fare_transfer_rules.txt` file that we tried to avoid by using `fare_leg_distance_rules.txt`.
- We suggest focusing only on modelling the first case (total journey pricing) and implementing it in `fare_leg_join_rules.txt`. This will simplify the solution.
- As for the Netherlands, it would be valuable to consider a specific extension that models its fare structure.

The full investigation is available [here](https://docs.google.com/document/d/13zcemcolnFamC5MQ_8F_oK9xaV8GF50-Pc5S9mtRr8A/edit?usp=sharing). This is also relevant for @skinkie to take a look at.

We can discuss this at the next working group meeting as well.

skalexch:
Update: The addition of `duration_limit` and `duration_limit_type` to `fare_leg_join_rules.txt` is now in its own proposal.
https://github.com/google/transit/pull/639

PR #558, opened 2025-04-29
[Same as #528] Clarification on continuous pickup/dropoff value

tzujenchanmbd:
[PR#528](https://github.com/google/transit/pull/528) voting passed on Apr 7. This is the exact same change just to avoid unsuccessful check in #528.

@Sergiodero could you help check if this is exact same as 528 and we can merge.


Sergiodero:
Changes LGTM (same as #528), I'll merge them now.

PR #561, opened 2025-05-09
[GTFS Fares v2] Semantics clarification

tzujenchanmbd:
Based on recent discussions in the gtfs-fares Slack channel and working group, clarify some fares v2 related semantics including:

- Introduce a `Local Time` data type to eliminate potential confusion caused by the existing `Time` data type in timeframes.txt.
- Add "effective fare leg" to the general description of fare_transfer_rules.txt
- Add "timer should start from the first matched leg" clarification

Tagging active working group members for review🙏
@felixguendling @westontrillium @jfabi @skinkie @halbertram @evansiroky @miklcct @jll01 @bdferris-v2


gtfs/spec/en/reference.md
```diff
@@ -538,7 +539,7 @@ Fare rules for transfers between legs of travel defined in [`fare_leg_rules.txt`
 
 To process the cost of a multi-leg journey:
 
-1. The applicable fare leg groups defined in `fare_leg_rules.txt` should be determined for all individual legs of travel based on the rider’s journey.
+1. The applicable fare leg groups defined in `fare_leg_rules.txt` should be determined for all individual legs or effective fare legs of travel based on the rider’s journey.
```
felixguendling, line 542:
```suggestion
1. The applicable fare leg groups defined in `fare_leg_rules.txt` should be determined for all effective fare legs of travel based on the rider’s journey.
```

Depending on how an "effective fare leg" is defined, it could also be an individual leg. It's a list of 1 to N joined legs. So always just referring to "effective fare leg" (without mentioning "individual legs") would be fine and would allow for a shorter documentation (which usually equals to better readability IMO).

gtfs/spec/en/reference.md
```diff
@@ -559,7 +560,7 @@ To process the cost of a multi-leg journey:
...
+| `duration_limit_type` | Enum | **Conditionally Required** | Defines the relative start and end of `fare_transfer_rules.duration_limit`.<br><br>Valid options are:<br>`0` - Between the departure fare validation of the current leg and the arrival fare validation of the next leg.<br>`1` - Between the departure fare validation of the current leg and the departure fare validation of the next leg.<br>`2` - Between the arrival fare validation of the current leg and the departure fare validation of the next leg.<br>`3` - Between the arrival fare validation of the current leg and the arrival fare validation of the next leg.<br><br>When a transfer rule with the same `from_leg_group_id` and `to_leg_group_id` is matched multiple times consecutively within a multi-leg journey, the `duration_limit` specified by the rule should be measured starting from the first matched leg.<br><br>Conditionally Required:<br>- **Required** if `fare_transfer_rules.duration_limit` is defined.<br>- **Forbidden** if `fare_transfer_rules.duration_limit` is empty. |
```
felixguendling, line 563:
```suggestion
| `duration_limit_type` | Enum | **Conditionally Required** | Defines the relative start and end of `fare_transfer_rules.duration_limit`.<br><br>Valid options are:<br>`0` - Between the departure fare validation of the first matched leg and the arrival fare validation of the next leg.<br>`1` - Between the departure fare validation of the first matched leg and the departure fare validation of the next leg.<br>`2` - Between the arrival fare validation of the first matched leg and the departure fare validation of the next leg.<br>`3` - Between the arrival fare validation of the first matched leg and the arrival fare validation of the next leg.<br><br>When a transfer rule with the same `from_leg_group_id` and `to_leg_group_id` is matched multiple times consecutively within a multi-leg journey, the `duration_limit` specified by the rule should be measured starting from the first matched leg.<br><br>Conditionally Required:<br>- **Required** if `fare_transfer_rules.duration_limit` is defined.<br>- **Forbidden** if `fare_transfer_rules.duration_limit` is empty. |
```

I would eliminate all occurrences of "current leg" and replace them with "first matched leg" to make it clear from the beginning.

gtfs/spec/en/reference.md
```diff
@@ -403,14 +404,14 @@ There must not be overlapping time intervals for the same `timeframe_group_id` a
...
+|  `end_time` | Local time | **Conditionally Required** |  Defines the end of a timeframe. The interval does not include the end time.<br> Values greater than `24:00:00` are forbidden. An empty value in `end_time` is considered `24:00:00`. <br><br> Conditionally Required:<br> - **Required** if `timeframes.start_time` is defined.<br> - **Forbidden** otherwise |
 | `service_id` | Foreign ID referencing `calendar.service_id` or `calendar_dates.service_id` | **Required** | Identifies a set of dates that a timeframe is in effect. |
 
 #### Timeframe Local Time Semantics
 - When evaluating a fare event’s time against [timeframes.txt](#timeframestxt), the event time is computed in local time using the local timezone, as determined by the `stop_timezone`, if specified, of the stop or parent station for the fare event. If not specified, the feed’s agency timezone should be used instead.
```
felixguendling, line 412: This would require that the feed has only one agency or at least all agencies to have the same timezone to be non-ambiguous in case there's no stop timezone.

felixguendling (commented):
Thank you for addressing this and improving the standard! :+1:

gtfs/spec/en/reference.md
miklcct: The specification falls back to the agency time zone, so if there is no stop time zone, it should fallback to the actual agency used.

For example, if there are two agencies, A work on UTC+0, B work on UTC+1, and the current time is UTC 00:30 with a timeframe which start at 01:00 local time, my understanding is that it is in the timeframe if you use a service of B, but not a service of A.

gtfs/spec/en/reference.md
bdferris-v2: Effective Fare Leg is currently defined as "A sub-journey of two or more legs" so indeed, this would need to change to support the semantics you've proposed.  That said, I think I tend to lean towards the status-quo definition.  I think there is value in distinguishing between an individual leg and an effective leg in the spec, as it usually more directly highlights in the text where you need to thing about how an individual leg vs effective leg impacts semantics.

gtfs/spec/en/reference.md
bdferris-v2: Per the spec for `agency_timezone`, if multiple agencies are specified in the dataset, each must have the same agency_timezone.  So there should only ever be a single "feed"-level timezone either way.

gtfs/spec/en/reference.md
bdferris-v2: I'd go further and replace both "current"+"next" with "first"+"last".  "First matched leg" is ok but wondering if "first leg in transfer sub-journey" would clearer?  (Not sure if I feel too strongly on that point)

felixguendling:
+1 MOTIS

jll01:
+1 Transit

bdferris-v2:
+1 Google

felixguendling:
+1 MOTIS

PR #562, opened 2025-05-17
Update stop_times.txt in sample-feed-1 to be RFC 4180-compliant

helomri:
According to [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180#section-2), the spec used in the GTFS format, every single record in a CSV file must contain the same amount of fields.

The following file has records of length 9 but from line 17 onward, only 5 fields are defined:
https://github.com/google/transit/blob/daa60b4c500534b1a193a499793da3624c3eeb82/gtfs/spec/en/examples/sample-feed-1/stop_times.txt#L1-L29

This (rudimentary) example feed is part of my test suite for my library which makes some of them fail because of this odd formatting. Since it is unusual behavior and is directly against reference documents, I prefer not to fit this capability in my software as it could accept partial GTFS feeds as valid and provide erroneous data in consequence.


Sergiodero:
Hi @helomri 👋🏼 Thanks for this contribution, the changes look good! We normally tend to leave editorial changes open for a week in case anyone else wants to chime in, so we'll be merging this change early next week. Thanks again!

PR #567, opened 2025-06-15
Add addition best practice on hosting GTFS files

doconnoronca:
I sometimes run into problems on TransSee where my system to automatically download GTFS files with libcurl fails because the web server hosting it is blocking it.

Two of the reason this happens is because the web server only allow actual browsers to download files and because I am blocked from downloading it because they are excluding access to IP addresses other then ones in the United States from accessing it. I am in Canada.

This change addresses this by adding best practices advising against this.

### DIFF
--- a/gtfs/spec/en/reference.md
+++ b/gtfs/spec/en/reference.md
@@ -178,6 +178,7 @@ The following example demonstrates how a field value would appear in a comma-del
  * Old services (expired calendars) should be removed from the feed.
  * If a service modification will go into effect in 7 days or fewer, this service change should be expressed through a GTFS-realtime feed (service advisories or trip updates) rather than static GTFS dataset.
  * The web-server hosting GTFS data should be configured to correctly report the file modification date (see [HTTP/1.1 - Request for Comments 2616, under Section 14.29](https://tools.ietf.org/html/rfc2616#section-14.29)).
+ * The web-server should not attempt to prevent automated tools from downloading the GTFS files as they are often downloaded with automated tools rather than downloaded with a browser. It should not restrict access to a limited geographic area as organizations that work with GTFS data can be anywhere in the world. The web-server can still block requests that appear to be abusive but should not limit access excessively.
 
 ## Field Definitions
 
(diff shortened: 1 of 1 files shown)


skinkie:
While supportive, the statements have to do with reuse and open data. Is that within scope of the best practises?
How does your libcurl implementation handle last modified and etags? Also part of the best practise.

doconnoronca:
> While supportive, the statements have to do with reuse and open data. Is that within scope of the best practises?

There are other best practises related to how the file is hosted, like not requiring a login and supporting the file modification date.

 > How does your libcurl implementation handle last modified and etags? Also part of the best practise.

I use CURLOPT_TIMECONDITION and CURLOPT_TIMEVALUE to set a If-Modified-Since http header, which works for most sources.

tzujenchanmbd:
I wonder if the current use of negative phrasing (e.g., "should not") might come across as a bit strong, especially considering that the final decision rests with data producers. In some regions, producers may have "compelling reasons" (e.g., political considerations).

Would it make sense to rephrase these statements using more positive, recommendation-based language? This might better respect the position of producers while still conveying the best practices valued by the community.

skinkie:
If you are creating a _standard_ you must use standard language not "do whatever you want to do".

doconnoronca:
"Should not" is a phase with a [specific definition](https://gtfs.org/documentation/schedule/reference/#document-conventions) in the standard.

doconnoronca:
I'm going to call for a vote. The voting period will end on Aug 21, 2025 at 23:59:59 UTC.

nighthawk:
+1 SkedGo

stevenmwhite:
I agree with @skinkie on the delineation between the technical spec and the practice of hosting data in an open way... but isn't that why this is in the "best practices" and not in the spec itself? This seems to be like a reasonable best practice.

skinkie:
> I agree with @skinkie on the delineation between the technical spec and the practice of hosting data in an open way... but isn't that why this is in the "best practices" and not in the spec itself? This seems to be like a reasonable best practice.

Not trying to be stuborn. But from this point https://gtfs.org/documentation/schedule/schedule-best-practices/#practice-recommendations-organized-by-file onward the best practice becomes a guideline, with examples. Read: better than the "schema".

doconnoronca:
> Not trying to be stuborn. But from this point https://gtfs.org/documentation/schedule/schedule-best-practices/#practice-recommendations-organized-by-file onward the best practice becomes a guideline, with examples. Read: better than the "schema".

The whole page is best practices, which would be guidelines by definition. The spot you point out is just where best practices for individual files begin.

The best practices used to be a separate document but where merged into the specification to give them more prominence.

Issue #358, opened 2022-11-15
Defining a route type at the trip level


vpaturet:
The current GTFS specification allows for defining a route type ("route_type" in routes.txt) at the route level which applies to every trips in that route.
There are cases however where some trips in a given route are scheduled with an alternate transport mode. A typical example is replacement bus, where train trips are temporarily replaced by bus trips due to construction work on the railways.
From a passenger's perspective, these replacement bus trips belong to the original route.
There is no way to model this with the current GTFS specification.

MBTA came up with a solution for this use case: adding a custom field "trip_route_type" in trips.txt
See:
- https://groups.google.com/g/massdotdevelopers/c/M35VgMBM2qk/m/up6lYUTpCQAJ
- https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#tripstxt

Would it make sense to standardize this custom field "trip_route_type" in the GTFS specification?

jspetrak:
From the rail perspective, replacement bus is technically a train with commercial category bus, and it calls train stations, not bus stations. I know it is confusing.

If we allow to modify route data on trip level, I would think also about routeless trips which most of the trains are, at least in Europe. And there usually less than more successful workarounds for that.

hannesj:
Another possibility would be to mark a route to be part/replacement of another route. That way we could link those together and use the information to link the real-time alerts to both routes.

barbagus:
The French regional railways [TER](https://en.wikipedia.org/wiki/Transport_express_r%C3%A9gional) operates both trains and buses on regular routes all over the country. They do publish a GTFS dataset daily: [https://eu.ftp.opendatasoft.com/sncf/gtfs/export-ter-gtfs-last.zip](https://eu.ftp.opendatasoft.com/sncf/gtfs/export-ter-gtfs-last.zip)

Some routes are designed to be operated with both trains and buses depending on the trip. You can download a human readable (PDF) schedule for such a route [here](https://ter-fiches-horaires.sncf.fr/publish/A19%20et%20cars%20Colmar%20Metzeral%20%28du%2021%20ao%C3%83%C2%83%C3%82%C2%BBt%20au%209%20d%C3%83%C2%83%C3%82%C2%A9cembre%202023%29%20MAJ%201008.pdf) (fr).
[...]

PR #572, opened 2025-07-16
add trip_route_type into trips.txt GTFS static

miklcct:
## Summary
This proposal adds a new optional field, called `trip_route_type`, into `trips.txt`. It is currently in use by [MBTA](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#tripstxt) to denote trips inside a route where the type of transportation is not the same as the route itself.

## Describe the Problem
It is impossible, in the current GTFS standard, to denote a trip using a different type of transportation within a route, where NeTEx can describe it well by assigning a mode of transport at the trip level. For example, it is possible in NeTEx, but not GTFS, to denote a bus trip running on a tram route.

## Use Cases
If a rail replacement bus is associated as to a rail route, it can then be shown when querying the trips of the route such that travellers are aware of it. For example, in Stockholm, bus departures can be occasionally found on route 7, which is advertised as a tram route, intervened between tram departures.

It is not to be used when the rail replacement bus has a separate identity to the rail route (for example, in the London Underground, the replacement buses have separate codes, such as ML-1 when the tube line is closed, in this case they should be modelled as a separate route using the existing method)

## Proposed Solution
Add a field `trip_route_type` into `trips.txt` which specifies the type of transportation for the particular trip.

## Type of change

 

GTFS Schedule
- [x] Functional Change
- [ ] Non-Functional Change

GTFS Realtime
- [ ] Specification Change

## Additional Information
OpenTripPlanner is now [facing difficulty in achieving a standard-neutral treatment](https://github.com/opentripplanner/OpenTripPlanner/pull/6676#issuecomment-3018113097) of rail replacement buses because of the inability in GTFS to specify the same concept in NeTEx.

## Proposed Discussion Period
This is a small, non-breaking change, so 10 days should be enough for discussion.

## Testing Details

- Consumer(s): OpenTripPlanner (to be confirmed)
- Producer(s): MBTA
- Estimated Testing Period: 14 days

## Proposal Update Tracker 

| (YYYY-MM-DD) | (Brief description of the update) |

## Checklist


skinkie:
This is poor transport design, does not even work with _Transmodel_. The public name may be 7, it is likely never the system number.

miklcct:
> This is poor transport design, does not even work with _Transmodel_. The public name may be 7, it is likely never the system number.

GTFS is for passenger information and how the bus trips and the tram trips being represented in the internal system has completely no relevance that they are seen by the passengers and the operators as within a single route.

In NeTEx, it is modelled as a _Line_ with _TransportMode_ being `tram`, and some of the `Journey`s on the `JourneyPattern`s in the `Route`s belonging to the `Line` having the _TransportMode_ being `bus`.

miklcct:
> > GTFS is for passenger information and how the bus trips and the tram trips being represented in the internal system has completely no relevance that they are seen by the passengers and the operators as within a single route.
> 
> And for this reason, you could just make a new route_id, with the same route_short_name.

And this is a distortion of the reality where the operator and passenger both consider them to be the same route.

Sergiodero:
Thanks @miklcct, I noticed that you refer to MBTA as the producer for this change, but in issue #358 that you pointed out, @jfabi explained [why MBTA moved away from using this solution](https://github.com/google/transit/issues/358#issuecomment-1871315144) and the challenges associated with it. I wonder if this change is something MBTA would still support for the spec (not only for their feed and/or internal use), and if they would be ok formally playing the producer role in this PR.

Regardless, I would also strongly suggest to have this conversation at the issue level, since it seems like there is still some consensus needed to align on a potential solution based on the discussion in #358. 

As mentioned in the Governance framework:
> The Issue Stage is meant for discussing new ideas, identifying needs, and proposing improvements to the specification. Issues help evaluate the necessity and support for a change, while organizing the resources required to proceed to the Pull Request Stage. It’s recommended to start at the Issue stage to build consensus around new ideas. 


Maybe continuing the conversation in #358 could be the best option.

jfabi:
> Thanks @miklcct, I noticed that you refer to MBTA as the producer for this change, but in issue #358 that you pointed out, @jfabi explained [why MBTA moved away from using this solution](https://github.com/google/transit/issues/358#issuecomment-1871315144) and the challenges associated with it. I wonder if this change is something MBTA would still support for the spec (not only for their feed and/or internal use), and if they would be ok formally playing the producer role in this PR.
> 
> Regardless, I would also strongly suggest to have this conversation at the issue level, since it seems like there is still some consensus needed to align on a potential solution based on the discussion in #358.

Thanks to both @miklcct and @Sergiodero here—the comment I offered in https://github.com/google/transit/issues/358 still accurately describes the MBTA's position on this issue. While we still publish `trips.trip_route_type` in our feed, we no longer use it to its original intent; that is, we do not publish any cases where a trip's `trip_route_type` would not match the respective route's `route_type`.

We could be open to re-discussing our use if there was a wide consensus on an approach across consumers, but at this time we would not feel comfortable acting as the first producer.

miklcct:
> > Thanks @miklcct, I noticed that you refer to MBTA as the producer for this change, but in issue #358 that you pointed out, @jfabi explained [why MBTA moved away from using this solution](https://github.com/google/transit/issues/358#issuecomment-1871315144) and the challenges associated with it. I wonder if this change is something MBTA would still support for the spec (not only for their feed and/or internal use), and if they would be ok formally playing the producer role in this PR.
> > Regardless, I would also strongly suggest to have this conversation at the issue level, since it seems like there is still some consensus needed to align on a potential solution based on the discussion in #358.
> 
> Thanks to both @miklcct and @Sergiodero here—the comment I offered in #358 still accurately describes the MBTA's position on this issue. While we still publish `trips.trip_route_type` in our feed, we no longer use it to its original intent; that is, we do not publish any cases where a trip's `trip_route_type` would not match the respective route's `route_type`.
> 
> We could be open to re-discussing our use if there was a wide consensus on an approach across consumers, but at this time we would not feel comfortable acting as the first producer.

OK, I will further discuss with @tkalvas (HSL) to see if there are any examples appropriate for this PR for them to take on as a producer, or if HSL's need can be delivered via a better alternative way. The example in Stockholm is that both trams and buses call at the exact same stops and run on the exact same shape.

eliasmbd:
> OK, I will further discuss with @tkalvas @vpaturet (HSL) to see if there are any examples appropriate for this PR for them to take on as a producer,

This is a good consideration. Make sure to involve the first producers, they must represent themselves in this comment section when testing the changes and showing proof of testing.

Bertware:
> The bus trips of the Stockholm route 7 and the tram trips of the Stockholm route 7 are not in different routes.

The tram line is line 7, the bus line is line 970 designated 7. This PR would not affect this example, since they are different different lines. They are not grouped together either, the bus line belongs to a different group of lines (replacement buses) than the tram line (which is in its own group). It is not technically feasible for us to join these two lines into one GTFS route. From a policy point of view, Trafiklab does not want to deviate from the source structure defined by the PTA when converting to GTFS/NeTEx, which is another reason why we would not implement this. 

From a producer point of view, @Brickblock1 s note on split journeys due to different short_names would be more relevant, and a valid addition to any discussion on routes and trips. I'd say however that the current situation, where we as a producer split transmodel lines into different routes due to different designations is a good thing for data consumers, as these X, S, .., variants of lines are often treated as different lines when communicating with travellers.

Bertware:
That bus is technical line 970, designated 7 (in the same way one agency can have "bus line 1" in 5 different cities by using different technical line numbers, for example 1, 101, 201, 301, 401). The bus shown in the picture is, in the PTA data, on a different line than the trams. 

The bus is replacement/reinforcement traffic making up for a lack in available seating on the trams compared to the expected number of riders. They usually are used as replacement buses, but during the summer months they are scheduled continuously. Given that the replacement bus line already exists, it likely makes more sense (and is technically/operationally easier) to run the reinforcement traffic on the replacement line. Additionally, the reinforcement traffic is using the replacement bus line color. This means that, if the color were to be included in the GTFS feed, there would be a need for two different routes with their own `route_color`. even if route types could be defined on trip level.
{A27C8AC8-7EEE-42EC-88B6-6493A05119EB}


A deviation message is available in the realtime feed, and on the [PTA website](https://sl.se/reseplanering/trafiklaget?locationType=LINE&name=Sp%C3%A5rv%C3%A4g+city&number=7&designation=7&transportType=TRAM&transportSubType=SPARVAG_CITY&start=T-Centralen&end=Waldemarsudde).

miklcct:
> That bus is technical line 970, designated 7 (in the same way one agency can have "bus line 1" in 5 different cities by using different technical line numbers, for example 1, 101, 201, 301, 401). The bus shown in the picture is, in the PTA data, on a different line than the trams.
> 
> The bus is replacement/reinforcement traffic making up for a lack in available seating on the trams compared to the expected number of riders. They usually are used as replacement buses, but during the summer months they are scheduled continuously. Given that the replacement bus line already exists, it likely makes more sense (and is technically/operationally easier) to run the reinforcement traffic on the replacement line. Additionally, the reinforcement traffic is using the replacement bus line color. This means that, if the color were to be included in the GTFS feed, there would be a need for two different routes with their own `route_color`. even if route types could be defined on trip level. {A27C8AC8-7EEE-42EC-88B6-6493A05119EB}
> 
> A deviation message is available in the realtime feed, and on the [PTA website](https://sl.se/reseplanering/trafiklaget?locationType=LINE&name=Sp%C3%A5rv%C3%A4g+city&number=7&designation=7&transportType=TRAM&transportSubType=SPARVAG_CITY&start=T-Centralen&end=Waldemarsudde).

Thanks. This is convincing enough for me they are operationally two different routes, but this runs against the GTFS Best Practice that

> All trips on a given named route should reference the same route_id.
> Different directions of a route should not be separated into different route_id values.
> Different spans of operation of a route should not be separated into different route_id values. i.e. do not create different records in routes.txt for “Downtown AM” and “Downtown PM” services).

miklcct:
However it is extremely confusing to passengers if there are two different routes with the same public code in the same geographical area.

miklcct:
> Passengers have absolutely 0 to do with routes.txt.

It will result in your journey planner telling your partners to take _route 7 or route 7_ (with different background colours) to your destination, if both route 7 and route 7 serve the journey, and that becomes confusing.

skinkie:
It won't.
 1. If the operator chose the same name, why wouldn't they choose the same colours? 
 2. Generally don't operate on the same time (unless the new trips are actually reinforcement trips), they they don't compete.
 3. Routes never compete, trips compete.

PR #585, opened 2025-10-20
Add images.txt + agency logo

felixguendling:
## Summary
It would be nice to show brand logos for transit agencies in apps. Many apps already show logos but have to use a side channel because GTFS does not support logos yet. This proposal is based on the way it's modeled in GBFS (except that I propose to make it optional in GTFS). Since it works well there, it should work for GTFS as well.

For Transitous we are planning to add logos for agencies and my idea was to not build yet another side channel to do this (like probably all existing apps) but rather support the ecosystem by proposing a standard way.

Usually, the discussion happens in an issue, but new guidelines state that

> if the proposal’s scope is already well-defined, beginning directly at the Pull Request Stage is appropriate.

Since the scope is well-defined, I decided to follow @etienne0101's [request](https://github.com/google/transit/pull/76#issuecomment-3422531103).

**This is only a first draft.** More requirements like transparent background (which is basically the default for SVG, that's why I think this requirement might not be needed here), etc. or an additional URL for 16x16 rasterized images can be added as well (but this could also be done in a separate PR to keep discussion + scope small for this). Just let me know.

## Describe the Problem
Currently, apps can't show brand logos for transit agencies based on the GTFS feed and have to work around this via side channels.

## Use Cases
Apps showing brand logos to customers 

## Proposed Solution
Based on how GBFS handles it. Add a new `agency_brand_image_url` column to `agency.txt`.

## Type of change

 

GTFS Schedule
- [x] Functional Change
- [ ] Non-Functional Change

GTFS Realtime
- [ ] Specification Change

## Proposed Discussion Period
Since the issue https://github.com/google/transit/pull/76 didn't come to a consensus, discussion should take as long as we don't have consensus.

## Testing Details

- Consumer(s): TBD
- Producer(s): TBD
- Estimated Testing Period: TBD

## Proposal Update Tracker 

| (YYYY-MM-DD) | Add agency_brand_image_url to agency |


kona314:
Thanks for the PR, this would be a great addition. As you point out, lots of apps are having to invent their own versions of this, so there's clearly need. 

I think it's really important to define more requirements for using these fields: 
- Most agencies (if not all) have multiple versions of their logo. Which one should be used? 
  - Presumably, whatever is considered "main" or "primary" is the answer. My local transit agency (also my employer, but I'm here speaking for myself) has a very wide, rectangular logo, but the spec requires a square or a circle. Does this mean they should use the secondary, 1:1 aspect ratio variant of the logo? Or squeeze the primary, wide one into a square with a bunch of empty space on the top and bottom? 
  - Is it expected that it would usually/always be presented in a small format? This affects the amount of detail visible and what version of the logo would be appropriate. Many agencies have simplified versions for small presentation (e.g. favicons). 
  - To this end, is it appropriate to require a square or circle at all? 
- Some brand guidelines impose rules around the background for a logo, so we may see agencies insist on putting a non-transparent background in these images. I'm open to ideas, but I think we should at least state that it's recommended for agencies to publish with a transparent background. 
- How should caching work? 
  - I would argue the last modified column can/should be dropped entirely, and we modify the URL column's description to state that consumers should be expected to (1) check this URL at least as frequently as when ingesting a new version of the feed (though some may opt for more often) and (2) rely on standard HTTP headers for caching, just like for the feed itself. 
  - I recognize GBFS uses the last modified column, I don't know much about GBFS or why they would have chosen to adopt the approach they did, but this would seem to work well in a GTFS context. 
- I notice GBFS also has a field for a dark mode URL. Should we include something similar? 
- Should there be some mention of what the terms are/are not recommended to include? 
  - Presumably, if the agency is including a logo URL, there should be some understanding that developers are allowed to use it for _something_, but we also don't want consumers assuming they have full rights to the image when the terms may be more nuanced. I might even go so far as to say including a logo URL should imply certain terms, but I recognize this quickly gets messy to define.

felixguendling:
I removed the last modified column and stated that consumers should respect standard HTTP caching headers. I think that makes it clear that files should not be downloaded before `max-age` expired?

Thank's again for your feedback.

Let me know if there's anything else!

felixguendling:
- I added the following to both image URLs (default + dark)
  > A data consumer MUST store the brand image on their own servers to avoid direct traffic from end user devices to the agencies web servers.

  Regarding which web server serves the icon: I would propose to put the burden on the data consumer. I would like to make it as easy as possible for the producer and eliminate all barriers that could be used as an argument against publishing feeds with logos. Since traffic can always be an argument against adding the logo, the consumer should host the logos. But I am open to other opinions.
- Thank you very much for your proposal for usage conditions! I took it verbatim. We can change it in case someone has ideas for improvement. Maybe your proposed terms could be states as "this is the default if you don't set `agency_brand_terms_url`"? Something like the following?
  > If `agency_brand_terms_url` is not set, this permits feed consumers, without taking additional actions, to use the logo in reference to the agency's services broadly throughout their products
- Regarding extending to
  - `agency_brand_image_url_main`
  - `agency_brand_image_url_square`
  - `agency_brand_image_url_16x16`
  - `agency_brand_image_url_dark_main`
  - `agency_brand_image_url_dark_square`
  - `agency_brand_image_url_dark_16x16`
 
  I would like to wait for more feedback (e.g. experiences from those who already developed apps with logos) to get a broader view on this topic.

  I think that already the 1:1 SVG with a light+dark version would bring a lot of value to client apps and having six different URLs could become overwhelming for small agencies. But I definitely see value of a 16x16 logo that can be displayed for example on a map.

skinkie:
-1

Introduce a branding.txt, have a branding_id, give that all the properties that you propose, refer branding_id from agency.txt, routes.txt, trips.txt

Make it clear what branding is, it is rider facing, signatures, colours, names, etc.

That will make agency.txt a true _operator_ thing again.

miklcct:
> -1
> 
> Introduce a branding.txt, have a branding_id, give that all the properties that you propose, refer branding_id from agency.txt, routes.txt, trips.txt
> 
> Make it clear what branding is, it is rider facing, signatures, colours, names, etc.
> 
> That will make agency.txt a true _operator_ thing again.

agency.id should not be an operator thing. For example, London Buses are operated by a large number of operators but it is clearly one agency.

skinkie:
> Your point of introducing a new file is valid but can you please don't confuse the matter by mentioning operators?

The problem is GTFS combining distinct concepts. When we discuss these concepts we must make a distinction. Currently agency.txt is used as a way to differentiate on _branding_. For example FlixTech is an operator, their GTFS consists of FlixBus and FlixTrain. That is something artificial to facilitate branding on Google Maps.

> London Buses have a different brand for its SL and BL prefixed routes compared to ordinary routes, and it is completely irrelevant to which operator runs any of its routes.

Who do you call for _lost and found_?

gtfs/spec/en/reference.md
```diff
@@ -182,6 +182,22 @@ The following example demonstrates how a field value would appear in a comma-del
...
+|  `brand_url` | URL | Optional |  URL of the transit brand. |
+|  `brand_phone` | Phone number | Optional | A voice telephone number for the specified brand. This field is a string value that presents the telephone number as typical for the brand's service area. It may contain punctuation marks to group the digits of the number. Dialable text (for example, TriMet's "503-238-RIDE") is permitted, but the field must not contain any other descriptive text. |
+|  `brand_terms_url` |  URL | **Conditionally Forbidden** | A fully qualified URL pointing to the location of a page that defines the license terms of the brand icons or other trademark information. <br><br>It is recommended that the terms permit feed consumers, without taking additional actions, to use the logo in reference to the agency's services broadly throughout their products, provided an attribution citing these terms is visible to users somewhere in these products. <br><br> Consumers SHOULD respect standard HTTP caching headers when fetching this file. <br><br>Conditionally Forbidden:<br>- **Forbidden** when `agency_brand_image_url` is not set <br>- Optional otherwise |
```
johnhelmuth-st: `agency_brand_image_url` should be `brand_image_url` I think?

gtfs/spec/en/reference.md
```diff
@@ -182,6 +182,22 @@ The following example demonstrates how a field value would appear in a comma-del
...
+|  `brand_image_url_dark` | URL  | Optional | A fully qualified URL pointing to the location of a graphic file representing the agency's brand for the service for use in dark mode applications. The file SHOULD have a transparent background and contain a version of the brand image optimized to fill square or round formats larger than 16x16. Consumers SHOULD respect standard HTTP caching headers when fetching this file. File MUST be in SVG V1.1 format and MUST be either square or round. <br><br>A data consumer MUST store the brand image on their own servers to avoid direct traffic from end user devices to the agencies web servers. |
```
johnhelmuth-st: Can a feed have `brand_image_url_dark` but not `brand_image_url`?

felixguendling: I was thinking about this, too but didn't really come to a conclusion. Any ideas (also on wording)?

I think the default logo should always be set as it's more important than the dark mode logo and can in most cases even be shown in dark mode apps without looking horrible.

johnhelmuth-st: I can see 2 options.

1. Make `brand_image_url` Required and leave `brand_image_url_dark` Optional.  (Not a fan of this, as some agencies may want to expose diff. phone#s, urls, etc, without icon.)
2. Make `brand_image_url_dark` Conditionally Forbidden, meaning: forbidden unless `brand_image_url` is set.

-# Edited for correct "Presense" value.

felixguendling: I made it Conditionally Forbidden like you prose in Option 2. I think this makes sense.

felixguendling:
> No. The route short name should still be something like S6

Only showing route name does not work for the other examples I made (which you seem to have missed):

- "RB 75 (28723)" (route name + trip name relevant)
- "ICE 1034" (only trip name shown to customers, route name / line exists but is not commonly used in customer information such as tickets, departure boards or official booking apps). 

Therefore, the "only showing `route_short_name`" approach is not sufficient for the examples you didn't mention. Even if you also show the trip name somewhere in the details it's just plain wrong for the "ICE 1034" example to show it's line number instead of the train number.

We introduced a `display_name` property on trip level and scripting in MOTIS to overwrite this field. So we have something that works most of the time and matches what customers will find on real-world departure boards quite often. However, this is not possible with vanilla GTFS at the moment.

miklcct:
> > In this case the 123 is the route number. The same as announcing "bus 52 to Willesden", where 52 is a route number.
> 
> And in The Netherlands we don't have routes (lines) at all for trains. So it is only IC Amsterdam.

In GB the information is shown in form as, for example, "a Mildmay line service", on the departure boards and announcements, where "Mildmay line" is the route name.

However, the line name can be as generic as the whole operator called the same line, such as "South Western Railway".

kona314:
I'm thinking we should refactor this a bit (again).

The brand vs. agency distinction is a worthwhile discussion for another context. I believe the intent of this PR is simply to be able to show images and iconography that riders are familiar with in contexts that appropriate--an agency icon, a route symbol, etc.. 

To this end, I'm wondering if we should simplify this down to an `image_assets.txt` file. In `agency.txt`, `routes.txt`, etc. we could have a `icon_image_id` (and/or `logo_image_id`, `icon_16x16_image_id`, etc.) referencing IDs in this file. Columns could require that they reference a square image or not, but this would be constrained or specified on the column, not the asset. 

The base of the file would look something like the below. Each row represents an asset and variants of each are in the columns. I didn't add a mono column but we could throw that in too, though we'd probably have to rework the presence fields to require at least one of `image_url` or `image_url_mono`. We could also add an aspect ratio column but I'm not sure this is valuable, consumers can calculate that on import. A future reworking of brands and agencies could reference images in this file as appropriate.

### image_assets.txt

**Conditionally Required**

Primary key (`image_id`)

|  `image_id` | Unique ID | **Required** | Identifies an image referenced in the feed. |
|  `image_terms_url` |  URL | Optional | A fully qualified URL pointing to the location of a page that defines the license terms of the image. <br><br>It is recommended that the terms permit feed consumers, without taking additional actions, to use the image in reference to services contained in the feed broadly throughout their products, provided an attribution citing these terms is visible to users somewhere in these products. <br><br> Consumers SHOULD respect standard HTTP caching headers when fetching this file. |
|  `image_url` | URL | **Required** | A fully qualified URL pointing to the location of a graphic file. The file MUST be in SVG v1.1 format and SHOULD have a transparent background. Consumers SHOULD respect standard HTTP caching headers when fetching this file.<br><br>A data consumer MUST store the brand image on their own servers to avoid direct traffic from end user devices to the agencies' web servers. |
|  `image_url_dark` | URL  | Optional | A fully qualified URL pointing to the location of a graphic file, for use in dark mode applications. The file MUST be the same size as the one in `image_url` and represent an otherwise identical image, except for any dark mode optimizations. When this value is specified and a consuming application is in dark mode, the application SHOULD use this variant of the image, otherwise they should use the standard image. The file MUST be in SVG v1.1 format and SHOULD have a transparent background. Consumers SHOULD respect standard HTTP caching headers when fetching this file.<br><br>A data consumer MUST store the brand image on their own servers to avoid direct traffic from end user devices to the agencies' web servers. |

gcamp:
I think having a dedicated images file is a good idea, because images aren't limited to agency/brands, they can also apply to routes, stops, vehicles, etc.

felixguendling:
Ok, then let's split this from `brands.txt` discussion and only focus on bringing logos into end user apps. I think you're right that `brands.txt` can be handled as a orthogonal topic and we don't "destroy" something now that can't be fixed later (at least from what I see now).

I changed it to:

- `images.txt` instead of `image_assets.txt` is maybe sufficient? Also makes it consistent with key names `image_id` instead of `image_asset_id`.
- added alternative text

Questions:
- If this is a general image database used throughout the feed, should we drop the SVG requirement there and add it to every place where it should be a logo (agency/brand/product logo). Maybe at one point we want to add photos of stops so users can easily identify them or something else where raster images are a better fit?
- Should we add a `image_16x16_url` and `image_monochrome_url` to the image table or is this something for `logo_16x16_image_id` and `logo_monochrome_image_id` in the agency table?

Edit: I was thinking to maybe even **remove** the terms completely. For consumers of large amounts of feeds it's a lot of effort to read all those legalese to check if you can really use it in your app. Maybe we should just enforce @kona314's fall-back terms. What is a logo worth in GTFS if those minimum conditions (i.e. I am allowed to used the logo in my app in reference to their services given I credit the source) are not met?!

kona314:
**On raster images**
I'm torn on this one. I see the value in this flexibility, but I worry about the complexity of mixing very different kinds of images/the likelihood of things getting mixed up. I also think the needs of brand imagery are very different from those of raster images like stop photos, like with the use of different variants (dark, mono) of the "same" image—perhaps a different filename is in order to distinguish the two. `iconography.txt`? This was my thinking behind the use of "assets" since it's commonly used in design, but I know that's not the most recognizable. 

**On additional variants**
I view each column as a different variant of what is otherwise recognizably the same image. To this end, I feel that `image_monochrome_url` is appropriate in the image file, but specific sizes or aspect ratios should be a separate image referenced with separate columns in the specific object file (e.g. `logo_16x16_image_id` in `agency.txt`). All variants should always be the same, as specified in `image_dark_url`. This would also allow full buildout of light, dark, and mono variants of both the full size and smaller images. 

**On terms**
I hear where you're coming from on this, and as a developer, I would very much appreciate that simplicity. To your point, it also wouldn't make sense to populate the field but not enable use. However, I think this would block a lot of agencies from utilizing this functionality. Many agencies already have specific needs or requirements for licensing their registered trademarks, and it's important to have the flexibility in the spec for them to express that, otherwise they may not bother at all.

There might be a middle ground with the use of a required `image_terms` enum column. One value would specify that custom terms in the URL field apply, indicating developers must take action before enabling the image to appear in their apps. Other values would apply pre-defined terms developers can account for in advance. My instinct is that we should leverage Creative Commons licenses and _strongly recommend_ BY-ND, but I fear CC would spook some agencies with registered trademarks and they'd leverage their own terms, so perhaps a basic "GTFS terms" they can opt into would be a worthwhile enum value as well. We would need to make sure any changes to the terms get their own enum value so terms don't change without consent of the publisher. 

**Other**
I noticed you also retained `image_url` as optional and `image_dark_url` as conditionally forbidden—this would make it valid to produce a row with an ID and alt text, but no actual image. My proposal above has these are required and optional respectively. If we're adding mono into the mix, it might be best to just require that one of the three variants be populated, so it's not legal to produce an image with no images.

Issue #606, opened 2026-01-30
Adding NewRoutes to GTFS Realtime


cleveramarquet:
### Describe the problem

Currently, GTFS Realtime lacks a standard way to provide information about new transit routes that do not yet exist in the static GTFS schedule. This creates a gap for riders when agencies introduce temporary or brand-new service that hasn't been included in the latest static feed update.

### Use cases

1. Providing information for new routes that have been launched but are not yet present in the current static GTFS dataset.

2. Displaying route details like names, descriptions, and colors for these new services to ensure riders have complete information.

### Proposed solution

Introduce a NewRoutes message to the GTFS Realtime FeedEntity. This message will include a list of Route objects, each containing fields like route_id, agency_id, route_short_name, route_long_name, route_desc, route_type, route_url, route_color, and route_text_color. This allows real-time feeds to define the properties of new routes dynamically.

[gtfs-realtime.proto.txt](https://github.com/user-attachments/files/24969911/gtfs-realtime.proto.txt)

### Additional information

As part of a joint effort between Clever Devices and Transit App, we would like to move forward with the NewRoutes use case specifically of the [GTFS-ServiceChanges_v3.1](https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit?tab=t.0#heading=h.mm6bblup6h8d) proposal. We saw the need for a more focused specification tailored to consistent and accurate representation of bus bridges or additional service. This change allows producers to define new, temporary, or dynamic routes entirely within the real-time feed.
[...]

Issue #310, opened 2022-03-22
Expand / Add definition around route_type 


e-lo:
## Customer Need

*On behalf of Amtrak...*

*As a customer, I'd like to know (and potentially select a route by) if I can expect the bus route I've chosen to provide a comfortable amount of space to sit and some space for my personal items - so that I can evaluate if it is an appropriate route for me or if is worth any difference in costs.*

*As a customer, I'd like to know (and potentially select a route by) if I can expect the bus route I've chosen to requires me to navigate across vehicle rows where I cannot stand and sit, next to individuals that I may not know without the option to move during the trip - so that I can evaluate if it is an appropriate route for me (my ability and comfort level in being in close proximity to others) or if is worth any difference in costs.*

## Current Spec

[`routes.route_type`](https://gtfs.org/reference/static#routestxt) :
> Indicates the type of transportation used on a route. Valid options are:
> 
> 0 - Tram, Streetcar, Light rail. Any light rail or street level system within a metropolitan area.
> 1 - Subway, Metro. Any underground rail system within a metropolitan area.
> 2 - Rail. Used for intercity or long-distance travel.
> 3 - Bus. Used for short- and long-distance bus routes.
> 4 - Ferry. Used for short- and long-distance boat service.
> 5 - Cable tram. Used for street-level rail cars where the cable runs beneath the vehicle, e.g., cable car in San Francisco.
> 6 - Aerial lift, suspended cable car (e.g., gondola lift, aerial tramway). Cable transport where cabins, cars, gondolas or open chairs are suspended by means of one or more cables.
> 7 - Funicular. Any rail system designed for steep inclines.
> 11 - Trolleybus. Electric buses that draw power from overhead wires using poles.
> 12 - Monorail. Railway in which the track consists of a single rail or a beam

#### 1. Add `coach` as a `route_type` | Update `bus`

3:  Bus, 
- customers may be seated or standing
- differentiated from a van by customers being able to navigate across rows via an aisle that is standing-height

9: Coach, 
- a bus with premium features such as:  
  - more comfortable, dedicated seating. 
 -  May or may not include: space underneath and above the rider for luggage, individual climate control, on-board restrooms, entertainment, and connectivity.

#### 2. Add `van` as a `route_type`

???: van, a slightly larger auto where 
- a single or pair of doors provides access to multiple rows of seats, access to which is typically provided via a small side aisle or folding seats. 
- Interior ceiling is below typical standing height. 

## Previous discussion

[...]

PR #621, opened 2026-04-09
Add NewRoutes to GTFS-RT to support dynamic service (#606)

cleveramarquet:
### Description
This proposal introduces a `NewRoutes` message to the `FeedEntity` in GTFS Realtime. As discussed in [Issue #606](https://github.com/google/transit/issues/606), the current specification lacks a standard way to define route metadata for services that do not exist in the static GTFS schedule.

The full technical specification and field definitions can be found in the [GTFS-NewRoutes Google Doc](https://docs.google.com/document/d/1Dv23vnycQcPBk1BRvDq5a_fL827up5SoYRZcnvar6m4/edit).

By allowing a `Route` to be defined directly within the real-time feed, consumers can provide a complete and accurate experience for riders without requiring agencies to "reuse" unrelated static IDs.

### Implementations
* **Producer:** [Clever Devices Test Feed](http://66.216.13.125:8080/gtfsrt/)
* **Consumer:** [Transit App](https://transitapp.com/) (Development in progress)

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -1266,6 +1266,14 @@ message Route {
   // See definition of routes.route_sort_order in (CSV) GTFS.
   optional uint32 route_sort_order = 10;
 
+  // See definition of calendar.start_date in (CSV) GTFS.
+  // The first day that the new route is active, in YYYYMMDD format.
+  optional string start_date = 11;
+
+  // See definition of calendar.end_date in (CSV) GTFS.
+  // The last day that the new route is active, in YYYYMMDD format.
+  optional string end_date = 12;
+
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime Specification in order to add and evaluate new features and
   // modifications to the spec.
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -1252,7 +1252,7 @@ message Route {
 
   // See definition of routes.route_type in (CSV) GTFS.
   // Required
-  optional RouteType route_type = 6;
+  optional int32 route_type = 6;
 
   // See definition of routes.route_url in (CSV) GTFS.
   optional TranslatedString route_url = 7;
@@ -1273,22 +1273,6 @@ message Route {
 
   // The following extension IDs are reserved for private use by any organization.
   extensions 9000 to 9999;
-}
-
-// NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
-enum RouteType {
-  // route type as defined in the (CSV) GTFS
-
-  LIGHT_RAIL = 0;
-  SUBWAY = 1;
-  RAIL = 2;
-  BUS = 3;
-  FERRY = 4;
-  CABLE_CAR = 5;
-  GONDOLA = 6;
-  FUNICULAR = 7;
-  TROLLEY_BUS = 11;
-  MONORAIL = 12;
 }
 
 // NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -113,6 +113,7 @@ message FeedEntity {
   optional Shape shape = 6;
   optional Stop stop = 7;
   optional TripModifications trip_modifications = 8;
+  optional Route route = 9;
 
   // The extensions namespace allows 3rd-party developers to extend the
   // GTFS Realtime Specification in order to add and evaluate new features and
@@ -423,6 +424,10 @@ message TripUpdate {
     // Specifies the name for this trip when it differs from the original.
     // NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
     optional string trip_short_name = 6;
+
+    // Specify the route associated with the trip, if the route has been added dynamically. The associated `Route` message should be included in the same feed.
+    // Optional if schedule_relationship is NEW, forbidden otherwise. 
+    optional string route_id = 7;
 
     // The extensions namespace allows 3rd-party developers to extend the
     // GTFS Realtime Specification in order to add and evaluate new features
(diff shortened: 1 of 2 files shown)


gtfs-realtime/proto/gtfs-realtime.proto
```diff
@@ -1222,6 +1229,68 @@ message TripModifications {
...
+  // See definition of routes.route_long_name in (CSV) GTFS.
+  // Conditionally Required, see reference
+  optional TranslatedString route_long_name = 4;
+
+  // See definition of routes.route_desc in (CSV) GTFS.
+  optional TranslatedString route_desc = 5;
+
+  // See definition of routes.route_type in (CSV) GTFS.
+  // Required
+  optional RouteType route_type = 6;
```
leonardehrenfried: As I did in the Google doc, I'm requesting this to be changed to an int so it can reflect the full range of extended route types. As a benefit this enum would then not have to be kept in sync with static GTFS.

leonardehrenfried: cc @hbruch

cleveramarquet: I see your point about the extended route types. However, I used an enum here to stay consistent with how the rest of the current GTFS-RT specification handles route types. Perhaps we could stick with the enum for this PR to keep the scope focused on NewRoutes, and then address a global move to int32 for route types across the entire spec in a separate proposal?

leonardehrenfried: Well, since GTFS has ironclad backwards-compatibility guarantees, I don't think I will be able to convince the community that there are two ways to define the mode.

felixguendling: 100% agree with @leonardehrenfried.
We need extended route types support from the beginning so the feature can be useful.

(off topic: IMO the current GTFS route types should be deprecated and replaced with extended route types)

cleveramarquet: I've updated route_type to int32 to support extended types as suggested. Thanks for the catch!

gcamp:
@cleveramarquet You're missing the reference.md updates!

gtfs-realtime/proto/gtfs-realtime.proto
gcamp: The move to `int32` (should probably be `uint32`?) doesn't bother me, but it's really not obvious to me that extended route type is the way to go. There's been discussion in the past about this and it was contentious

BethBeveridge (approved):

leonardehrenfried:
Thanks, @cleveramarquet, for yielding to my request.

gtfs-realtime/proto/gtfs-realtime.proto
leonardehrenfried: I agree that the status of the extended route types have been pretty frustrating for many years.

I'm aware that they are pretty unimportant in North America (correct?) but in Europe people very much like to separate their services into distinct types. From the perspective of a routing algorith, it may not matter much, but regular commuters care quite a lot about the difference between regional rail and suburban rail, for example.

I'm in favour of these distinctions, too, and would like to hear your opinions. If @gcamp, @felixguendling and I agree that it's worthwhile, we could try another attempt at standardising.

felixguendling: > we could try another attempt at standardising

I would support this as it's a very obvious shortcoming of GTFS compared to NeTEx and commercial proprietary data formats like HRDF, DIVA, ISA, etc. And it should be easy to fix because extended route types are used in practice in a lot of places already.

tzujenchanmbd: Previous discussions: 
[issue#310](https://github.com/google/transit/issues/310), [PR#279](https://github.com/google/transit/pull/279)

We can continue route type discussion in [#310](https://github.com/google/transit/issues/310)

tzujenchanmbd:
Hi @cleveramarquet, Just a quick reminder based on [CHANGES.md](https://github.com/google/transit/blob/master/gtfs-realtime/CHANGES.md), a few conditions need to be met before initiating a vote:

> The discussion lasts for as long as the advocate feels necessary, but **must be at least 7 calendar days**.

_(Counted from when the PR is opened)_

> Before calling for a vote, at least one GTFS-realtime producer and one GTFS-realtime consumer should implement the proposed change. It is expected that the GTFS-realtime producer(s) include the change in a public-facing GTFS-realtime feed and provide a link to that data within the pull request comments, and that the GTFS-realtime consumer(s) provides a link in the pull request comments to an application **that is utilizing the change in a non-trivial manner** (i.e, it is supporting new or improved functionality).

_("Development in progress" does not meet this requirement)_

Given that these conditions are not currently met, we consider the current vote invalid. You may initiate another vote once all requirements are met.

cleveramarquet:
Thanks [@leonardehrenfried](https://github.com/leonardehrenfried) and [@felixguendling](https://github.com/felixguendling) for the feedback. I have updated route_type to int32 in the .proto file and updated the documentation in reference.md to reflect this. This ensures we support the full range of extended route types from the start.

[@gcamp](https://github.com/gcamp) I’ve just pushed an update that includes the reference.md changes. It now contains the full definitions for the Route message, including the route_sort_order, start_date, and end_date fields.

[@tzujenchanmbd](https://github.com/tzujenchanmbd) Thank you for the reminder on the [CHANGES.md](https://www.google.com/search?q=https://github.com/google/transit/blob/master/CHANGES.md) requirements. I’ve retracted the formal 'Call for Vote' for now.

I will leave the PR open for the required discussion period. In the meantime, I am working with the Transit App team to move their implementation from 'in-progress' to a state that meets the voting requirements. I'll provide the updated links once those are ready.

Regarding the route_type discussion, I’ve updated my PR to int32 to ensure NewRoutes is compatible with extended types immediately. However, to avoid overloading this proposal with a broader debate on global transit mode definitions, I suggest we move any further discussion on standardizing route_type across the rest of the spec to [Issue #310](https://github.com/google/transit/issues/310) as suggested by [@tzujenchanmbd](https://github.com/tzujenchanmbd).

felixguendling:
`route_type` should not stop us from making progress on this one, right?

The `route_type` field will stay, we can still standardize more values in a separate step together with GTFS-static `route_type`.

cleveramarquet:
> `route_type` should not stop us from making progress on this one, right?
> 
> The `route_type` field will stay, we can still standardize more values in a separate step together with GTFS-static `route_type`.

That's correct @felixguendling, this should not stop us from making progress here. We can take care of standardizing more values for route_type in another pull request.

We are very close to having both public endpoints in between Hitachi (Clever Devices) and Transit running so we can officially put this pull request for a vote soon.

Issue #620, opened 2026-04-06
Proposal: Additional (Trip) Notices


sebastianknopf:
### Describe the problem

Many data standards like NeTEx and HRDF offer data structures to provide additional human-readable notes assigned to trips in order to inform about specials not covered in technical properties or which shall be communicated in a certain, explicit wording to the passener. GTFS does this actually not, some data providers use informational service alerts for that (e.g. travic.app states additional notes under 'service alerts'). However, service alerts belong to the realtime data and are (for my understanding) meant to inform about unplanned changes. The additional notes I mean are already known during conceptual planning and belong not really to the scope of realtime data.

### Use cases

Adding any arbitrary, non-technical, but already known notice to a trip, e.g.:
- "Bike reservation recommended"
- "Special Train - Dedicated fare is required"
- "Express Bus - Please note the special trip times"
- "Sesonal Trip - Runs during the XYZ festival"

### Proposed solution

First draft:
- Add a file notices.txt (notice_id, display_text, [tts_text])
- Add a file trip_notices.txt (trip_id, notice_id)

Before I investigate more time here, I'd like to hear some of your opinions regarding this topic. I believe, I am not the only one having this issue.

### Additional information

We must be careful when imolementing an extension like that, because it might lead to 'misuse' of the notices instead of filling already existing technical fields properly.

skinkie:
What about doing this similar to translations.txt?

doconnoronca:
Why not add new columns to trips rather then create new tables?

skinkie:
> Why not add new columns to trips rather then create new tables?

Because notices are not limited to trips. Because the cardinality does not match. So many reasons...

sebastianknopf:
> What about doing this similar to translations.txt?

Good point! That could look like this: 

| Column | Type | Required | Comment |
| ---- | ---- | ---- | ---- |
| notice_id | Text | Yes | Required for uniquely identifying a notice for translations. |
| table_name | Enum | Conditionally Required | Identifies a table for which the notice is valid. Valid for `routes`, `trips` and `stops`. Required if an record_id is set. |
| record_id | Text | Optional | Primary key for the table specified in `table_name`. For `routes` use `route_id`, for `trips` use `trip_id` and for `stops` use `stop_id`. Can be left out for having a 'general notice'  for the whole timetable. |
| display_text | Text | Yes | Text to be displayed within the notice. |

This way, we'd have a simple way to...
- ... add notices for routes, trips and stops; could also be extended in future for other entity types
[...]

Issue #610, opened 2026-02-11
Proposal: Add cyclist position availability fields to VehiclePosition and CarriageDetails


oliver-sturrock:
## Problem

Transit passengers traveling with bicycles have no way to know whether a bike rack or storage area will be available on an approaching vehicle. Typical transit buses have only 2–3 external bike rack positions. When full, cyclists are denied boarding and must wait for the next vehicle or abandon their transit trip.

The Santa Clara VTA confirmed this problem in a USDOT SMART Grant pilot (FAIN: 69A3552341049), finding that denied boardings are a measurable service quality issue and that no transit agency in North America currently provides this data through standard feeds.

## Proposed Solution

Add two new `optional uint32` experimental fields to **VehiclePosition** and **CarriageDetails**:

| Field | Message | Description |
|-------|---------|-------------|
| `cyclist_positions_available` | `VehiclePosition` (field 12), `CarriageDetails` (field 6) | Positions currently available |
| `total_cyclist_positions` | `VehiclePosition` (field 13), `CarriageDetails` (field 7) | Total capacity |

This follows the same pattern used for `occupancy_status`, `occupancy_percentage`, and `multi_carriage_details`.

## Working Producer Implementation

Sportworks Global LLC has a public GTFS-RT feed demonstrating the proposed fields:

- **Feed URL:** https://cascadia-metro-gtfs-rt.sportworks.com/gtfs-rt/vehicle-positions
- **Static GTFS:** https://cascadia-metro-gtfs-rt.sportworks.com/gtfs-rt/gtfs.zip

The feed has been validated against `protoc`, Python `gtfs-realtime-bindings`, and Node.js `gtfs-realtime-bindings` — all pass without errors. Full validation results will be included in the pull request.

## Pull Request

A PR with the proto and reference.md changes will follow shortly.

**Contact:** Oliver Sturrock, olst+gtfs@silverfallscapital.com, Sportworks Global LLC / Silver Falls Capital

doconnoronca:
I've been working on my proposed [vehicles.txt](https://github.com/doconnoronca/transit/blob/vehicles/gtfs/spec/en/reference.md#vehiclestxt) addition to GTFS that includes a bicycle capacity value, but until that is widely adopted including a total_cyclist_positions value would be quite useful in the real time information.

Issue #458, opened 2024-05-22
Addition of vehicles.txt to GTFS static


doconnoronca:
### Describe the problem

Real time prediction tools could use additional information to describe what type of vehicle is approaching and its capacity and accessibility features.

### Use cases

Real-time data usually includes the vehicle number which could be used to look up more details about the vehicle then it would be practical to include with the real-time data feed. Capacity would be used to indicate how busy a vehicle is if the real-time passenger load is provided. The vehicle type would be useful to indicate when a rail vehicle has been replaced by a bus.

This could also be used for reporting fleet information in a standard format for analyses by other levels of government and the public.

### Proposed solution

A new vehicles.txt file in the GTFS static file to describe the capacity, accessibility and features of individual vehicles. It would apply to buses, LRT, subway cars, rail cars and even ferries.

Fields could include:

- Id: Corresponds to the id of the vehicle in the real-time data. Required and unique
- Label: Corresponds to the label of the vehicle in the real-time data.
- Vehicle_Type: Could match the values of [Route_Type](https://gtfs.org/schedule/reference/#routestxt)
- Manufacturer
- Model
- Year: Year delivered
- Length_Feet
- Length_Meters
- Articulations
- Number_of_Levels
- Seated_Capacity
- Standing_Capacity
- Wheelchair_Capacity
- Interior_Bicycle_Capacity
- Exterior_Bicycle_Capacity
- Restrooms
- Boarding_Level: 1 - High platform level boarding; 2 - Low platform level boarding; 3 - Steps to board
- Boarding_Steps: Number of steps to board
- Accessible_Boarding: 0-Not accessible; 1-Level boarding; 2- Ramp can be deployed; 3- Lift can be deployed
- Propulsion: 1- Unpowered; 2- Overhead wire; 3- Third rail electric; 4- Battery electric; 5- Hydrogen; 6- ICE-battery hybrid; 7- Internal combustion engine; 8- Cable pulled; 9- Stream engine!?!?!

gcamp:
Related : https://github.com/google/transit/pull/200

doconnoronca:
I reviewed the [GTFS-Vehicles proposal](https://github.com/google/transit/pull/200). I added a few things inspired by that. My proposal is close to what @mgilligan and @botanize suggested.

This is a much simpler. It just has a list of vehicles and describes them. I have suggested one row per vehicle. The largest agencies would have thousands of rows, which isn't a lot these days, but I would be open to having ranges of Ids and Labels instead.

[...]

PR #607, opened 2026-02-03
Add license information in the GTFS schedule feed

skalexch:
## Summary
This proposal allows for describing license information in GTFS feeds. 
1. It adds a field type, _SPDX ID_, that reference the SPDX identifier for standard licenses.
2. It adds a file `licenses.txt` that describes the SPDX license or provides a custom license URL for the whole feed, with the possibility of specifying the license information for specific files in other rows.

## Describe the Problem
There has been a lot of conversation in the community regarding license information. It seems like most people agree on adding license information, and the main discussion was around the scope, i.e. can some files have different terms?

Previous conversations about license information, and there were some previously suggested options, mainly:
- A column in  `feed_info.txt` that can be SPDX or URL if the license is custom.
- Multiple columns that can be specified by operator (`agency_id`) and even entity (filename).
- A column in `feed_info.txt` + a traditional `LICENSE.txt` file.

The full conversations can be found in the [license proposal issue](https://github.com/google/transit/issues/519) and a [Google Groups thread](https://groups.google.com/g/gtfs-changes/c/iDmt-vjs180/m/wMJI7LHGAQAJ).

## Proposed Solution
The solution entails the addition of a file called `licenses.txt`, which lists the licenses for the whole feed as well as any specific files that might have a different license (e.g. shapes.txt). The license is specified either as an SPDX ID or as a custom URL.

The producer can choose to either:
- Create rows that specify the license for a specific file each.
- Create a row to specify the license for the whole feed, except for the files that were mentioned in other rows.

## Type of change

 

GTFS Schedule
- [x] Functional Change
- [ ] Non-Functional Change
- [ ] Documentation Maintenance

GTFS Realtime
- [x] Specification Change
- [ ] Specification Change (Experimental Field)

## Proposed Discussion Period
To put more eyes on this proposal, we propose a **discussion period of at least two weeks**.

## Testing Details

- Consumer(s): 
- Producer(s): 
- Estimated Testing Period: 

## Proposal Update Tracker 

| Date       | Update Description |
|------------|--------------------|
| (2026-09-17) | Vote-to-test opened |

## Checklist

- [x] I have read and understood the [GTFS Change process](https://github.com/google/transit/tree/master/gtfs/Governance/change-process.md).
- [x] I have read and understood the [GTFS Guiding Principles](https://github.com/google/transit/tree/master/gtfs/Governance/guiding-principles.md). 
[...]

PR #636, opened 2026-05-11
Addition of vehicles.txt to GTFS schedule

doconnoronca:
## Summary
Adds new file vehicles.txt to describe the capacity, accessibility, appearance and features of individual vehicles or vehicle ranges.

## Describe the Problem
Based on issue [458](https://github.com/google/transit/issues/458). When combined with the vehicle label in GTFS real-time or other real-time APIs, this allows consumers to provide a variety of information about the approaching vehicle to riders.

## Use Cases

- Inform riders about the type of vehicle approaching and if it is the same as the scheduled type of vehicle, such as when streetcars are replaced by buses.
- Inform riders of the size, color and appearance of the approaching vehicle.
- Inform consumers about the capacity of vehicles so that a better passenger load estimate can be calculated.
- Allow analysts to calculate the actual capacity of a route based on vehicle capacity.
- Inform riders of the availability of features including bike storage, luggage space, restrooms, changing points, WiFi.
- Provide disabled riders of the accessibility features of the approaching vehicles and how the use them.

## Proposed Solution
The addition of a new table called vehicles.txt that describes the vehicles of a transit agency. Also, the addition of vehicle_class columns to routes.txt and trips.txt which allows routes and trips to include the vehicles that typically operate for that route or trips and allow vehicles.txt information to be used if real-time data isn't available.

## Type of change

 

GTFS Schedule
- [X] Functional Change
- [ ] Non-Functional Change

## Additional Information
There is currently a [vehicles](https://github.com/TIDES-transit/TIDES/blob/main/spec/vehicles.schema.json) file in TIDES as well as [TODS](https://github.com/MobilityData/transit-operational-data-standard/blob/main/docs/en/spec/index.md#vehiclestxt). This proposal expands on TODS with additional details that are useful for riders.

The [Proposal: Add cyclist position availability fields to VehiclePosition and CarriageDetails](https://github.com/google/transit/issues/610) includes a total_cyclist_positions value, which wouldn't be needed if the data was available from vehicles.txt.

Some of the [examples](https://github.com/google/transit/issues/620#issuecomment-4194557008) for the use of Proposal: Additional (Trip) Notices could be handled by vehicles.txt including the availability of WiFi, food and bicycle storage.

I have created a [Google Sheet](https://docs.google.com/spreadsheets/d/1G6NfQpll1nT4qVa6S3ec6QhmeT8kZs5acq66diYX5zY/edit?usp=sharing) with vehicles data for Toronto TTC, Los Angeles Metro, New York Metro North rail and Toronto GO Train. These can be downloaded as Comma Separated Values to generate a vehicles.txt file. These have been loaded into [TransSee](https://www.transsee.ca/vehicles?a=ttc).

Some transit enthusiasts who use TransSee have contributed files as well. Transit enthusiasts could crowd source data in places where transit agencies don't generate it on their own. There could be TODS like extensions to this structure that would include data that is important to them.

[BusTimes.org](https://bustimes.org/) provides some of the same information for [bus fleets](https://bustimes.org/api/vehicles/?id=&slug=&operator=BNSM&vehicle_type=&livery=&withdrawn=false&search=&fleet_code=&reg=&code=&ordering=) in the UK.

## Proposed Discussion Period
As a large proposal, I suggest at least three months of discussion.

## Testing Details

- Consumer(s): TransSee
- Producer(s): TransSee has produced some vehicles.txt file based on public data and some transit enthusiasts have contributed some. I would like to see a transit agencies start generating the file.
- Estimated Testing Period: At least three months.

## Proposal Update Tracker 

| 2026-05-10 | Create PR draft |

## Checklist

- [X] I have read and understood the [GTFS Change process](https://github.com/google/transit/tree/master/gtfs/Governance/change-process.md).
- [X] I have read and understood the [GTFS Guiding Principles](https://github.com/google/transit/tree/master/gtfs/Governance/guiding-principles.md). 
- [X] I understand the [Role of Advocate and the Responsibilities](https://github.com/google/transit/tree/master/gtfs/Governance/roles.md) attached to it. 
- [X] I have signed the [Contributor License Agreement (CLA)](https://cla.developers.google.com/about/google-individual).
- [X] I have ensured that this proposal is not covered by an existing Pull Request.


skinkie:
You are actually introducing Transmodel like functionality with newly invented terms. My suggestion would be to fully align the attribute names with NeTEx.

kona314:
- TODS [defines a `vehicles.txt`](https://github.com/MobilityData/transit-operational-data-standard/blob/main/docs/en/spec/index.md#vehiclestxt) with much more limited information. This would be a breaking conflict for producers attempting to provide both. 
  - To resolve this, I'd suggest moving the bulk of the attributes in this file to a `vehicle_groups.txt` file (or other similar name), then define `vehicles.txt` as identical to TODS, but with an extra `vehicle_group_id` column referencing an ID in `vehicle_groups.txt`.
    - This would also handle the concerns in the next two bullets. 
- If the corresponding RT field is `label`, should `vehicle_id` be called `vehicle_label` instead? Same for `vehicle_id_high`, which would be `vehicle_label_high`. 
- Some agencies have letter prefixes/suffixes on their coaches, e.g. B0747 or 1312E. How would this work with defining ranges? 
  - Are there libraries to expand e.g. B0741-B0749 and 1301E-1325E into 9 and 25 labels respectively? This is just not a problem I'm familiar with. 
- `agency_id` should be added and part of the primary key, with the same semantics as `routes.agency_id`, for handling feeds with multiple agencies. 
  - If moving to `vehicle_groups.txt`, I don't think it would need to be part of the PK, since the group ID column would handle that. 
- Should `cab` be explicitly forbidden for non-rail vehicles? 
- To me, "standing capacity" is subtly different from "fully loaded capacity," as agencies often differentiate between comfortable standing room and crush loading. Should the description for `capacity_standing` draw a clearer line here? 
  - I think my preference would be to rename to `capacity_full` and have the description explicitly tie this to the RT `OccupancyStatus.FULL` value and definition, which it feels like it's already alluding to. 
- Why are AC/USB-A/USB-C plugs defined the way they are? I think it's confusing to have `1` mean an unknown number, but all other numbers mean there are that many. What if there's one plug? (Most likely on small vehicles, e.g. vans or gondolas.) 
  - Can `-1` be used for an unknown quantity? This is more intuitive to me since you can't have negative plugs. 
    - Bikes, wheelchairs, strollers, restrooms/accessible restrooms (maybe?),  could also benefit from `-1` as unknown quantity. 
- `vehicle_type` and `fuel_type` both reference US FTA values. I believe there's been a historic intent in GTFS to avoid endorsing any specific country's systems, which I would agree with. 
  - I don't have an alternative proposal here. There are of course the GTFS route types, which are more limited, especially when not counting the extended route types. 
- I like the inclusion of `vehicle_class`. This almost invites a `vehicle_classes.txt`, which could define rider-facing values for this info like a name? This would also help the above with overcoming shortcomings of the GTFS route types (e.g. double-decker buses would still be `vehicle_route_type = 3`, but with a `vehicle_class_name = "Double-decker"`).
- Buses at my agency announce the route automatically only when the stop is served by multiple routes, otherwise no. Would this be value `2` of `route_announcements`? 
- Nitpicky naming consistency items: 
  - `capacity_bike` and `car_capacity` having the word "capacity" switched seems unnecessary. 
  - `strollers` should be `capacity_strollers`.
  - Some capacity items are plural (strollers) and others are not (bike, car).
  - `bicycle_notes` should be `bike_notes` for consistency with `capacity_bike` and `trips.bikes_allowed`.

doconnoronca:
> You are actually introducing Transmodel like functionality with newly invented terms. My suggestion would be to fully align the attribute names with NeTEx.

Based on the [NeTEx standard](http://netex.uk/netex/doc/2019.05.07-v1.1_FinalDraft/prCEN_TS_16614-PI_Profile_FV_%28E%29-2019-Final-Draft-v3.pdf) these are the relevant fields.

- SelfPropelled: Covered by fuel_type
- TypeOfFuel: Same as fuel_type, but uses a different format. Also GTFS uses underscores between words.
- EuroClass: Seems to be specific model of bus.
- LowFloor: Covered by boarding_steps,
- HasLiftOrRamp: Covered by accessible_boarding enum with more detail,
- HasHoist: Also covered by accessible_boarding,
- Length: Matches.
- Width, Height, Weight: Not included, but could be added if there is interest.

doconnoronca:
> * TODS [defines a `vehicles.txt`](https://github.com/MobilityData/transit-operational-data-standard/blob/main/docs/en/spec/index.md#vehiclestxt) with much more limited information. This would be a breaking conflict for producers attempting to provide both.
>       
>       * To resolve this, I'd suggest moving the bulk of the attributes in this file to a `vehicle_groups.txt` file (or other similar name), then define `vehicles.txt` as identical to TODS, but with an extra `vehicle_group_id` column referencing an ID in `vehicle_groups.txt`.
>         
>         * This would also handle the concerns in the next two bullets.

Yes, I can see the benefit from creating two tables, one based on the TODS vehicles.txt plus a `vehicle_group_id` and `year_manufactured` and moving most of the other columns into a `vehicle_groups.txt` file. However creating two tables makes the structure somewhat more complex.

I could just update vehicles.txt with all the columns to follow TODS instead of TIDES.

>     * If the corresponding RT field is `label`, should `vehicle_id` be called `vehicle_label` instead? Same for `vehicle_id_high`, which would be `vehicle_label_high`.
> 
>     * Some agencies have letter prefixes/suffixes on their coaches, e.g. B0747 or 1312E. How would this work with defining ranges?
>       
>       * Are there libraries to expand e.g. B0741-B0749 and 1301E-1325E into 9 and 25 labels respectively? This is just not a problem I'm familiar with.

In TransSee, for an individual vehicle, it finds the record where the vehicle's label is between the low and high range, so a library to expand the values isn't really needed.

>     * `agency_id` should be added and part of the primary key, with the same semantics as `routes.agency_id`, for handling feeds with multiple agencies.

This would help for some places with overlapping fleet numbers.

>     * Should `cab` be explicitly forbidden for non-rail vehicles?

The Niagara Falls People Mover used to have unpowered bus trailers with no cabs.

>     * To me, "standing capacity" is subtly different from "fully loaded capacity," as agencies often differentiate between comfortable standing room and crush loading. Should the description for `capacity_standing` draw a clearer line here?
>       
>       * I think my preference would be to rename to `capacity_full` and have the description explicitly tie this to the RT `OccupancyStatus.FULL` value and definition, which it feels like it's already alluding to.

That would probably be better.

>     * Why are AC/USB-A/USB-C plugs defined the way they are? I think it's confusing to have `1` mean an unknown number, but all other numbers mean there are that many. What if there's one plug? (Most likely on small vehicles, e.g. vans or gondolas.)
>       
>       * Can `-1` be used for an unknown quantity? This is more intuitive to me since you can't have negative plugs.
>
>         * Bikes, wheelchairs, strollers, restrooms/accessible restrooms (maybe?),  could also benefit from `-1` as unknown quantity.

It's worth considering. Looks like `transfer_count` uses -1 for unlimited transfers.

>     * `vehicle_type` and `fuel_type` both reference US FTA values. I believe there's been a historic intent in GTFS to avoid endorsing any specific country's systems, which I would agree with.
>       
>       * I don't have an alternative proposal here. There are of course the GTFS route types, which are more limited, especially when not counting the extended route types.
> 
>     * I like the inclusion of `vehicle_class`. This almost invites a `vehicle_classes.txt`, which could define rider-facing values for this info like a name? This would also help the above with overcoming shortcomings of the GTFS route types (e.g. double-decker buses would still be `vehicle_route_type = 3`, but with a `vehicle_class_name = "Double-decker"`).

I think that `vehicle_group` would be more specific than `vehicle_class`. A route could be in the double-decker class but have a few different models with different features from different `vehicle_group`s.

>     * Buses at my agency announce the route automatically only when the stop is served by multiple routes, otherwise no. Would this be value `2` of `route_announcements`?

Interesting feature.

>     * Nitpicky naming consistency items:
>       
>       * `capacity_bike` and `car_capacity` having the word "capacity" switched seems unnecessary.
>       * `strollers` should be `capacity_strollers`.
>       * Some capacity items are plural (strollers) and others are not (bike, car).
>       * `bicycle_notes` should be `bike_notes` for consistency with `capacity_bike` and `trips.bikes_allowed`.

Good suggestions.

kona314:
> Yes, I can see the benefit from creating two tables, one based on the TODS vehicles.txt plus a vehicle_group_id and year_manufactured and moving most of the other columns into a vehicle_groups.txt file. However creating two tables makes the structure somewhat more complex.
> 
> I could just update vehicles.txt with all the columns to follow TODS instead of TIDES.

To follow TODS would eliminate the ability to define things with high/low ranges, which to me would be massively more complex than having a separate table since producers have to duplicate this data across a ton of individual vehicles. In my experience, operators already understand this concept and would have no problem mimicking it. 

> In TransSee, for an individual vehicle, it finds the record where the vehicle's label is between the low and high range, so a library to expand the values isn't really needed.

Pantograph does this currently but is moving away from it, instead defining every single vehicle ID. I think it's also a pretty basic expectation that a consumer could determine all the vehicles an agency has based on this file. 

>This would help for some places with overlapping fleet numbers.

There are also cases where this would be useful to riders. In Puget Sound, bus routes from Sound Transit are operated by one of the three county partners, and ST routes appear in those agencies' GTFS feeds. The agencies usually operate ST service with ST buses and their own service with their own buses, but this does get mixed up from time to time. Similar to the above, I think it's also a reasonable expectation that a consumer could know who owns a vehicle in this file. The semantics should work just like `routes.agency_id`, so it would only be required if there's more than one agency in the feed. 

> The Niagara Falls People Mover used to have unpowered bus trailers with no cabs.

The description for this field is ambiguous as written and implies it's only for rail vehicles. If the intent is to cover all vehicles then it should be updated accordingly. 

> I think that `vehicle_group` would be more specific than `vehicle_class`. A route could be in the double-decker class but have a few different models with different features from different `vehicle_group`s.

I agree. My agency has `route_type = 3` for everything since we operate bus service. We have a few different classes of buses—30s, 40s, artics, BRT artics, double deckers—which map directly to scheduled vehicle type assignments. Then groups would be used to define specific models and series. 

> Interesting feature.

The ADA requires this at stops served by multiple routes, so our system is configured to do it only at those stops. I'm sure other agencies have similar functions. (I do not know why one wouldn't just always announce the route, but this is the reality of the situation at the moment.) Point being, `route_announcements` as-is seems to imply all or nothing. Separately, I wonder if there are cases where some stations announce the route within the station but others depend on the vehicle exterior announcements. All of this also seems to be wandering into the category of stop amenities rather than vehicle amenities. I think this field might need some additional work and would be interested in feedback from those more familiar with this topic.

doconnoronca:
I have updated the fields to match TODS field names.

The question is should this data be split into two tables, vehicles.txt with the `vehicle_label`, `vehicle_label`, `license_plate` from [TODS](https://github.com/MobilityData/transit-operational-data-standard/blob/main/docs/en/spec/index.md#vehiclestxt) as well as `year_manufactured`, `nick_name`, `decoration` from my proposal and a `vehicle_group` field which links to the vehicle_group.txt table with the rest of the columns.

There would still be a `vehicle_class` column in the vehicle_group.txt table to link to the routes and trips, and would be less specific that the `vehicle_group`. The `vehicle_group` would generally be a series of vehicles bought in the same order, while the `vehicle_class` would be type of vehicle, like articulated buses.

Is it better to split the data into table tables and making the structure somewhat more complex or to keep it in one table with a `vehicle_label_high` column for ranges, but to be TODOs compatible they wouldn't be able to use the `vehicle_label_high` column?

Brickblock1:
> Is it better to split the data into table tables and making the structure somewhat more complex or to keep it in one table with a `vehicle_label_high` column for ranges, but to be TODOs compatible they wouldn't be able to use the `vehicle_label_high` column?

I don't see using the vehicle label as the primary key working. In Stockholm for example there are multiple buses with the same label used by different operators which all run under the same agency. This can not be solved without using unique vehicle_ids.

doconnoronca:
I have added the agency_id column, changed the USB/AC plug fields to use `-1` for an unknown number of plugs available and made some fixes to the field name and descriptions following @kona314 recommendations.

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,63 @@ The file defines the attributions applied to the dataset.
...
+
+File: **Optional**
+
+Primary key (`vehicle_label`. `agency_id`)
+
+This file describes the capacity, accessibility and features of individual vehicles or vehicle ranges.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+|  `vehicle_label` | Text | **Required** | A label that is unique to the vehicle and matches the label shown in the GTFS RealTime label value in the VehicleDescriptor or CarriageDetails or the vehicle identifier in other real-time data sources. If `vehicle_label_high` is provided then this value is the low end of the range. |
```
rudiejd: since `vehicle_label` is not guaranteed to be unique within a fleet, as already mentioned in other comments, could we instead introduce a new field here that is guaranteed to be unique within an agency's fleet? That would make reverse-compatibility easier, and allow `label` to still represent a public facing identifier that a rider might see. This field could then be added to realtime for more exact correlation to `vehicles.txt`

doconnoronca: I have added vehicle_id and vehicle_id_high as the new primary key. They correspond to the id value of the VehicleDescriptor and CarriageDetails.

rudiejd: Te problem with using `vehicle_id` here is that, in general, `vehicle_id` need not be unique across multiple service dates, and `vehicle_id` need not correspond to any characteristics of the vehicle - for example, at MBTA our `vehicle_id` auto increments within a service date, and the same vehicle can have 7 different `vehicle_id` values in the same week. For this reason and reverse compatibility, you might consider introducing a new field that is guaranteed to have those properties.

phil-swiftly:
Hi, just chiming in on behalf of Swiftly - we really like the proposal, great work so far. 

We're curious to hear if anyone has run into a case where an agency uses vehicle_ids in a more complicated manner than an ASCII string comparison can handle. Thinking of theoretical examples like `abc123_9`, `abc124_9`, `abc123_7` where an agency would want to group by the end numbers. I think the closest I've seen is how the NYCMTA handles vehicle ids in their [GTFS-RT feeds ](https://bt.mta.info/developers/gtfs-realtime/) that belong to different bus companies, although that does not directly violate the ASCII string comparison bc the companies are not designated numerically. If we think this might be a real concern, something like regex may be more appropriate for facilitating vehicle grouping.

doconnoronca:
> Hi, just chiming in on behalf of Swiftly - we really like the proposal, great work so far.
> 
> We're curious to hear if anyone has run into a case where an agency uses vehicle_ids in a more complicated manner than an ASCII string comparison can handle. Thinking of theoretical examples like `abc123_9`, `abc124_9`, `abc123_7` where an agency would want to group by the end numbers. I think the closest I've seen is how the NYCMTA handles vehicle ids in their [GTFS-RT feeds ](https://bt.mta.info/developers/gtfs-realtime/) that belong to different bus companies, although that does not directly violate the ASCII string comparison bc the companies are not designated numerically. If we think this might be a real concern, something like regex may be more appropriate for facilitating vehicle grouping.

Glad to hear from Swiftly.

There are cases where the fleet ranges won't work. In those cases they don't have to use the ranges and they can provide one row for each vehicle.

There are cases where regular expressions aren't well suited, like Miami has a bus range from 19101 to 19231 that are New Flyer buses and another range from 19232 to 19265 that are Gillig buses.

New York has values like "MTA NYCT_1234" or "MTABC_1234" which should work with the ASCII string comparisons.

gtfs/spec/en/reference.md
doconnoronca: Look at the MBTA vehicle location feed, it just add a "y" in front of the vehicle id for buses, I assume to help make them unique.

Adding a third identifier seems like over kill.

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,65 @@ The file defines the attributions applied to the dataset.
...
+|  `articulated_sections` | integer | Optional | The number of articulated sections. If there are no articulations the number should be 1. |
+|  `levels` | integer | Optional | The number of levels available to passengers. Only levels directly above or below other levels count, not where the height of the floor changes within the vehicle. |
+| `cab` | enum | Optional | Options are:<br><br>`0` - No cab in the vehicle.<br>`1` - In addition to a passenger area the car has a cab for controlling the vehicle or train.  |
+|  `image_url` | URL | Optional | Fully qualified URL of an image of the vehicle. The image should have as aspect ratio between 1:1 and 2:1 and a have a vertical resolution of at least 500. The image should show the front of the vehicle (except for train cars) and the side the passengers baord. The rights to the image should also free commerical reproduction of the image, for example licenced under [CC-BY](https://creativecommons.org/licenses/by/4.0/). Consumers should only download this image one time for each GTFS update and always serve it to end users themselves. End users should never be automatically downloading directly from the URL provided. |
```
davidmurray, line 943: Typo here: baord

ttaylor-avail:
> Hi, just chiming in on behalf of Swiftly - we really like the proposal, great work so far.
> 
> We're curious to hear if anyone has run into a case where an agency uses vehicle_ids in a more complicated manner than an ASCII string comparison can handle. Thinking of theoretical examples like `abc123_9`, `abc124_9`, `abc123_7` where an agency would want to group by the end numbers. I think the closest I've seen is how the NYCMTA handles vehicle ids in their [GTFS-RT feeds ](https://bt.mta.info/developers/gtfs-realtime/) that belong to different bus companies, although that does not directly violate the ASCII string comparison bc the companies are not designated numerically. If we think this might be a real concern, something like regex may be more appropriate for facilitating vehicle grouping.

When combining agencies into a single feed, we append the agency_id to the end of most entity ids in the feed. Route, Stop, Trip, Vehicle, and etc.  "{vehicle_id}-a{agency_id}"

doconnoronca:
> When combining agencies into a single feed, we append the agency_id to the end of most entity ids in the feed. Route, Stop, Trip, Vehicle, and etc. "{vehicle_id}-a{agency_id}"

Doesn't that violate the requirement that those ids match the GTFS schedule values.

Seems like GTFS real-time could use an agency_id field at the FeedEntity level.

ttaylor-avail:
My two cents after my quick review of the pull-request.

I like the idea of having more vehicle info in the standard feed. However, I think it would be more applicable to have a file for vehicle types / groups that define the information you need, instead of the individual vehicles themselves.

GTFS Schedule: Perhaps a vehicle_types.txt
GTFS RT VehiclePositions: Perhaps a field that ties the vehicle to a vehicle type. 

Why?
1) In the case of Transit fleets with busses, vans, and etc. What happens when a vehicle goes offline for unexpected maintenance and the agency swaps with a different vehicle and possibly type? Publish a new GTFS export?
2) Let's say you have a single vehicle type. You could end up with a vehicles file with 200 - 1000+ vehicles listed in it, which are duplicates of each other with the exception of a couple fields.
3) I like the idea of having the GTFS RT feed being able to define the vehicle type, which would also handle the case of vehicles switching in real-time.

ttaylor-avail:
> > When combining agencies into a single feed, we append the agency_id to the end of most entity ids in the feed. Route, Stop, Trip, Vehicle, and etc. "{vehicle_id}-a{agency_id}"
> 
> Doesn't that violate the requirement that those ids match the GTFS schedule values.
> 
> Seems like GTFS real-time could use an agency_id field at the FeedEntity level.

The agencies would also need to be merged into a single GTFS schedule export.

doconnoronca:
> I like the idea of having more vehicle info in the standard feed. However, I think it would be more applicable to have a file for vehicle types / groups that define the information you need, instead of the individual vehicles themselves.

The [TIDES version](https://github.com/TIDES-transit/TIDES/issues/269) of vehicles.txt includes a `vehicle_groups` file that has most of the descriptions and a `vehicles` file that has records for individual vehicles and which group they belong to.
 
> GTFS Schedule: Perhaps a vehicle_types.txt 
> GTFS RT VehiclePositions: Perhaps a field that ties the vehicle to a vehicle type.

I am hesitant to make it a requirement to change GTFS schedule and GTFS real-time to fully enabled this feature, as changing both makes implementation take much longer.
 
> 1. In the case of Transit fleets with busses, vans, and etc. What happens when a vehicle goes offline for unexpected maintenance and the agency swaps with a different vehicle and possibly type? Publish a new GTFS export?

Out of service vehicles could still be listed. If a different vehicle is swapped in it would show up in the real-time data.

> 2. Let's say you have a single vehicle type. You could end up with a vehicles file with 200 - 1000+ vehicles listed in it, which are duplicates of each other with the exception of a couple fields.

The fleet ranges can handle that for most cases or the table could be split into `vehicle_groups` and `vehicles` files.

> 3. I like the idea of having the GTFS RT feed being able to define the vehicle type, which would also handle the case of vehicles switching in real-time.

ttaylor-avail:
> > When combining agencies into a single feed, we append the agency_id to the end of most entity ids in the feed. Route, Stop, Trip, Vehicle, and etc. "{vehicle_id}-a{agency_id}"
> 
> Doesn't that violate the requirement that those ids match the GTFS schedule values.
> 
> Seems like GTFS real-time could use an agency_id field at the FeedEntity level

> **[doconnoronca](/doconnoronca) **

@doconnoronca we ran into an issue with the few feeds we have which require combining 2 or more agencies. It seemed the GTFS spec didn't account for the sharing of certain ids like Route and Stop between the agencies. Appending the agency_id to the various entity ids solved the problem, and made the MobilityData validator happy.  Ideally, I would love to have the agency_id carried into the rest of the schedule files, and the RT feed.

lauriemerrell:
A couple of comments, just from myself as an individual:

1. This seems like it would be the first time that direct references to FTA categories are introduced into the GTFS Schedule spec (for `vehicle_type` and `fuel_type`). I am not sure if that's appropriate? For other fields (e.g. `route_type`) GTFS uses its own categories. 
2. I wonder about `facility_name` which seems to be more about operational considerations than rider-facing display. It seems to me that there could be a risk of false precision if this field is populated in GTFS schedule and then consumers display information to riders based on inferences related to this.

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,65 @@ The file defines the attributions applied to the dataset.
...
+|  `vehicle_label_high` | Text | Conditionally Required | If a range of vehicles is represented this is the high end of the range for the vehicle label. The number of characters must be the same as `vehicle_label` and the range covers values with the same number of charactors and are between `vehicle_label` and `vehicle_label_high` when using an ASCII string comparison. Required if both `vehicle_label` and `vehicle_id_high` are included. |
+|  `agency_id` | Foreign ID referencing `agency.agency_id` | Recommended | Agency for the specified vehicles. |
+|  `license_plate` | Text | Optional | The license plate or number plate of the vehicle. Does not apply if a range of vehicles is being used. |
+|  `vehicle_class` | ID | Recommended | Identifies a class of vehicle that usually runs on certain routes. Ment to link with vehicle_class fields in routes.txt and trips.txt. Can be used to provide vehicle information when label isn't available or highlight when the vehicle doesn't match the expected class and what the major differences are. |
```
lauriemerrell, line 932: typo `Ment` --> `Meant`

gcamp:
I didn't mention anything on this PR, but I generally approve of the approach. I have a hard time seeing a use for all the different fields proposed, but if there's a consumer/producer of data, all good for me. 

Note that I just stared https://github.com/google/transit/pull/650 which would potentially interact if both PRs get approved. I made sure they are compatible!

manfredbrandl:
I think we should start with vehicle_types.txt because it should be enough to inform passengers of the utilities provided by the vehicle. Information about individual vehicles dont help passengers as long as we dont have realtime-info about utilities unusable on the individual vehicle.

doconnoronca:
> I think we should start with vehicle_types.txt because it should be enough to inform passengers of the utilities provided by the vehicle. Information about individual vehicles dont help passengers as long as we dont have realtime-info about utilities unusable on the individual vehicle.

Even without realtime information on what might not be working on an individual vehicle, it can be very useful to know if a vehicle that is supposed to have certain features is operating or not.

skinkie:
> > I think we should start with vehicle_types.txt because it should be enough to inform passengers of the utilities provided by the vehicle. Information about individual vehicles dont help passengers as long as we dont have realtime-info about utilities unusable on the individual vehicle.
> 
> Even without realtime information on what might not be working on an individual vehicle, it can be very useful to know if a vehicle that is supposed to have certain features is operating or not.

An operator has hunderds of vehicles, typically all of the same type. Without knowing which exact vehicle a trip / block (which is not in GTFS) is going to execute it, it will be just bloat.

manfredbrandl:
> Even without realtime information on what might not be working on an individual vehicle, it can be very useful to know if a vehicle that is supposed to have certain features is operating or not.

I do not agree!  Passengers are not intrested to know if vehicle x is operating today or not. Passengers are intrested if the vehicle which operates on their trip does have certain features. For that it is enough to know the vehicle type.

doconnoronca:
So do @manfredbrandl and @skinkie prefer only to have vehicle descriptions by vehicle type and not provide fleet number or fleet number ranges to possibly connect to real time data?

manfredbrandl:
@doconnoronca I would like to display the planned vehicle specifications to passengers for future journeys. However, only the planned vehicle type and its planned specifications are known in advance. The actual vehicle used, including its vehicle type and the features it actually possesses (and which are operational), is only known in the real-time data at the start of the service. The specific vehicle is not required in the planning data; similarly, in the real-time data, only the features that are actually present are required.

vikiawv:
I think passengers could also benefit from also knowing the exact vehicle class or vehicle number.

For instance ÖBB (from Austria) has a gigantic fleet of so called "Cityjets". Problem beeing: Every Cityjet under number 4746 158 has no power socket for USB-A or USB-C
Without the vehicle number there is no way on telling beforehand if the train is equipped with such a socket or not.

Additionally any Unit below Number 4746 158 is not equipped according to the standards for blind people. They can only travel with train 4746 158 and upwards allone, else they need company by either a seeing-eye-dog or another person that can see.

If in this case for example the Provider wants to submit this data on which exact carriage is used, it would be very great to know for blind people basically.

I am neither for nor against adding the vehicle number. I just think it adds value to specific usecases.

doconnoronca:
> @doconnoronca I would like to display the planned vehicle specifications to passengers for future journeys. However, only the planned vehicle type and its planned specifications are known in advance. The actual vehicle used, including its vehicle type and the features it actually possesses (and which are operational), is only known in the real-time data at the start of the service. The specific vehicle is not required in the planning data; similarly, in the real-time data, only the features that are actually present are required.

The proposal includes vehicle_class field added to the trips, routes and vehicles tables allows the planned vehicle type and its features to be known in advance.

manfredbrandl:
> The proposal includes vehicle_class field added to the trips, routes and vehicles tables allows the planned vehicle type and its features to be known in advance.

This proposal assumes that all vehicles are exported along with the timetable data. This means that new timetable data must be generated and imported, even if the timetable service (including vehicle types and their properties) remains unchanged. In the timetable data, we only have the vehicle types, whilst the individual vehicles are only available in the real-time system. How can we use this proposal to make vehicle properties available in both the timetable data and the real-time data?

doconnoronca:
> This proposal assumes that all vehicles are exported along with the timetable data. This means that new timetable data must be generated and imported, even if the timetable service (including vehicle types and their properties) remains unchanged.

I'm not sure why you are saying new GTFS schedule data must be generated if the time table and vehicle data hasn't changed. When new vehicles enter service?

> In the timetable data, we only have the vehicle types, whilst the individual vehicles are only available in the real-time system. How can we use this proposal to make vehicle properties available in both the timetable data and the real-time data?

The purpose of the vehicle ids and labels is to link the vehicles real-time to the vehicle properties in the schedule data.

The vehicle_class can be used when real-time data is unavailable.

If both are available consumers can highlight an unscheduled differences in the vehicle for users.

skinkie:
> > This proposal assumes that all vehicles are exported along with the timetable data. This means that new timetable data must be generated and imported, even if the timetable service (including vehicle types and their properties) remains unchanged.
> 
> I'm not sure why you are saying new GTFS schedule data must be generated if the time table and vehicle data hasn't changed. When new vehicles enter service?

That is not how it works. You expect there is a scheduled relation to an exact instance of a `vehicle_type`. That is not the case.


> > In the timetable data, we only have the vehicle types, whilst the individual vehicles are only available in the real-time system. How can we use this proposal to make vehicle properties available in both the timetable data and the real-time data?
> 
> The purpose of the vehicle ids and labels is to link the vehicles real-time to the vehicle properties in the schedule data.
> 
> The **vehicle_class** can be used when real-time data is unavailable.
> 
> If both are available consumers can highlight an unscheduled differences in the vehicle for users.

Again, you invent new terms. Don't do that, this is a `vehicle_type`.

gtfs/spec/en/reference.md
```diff
@@ -247,6 +248,7 @@ Primary key (`route_id`)
...
+|  `vehicle_class` | ID | Optional | Matches the `vehicle_class` in vehicles.txt to identify the type of vehicle that usually operates on this route. Can be used for information on what type of vehicle to expect when real time data is not available and identify where the vehicle is not the expected type when real time data is available. If trips.txt has a `vehicle_class`, that value overrides the value in routes.txt. |
```
skinkie, line 251:
```suggestion
|  `vehicle_type` | ID | Optional | Matches the `vehicle_type` in vehicles.txt to identify the type of vehicle that usually operates on this route. Can be used for information on what type of vehicle to expect when real time data is not available and identify where the vehicle is not the expected type when real time data is available. If trips.txt has a `vehicle_type`, that value overrides the value in routes.txt. |
```

gtfs/spec/en/reference.md
```diff
@@ -270,14 +272,14 @@ Primary key (`trip_id`)
...
 | `safe_duration_factor` | Float | **Optional** | Multiplier applied to travel time estimates calculated for on-demand trips.<br><br>See "Calculating on-demand trip time estimates with safe duration fields" section below for guidance on how to use this and the `safe_duration_offset` fields. |
 | `safe_duration_offset` | Float | **Optional** | Fixed offset value in seconds applied to travel time estimates calculated for on-demand trips.<br><br>See "Calculating on-demand trip time estimates with safe duration fields" section below for guidance on how to use this and the `safe_duration_factor` fields. |
+|  `vehicle_class` | ID | Optional | Matches the `vehicle_class` in vehicles.txt to identify the type of vehicle that usually operates on this trip. Can be used for information on what type of vehicle to expect when real time data is not available and identify where the vehicle is not the expected type when real time data is available. This value overrides the value in routes.txt. |
```
skinkie, line 275:
```suggestion
|  `vehicle_type` | ID | Optional | Matches the `vehicle_type` in vehicles.txt to identify the type of vehicle that usually operates on this trip. Can be used for information on what type of vehicle to expect when real time data is not available and identify where the vehicle is not the expected type when real time data is available. This value overrides the value in routes.txt. |
```

gtfs/spec/en/reference.md
```diff
@@ -270,14 +272,14 @@ Primary key (`trip_id`)
...
 
 
 #### Calculating on-demand trip time estimates with safe duration fields
 Together, `safe_duration_factor` and `safe_duration_offset` allow an estimation of the longest amount of time a rider can expect the on-demand trip to take, for 95% of cases. Data consumers are expected to use `safe_duration_factor` and `safe_duration_offset` to make the following calculation:<br>`SafeTravelDuration (seconds) = safe_duration_factor × DrivingDuration (seconds) + safe_duration_offset (seconds)`<br>where `DrivingDuration` is the time it would take a private car to travel the distance being calculated for the on-demand service, and `SafeTravelDuration` is the longest amount of time a rider can expect the on-demand trip to take.<br><br>
 
 This calculation should only apply to the portion of a trip that is on-demand. If a service is a deviated-fixed service, or if a rider trip includes a transfer from an on-demand to a fixed-route service, the duration of the fixed-route portion of the trip should be calculated according to the `departure_time` and `arrival_time` fields.
 
-
```
skinkie, line 280: Unrelated change.

gtfs/spec/en/reference.md
skinkie, line 932:
```suggestion
|  `vehicle_type` | ID | Recommended | Identifies a type of vehicle that usually runs on certain routes. Meant to link with vehicle_type fields in routes.txt and trips.txt. Can be used to provide vehicle information when label isn't available or highlight when the vehicle doesn't match the expected class and what the major differences are. |
```

gtfs/spec/en/reference.md
skinkie, line 943:
```suggestion
|  `image_url` | URL | Optional | Fully qualified URL of an image of the vehicle. The image should have as aspect ratio between 1:1 and 2:1 and a have a vertical resolution of at least 500. The image should show the front of the vehicle (except for train cars) and the side the passengers board. The rights to the image should also free commerical reproduction of the image, for example licenced under [CC-BY](https://creativecommons.org/licenses/by/4.0/). Consumers should only download this image one time for each GTFS update and always serve it to end users themselves. End users should never be automatically downloading directly from the URL provided. |
```
I think the licensing should be solved via https://github.com/google/transit/pull/607

skinkie (changes requested):
Rename `vehicle_class` to `vehicle_type`. Incorporate typo fixes.

skinkie:
Futhermore I am seeking for full alignment with attributes.

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,65 @@ The file defines the attributions applied to the dataset.
...
+File: **Optional**
+
+Primary key (`vehicle_id`, `agency_id`)
+
+This file describes the capacity, accessibility and features of individual vehicles or vehicle ranges.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+|  `vehicle_id` | Text | **Required** | An id that is unique to the vehicle and corresponds to the value in the GTFS Realtime `id` value in the `VehicleDescriptor` or `CarriageDetails` or the vehicle identifier in other real-time data sources. If `vehicle_id_high` is provided then this value is the low end of the range. |
+|  `vehicle_id_high` | Text | Optional | If a range of vehicles is represented this is the high end of the range. The number of charactors must be the same as `vehicle_id` and the range covers values with the same number of characters and are between `vehicle_id` and `vehicle_id_high` when using an ASCII string comparison. Not all vehicles covered by the range need to actually exist. |
```
gcamp, line 927: What is the expected behaviour when using UTF-8 strings?

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,65 @@ The file defines the attributions applied to the dataset.
...
+
+File: **Optional**
+
+Primary key (`vehicle_id`, `agency_id`)
+
+This file describes the capacity, accessibility and features of individual vehicles or vehicle ranges.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+|  `vehicle_id` | Text | **Required** | An id that is unique to the vehicle and corresponds to the value in the GTFS Realtime `id` value in the `VehicleDescriptor` or `CarriageDetails` or the vehicle identifier in other real-time data sources. If `vehicle_id_high` is provided then this value is the low end of the range. |
```
gcamp, line 926: It's a bit unclear to me what is the goal with CarriageDetails here. There doesn't seem to be a way to specify in the static that a vehicle is going to have multiple carriage, but you can define it in real time? 

Trying to see how we could potentially specify carriage label when a train has numbered/lettered carriage like google does here 

Image

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,65 @@ The file defines the attributions applied to the dataset.
...
+|  `capacity_seated` | integer | Optional | The number of seats available to passengers, including ones that may become unavailable due to making space for wheelchairs or bikes. |
+|  `capacity_full` | integer | Recommended | The number of passengers that is considered a full load for a vehicle. If passenger occupancy is provided, this should be the number of passengers that is considered Full or 100%. |
+|  `capacity_wheelchairs` | integer | Recommended | The number of spots available for passingers with wheelchairs. |
+|  `capacity_strollers` | integer | Optional | The number of spots available for strollers. Would include spots used by wheelchairs or strollers. |
+|  `capacity_bikes` | integer | Recommended | The number of bicycles the vehicle can carry, either on the inside or outside of the vehicle. |
+|  `bike_notes` | Text | Optional | A decription of the location, procedures and restriction of bringing a bicycle on the vehicle. |
+|  `capacity_cars` | integer | Optional | The number of standard size cars the vehicle can carry. Primarily for farries. |
```
gcamp, line 953:
```suggestion
|  `capacity_cars` | integer | Optional | The number of standard size cars the vehicle can carry. Primarily for ferries. |
```

gtfs/spec/en/reference.md
```diff
@@ -910,3 +912,65 @@ The file defines the attributions applied to the dataset.
...
+|  `agency_id` | Foreign ID referencing `agency.agency_id` | Recommended | Agency for the specified vehicles. |
+|  `license_plate` | Text | Optional | The license plate or number plate of the vehicle. Does not apply if a range of vehicles is being used. |
+|  `vehicle_class` | ID | Recommended | Identifies a class of vehicle that usually runs on certain routes. Ment to link with vehicle_class fields in routes.txt and trips.txt. Can be used to provide vehicle information when label isn't available or highlight when the vehicle doesn't match the expected class and what the major differences are. |
+| `facility_name` | Text | Optional | The name of the facility where the vehicle is normally stored when not in service, such as the bus garage or rail yard. |
+|  `vehicle_type` | Text | Recommended | Should be one of the two character codes in the Vehicle Type column on the Reference Sheet tab in [A-30 Template](https://www.transit.dot.gov/ntd/30-template) from the US Federal Transit Administration or a one or two word description of the type of vehicle. Examples:<br><br>`BU` - Bus<br>`LR` - Light Rail Vehicle<br>`RP` - Commuter Rail Passenger Coach |
```
skalexch: When downloading the sheet from https://www.transit.dot.gov/ntd/30-template, I cannot find the codes.

In addition, it is better to have a standardized list or to have some values enumerated in the description here. To make it even simpler, it could just either use a list of standardized code OR a text description, not both. If the purpose of the field is to provide more details on what the vehicle is, then better have it as text and not a list of specific values.

gtfs/spec/en/reference.md
doconnoronca: I could remove the "ASCII" so just says "string comparison".

gtfs/spec/en/reference.md
doconnoronca: Tthe GTFS real-time has id and label fields in the CarriageDetails structure and the id is supposed to "be unique per vehicle". That should allow you to find an associated vehicle description for that train car. If the GTFS real-time only has generic letters or numbers then it can't be used for this.

doconnoronca:
@skinkie `vehicle_class` is supposed to be more specific than `vehicle_type`.  `vehicle_type` 3 is bus, but there are many different sizes and configurations of buses, including trolley buses, articulated buses, double decker buses, minibuses, etc. which significantly effect the capacity and features riders expect.

skinkie:
> @skinkie `vehicle_class` is supposed to be more specific than `vehicle_type`. `vehicle_type` 3 is bus, but there are many different sizes and configurations of buses, including trolley buses, articulated buses, double decker buses, minibuses, etc. which significantly effect the capacity and features riders expect.

No. route_type = 3 is bus.
vehicle_type is the generic form of an instance of a vehicle.

Again: don't invent terms.

doconnoronca:
> No. route_type = 3 is bus. vehicle_type is the generic form of an instance of a vehicle.
> 
> Again: don't invent terms.

I'm inventing terms to avoid confusion with existing terms.

skinkie:
This is creating confusion. vehicle_type is the used term for it.

gtfs/spec/en/reference.md
gcamp: And there's no way to specify multiple carriage in the static GTFS?

doconnoronca: You mean to be able to provide the expected vehicle_class/vehicle_type for the carriages where real time data is not available?

gcamp: Yes, or even if real time is available, you might want to know in advance.

doconnoronca: Information on schedule consists is something that could be added to GTFS, but that would beyond the scope of this pull request.

There is [a structure](https://github.com/TIDES-transit/TIDES/blob/main/spec/vehicle_train_cars.schema.json) for that in TIDES, which is subject to a [proposal](https://github.com/TIDES-transit/TIDES/pull/272) to significantly revise it.

gcamp: That's fine that we add this capability later but I think we should make sure we design for it as a possible extension. Right now a `vehicle_id` can both mean a single carriage (via real time) or a train of multiple carriage which I don't like.

PR #638, opened 2026-05-21
Notices

leonardehrenfried:
## Summary

Add a few files to allow attaching notices to the following entities:

- trip
- segments of a trip
- route
- stop

Notices are short messages that provide additional non-structured information about a the entity.

## Describe the Problem

The problem is described in #620

## Use Cases

To add additional custom messages to passengers that do not fit into other places in the specification.

Examples:

- "Bike reservation recommended"
- "Special Train - Dedicated fare is required"
- "Express Bus - Please note the special trip times"
- "Seasonal Trip - Runs during the XYZ festival"
- "Dining car available"
- "Pre-booking required for bicycle conveyance"

In apps this shows up like this:

image

## Proposed Solution

I added three files:

- notices.txt
- notice_assignment.txt
- trip_segments.txt

The first two allow you to create and assign the messages to the relevant entities.

The third is a new type of entity, a "trip segment". It allows you to define parts of a trip that you can then apply a notice to. This is useful if the message doesn't apply to the entire trip.

## Type of change

GTFS Schedule
- [x] Functional Change
- [ ] Non-Functional Change

GTFS Realtime
- [ ] Specification Change

## Proposed Discussion Period

I recommend reserving at least one month for discussion to ensure everyone has sufficient time to discuss the proposal.

## Testing Details

- Consumer(s): OTP, MOTIS (tbc)
- Producer(s): MBTA (tbc) 
- Estimated Testing Period: 6 months

## Proposal Update Tracker 

| 2026-05-21 | Initial version |

## Checklist

cc @sebastianknopf @alec-georgoff


sebastianknopf:
If this is interesting for you, I actually convert GTFS data from raw ISA (a widely used proprietary format defined by IVU Traffic Technologies), which also contain notices. I could extend this converter for also rendering the notices according to this draft here. Just give me a sign.

leonardehrenfried:
Well, at some stage we need some real data to implement this. If you are interested in producing it, that would be great.

I have to read up how to move this proposal into the testing stage but since discussion appears to have died down, I think I am calling a vote to test soon.

leonardehrenfried:
So far I have only received heart emojis but no comments, presumably because we have disscussed this quite a bit in https://github.com/google/transit/issues/620.

Therefore I would like to move forward by getting a maintainer for this proposal. Can someone provide a hint how I get one?

I myself can be the advocate as well as a consumer contributor with OTP. It sounds like @felixguendling might be a consumer, too.

What I am looking for is a producer advocate. @sebastianknopf and @jfabi have indicated that they can provide data for it. Are you interested in the role?

sebastianknopf:
I think, the main discussion is already done in #620. In my option this is mainly a consumer driven topic and all representatives of consumers have mentioned their implications and we found a good solution.

I have a converter [isa2gtfs](https://github.com/sebastianknopf/isa2gtfs) which could be extended easily to generate some GTFS data according to this spec. Do you have a certain time, when you prefer to have the testing data? So I could spend some energy into generating some sample data.

leonardehrenfried:
Well, there is a north American agency that is interested in producing data for it, but I suspect it will take them a few month to actually get around to doing it.

If you're faster than that, I would love to implement your feed.

sebastianknopf:
I'm in the implementation yet. However, I found some restrictions in the base data: 
- If have no data for generating `trip_segments.txt` automatically. I could only create some trip segments manually for testing.
- I also have no data for stop-assigned notices. I also could create manually, if needed.
- I have some 'general notices' in the data, which would be mapped to the routes then, as they're related to the routes at all.

I hope thats fine for you anyway. All other notices which will be generated are related to trips. I think, I will be ready until next week with the implementation of the converter.

sebastianknopf:
Here you go... https://opendata.vpe.de/gtfs_notices.zip Was a little faster than I thought. This feed is a derivate of the Verkehrsverbund Pforzheim-Enzkreis GTFS feed. You can use if for testing as long as VPE is still existing. If you need an adaptions, just tell me, I'll try my best. The feed contains:

- Notices related to trips
- General notices related to routes

If you need additional realtime data, just take a look at opendata.vpe.de.

leonardehrenfried:
Thanks, @sebastianknopf. I will use that to build an implementation for OTP.

Regarding the missing notices for trip segments and stops: My current design makes it easy to add them later should we not be able to find a data producer for them. Therefore I'm relaxed about that.

leonardehrenfried:
So far we have myself as the advocate and data consumer and @sebastianknopf as the data producer.

What I need is a maintainer. How do I get one @etienne0101?

tzujenchanmbd:
Sorry for joining the discussion a bit late.

As mentioned in #620, some data producers currently use Service Alerts to communicate this type of information. Could we add language directly to this proposal to clarify the distinction between the two, similar to what @sebastianknopf [mentioned here](https://github.com/google/transit/issues/620#issue-4210438651)?

leonardehrenfried:
@tzujenchanmbd I adjusted the language a bit.

leonardehrenfried:
@etienne0101 @tzujenchanmbd Can any of you be the maintainer? Is that required to start a vote to test?

So far no-one has committed to produce data for stops or trip segments, so I'm leaning towards leaving that out of the first vote.

etienne0101:
Hi @leonardehrenfried  
Thank you for moving this proposal forward.

According to the [new governance](https://github.com/google/transit/blob/master/gtfs/Governance/change-process.md), this functional change requires an official review period, and two votes.

- The review period is at least **one week**. Despite having started the review here, there needs to be an **officially announced review period for any final changes.** You can start and announce it now.
- After that, the first vote is called to allow for testing the change. 
- If the vote passes, we proceed to the testing phase.
- The second vote happens after the testing phase and leads to adoption when it passes.

We can be the maintainer and assist you through the process.

leonardehrenfried:
I hereby announce the review period for this proposal.

Like that?

leonardehrenfried:
I was asked to add the following: This review period will end on Thursday, June 18th at 23:59 UTC, after which we'll initiate a vote to test.

ffisherBART:
I'm a proponent of including pathways as another entity to consider to allow this for. (and extending alerts to be assigned to pathways) Conceptually, that should be even more straightforward than creating a new trip_segment entity.

leonardehrenfried:
@ffisherBART Great. Will you be able to create data for it when the time comes?

leonardehrenfried:
Experimental support under review in OTP: https://github.com/opentripplanner/OpenTripPlanner/pull/7726

gtfs/spec/en/reference.md
```diff
@@ -849,6 +855,64 @@ Defines the booking rules for rider-requested services
...
 
+### notices.txt
+
+File: **Optional**
+
+Primary key (`notice_id`)
+
+Defines notices that can be displayed to riders. These can be attached to one or more entities with `notice_assignments.txt`.
+
+These notices are additional textual messages that can improve results with human-readable information. As opposed to GTFS-Alerts, they are already known during conceptual planning and therefore are not "real-time".
```
tzujenchanmbd:
```suggestion
These notices are additional textual messages that can improve results with human-readable information. As opposed to GTFS Realtime Service Alerts, they are already known during conceptual planning and therefore are not "real-time".
```

tzujenchanmbd (commented):
Suggestion to align the wording commonly used in the community.

leonardehrenfried:
The review period ended on June 18th 2026. There have not been substantial comments.

Are we ready to open the vote to test? Have I overlooked an aspect of the new governance?

gtfs/spec/en/reference.md
```diff
@@ -849,6 +855,64 @@ Defines the booking rules for rider-requested services
...
+- trip segment: the notice applies to part of a trip.
+- stop: the notice applies to a stop but completely decoupled from a specific trip. It is not intended for notices that target stop times.
+  
+Conditionally Required:
+- **Required** if [notices.txt](#noticestxt) is provided.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+| `notice_id` | Foreign ID referencing `notices.notice_id` | **Conditionally Required** | Identifies the notice to assign.<br><br>Conditionally Required:<br>- **Required** if `notice_group_id` is not defined.<br>- **Forbidden** if `notice_group_id` is defined. |
+| `notice_group_id` | ID referencing `notices.notice_group_id` | **Conditionally Required** | Identifies the notice group to assign. All notices sharing this `notice_group_id` in [notices.txt](#noticestxt) are assigned.<br><br>Conditionally Required:<br>- **Required** if `notice_id` is not defined.<br>- **Forbidden** if `notice_id` is defined. |
```
skalexch:
```suggestion
| `notice_group_id` | Foreign ID referencing `notices.notice_group_id` | **Conditionally Required** | Identifies the notice group to assign. All notices sharing this `notice_group_id` in [notices.txt](#noticestxt) are assigned.<br><br>Conditionally Required:<br>- **Required** if `notice_id` is not defined.<br>- **Forbidden** if `notice_id` is defined. |
```

gtfs/spec/en/reference.md
```diff
@@ -849,6 +855,64 @@ Defines the booking rules for rider-requested services
...
+Specifically, attaching them means the following:
+
+- route: the notice applies to _all_ trips in the route. If you want to assign to a specific trip use a `table_name` with value `trip`
+- trip: the notice applies to a specific trip in its entirety, not to individual hops.
+- trip segment: the notice applies to part of a trip.
+- stop: the notice applies to a stop but completely decoupled from a specific trip. It is not intended for notices that target stop times.
+  
+Conditionally Required:
+- **Required** if [notices.txt](#noticestxt) is provided.
+
```
skalexch:
```suggestion
```

skalexch (changes requested):
@leonardehrenfried just a couple structural suggestions only.

You can initiate a [vote-to-test](https://github.com/google/transit/blob/master/gtfs/Governance/change-process.md#step-24-vote-to-test) now.

leonardehrenfried:
I'm calling for a vote to test on this proposal. The voting ends on July 13 2026 at 23:59:59 UTC.

According to the new rules, please indicate if you're a producer, consumer or general contributor.

felixguendling:
+1 MOTIS (consumer)

leonardehrenfried:
Thanks for the vote, @felixguendling. Can you please edit your post to say that MOTIS is a consumer?

(I updated my post, too).

PatrickSteil:
+1 Patrick Steil, General Contributor (and consumer for some freelancer work)

PatrickSteil:
I really like the idea of having "static" notices, such as "no bikes allowed during rush hour" or "on-board Wi-Fi available". This kind of passenger-facing information can't (and shouldn't) be modeled as real-time alerts, so static notices seem like a great addition to GTFS.

optionsome:
+1 OpenTripPlanner (Consumer)

sebastianknopf:
+1 @sebastianknopf (Consumer mainly with OTP and some customer related apps, Producer with customer related projects)

MikeSki612:
+1 Metro Transit St. Paul-Minneapolis, MN (Producer)

We currently produce a similar custom file in a version of our GTFS package for some internal applications. We have already done some work to translate these trip 'comments' into the proposed notices format. I would like to echo some of the enthusiasm from other commentors about how this type of qualitative information could be valuable for customers.

skinkie:
+1 OpenGeo

leonardehrenfried:
@skalexch Since we have two consumers, two producer and more than five votes in total, can you confirm that this vote will pass, provided that we don't get any negative votes?

skalexch:
@leonardehrenfried [you're correct](https://github.com/google/transit/blob/master/gtfs/Governance/change-process.md#:~:text=Before%20testing%2C%20the,least%2014%20days.). As long as no negative votes come in before the end, the vote should pass. Keep in mind, a negative vote should come with actionable feedback, otherwise it's invalid.

leonardehrenfried:
The vote to test has passed with 6 votes in favour: 

- 2 consumers: MOTIS, OpenTripPlanner
- 2 producers: @sebastianknopf, Metro Transit
- 2 general contributors

leonardehrenfried:
The testing period has now started. OTP support is being reviewed and I believe that it will end up in a client website in a few weeks time.

I will therefore aim for a vote in early September.

HSL (Helsinki) was thinking about using this feature but they would like to add a title and URL field. So far their commitment to actually producing data for it is weak so I will not (for now) add those fields to the proposal. 

@skalexch if they do get around to producing this data, will this need another test vote?

skalexch:
@leonardehrenfried I just deleted my previous comment because only now I was able to read that they want to add fields to the proposal.

So according to the [governance](https://github.com/google/transit/blob/master/gtfs/Governance/change-process.md#step-24-vote-to-test:~:text=Problem%20identification%20during,consider%20withdrawing%20it.), if a change in the scope of the proposal happens during testing period, you can go back to the discussion step (which would inevitably require another vote-to-test), or abandon the proposal.

If it were a change in requirement or wording, it wouldn't have changed much. However, adding two fields seems to be a significant change in the scope.

In that case, what I suggest is either

1) Redoing the vote-to-test with the additions, or
2) Sticking to the proposal and only using the proposal-compliant subset of their implementation for testing. Then later on adding the other two fields in another proposal.

leonardehrenfried:
Notices are now exposed as a field in the OTP API. I also received a feed from Metro Transit (St. Paul, Minnesota) and have deployed both.

This is what it looks like in our API: https://t.ly/HG6of

This is an API explorer, not a passenger UI. Is the latter required to initiate a test to adopt, @skalexch ?

(The UI will come in the future, but we're not quite there yet.)

felixguendling:
The governance document says:

> First Adopters must apply, test, and share changes in a public-facing environment:
> - a link to a public-facing GTFS feed for Producers
> - a public-facing link to a GTFS consuming application for Consumers.

I think an API endpoint fulfills this requirement "a public-facing link to a GTFS consuming application" 100%.

Same here: https://github.com/google/transit/pull/546#issuecomment-4607111116

For MOTIS, I would usually wait for a release to announce that a new feature was added (as features on the development branch might still change significantly): https://github.com/google/transit/pull/546#issuecomment-4606256781

leonardehrenfried:
Metro Transit (Minneapolis) were kind enough to publish a feed for spec verification: https://svc.metrotransit.org/mtgtfs/next/GTFS_OTP.zip

Since we are getting closer to a vote, I would like to have a discussion about the following points:

#### Notice groups & stop notices

So far I have not found a feed that contains notice groups or notices assigned to a stop. Can they stay in this proposal or should they be taken out?

#### Trip segments

I only have data about individual stop_times, not actual segments. This means that from_stop_sequence and to_stop_sequence are identical.

I would like to be crystal clear about what exactly a segment is: does it apply to the visit at the stop _and_ the hop in between them or just to the hop. In such a case we have to redesign a little and create a separate entity for an individual stop time.

I am asking this, because I think this will be used for things like "dining car open from stop A to B". In the current definition the dining car will be open the second we arrive at stop A, when in reality it's probably not.

It's not a huge issue, so we can also agree to live with a bit of ambiguity.

felixguendling:
I'm not sure if that helps, but I would start from a use case perspective: Is it clear which notices should be displayed for
- departure and arrival boards
- full trip
- trips inside a journey, so only a partial trip (from departure of enter stop to arrival at exit stop)

Given that a trip segment covers always the interval from departure to arrival `[dep, arr]`, this does not make sense for a stop. So a definition for a notice that covers one stop would have to invert it to `[arr, dep]`. This would still not allow to set a notice to only `[arr, arr[` or only the departure `[dep, dep[`. Maybe I miss something, but I think this is currently not covered by the spec because the spec uses stop and not event granularity. Not sure if it's worth it to change that.


> Notice groups & stop notices

It's mostly a measure to keep feeds with many repetitive notices compact but if that's not important, we can also remove it. But since it's a feature that would be hard to add once implementations support ungrouped notices, we would probably have to live with larger feeds.

leonardehrenfried:
I think the usecases "full trip" and "partial trip" can be successfully implemented, but I'm not sure the proposal is good for the "departure board" case.

Here are a few ideas how to tackle it:

### 1: Change definition

When from_sequence_number and to_sequence_number are identical it means that it applies to a stop_time (stop visit) only and not a segment of the trip.

Not a lot of work, but smells a bit.

### 2: Allow assignment to an individual stop_time

Perhaps by writing a new column to stop_times that allows you to reference a notice or notice group.

Feels a bit inconsistent because the rest goes through notice_assignment.

### 3: Add information to trip_segment

Allow specifying that a trip segment is actually not a "segment" but refers to an individual stop visit.

This could be done by adding more columns to trip_segments.

I will be doing a bit of thinking - perhaps I have other ideas. If anyone else has, I am all ears.

PR #578, opened 2025-09-22
[GTFS Fares v2] Add network sets and relax constraint on networks in fare_leg_join_rules.txt

skalexch:
## Summary
This proposal:
1) Adds network set predicates which allow for matching with collections of networks in `fare_leg_rules.txt`
2) Relaxes the constraint that only legs with the same network can be joined in `fare_leg_join_rules.txt`, thus allowing joins across networks.

## Describe the Problem
1) There needs to be a method to define how a fare leg rule should match an effective fare leg (resulting from a `fare_leg_join_rule.txt` join) that spans multiple networks.
2) The current spec says in `fare_leg_join_rules.txt` that a leg with a `network_id` can only be joined to another leg with the same `network_id`, which is restrictive and does not represent multiple real-world cases,

## Use Cases
Network sets can be used in cases where routes from different networks can be priced together under the same fare scheme.

Three real-world cases are detailed in the [Network sets real-world case research](https://docs.google.com/document/d/16k9Ljaj9ej0aD1VocYR8PL4nYYcecwfRCE_UeAfHZ8A/edit?usp=sharing) document, including STM's Airport bus, the ORCA fare scheme in Seattle, and the multiple fare tiers of LeCar in the Aix-Marseille-Provence region.

## Proposed Solution
The full proposal can be found [here](https://docs.google.com/document/d/16IIvlVCqdHw3z84Xv_wGspRcl5uKHsvVKqdrZTBk1GI/edit?usp=sharing). It includes:
1) Adding Network Set Predicates to match effective fare legs to rules in `fare_leg_rules.txt`. These network sets are defined in `network_sets.txt` and mapped to routes in `network_set_elements.txt`.
2) For Network Sets to be usable, it is necessary to relax the constraint that only legs with the same network can be joined in `fare_leg_join_rules.txt`. By allowing leg joins across different networks, the effective fare leg will contain multiple `network_ids`.

This proposal was discussed and its scope was defined in multiple Fares v2 Working Group meetings. For further details, please check:
- [Meeting notes - 2025/05/27
](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit?tab=t.0#heading=h.yrj8wb4jpc5u)
- [Meeting notes - 2025/07/22](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit?tab=t.0#heading=h.awz1jbcydmhi)

[...]

PR #639, opened 2026-06-03
[GTFS-Fares v2] Add duration limit to fare_leg_join_rules.txt

skalexch:
## Summary
This proposal:
- Adds `duration_limit` and `duration_limit_type` to join rules in `fare_leg_join_rules.txt`.
- Suggests situations where it's possible to use either `fare_leg_join_rules.txt` or `fare_transfer_rules.txt`.
- Forbids overlap between `fare_leg_join_rules.txt` and `fare_transfer_rules.txt` for joins based on `network_id`.

## Describe the Problem
This was part of the [distance-based fares proposal](https://github.com/google/transit/pull/556). We decided to separate it in its own proposal for a few reasons:
1) The addition can actually help the fare implementation of many fare structures that are not distance-based (mentioned in Use Cases).
2) The distance-based fares proposal is moving slower than is required by the need for adding duration parameters to `fare_leg_join_rules.txt`.

## Use Cases
This is mainly to allow the usage of `fare_leg_join_rules.txt` as a simpler way to define transfers in fare structures such as:
- Regions with complex point-to-point or zone-based fares in which only the first and last stop determine the fare.
- (_Upcoming_) Regions with [cumulative zonal-fares](https://github.com/google/transit/pull/483) that have timed passes.
- (_Pending [Network Sets and the relaxation of the single network join constraint](https://github.com/google/transit/pull/578)_) Integrated fare systems with multiple agencies.
- (_Upcoming_) Regions with distance-based fares.

## Proposed Solution
Adding `duration_limit` and `duration_limit_type` to join rules in `fare_leg_join_rules.txt` and forbidding overlap between `fare_leg_join_rules.txt` and `fare_transfer_rules.txt` for joins based on `network_id`.

## Type of change

 

GTFS Schedule
- [x] Functional Change
- [ ] Non-Functional Change

GTFS Realtime
- [x] Specification Change

## Additional Information
 More research can be found in the [research document](https://docs.google.com/document/d/1PrEUBOmf3R2pZXzNf32A02WcawYfKTo-__nm3bpG-N4/edit?usp=sharing).

## Proposed Discussion Period
We propose a discussion period of **at least 2 weeks** before starting an official review period. We will start the review period later in June.

## Testing Details

- Consumer(s): 
- Producer(s): 
- Estimated Testing Period: 

## Proposal Update Tracker 

| (2026-08-06) | (Review period opened) |

## Checklist


etienne0101:
Hi @BodoMinea @jfabi, @felixguendling, @jasonad123, @halbertram @skinkie @bdferris-v2 @miklcct @jll01 @davidlewis-ito 

Following previous Fares Working Group discussions in 2025, we have decoupled the duration limit fields from the distance-based fares proposal into a standalone Pull Request.

We believe this addition is broadly useful for various fare structures beyond distance-based ones, especially for simplifying transfer definitions.

We are planning to initiate a "vote-to-test" in mid-July.
Please take a look at the PR and let us know if you have any feedback or if you are interested in testing these new fields.

Thank you in advance for your contribution!

gtfs/spec/en/reference.md
```diff
@@ -541,6 +541,20 @@ For a sub-journey of two consecutive legs with a transfer, if the transfer match
...
...limit_type` | Enum | **Conditionally Required** | Defines the relative start and end of `fare_leg_join_rules.duration_limit`.<br><br>Valid options are:<br>`0` - Between the departure fare validation of the first leg in the effective leg and the arrival fare validation of the last leg in the effective leg.<br>`1` - Between the departure fare validation of the first leg in the effective leg and the departure fare validation of the last leg in the effective leg.<br>`2` - Between the arrival fare validation of the first leg in the effective leg and the departure fare validation of the last leg in the effective leg.<br>`3` - Between the arrival fare validation of the first leg in the effective leg and the arrival fare validation of the last leg in the effective leg.<br><br>When an effective leg with the same `from_network_id` and `to_network_id` is matched multiple times consecutively within a multi-leg journey, the `duration_limit` specified by the effective leg should be measured starting from the first matched leg.<br><br>Conditionally Required:<br>- **Required** if `fare_leg_join_rules.duration_limit` is defined.<br>- **Forbidden** if `fare_leg_join_rules.duration_limit` is empty. |
```
felixguendling, line 545: The router doesn't know when exactly the ticket will be validated (i.e. walking time between fare gate and departure/arrival of the vehicle vice versa).

So I would change `departure fare validation` to `departure time` (same for arrival).

skalexch: Thanks for sharing your feedback @felixguendling. In fact, the wording is copied verbatim from the definition of `duration_limit_type` in `fare_transfer_rules.txt`. Therefore, both need to be changed if need be.

However, this was talked about in a few [2023 Fare Working Group Meetings](https://docs.google.com/document/d/1d3g5bMXupdElCKrdv6rhFNN11mrQgEk-ibA7wdqVLTU/edit?tab=t.0#heading=h.4srmxkxsaix4) when timeframes were being discussed.

The consensus seemed to be that in a real-world situation, the timeframe (or fare product transfer duration since it's similar) starts when tapping on the bus (0 min of delay) or at a fare gate (a few minutes of delay based on the size of the station, how long the rider will linger or if they are taking the first train). So the time limit for a fare product starts counting down when the rider taps. Therefore, the wording "fare validation" accurately describes the mechanism in real life.

In terms of router behaviour, there are ways of approximating the duration of the journey's walking-legs to be able to compare directly with `duration_limit`, among others:
- Using pathways if they exist.
- Approximating in-station walking time using OSM.
- Providing default in-station walking times.

This is left to trip planners to choose how they approximate it.

felixguendling: > This is left to trip planners to choose how they approximate it.

A good standard doesn't leave room for interpretation.

So I would propose to:

- either state that the timetable times (arrival / departure) can be used as they are. E.g. because the ticketing system itself leaves some room for walking by having more permissive limits or an average walking time is already subtracted from the duration limit
- or it's clearly stated that in-station walking times have to be calculated using GTFS pathways from fare gates (probably `pathway_mode=6`?) to tracks (or vice versa for arrival) - which would require a Fares v2 feed to always come with proper GTFS pathways.

I think that adding a `duration_limit` without a clearly defined way to measure durations (e.g. based on GTFS pathways; OSM is out-of-scope here I guess) isn't helpful if you want to consume the data.

skalexch:
@felixguendling I am planning to add this to the proposal both in the `fare_transfer_rules.txt` and `fare_leg_join_rules.txt` sections. We wanted to clarify that the `duration_limit` strictly adheres to the time limit on transfers defined by the agency. It is better to not introduce any approximations to the field itself. Therefore, `pathways` is the only way provided by the spec to calculate the station crossing time by trip planners for large stations.

Snippet to be added:
> #### Calculation of `duration_limit`
> 
> As implied by `duration_limit_type`, `duration_limit` starts and ends at fare validation events (e.g. a tap at a fare gate or onboard validator), not at the moment a rider physically boards or alights a vehicle. The `duration_limit` does not account for the time it takes a rider to cross between a fare gate and a platform.
> 
> This means `duration_limit` should be set to the same theoretical transfer time limit that the agency publishes to riders (e.g., "transfers must be completed within 90 minutes"), rather than to a measured or estimated travel time.
> 
> If the feed contains [pathways.txt](#pathwaystxt), data consumers can estimate this crossing time using `pathways.traversal_time`, `pathways.length` or `pathways.stair_count`.
> 

I'd like to hear your feedback on this!

felixguendling:
This is way better. However, it's still not on a level that would guarantee that if route planner (or any other consumer like a ticketing system) A and B both follow this definition it's guaranteed that the same ticket options will be shown in all cases. IMO this should be the goal with such a definition.

skalexch:
@felixguendling thanks for the feedback! I've updated the description and committed the change. Aside from keeping the same definition of `duration_limit` as the theoretical transfer validity time, I emphasized that consumers should use `traversal_time` when pathways exist. This guarantees that if producers and consumers follow this best practice, they'll arrive at the same result in the journey planner.

We didn't restrict consumers from using other fields when `traversal_time` isn't available, for two reasons: 1) we wanted to preserve that flexibility for consumers, and 2) `traversal_time` is not always populated unfortunately. In an ideal state, `traversal_time` would be used universally to estimate.

skalexch:
We are starting a one-week review period today before beginning the vote-to-test. The review period will conclude on 13 August 2026 at 23:59:59 UTC.
Please review the proposed changes and the formatting and share your feedback here.

skalexch:
We are starting a **vote-to-test** for this feature. The voting period will last 2 weeks and **ends on 01-09-2026 at 23:59:59UTC**.

skalexch:
The vote is concluded and failed. No votes were cast. We will see what can be improved and come back to this soon.

miklcct:
Sorry I completely missed this. I think that the definition should stay as the "fare validation event" while a router, lacking other information, can use the departure time and arrival time to approximate them.

PR #640, opened 2026-06-03
fix(transfers): min_transfer_time is conditionally required for timed transfers

rudiejd:
## Describe the Problem
 
The wording of `min_transfer_time` vaguely indicates that it might be required for timed transfers (`transfer_type` = 2). This is the interpretation that some consumers, such as Google, accept, but the column is specified as optional in the spec so the requirement is not a formal one.

## Proposed Solution
 
Make `min_transfer_time` conditionally required for timed transfers. 

## Type of change

 

GTFS Schedule
- [ ] Functional Change
- [x] Non-Functional Change

GTFS Realtime
- [ ] Specification Change

## Additional Information

MBTA had our feed rejected by Google because this requirement was not met. The requirement seems reasonable, but it is not specified strongly enough so as to be an unambiguous rule. 

## Checklist


huangh (approved):

rudiejd:
I'm announcing the start of the review period for this PR.

The review period will start today, 7/6/26, and will end 7/13/26 at 16:00 UTC.

gcamp:
+1 Transit

rudiejd:
a bit late, but the review period is over and we have started the vote to adopt!

The vote to adopt will conclude 7/27 at 16:00UTC. I do not have permission to post this to the google group, so I would appreciate if someone else could!

felixguendling:
+1 MOTIS

skinkie:
+1 OpenGeo

gcamp:
+1 Transit

rudiejd:
The vote has concluded. Results

For: 3
Against: 0

Since I didn't hit the required 5 vote / 2 producer 2 consumer threshold, this vote has failed. However, since no one seemed opposed, I will restart the vote and advertise it on slack. 

The new voting period begins now, and will conclude 8/14 at 4PM UTC

gcamp:
+1 Transit

skinkie:
+1 OpenGeo

felixguendling:
+1 MOTIS

leonardehrenfried:
I'm voting +1 OTP for testing, but what does the min_transfer_time actually mean for the timed transfer?

Does it mean that in the case of a delayed arrival the departure of next vehicle happens at "real-time arrival plus the min_transfer_time"? If that is the case, I would like to see that reflected in the spec.

rudiejd:
@leonardehrenfried 

yeah, I think that could be handled separately, and might entail a functional change depending on what the clarification ends up being. My assumption based on the wording:

>Transfer requires a minimum amount of time between arrival and departure to ensure a connection

was the minimum amount of time that it would take a passenger to get from the end `stop_id` of their current trip to the first `stop_id` of the next one - so in the case of a delayed arrival, a passenger might not have the time available to make the transfer. MBTA calculates the `min_transfer_time` value by using our pathways data to determine how long it would take a passenger to traverse from one `stop_id` to the next.

leonardehrenfried:
What does a timed transfer then guarantee in case the first vehicle is delayed?

If you can derive the time it takes to walk from the from_stop to the to_stop from the pathways, why do you need to add the time again in the transfer? Is it because there might be feeds without pathways?

leonardehrenfried:
Is it really _required_ to add the min_transfer_time for a timed transfer or should it just be allowed?

EDIT: Ah, sorry I just realised that that is exactly what you are aiming for with this PR.

rudiejd:
>What does a timed transfer then guarantee in case the first vehicle is delayed?

I would expect that consumers would not show a transfer when the difference between the current time and the `departure_time` of the transfer target is greater than `min_transfer_time` - I can't speak to how this is actually implemented on the consumer side, though.

>If you can derive the time it takes to walk from the from_stop to the to_stop from the pathways, why do you need to add the time again in the transfer? Is it because there might be feeds without pathways?

yep, there are producers that publish feeds without pathways - and there are some consumers that might support transfers but not pathways e.g. Aubin, Caternary per https://mobilitydatabase.org/gtfs-feature-tracker

sberkley:
+1 Sound Transit (Producer) https://gtfs.sound.obaweb.org/prod/40_gtfs.zip

brodyFlannigan:
+1 Transcollines / MRC des Collines-de-l'Outaouais (Producer)

rudiejd:
I forgot to post, but voting closed 8/14 (last friday) at 12:00 UTC

Vote results:

For: 6 (4 producers, 2 consumers)

Against: 0

This vote to adopt has succeeded. I don't have permission to notify the mailing list, but I'll make a post in MobilityData slack and ask for this PR to be merged. After this is merged, I would like to update (or have someone else update) to canonical schedule validator to make sure that `min_transfer_time` is populated for timed transfers.

skalexch:
@rudiejd congratulations! We will merge this PR soon. I will also ask around to see if we could give you access to be able to post the update regarding the proposal.

skalexch:
@rudiejd did you post and you're waiting for approval or you can't even draft anything to begin with?

rudiejd:
@skalexch I don't think I can send an email at all. I don't really have an active google account, but my email is jd@rudiejd.io

skalexch:
@rudiejd I think you need to link it somehow to a Gmail account, for this case, I posted on your behalf. @fredericsimard is there a way to solve this?

fredericsimard:
> rudiejd I think you need to link it somehow to a Gmail account, for this case, I posted on your behalf. 
> fredericsimard is there a way to solve this?

@skalexch @rudiejd Apologies, I'm not sure I ascertain what the issue is exactly? Could you tell me more please?

skalexch:
@fredericsimard Rudie cannot post to Google Groups. Not sure if there needs to be an addition to GG or because the email is not a gmail.

fredericsimard:
@skalexch I see. Groups have a setting to accept people from outside the org, but nothing that requires a Google account to be in the group, therefore any email address can be part of it. It is a mailing list system after all.

Someone needs to add him to the group, alas as I am not a manager on it I cannot do that for him, someone at Google presumably can. Who? No idea, sadly.

skalexch:
@etienne0101 could you add @rudiejd to GG? [jd@rudiejd.io](mailto:jd@rudiejd.io)

PR #648, opened 2026-07-13
Clarify spec: Add definitions to Cause and Effect values and clarify Trip Update and Service Alerts usage for consumer routing decisions

skalexch:
## Summary

MobilityData is currently developing Service Alerts guidance to help the community use Service Alerts more consistently. As part of this work, we are gathering feedback through the Service Alerts Working Group meeting.

In the Apr 28th, 2026 meeting, we presented and discussed detailed definitions of values for the Cause and Effect fields, drafted [here](https://docs.google.com/document/d/1U_al9cyN8OVWJqzyiENB1imWgX9KrsUL5AViUpHswPs/edit?usp=sharing),  to make sure that producers are attributing the correct values to real-world cases.

In addition, we discussed how consumers should interpret cases where both Trip Updates and Service Alerts apply but provide conflicting information. The group reached consensus that consumers should ~~by default give precedence to Trip Updates for routing decisions.~~ use both to inform routing decisions, and that producers should have the highest quality possible in their trip updates feeds.

## Proposed Solution
 

This PR
1) Adds a clarification to `feed-entities.md` describing that both trip updates and services alerts should be used to inform routing decisions. And that trip updates must always be accurate to what is happening in the transit network.
While this clarification will also be included in the Service Alerts guidance, we believe it is appropriate to include it directly in the specification as well, in order to make this shared expectation clearer.

2) Details the definitions for values of Cause and Effect, to make sure that service alerts are correctly categorized.

These changes are intended to clarify definitions and to clarify a principle for expected consumer behaviour, and they do not introduce a new field or functional change. Therefore, I propose that it can proceed to a vote without requiring implementation testing.

## Type of change

 

GTFS Schedule
- [ ] Functional Change
- [ ] Non-Functional Change

GTFS Realtime
- [x] Specification Change

### Additional information

[Here is the working group meeting notes](https://docs.google.com/document/d/1rL04SPOPzGxpAO2KikzH6F1QV-Zr25iZvdwCmRGaI5w/edit?tab=t.0)

We also plan to add an additional Service Alerts meeting in July to continue discussing guidance around Service Alert usage. [Registration page](https://community.mobilitydata.org/working-groups#events)

### DIFF
--- a/gtfs-realtime/spec/en/feed-entities.md
+++ b/gtfs-realtime/spec/en/feed-entities.md
@@ -32,6 +32,13 @@ A service alert will usually consist of some text which will describe the
 problem, and we also allow for URLs for more information as well as more
 structured information to help us understand who this service alert affects.
 
+#### Interaction between Trip Updates and Service Alerts
+
+If data consumers use Service Alerts to affect routing decisions, and both Trip Updates
+and Service Alerts apply but provide conflicting information, consumers should
+give precedence to Trip Updates for routing decisions, unless there is clear
+evidence that the Trip Updates is incorrect.
+
 [More about Service Alerts...](service-alerts.md)
 
 ## Vehicle Positions
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -413,7 +413,7 @@ The effect of this problem on the affected entity.
 | **NO_SERVICE** | No transit service to the specified entity(-ies).<br>- For stations or stops, the rider will not be able to board or alight.<br>- For routes, the route will not run.<br>- For trips, those specific trips are cancelled. |
 | **REDUCED_SERVICE** | The number or frequency of trips is reduced. |
 | **SIGNIFICANT_DELAYS** | The route will consistently run late (insignificant delays should only be provided through [Trip updates](trip-updates.md)). |
-| **DETOUR** | The route changes its shape, resulting in the route not serving one or multiple stops or picking up/dropping off at other locations. |
+| **DETOUR** | The route changes its shape, resulting in the route not serving one or multiple stops, or picking up/dropping off at other locations. |
 | **ADDITIONAL_SERVICE** | The number or frequency of trips is increased. For example, more buses are running to cover for a special event. |
 | **MODIFIED_SERVICE** | Operations are different from what the rider would normally expect. An example is an alert that reminds riders of an upcoming holiday schedule that is different from normal service on that day of the week. |
 | **OTHER_EFFECT** | Not represented by any of these options. |
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -410,10 +410,10 @@ The effect of this problem on the affected entity.
 
 | _**Value**_ | _**Comment**_ |
 |-------------|----------------|
-| **NO_SERVICE** | No transit service to the specified entity(-ies). For stations or stops, the rider will not be able to board or alight. For routes, the route will not run. For trips, those specific trips are cancelled. |
+| **NO_SERVICE** | No transit service to the specified entity(-ies).<br>- For stations or stops, the rider will not be able to board or alight.<br>- For routes, the route will not run.<br>- For trips, those specific trips are cancelled. |
 | **REDUCED_SERVICE** | The number or frequency of trips is reduced. |
-| **SIGNIFICANT_DELAYS** | The route will consistently run late (insignificant delays should only be provided through Trip updates). |
-| **DETOUR** | The route changes its shape, resulting in one or multiple stops being moved. |
+| **SIGNIFICANT_DELAYS** | The route will consistently run late (insignificant delays should only be provided through [Trip updates](trip-updates.md)). |
+| **DETOUR** | The route changes its shape, resulting in the route not serving one or multiple stops or picking up/dropping off at other locations. |
 | **ADDITIONAL_SERVICE** | The number or frequency of trips is increased. For example, more buses are running to cover for a special event. |
 | **MODIFIED_SERVICE** | Operations are different from what the rider would normally expect. An example is an alert that reminds riders of an upcoming holiday schedule that is different from normal service on that day of the week. |
 | **OTHER_EFFECT** | Not represented by any of these options. |
--- a/gtfs-realtime/spec/en/reference.md
+++ b/gtfs-realtime/spec/en/reference.md
@@ -386,21 +386,21 @@ Cause of this alert.
 
 **Values**
 
-| _**Value**_ |
-|-------------|
-| **UNKNOWN_CAUSE** |
-| **OTHER_CAUSE** |
-| **TECHNICAL_PROBLEM** |
-| **STRIKE** |
-| **DEMONSTRATION** |
-| **ACCIDENT** |
-| **HOLIDAY** |
-| **WEATHER** |
-| **MAINTENANCE** |
-| **CONSTRUCTION** |
-| **POLICE_ACTIVITY** |
-| **MEDICAL_EMERGENCY** |
-| **SPECIAL_EVENT** |
+| _**Value**_ | _**Comment**_ |
+|-------------|----------------|
+| **UNKNOWN_CAUSE** | Used when the cause of the disruption is not known or has not been determined. Should be avoided if any more specific cause can be reasonably identified. |
+| **OTHER_CAUSE** | Used for causes that do not represented by any of predefined options. |
+| **TECHNICAL_PROBLEM** | Issues related to vehicles or infrastructure malfunctioning. Examples include mechanical failure and information system bugs. |
+| **STRIKE** | Service disruptions caused by a labor strike involving the transit agency or operator. |
+| **DEMONSTRATION** | Service disruptions caused by public demonstration, protest, or gathering. |
+| **ACCIDENT** | Incidents such as collisions or derailments that disrupt services. |
+| **HOLIDAY** | Any public holiday that might result in a change of service. |
+| **WEATHER** | Disruptions caused by weather conditions such as snow, storms, flooding, etc. |
+| **MAINTENANCE** | Planned maintenance related to the services. Examples include track maintenance, elevator maintenance and vehicle maintenance. |
+| **CONSTRUCTION** | Any construction that affects the service, regardless whether it relates to the services or not. Examples include station construction and roadworks that affect bus routes. |
+| **POLICE_ACTIVITY** | Disruptions due to law enforcement activity or security incidents. Examples include investigations or security threats. |
+| **MEDICAL_EMERGENCY** | Incidents involving passenger or staff medical emergencies. Examples include an on-board medical problem that forces a train to stop. |
+| **SPECIAL_EVENT** | A special one-time or recurring event such as a parade, festival, performance, farmers market, or sporting event. |
 
 ## _enum_ Effect
 
@@ -408,19 +408,19 @@ The effect of this problem on the affected entity.
 
 **Values**
 
-| _**Value**_ |
-|-------------|
-| **NO_SERVICE** |
-| **REDUCED_SERVICE** |
-| **SIGNIFICANT_DELAYS** |
-| **DETOUR** |
-| **ADDITIONAL_SERVICE** |
-| **MODIFIED_SERVICE** |
-| **OTHER_EFFECT** |
-| **UNKNOWN_EFFECT** |
-| **STOP_MOVED** |
-| **NO_EFFECT** |
-| **ACCESSIBILITY_ISSUE** |
+| _**Value**_ | _**Comment**_ |
+|-------------|----------------|
+| **NO_SERVICE** | No transit service to the specified entity(-ies). For stations or stops, the rider will not be able to board or alight. For routes, the route will not run. For trips, those specific trips are cancelled. |
+| **REDUCED_SERVICE** | The number or frequency of trips is reduced. |
+| **SIGNIFICANT_DELAYS** | The route will consistently run late (insignificant delays should only be provided through Trip updates). |
+| **DETOUR** | The route changes its shape, resulting in one or multiple stops being moved. |
+| **ADDITIONAL_SERVICE** | The number or frequency of trips is increased. For example, more buses are running to cover for a special event. |
+| **MODIFIED_SERVICE** | Operations are different from what the rider would normally expect. An example is an alert that reminds riders of an upcoming holiday schedule that is different from normal service on that day of the week. |
+| **OTHER_EFFECT** | Not represented by any of these options. |
+| **UNKNOWN_EFFECT** | Used for alerts whose effect has not been identified yet. Do not use "Unknown effect" if the effect of the alert is known or can be deduced from the alert header or description. |
+| **STOP_MOVED** | A stop location is changed temporarily or permanently (if known to be permanent, ensure the new location is reflected in the schedule data). |
+| **NO_EFFECT** | The alert provides information to riders but does not affect operations. Examples include advertising public meetings and soliciting feedback via surveys. Unless necessary, it is advised not to use this effect. |
+| **ACCESSIBILITY_ISSUE** | The alert provides information about accessibility issues that affect step-free access. Examples include an out of service elevator or movable ramps, a bus that is unable to kneel, or an exceptional service with trains that have a big platform gap. |
 
 ## _enum_ SeverityLevel
 
--- a/gtfs-realtime/spec/en/service-alerts.md
+++ b/gtfs-realtime/spec/en/service-alerts.md
@@ -32,32 +32,35 @@ If you would like to represent an alert that affects more than one entity (e.g. 
 
 What is the cause of this alert? You may specify one of the following:
 
-*   Unknown cause
-*   Other cause (not represented by any of these options)
-*   Technical problem
-*   Strike
-*   Demonstration
-*   Accident
-*   Holiday
-*   Weather
-*   Maintenance
-*   Construction
-*   Police activity
-*   Medical emergency
-*   Special Event
+* **Unknown cause**: Used when the cause of the disruption is not known or has not been determined. Should be avoided if any more specific cause can be reasonably identified.
+* **Other cause**: Used for causes that do not represented by any of predefined options.
+* **Technical problem**: Issues related to vehicles or infrastructure malfunctioning. Examples include mechanical failure and information system bugs.
+* **Strike**: Service disruptions caused by a labor strike involving the transit agency or operator.
+* **Demonstration**: Service disruptions caused by public demonstration, protest, or gathering.
+* **Accident**: Incidents such as collisions or derailments that disrupt services.
+* **Holiday**: Any public holiday that might result in a change of service.
+* **Weather**: Disruptions caused by weather conditions such as snow, storms, flooding, etc.
+* **Maintenance**: Planned maintenance related to the services. Examples include track maintenance, elevator maintenance and vehicle maintenance.
+* **Construction**: Any construction that affects the service, regardless whether it relates to the services or not. Examples include station construction and roadworks that affect bus routes.
+* **Police activity**: Disruptions due to law enforcement activity or security incidents. Examples include investigations or security threats.
+* **Medical emergency**: Incidents involving passenger or staff medical emergencies. Examples include an on-board medical problem that forces a train to stop.
+* **Special Event**: A special one-time or recurring event such as a parade, festival, performance, farmers market, or sporting event.
 
 ## Effect
 
 What effect does this problem have on the specified entity? You may specify one of the following:
 
-*   No service
-*   Reduced service
-*   Significant delays (insignificant delays should only be provided through [Trip updates](trip-updates.md)).
-*   Detour
-*   Additional service
-*   Modified service: Operations are different from what the rider would normally expect.  An example is an alert that reminds riders of an upcoming holiday schedule that is different from normal service on that day of the week.
-*   Stop moved
-*   Other effect (not represented by any of these options)
-*   Unknown effect
-*   No effect: The alert provides information to riders but does not affect operations.  Examples include advertising public meetings and soliciting feedback via surveys.
-*   Accessibility issue: The alert provides information about accessibility issues that affects step-free access. Examples include an out of service elevator or movable ramps.
+* **No service**: No transit service to the specified entity(-ies).
+  * For stations or stops, the rider will not be able to board or alight.
+  * For routes, the route will not run.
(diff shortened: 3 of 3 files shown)


felixguendling:
> If data consumers use Service Alerts to affect routing decisions

Wouldn't it be nice to have a clear guidance on whether they should do this or not? At the moment it seems to be "do whatever you want"?

I also only see `NO_SERVICE` where this would be an option. Or are there others that should be considered for routing?

gcamp:
My comment [here](https://github.com/google/transit/pull/630#issuecomment-4374442536) still stands. From @tzujenchanmbd's response, the argument that there's flexibility in the guidelines doesn't remove the fact that I have the feeling that anyone considering NO_SERVICE in their routing will do the opposite of what the guidelines suggest. Might as well not have a guideline?

gtfs-realtime/spec/en/reference.md
```diff
@@ -386,41 +386,41 @@ Cause of this alert.
...
-| **ADDITIONAL_SERVICE** |
-| **MODIFIED_SERVICE** |
-| **OTHER_EFFECT** |
-| **UNKNOWN_EFFECT** |
-| **STOP_MOVED** |
-| **NO_EFFECT** |
-| **ACCESSIBILITY_ISSUE** |
+| _**Value**_ | _**Comment**_ |
+|-------------|----------------|
+| **NO_SERVICE** | No transit service to the specified entity(-ies).<br>- For stations or stops, the rider will not be able to board or alight.<br>- For routes, the route will not run.<br>- For trips, those specific trips are cancelled. |
```
gcamp, line 413: Was leaving out station entrances and pathway nodes done on purpose?

felixguendling:
> TU is generally a couple hours in the future and alerts are multiple weeks in advance.

What stops you from sending TU for multiple weeks in advance? Technically there's no blocker on the protocol level that would prevent this. It's a bit verbose, so that might be an indicator to improve TU rather than to (mis)use `NO_SERVICE`? The other option would be to clearly define that `NO_SERVICE` *has* to be considered during routing. The status-quo that everyone does what they want is something a standard should avoid. Most feeds are tailored to show the right thing for their "main" consumer (often that's Google Maps). But that's exactly what a good standard should prevent (tailoring the producer to a specific consumer).

leonardehrenfried:
I concur with @felixguendling: This dispute has been going on for so long that I frankly don't really care anymore what the standard position is as long as it is clear and the argument is over.

gtfs-realtime/spec/en/reference.md
skalexch: @gcamp yes. The topic of allowing alerts for pathways, entrances, exits is something that has been talked about for a while and we are planning to start that conversation soon. Building one what was mentioned on [issue#268](https://github.com/google/transit/issues/268#issuecomment-2784170466), the intent was to clarify some things in the realtime spec (such as here and in #546 ), create documentation to help with implementation and how to deal with use-cases (such as the spec vs RT Guidance and the RT Alerts Guidance), then eventually get to "alerts on pathway data". That conversation can involve all `location_types` and `pathway_modes`.

tzujenchanmbd:
As @leonardehrenfried mentioned, this has been an ongoing discussion for a long time. At the same time, we can also observe that Service Alerts are already widely used in practice, while some aspects of the documentation still need clarification. The current situation can lead to information that is difficult for riders to understand or even incorrect.

We (MobilityData) have been trying to find practical solutions that improve the current situation while remaining acceptable to different stakeholders within the community, including this PR, #546, and the [Service Alerts Guide](https://docs.google.com/document/d/14XIE0te4GDB0pYrLEJxNXa9SVUHps1ceFD6MMCSu-mw/edit?tab=t.0#heading=h.d917bcfd29f0), even if these compromise-based solutions are probably not perfect for everyone.

One observation from recent discussions is that the community seems to be increasingly recognizing the practical value of using NO_SERVICE as a routing signal. As @felixguendling suggested, perhaps we could explicitly indicate that NO_SERVICE has to be considered a valid input for routing decisions, which should align with what many trip planners already do today. We'd be interested in hearing what others think about making that expectation explicit.

However, once we clarify that NO_SERVICE can be used as a routing input, as well as its definition for different entities, a follow-up question is what consumers would do when NO_SERVICE conflicts with TU. This was discussed during the April Working Group Meeting, where the following direction emerged from the discussion:
- When SA is the only available data, consumers use NO_SERVICE as a routing signal.
- When both SA and TU are present **and conflict**, TU takes precedence by default.

During the discussion, this direction received support from the producers participating in the meeting, and representatives from both Google and Apple also expressed support in principle. ([Meeting notes](https://docs.google.com/document/d/1rL04SPOPzGxpAO2KikzH6F1QV-Zr25iZvdwCmRGaI5w/edit?tab=t.0#heading=h.tmzpm663puaw))

Even though there may be situations where consumers choose to prioritize SA instead (e.g. when there is clear evidence that TU is incorrect), I think defining a default behavior can still provide value by establishing a common expectation for handling conflicts. Does this seem like a reasonable balance that the community could align on?

leonardehrenfried:
My preference is to put everything routing-related in TU but I can get behind this definition.

felixguendling:
MOTIS is currently not considering `NO_SERVICE` relevant for routing and expects another TU. Currently TU=routing, SA=annotation. But as @leonardehrenfried, I'm open to change this to TU=routing, SA=routing+annotation if the GTFS standard clearly states that it has to be done this way.


> we clarify that NO_SERVICE can be used as a routing input

Please don't clarify that `NO_SERVICE` *can be used* as a routing input. Either clarify that it *has to be used* as a routing input or clarify that it *must not be used* as a routing input. If you say "*can*" be used as a routing input we are not making any progress. That's obviously already the status quo. I'm open for both options - as long as it's a clear "has to be used" (+ conflict resolution) or "must not be used".

skalexch:
@leonardehrenfried @felixguendling 
After the last RT Service Alerts Working Group Meeting, the group discussed this specific proposal. The main points concerned:

- The shorter timespan of Trip Updates which gives rise to the necessity of another disruption source for planning trips beyond that span.
- Risk of contradiction between Trip Updates and Service Alerts.
- Limited agency resources and consumers having to act quickly.

A consensus was reached in which:

1. Both Trip Updates and Service Alerts can be used to affect routing decisions. The group is also okay with using MUST and SHOULD for them. No priority order is put here.
2. Producers must have accurate Trip Update feeds. They also need to make sure that their alerts feed does not fully depend on their TU so that they can adjust it through other ways.
3. Producers must avoid contradictions between their feeds.
4. Consumers should use both feeds to inform routing decisions.

@gcamp @skinkie 

Your feedback is welcome!

leonardehrenfried:
I can live with this outcome and prefer it over the current ambiguity.

skalexch:
We are starting a **vote** to formally adopt these definitions and clarifications into the spec.

The voting period will span **1 week** and will end on **2026-08-20 at 23:59:59 UTC.**

gcamp:
+1 Transit

phil-swiftly:
+1 Swiftly

skalexch:
The vote concludes in one day. Tagging remaining participants from the previous working group meeting for visibility.
@bdferris-v2 @evansiroky @halbertram @jfabi @vicentecamposquezada

halbertram-ito:
+1 Ito

felixguendling:
> Data consumers SHOULD use both Trip Updates and Service Alerts to make routing decisions

Why only `SHOULD`? Why not `MUST`? For `SHOULD` a few examples should be listed in which situations you would not want to have service alerts influence routing.

```
SHOULD   This word, or the adjective "RECOMMENDED", mean that there
may exist valid reasons in particular circumstances to ignore a
particular item, but the full implications must be understood and
carefully weighed before choosing a different course.
```

Can we please state those "valid reasons in particular circumstances" explicitly? It's not obvious why the standard would not pin this down to "MUST".


> "such as cancelling a trip or closing a stop"

I would like to have an exhaustive list, not just examples. Currently I only see `NO_SERVICE` and the effect for routing should be clearly stated. For example for a stop: does `NO_SERVICE` mean I can't board/alight there or I can't even pass there, so all trips passing the stop are effectively cancelled? This leaves a lot of room for ambiguity. I don't see a good reason why the standard would leave this open?

----

Overall it leaves a lot of room for interpretation where it would help to be precise.

gtfs-realtime/spec/en/feed-entities.md
```diff
@@ -32,6 +34,21 @@ A service alert will usually consist of some text which will describe the
...
+
+If producers provide both Trip Updates and Service Alerts, producers SHOULD be able update their service alerts independently from their TripUpdates. These two feeds should not fully depend on each other.  
+
+Producers MUST also ensure that there is no conflict between their Trip Updates and Service Alerts feeds.  
+
+Examples of conflicts include:
+
+* A service alert informing of a stop closure while the trip updates for that stop are not set to `SKIPPED`.  
+* A service alert informing of a route closure while the trip updates for the cancelled trips are not set to `CANCELED`.  
+* A route closure spanning the whole day, for which the trip updates feed cancels trips over the next 90 minutes while no `NO_SERVICE` alert exists beyond those 90 minutes informing of the route closure.
```
leonardehrenfried, line 47: I find this sentence very hard to parse. Could you perhaps re-write it for maximum clarity (perhaps at the cost of some verbosity)?

gtfs-realtime/spec/en/feed-entities.md
```diff
@@ -32,6 +34,21 @@ A service alert will usually consist of some text which will describe the
...
+
+
+Data consumers SHOULD use both Trip Updates and Service Alerts to make routing decisions, such as cancelling a trip or closing a stop. In case there is a conflict between Trip Updates and Service Alerts, consumers SHOULD inform the agency/producer of the issue, and SHOULD exercise caution when applying either feed to affect routing decisions.
```
leonardehrenfried, line 50: So, cancelling a route or agency recursively cancels all trips of that entity? I think this would be a very valuable clarification.

gtfs-realtime/spec/en/service-alerts.md
```diff
@@ -32,32 +32,35 @@ If you would like to represent an alert that affects more than one entity (e.g.
...
-*   Modified service: Operations are different from what the rider would normally expect.  An example is an alert that reminds riders of an upcoming holiday schedule that is different from normal service on that day of the week.
-*   Stop moved
-*   Other effect (not represented by any of these options)
-*   Unknown effect
-*   No effect: The alert provides information to riders but does not affect operations.  Examples include advertising public meetings and soliciting feedback via surveys.
-*   Accessibility issue: The alert provides information about accessibility issues that affects step-free access. Examples include an out of service elevator or movable ramps.
+* **No service**: No transit service to the specified entity(-ies).
+  * For stations or stops, the rider will not be able to board or alight.
+  * For routes, the route will not run.
+  * For trips, those specific trips are cancelled.
```
leonardehrenfried, line 56: Can you set an agency to NO_SERVICE? If yes, what happens then?

leonardehrenfried:
I recognize that I'm coming in late but I have a few comments:

I agree with @felixguendling that the current wording is only a small improvement and would like to see it spelled out in more detail.

Questions:

- is it only NO_SERVICE that close stops or POLICE_ACTIVITY and CONSTRUCTION, too

felixguendling:
I'm clearly "pro clarification", so I support this effort, but in it's current status, I vote:

-1 MOTIS

I think we don't have to rush this one. And it's not much that's missing to make it a proper clarification.

I'm open to work on a clear table that maps "effect" to an equivalent TripUpdate (for everything that has a routing impact) with "for all" clauses. As long as TripUpdates are then clearly defined, this will give us a 100% clear guidance of how to apply ServiceAlerts for routing.

leonardehrenfried:
I hate to make this more painful than it already is but I agree with Felix.

-1 OTP

gtfs-realtime/spec/en/reference.md
```diff
@@ -386,41 +386,41 @@ Cause of this alert.
...
+| **REDUCED_SERVICE** | The number or frequency of trips is reduced. |
+| **SIGNIFICANT_DELAYS** | The route will consistently run late (insignificant delays should only be provided through [Trip updates](trip-updates.md)). |
+| **DETOUR** | The route changes its shape, resulting in the route not serving one or multiple stops, or picking up/dropping off at other locations. |
```
gcamp, line 416:
```suggestion
| **DETOUR** | The route changes its shape, which may affect pickup/drop off at some stops. |
```

Detour doesn't mean stops are cancelled

gcamp (commented):
On the SHOULD vs MUST, my understanding is that since we have this language
> Producers MUST also ensure that there is no conflict between their Trip Updates and Service Alerts feeds.  

The idea is that consumer don't have to consume both alerts and trip updates, if they don't want to. It's basically a way to have current consumer not taking alerts into account not having the obligation to update their implementation. I agree wording can be improved. 

My point at the meeting was that alerts are useful in the real world, since trip updates often lack the required cancellation, so the MUST ensure no conflict was a way that we all agreed. 

That said I agree wording could be improved

felixguendling:
> The idea is that consumer don't have to consume both alerts and trip updates, if they don't want to.

If the consumer happens to be a routing engine, then that routing engine will definitely produce wrong routing results if the producer happens to produce `NO_SERVICE` alerts and expects routing results to change without also encoding the same service change as TU?

Maybe some consumers can ignore ServiceAlerts but if we say that `NO_SERVICE` (and potentially others) should impact routing, I don't see another option than to make consuming ServiceAlerts mandatory for routing engines? If we say "if they don't want to", they can ignore SA for routing, then we are again at the starting point, i.e. the status "you can do whatever you want".

I think we should aim for more consistent behavior of routing results across routers. This can only be achieved by making those very fundamental decisions (like SA impacts routing or not) strict requirements. The other option is that GTFS is just pure syntax and every producer/consumer chooses the semantics on a "do whatever you want" basis making 1:1 alignment between producers+consumers necessary. IMO this is something a good standard should avoid. 100% compatibility = the default: you can't claim GTFS-RT support and then not be compatible in practice.

I would be interested to learn from someone who implemented a routing engine that already implements "SA impacts routing results" how they did this: is it only `NO_SERVICE` or also others? Is it possible to set `NO_SERVICE` on a complete agency which should then kill all trips of that agency in the complete impact period? That sounds dangerous in terms of "oops, wrong button -> nothing works" but could also be useful to model a strike?

gcamp:
> if the producer happens to produce NO_SERVICE alerts and expects routing results to change 
That's why there's the language that alerts and trip updates should match. If they don't, then it's against the spec. The proposal here is exactly what you're pushing for : a consistent behaviour if everyone follows the spec. 

Why not enforcing the alert consideration for consumers too? Because there was strong voice against that it should not be necessary. By putting `Producers MUST also ensure that there is no conflict between TU and SA` we solve the impasse while enforcing something everyone agrees on. 

> I would be interested to learn from someone who implemented a routing engine that already implements "SA impacts routing results" how they did this: is it only NO_SERVICE or also others? Is it possible to set NO_SERVICE on a complete agency which should then kill all trips of that agency in the complete impact period? That sounds dangerous in terms of "oops, wrong button -> nothing works" but could also be useful to model a strike?

Google and Transit only does it for NO_SERVICE yes. I think Google has some heuristics, and we also have feature flag depending on the agency. 

We support NO_SERVICE on an entire agency, but yes it's too high risk of false positive, so we again have a separate feature flag that we manually enable. We also support stop_id on a station entrance that is explicitly disallowed in this proposal.

felixguendling:
> By putting Producers MUST also ensure that there is no conflict between TU and SA we solve the impasse while enforcing something everyone agrees on.

Totally agree 👍 


> Google and Transit only does it for NO_SERVICE yes

I would be happy to see this as hard requirement in this PR. Since Google and Transit do it this way, it's proven to work and Google+Transit don't have to put any additional effort and the rest can align on this as well and know exactly what to do.


> We also support stop_id on a station entrance that is explicitly disallowed in this proposal.

I think disallowing it is a bad idea because any future change might become a compatibility issue. We can either keep this undefined (to clearly define it in the future) or just clearly define that all pathways through this entrance are considered infeasible, requiring the router to find new shortest paths (which might make a whole journey infeasible). It would also be interesting to allow referencing pathways to disable elevators in the future. We currently only support this for OSM+SIRI in MOTIS.

skalexch:
The vote has concluded and failed. The result is as follows:
+1: Transit, Swiftly, ITO (@gcamp @phil-swiftly @halbertram-ito )
-1: OTP, MOTIS (@leonardehrenfried @felixguendling )

The criticisms made above make sense. Based on the discussion so far, we will get back soon with some suggestions to make the proposal clearer and more practical.

leonardehrenfried:
I am sorry for voting -1. This has been the first time I ever did this.

I don't reject this feature outright and will vote in favour if the wording is clarified.

gtfs-realtime/spec/en/feed-entities.md
skinkie: The discussion was something like: some producers are unable to make new timetables. For months they are using their AVL system to skip a stop. But their AVL system has a horizon of just 90 minutes. Hence every trip planned after 90 minutes would still show the regular timetable.

For me it is also very hard to comprehend why people would _accept_ this quality of their vendors.

gtfs-realtime/spec/en/feed-entities.md
skinkie: My suggestion was more in the direction: there should be to cancellation flags. One from tripUpdates and one from ServiceAlerts.

gtfs-realtime/spec/en/service-alerts.md
skinkie: There is a strike :-)

gtfs-realtime/spec/en/feed-entities.md
felixguendling: Does it make sense to design the GTFS standard for the worst practices? I would rather design for good practices to produce great results.

In case they really can't produce a static GTFS file for months, I would rather write a little script that takes the GTFS file and the GTFS-RT protobuf and update the GTFS file accordingly. In most systems data coming in via GTFS-static will also be displayed differently than GTFS-RT and in this case you would probably want the GTFS-static version.

gtfs-realtime/spec/en/service-alerts.md
nlee-septa: I would expect that setting an agency to NO_SERVICE would be equivalent to setting all of that agency's routes to NO_SERVICE. Aside from a labor strike, this might also be used for extreme weather such as blizzards or typhoons/cyclones/hurricanes.

gtfs-realtime/spec/en/reference.md
skalexch: There was a previous comment from @ckraatz mentioning that if a detour does not change the behaviour at any stop, then it does not matter to the rider. And that can be communicated via a SIGNIFICANT_DELAYS alert.

I like @gcamp's wording, but would like to discuss whether the DETOUR alert "may affect" or "will affect" pickup/drop off at some stops.

One case I see for keeping "may" and including detours that only change the shape is in Montreal; where [STM buses can drop people off at night between stops](https://www.stm.info/en/info/advice/travelling-alone-night) if they are traveling alone or with a child, for safety reasons. If a detour were to change the shape but keep the same behaviour at the stops, it would be valuable to communicate the disruption as a DETOUR alert to let people know that the itinerary changed and that their "between-stops" drop-off might not be done.

skalexch:
Hello everyone, we have taken into account the feedback you provided. Our suggestion is to do the following:
- Changing wording to “Consumers MUST use both Trip Updates and Service..."
- Restricting the usage of alerts to affect routing for the effect “NO_SERVICE”: using MUST as well.
- As for "consumers MUST ensure that there is no conflict between TU and SA". How is that done in practice? If a consumer receives conflicting feeds, what should they do? @gcamp @felixguendling 

- @leonardehrenfried I will clarify that conflict example, what I meant there was: a route closure is communicated (via non-GTFS channels) for the whole day, TU has trips cancelled for the next 90min, beyond the scope of TU, there is no alert mentioning `NO_SERVICE` to be used for the remainder of the day.
This is not an example we can validate using GTFS tools, but it's a practical case of a conflict with real-world information.

If there's any missing detail, please share it here.

gtfs-realtime/spec/en/reference.md
gcamp: If we want to make a detour that doesn't affect stop SIGNIFICANT_DELAYS, that's probably ok, but we should be explicit in the spec. I think in most people's mind a detour is a detour, affecting stops or not. 

What you're referencing for the STM is more akin a deviated service than a detour.

felixguendling:
> As for "consumers MUST ensure that there is no conflict between TU and SA". How is that done in practice? If a consumer receives conflicting feeds, what should they do? @gcamp @felixguendling

I think the idea of @gcamp was the opposite?

> By putting Producers MUST also ensure that there is no conflict between TU and SA we solve the impasse while enforcing something everyone agrees on.

That means that a producer is not allowed to produce conflicting TU vs SA (i.e. `NO_SERVICE` on a stop/trip <=> `SKIP` on this stop/trip). If we make this rule, we don't have to provide a rule how consumers have to deal with conflicting feeds.

gtfs-realtime/spec/en/reference.md
tyler-m-cooper: I would agree here that the use of the DETOUR status alone should not mean stops are cancelled. Perhaps language should be added here which states: "The route changes its shape. Routing must only be affected if TripModifications are also provided which detail the new shape of the route and any removed or replaced stops."

gtfs-realtime/spec/en/feed-entities.md
tyler-m-cooper: I'll start by saying that yes, I agree that the standard should be built for good practices and not cater to bad vendor behavior. I also understand where you're coming from that in principle GTFS static should be updated in cases of known route changes. However, I just want to offer a slightly different framing with what this actually looks like IRL rather than just in digital. 
In the physical world, an agency may post a temporary sign on a bus stop that says a route will be diverted away from this stop for an amount of time due to construction. The bus stop sign itself still remains and lists that route as stopping there, since it will eventually return after the disruption is over.
The digital equivalent of updating GTFS static in this scenario is more like removing the bus stop sign entirely until the disruption is over. There hasn't been an official service "change" per say, just a temporary adjustment. Basically, by updating GTFS static there's no definitive way to link the change to a "why" without the service alert.

felixguendling: > there's no definitive way to link the change to a "why" without the service alert

I think this shortcoming of GTFS static will be solved by https://github.com/google/transit/pull/638 (no boarding + no alighting + notice)

skalexch:
Thanks for the clarification @felixguendling, I read it the other way.

Issue #614, opened 2026-02-21
Information which carriage is at what exit


vikiawv:
Include data on which carriage is close to what exit.

The idea emerged from a user in this discussion: https://codeberg.org/comaps/comaps/issues/2879
I have refined his idea to include sort of a prototype on what could be done. (I however have very basic knowledge only of GTFS) - There is a Libreoffice Calc document linked at the discussion as demo purpose if you want to take a look at it ;)

For now it is this basic idea:
`CarriageNumber;Exitname;Stationexits;(Lines to change[optional direction])`
Example: `A;Stubenring;Stubenring,Dr. Karl Lueger Platz;(2,2A,74A)`
For real use it probably needs to include more data about the line, line direction….

However two problems remain unsolved: What if it is a shorter train than usual? What if the train gets turned around at a stop not listed in the Lines destinations - can it then identify that it still is within the same direction.

If this data is not submitted by the PTA it is the idea that this data could be community created as sort of add-on for regular GTFS data.

- - - -

Is this all feasible within GTFS and does it follow the projects directions?

gcamp:
I have been thinking about this lately (good timing on this issue!), and this is what I came up with : https://gist.github.com/gcamp/f6adda87793f9e7e6f8d9da36ef7584b

This is not final in any way. http://bit.ly/gtfs-vehicles also had a version of this, a much more complete but also much more complex version.

gcamp:
@jfabi thanks for suggestions! 

I did the following modifications. 
- `carriage_count` added to the primary key of carriage_positions.txt as suggested 
- `route_id` and `direction_id` added to station_directions.txt — the platform side now applies as a default for
   all trips, with optional per-route and/or per-direction overrides. This covers bidirectional single-track need. I added these to
  station_directions.txt only (not carriage_positions.txt) since the mirroring logic already handles all the cases. The `route_id` was added by suggestion of @timMillet for some edge cases. 

Regarding accessibility: I think the existing facility_type field already covers this, let me know if I'm missing something. 

[Updated proposal](https://gist.github.com/gcamp/f6adda87793f9e7e6f8d9da36ef7584b). I might start a PR soon, might be easier to discuss there.

jfabi:
@gcamp In terms of the accessibility piece: I believe, how your proposal is currently framed, `facility_type` is to be empty when `to_stop_id` is provided (as a transfer platform or exit).

[...]

PR #650, opened 2026-07-29
GTFS Carriage Position

gcamp:
## Summary

This proposal adds two new files, `platform_directions.txt` and `carriage_positions.txt`, that let a feed express which carriage a rider should board to minimize walking at their destination.

## Describe the Problem

To speed up their trip, passengers want to know which carriage to board for a fast transfer or a fast exit at their destination. `pathways.txt` can give a rider turn-by-turn directions inside a station, yet nothing tells them where to stand on the platform before boarding. Agencies often have this information but GTFS has no way to encode it.

Discussion started in https://github.com/google/transit/issues/614. Thanks @vikiawv @jfabi for the comments and suggestions! 

## Proposed Solution

Two new files:

- `platform_directions.txt` defines which side of the platform the front carriage stops at. One record per platform covers all trips, and optional `route_id` and `direction_id` records take priority over that default so special configurations can be modeled, such as a platform on a bidirectional single-track section. Conditionally Required: required when `carriage_positions.txt` is provided, since a recommendation cannot be interpreted without knowing platform orientation.
- `carriage_positions.txt` maps an arrival platform, and optionally a specific transfer platform or station entrance/exit, to the recommended carriage to board. The optional `facility_type`, which reuses the `pathways.txt` `pathway_mode` values for stairs, escalator and elevator, covers cases where the elevator serving a destination is several carriages away from the stairs serving the same destination, so riders needing step-free access can be sent to a different carriage.

Recommendations are expressed relative to the front of the train at the arrival platform. When the departure platform is oriented the other way, planners mirror the carriage with `carriage_count - recommended_carriage + 1`, so a single record stays usable regardless of which end of the train the rider boards from.

`recommended_carriage` is numbered from `1` for the first carriage in the direction of travel, matching GTFS-realtime `CarriageDetails.carriage_sequence`. Consumers that already show per-carriage occupancy from real-time data can then combine both without translating between two numbering schemes for the same concept, which would be a likely source of off-by-one errors.

`carriage_count` is part of the primary key, which lets a platform carry different recommendations for trains of different lengths, and producers without exact carriage positions can use logical divisions instead: `carriage_count=3` with `recommended_carriage=1` simply means "board near the front".

Here is an example of a dataset I made from the Paris GTFS and the [open data carriage position CSV they have](https://prim.iledefrance-mobilites.fr/fr/jeux-de-donnees/positionnement-dans-la-rame). 

## Type of change

GTFS Schedule
- [X] Functional Change

## Additional Information

This proposal is deliberately a small subset of what [GTFS-VehicleBoardings](http://bit.ly/gtfs-vehicles) attempted: it does not model platform sections or train formations, and instead encodes only the boarding recommendation itself.

[PR #636](https://github.com/google/transit/pull/636) proposes `vehicles.txt` along with a `vehicle_class` on routes.txt and trips.txt. This proposal does not depend on it and stays usable on its own through `carriage_count`, but if #636 is adopted, an optional `vehicle_class` in `carriage_positions.txt` would be a natural follow-up to select the right configuration for a trip.

## Proposed Discussion Period

I recommend reserving one month for discussion, since this adds two new files and affects station modeling. _Extended an other month to end of September_

## Testing Details

- Consumer(s): Transit
- Producer(s): TBD (please mention if you want to produce!)
- Estimated Testing Period: TBD

## Proposal Update Tracker

| July 29 2026 | Initial PR submission |
| August 31 2026 | Extend discussion period |


skalexch:
@gcamp thank you for raising this proposal. Best carriage information is an important addition to the spec.
I have a few remarks/questions:

1. `front_car_position` should probably be renamed to `front_carriage_position`, a less regional term and more consistent with the file name.
2. `front_car_position` does not need to be specified for both the departure and arrival platforms in order to make a recommendation. If someone is going to alight at a platform, they should still be able to get a recommendation for where to board initially, using only the front carriage position at the arrival platform.
3. The primary key for `carriage_positions.txt` could be extended or made to be the entire row. Sometimes more than one carriage could be "best". For example, a pathway in the middle of the platform can be best accessed via carriages 4 and 5 on an 8-carriage train.
4. In the proposal, `to_stop_id` can only reference a stop/platform or an entrance/exit, generic nodes are not allowed. Generic nodes aren’t necessarily a bad thing: if the producer links all platforms to generic nodes, those pathways could be used to chain together a best-carriage recommendation (I.e. getting the best carriage that leads to the generic node which leads to the exit).
5. I conceptually like the fact that recommended_carriage and carriage_count could be logical enums (e.g. carriage_count=3 to denote front, middle and back). However, if a consumer uses them literally they might end up recommending `carriage_nb=2` while the rider is meant to take the middle carriage. There should be consistency there between what a producer and a consumer interpret. Better to introduce a separate field for categorical boarding positions.

gcamp:
@skalexch thank you for the comments! 

> 1. front_car_position should probably be renamed to front_carriage_position

👍 
> 2. front_car_position does not need to be specified for both the departure and arrival platforms in order to make a recommendation. 

~Can you explain me a bit more because I don't see how we can do without. The recommendation of where you board on the origin will matter if the train front carriage is left or right on that station. See this [line](https://github.com/google/transit/pull/650/changes#diff-3ecf0760eb54b4953728042a1e30586705dc2335807be94faae0de5829cd12a1R773)~ Edit : Never mind, I though a bit more about and you're right! Did the edit
> 3. The primary key for carriage_positions.txt could be extended or made to be the entire row. Sometimes more than one carriage could be "best". For example, a pathway in the middle of the platform can be best accessed via carriages 4 and 5 on an 8-carriage train.

👍 
> 4. In the proposal, to_stop_id can only reference a stop/platform or an entrance/exit, generic nodes are not allowed. Generic nodes aren’t necessarily a bad thing

I'm mixed on that one. I don't want to complexify the implementation too much, but I see the benefits. I'm going to change but if others think it's too much let me know. 
> 5. Better to introduce a separate field for categorical boarding positions.

I guess my counter question is why would a consumer care about if the num carriage is true or not in the context of boarding location? If we added a field I'm not sure how we would use it (re guiding principles of speculative features)

skalexch:
@gcamp thanks for the changes! For the two remaining questions:
4) +1, it will complexify but it's better to not prohibit the usage of generic nodes altogether because they can exist in a station's pathways graph and their use case could be valid. Prohibiting them now will open a versioning or spec correction discussion in the future if someone supporting generic nodes decides to implement best carriage info.

5) I'm not sure if I understood your question. The issue I see arises from this case:
Assume a consumer displays this phrase _"Take carriage `{recommended_carriage}` out of `{carriage_count}` from your right to exit easily"_.
The consumer gets a GTFS feed with `recommended_carriage=2` and `carriage_count=3`. The producer's implication is to take the MIDDLE (2) carriage out of [FRONT (1), MIDDLE (2), BACK (3)].
In this case, the consumer will show _"Take carriage 2 out of 3 from your right to exit easily"_. Because the consumer does not know that 3 is not the actual number of carriages, but a logical position enum. If the train in question has 8 carriages, that message might confuse the rider.

As it currently stands, the proposal works 100% clearly and correctly only if the consumer shows a graphic of the relative position of the rider after doing the calculation `recommended_carriage/carriage_count`. If the consumer wants to be specific on which exact carriage the rider should ride, then `carriage_count` has to always equate the real carriage count.

I like both concepts of being able to show either the actual carriage or a relative position [FRONT, MIDDLE, BACK]. But I think both of them being modelled using the same fields might cause misinterpretation.

gcamp:
>If the consumer wants to be specific on which exact carriage the rider should ride

My point is that is not something we want to do. I'm not against this use case, but it's a speculative feature unless somebody else ask. What we could do is rename the field and state that it's _only_ a virtual representation of the carriage length, and keep the door open for an other field in the future.

gtfs/spec/en/reference.md
```diff
@@ -756,6 +760,49 @@ Pathways are intended to exhaustively define the internal access graph of a stat
...
+
+[carriage_positions.txt](#carriage_positionstxt) `recommended_carriage` counts carriages from the front of the train, so on its own it does not tell a rider where to stand. Consumers pair it with `front_carriage_position` for the platform the rider boards at, from the perspective of a rider facing the track: `recommended_carriage=1` is the carriage stopping at the right end of the platform when `front_carriage_position=1`, and at the left end when `front_carriage_position=0`. Recommendations should only be applied when `front_carriage_position` is known for the boarding platform.
+
+|  Field Name | Type | Presence | Description |
+|  ------ | ------ | ------ | ------ |
+|  `stop_id` | Foreign ID referencing `stops.stop_id` | **Required** | Identifies the platform.<br><br>Must contain a `stop_id` that identifies a platform (`location_type=0` or empty).<br><br>Values for `stop_id` that identify stations (`location_type=1`), entrances/exits (`location_type=2`), generic nodes (`location_type=3`) or boarding areas (`location_type=4`) are forbidden. |
```
rudiejd, line 777: since `station_directions.txt` is actually concerned with only "stops" or "platforms" (`location_type = 0`), and actually forbids stations (`location_type = 1`), should the name of this file be changed? Possibly to `stop_directions.txt` or `platform_directions.txt`?

rudiejd, line 777: what would an empty `stop_id` in `station_directions.txt` signify? Would this select all stops in one direction? I think this behavior should be clarified

gcamp: Good point, `platform_directions.txt` makes sense to me

gcamp: That field is set as required, so you can't have it empty(?)

gtfs/spec/en/reference.md
```diff
@@ -756,6 +760,49 @@ Pathways are intended to exhaustively define the internal access graph of a stat
...
+|  `recommended_carriage` | Positive integer | **Required** | The recommended carriage to board, numbered from the first carriage in the direction of travel, which has a value of `1`, as in GTFS-realtime `CarriageDetails.carriage_sequence`. May represent a logical position rather than an exact carriage number. Must be less than or equal to `carriage_count`. |
+|  `carriage_count` | Positive integer | **Required** | Number of carriage positions in this configuration. May represent logical divisions (e.g., 3 for front/middle/back) rather than actual carriage count. |
```
nlee-septa, line 803:
```suggestion
|  `carriage_count` | Positive integer | **Required** | Number of carriage positions in this configuration. If `carriage_count_precision=0` or empty, represents logical divisions (e.g., 3 for front/middle/back). If `carriage_count_precision=1`, represents an exact carriage count. |
|  `carriage_count_precision` | Enum | Optional | Indicates if `carriage_count` represents relative divisions or an exact carriage count. Valid options are:<br><br>`0` or empty - `carriage_count` represents relative divisions (e.g., 3 for front/middle/back). <br>`1` - `carriage_count` is an actual count of discrete carriages. <br><br> If no values are provided, all `carriage_count` values should be considered to represent relative divisions. |
```
@skalexch I had a similar reservation about interpretations of logical vs. exact carriage counts. I think it's important that the producer be able to communicate to the consumer how exactly to interpret these values. What do you and @gcamp think of this option?

gtfs/spec/en/reference.md
```diff
@@ -756,6 +760,49 @@ Pathways are intended to exhaustively define the internal access graph of a stat
...
+|  `via_pathway_mode` | Enum | Optional | Mode of the pathway the rider takes to reach `to_stop_id`, for which `recommended_carriage` is the closest carriage. Valid options are a subset of [pathways.txt](#pathwaystxt) `pathway_mode`:<br><br>`2` - Stairs.<br>`4` - Escalator.<br>`5` - Elevator.<br><br>Riders with different accessibility needs take different pathways out of the same platform, so a producer may publish one record per mode. When empty, the recommendation applies regardless of how the rider continues. |
```
nlee-septa, line 804:
```suggestion
|  `via_pathway_mode` | Enum | Optional | Mode of the pathway (or series of successive pathways) the rider takes to reach `to_stop_id`, for which `recommended_carriage` is the closest carriage. Valid options are a subset of [pathways.txt](#pathwaystxt) `pathway_mode`:<br><br>`2` - Stairs.<br>`4` - Escalator.<br>`5` - Elevator.<br><br>Riders with different accessibility needs take different pathways out of the same platform, so a producer may publish one record per mode. When empty, the recommendation applies regardless of how the rider continues.<br><br>If `via_pathway_mode=2`, the connecting pathway(s) MAY have any `pathway_mode` value.<br>If `via_pathway_mode=4`, the connecting pathway(s), MUST NOT have `pathway_mode=2` (stairs).<br>If `via_pathway_mode=5`, the connecting pathway(s) MUST NOT have `pathway_mode=2` (stairs) or `pathway_mode=4` (escalator).|
```
Adding language like this will prevent a possible conflict where `carriage_positions.txt` indicates an elevator is available, but `pathways.txt` only contains stair pathways.

nlee-septa (commented):
Thank you for sharing this, @gcamp! This would be valuable information for riders, relatively low effort for producers to implement, and is already being implemented in different ways across consumers.

[SEPTA](https://www.septa.org/) is interested in producing this, with a few clarifications.

gtfs/spec/en/reference.md
skalexch: @nlee-septa +1 for any field that clarifies how the boarding position should be considered.

Just one thought: I'd change the name to something other than "precision" (if there is any word that conveys that the distinction is between _actual number_ and _category_). To me, "precision" give an impression of mathematical approximation.

gtfs/spec/en/reference.md
skalexch: I know that in most stations the hierarchy is this: Either stairs or stairs+escalator or stairs+escalator+elevator.

But to truly avoid conflicts, `pathways.txt` (if exists) should cover all possible pathways for that pair of station nodes.
So a cleaner logic would be:
If `pathways.txt` has at least one entry from `stop_id` to `to_stop_id` or a bidirectional entry between `stop_id` and `to_stop_id`, the value for `via_pathway_mode` should exist in `pathways.txt` for that pair of `stop_ids`.

That would lead to:
1) If pathways for that pair of nodes do not exist, no check is necessary. In this case, `carriage_positions.txt` is the direct source of truth.
2) If at least one pathway for that pair of nodes exists, then all possible pathways should be in `pathways.txt`. In this case, `carriage_positions.txt` is pulling from the COMPLETE source of truth.

gtfs/spec/en/reference.md
gcamp: `carriage_count_is_virtual` would be what I would suggest.  

I think this opens the door to having a label on each carriage, I asked the questions on how to model that in the vehicle proposal [here](https://github.com/google/transit/pull/636#discussion_r3813686720).

nlee-septa: Maybe `carriage_count_exactness` (or `carriage_count_relativeness`, since the default is "relative")? Or simply `carriage_count_type`, though I prefer more descriptive field names where possible.

Inspiration for this field comes from `stop_times.timepoint`, which conveys a similar meaning of "another field should be taken as exact or inexact". Unfortunately the nomenclature doesn't transfer well.

gtfs/spec/en/reference.md
nlee-septa: What if `carriage_positions.txt` indicates that an elevator pathway is available between `stop_id` and `to_stop_id`, but no such pathway is in `pathways.txt`? I think the spec should explicitly state that is not valid, because either `pathways.txt` does not contain all possible pathways, or `carriage_positions.txt` describes a pathway that does not exist (and it's not obvious which file has the error). Alternatively, instruction on which should prevail could be added.

gtfs/spec/en/reference.md
nlee-septa: Missed your earlier response, sorry! I like `carriage_count_is_virtual` or `carriage_count_is_relative`.

gtfs/spec/en/reference.md
rudiejd: ahh sorry, I incorrectly read "`location_type=0` or empty" as meaning that the stop could be a stop of `location_type=0` or `stop_id` could be empty. and Github isn't letting me resolve for some reason 🙃

gtfs/spec/en/reference.md
skalexch: @gcamp correct me if I am wrong, I think the point of `via_pathway_mode` is for accessibility? Because otherwise I agree with @nlee-septa and I think that its usage will only be valid when it is the only pathway between `stop_id` and `to_stop_id` and its entails having a 100% accurately mapped `pathways.txt`.

An issue I see, aside from the one mentioned by @nlee-septa, is if multiple pathways are between the platform and `to_stop_id`. If I have to take an elevator then a walkway then an escalator, the result would not be easy to distinguish. So if the point of the field is accessibility then it should be something like `accessible_navigation`. If the point is to show the user the first pathway they need to take, the field might be renamed, or a signposting field could be added.

gcamp: Yes it's for accessibility, but not only. Basically my process was 
1) [IDFM](https://prim.iledefrance-mobilites.fr/en/jeux-de-donnees/positionnement-dans-la-rame) has an "equipment type" value 
2) however the convention in GTFS is to use number enums that represent values, so I initially switched to that
3) If we use a number, might as well use the number we have already, which is the pathway_type

That value can also differentiate between stairs and escalators, even if yes the main purposes is elevators for accessibility purposes.

gtfs/spec/en/reference.md
gcamp: I'm a bit worried about data duplication with `vehicles.txt` if we do support "real" carriage count. However `vehicles.txt` right now [is not setup](https://github.com/google/transit/pull/636#discussion_r3863681478) to support carriage inside vehicles. 

To get out of the impasse, the only way out I see is : 
- only support virtual carriage count for now
- wait for the vehicles.txt specification to be merged before releasing this

I'd prefer the first option, but let me know if I'm missing some other better in between option.

gcamp:
Update today : 
- Extended the discussion period for an other month, since we're still in active discussion 
- Updated the file name from [station_directions.txt to platform_directions.txt](https://github.com/google/transit/pull/650/commits/2c00a9c8bb8000667515164f653bd55e9037375e) (suggested by @rudiejd)

orhazal:
Hello @gcamp,

Thank you for the proposal, this is valuable information for passengers and commuters.
We use the carriage positioning data provided by IDFM, and it would be great to have it standardized in GTFS.

I have a few questions:

1- For Transilien and RER, a single stop (monomodal stop place) with location_type=0 represents all the platforms, for all routes and directions in a station (e.g. Châtelet les Halles IDFM:monomodalStopPlace:45102 for RER A, B and D). Moreover, direction_id is not enough to pinpoint a "stable" direction within the station. A verifiable case is Pontoise and Saint-Ouen-l'Aumône (line H) : in these two stations, trips sharing the same direction_id can cross the station in opposite directions. Because of this, the (stop_id, route_id, direction_id) tuple from platform_directions.txt is not sufficient to model the direction in these stations.

We currently use a combination of (previous station, next station, terminus), but this approach turns out to be quite unreliable when operating patterns change (skipped stations, etc.).

The question : What information could we use to handle this case, which is problematic for several stations on the Transilien/RER network?

2- On the real vs relative carriage count discussion, I would like to bring up another case : the Seoul Metropolitan Subway. On this network, the boarding information goes one level deeper than the carriage : it is defined at the door level. Every boarding position on the platform is marked with a (car, door) pair, for example "5-1" for car 5, door 1, and journey planners like Naver tell passengers the exact door to board. Given the crowding levels on this network, this granularity is considered genuinely valuable for passengers to transfer or exit quickly in overcrowded stations. It is worth noting that this is not proprietary app data : Seoul Metro publishes it as [open data](https://data.seoul.go.kr/dataList/OA-22521/F/1/datasetView.do#).

The question : should the proposal anticipate a door-level granularity, for instance by keeping the carriage identification scheme compatible with an optional door field or a future extension, or is this considered too detailed for a first version?

Thank you.

vikiawv:
Since you Mention Seoul's Metro I also have to mention that in Vienna a somewhat similar system is used. It displays the exit directions for each individual door in the train carriage itself including departures of lines at the corresponding exit.
It is however not included in their App, only displayed in the Train itself.
https://de.wikipedia.org/wiki/U-Bahn_Wien_X#/media/Datei:Fahrgastinformationssystem_X-Wagen,_U-Bahn_Wien.jpg

The system itself is properiatory by Siemens Mobility it seems.

The local data provider "Wiener Linien" is probably willing to submit such data also for GTFS. They likely already have the entire system mapped out for the Siemens System.

orhazal:
@vikiawv Interesting ! Some subways in Paris have a similar display inside, but without door-level detail.

GR1YacHXIAAVUGz
GR1Yab0XgAECnFp
images

miklcct:
I have some doubt about this proposal:

1. If a route uses a mix of different length trains, can this spec specify that? For example, on a route where 8-carriage and 12-carriage trains mixes, at some stations, the 8-carriage train stops at the front, at others, the 8-carriage train stops at the middle, while at some stations having short platforms, the last 4 carriages or a 12-carriage train will not open their doors.

2. How can I specify that a train will open its doors on both sides at a specific stop?

orhazal:
> If a route uses a mix of different length trains, can this spec specify that? 

On the Transilien network in the Paris region, this information is only available in their realtime API (SIRI), which makes it unpredictable in the static GTFS schedule.

Maybe we could have one row per train composition variant, and the consumer could either show both of them to the traveller (short train -> BACK, long train -> MIDDLE) or pick the right one using the realtime data?

> How can I specify that a train will open its doors on both sides at a specific stop?

Is it really useful? In the end, the purpose of this information is to give the closest boarding position relative to the next connection or exit. Let's say subway A stops at a station and the doors open on both sides: the platform on the right has Exit 1 at the rear, and the platform on the left has a connecting tunnel to subway B in the middle. If the traveller wants to transfer to subway B, a single position (car 4 out of 8, or MIDDLE) is enough. If they want to leave through Exit 1, it's the same: car 8 out of 8, or BACK.
We had data on left/right door opening: it was hard to maintain and ultimately not that useful.

orhazal:
Another thing to consider here: what if we have a shared platform between two lines, where the transfer is just crossing the platform, meaning every position is the "best position"? How should the spec handle this? Do we list all the possible positions in the file (e.g. cars 1 to 8 out of 8 for this transfer), or do we put nothing at all? The first option is too verbose, and the second could be interpreted as missing data.

I lean towards the first option.

EDIT: a strong addition would be a transitive carriage position calculation. Example: I take the RER A to Châtelet les Halles. There, I cross the platform to take the RER B, so all positions are "best". Then at Aéroport CDG, I take the escalator located in the middle of the platform.

In this example, if we could somehow model an "ALL" best position, it would be easy for consumers to tell passengers to board in the middle of the RER A, then in the middle of the RER B, by transitivity. Otherwise, no boarding advice would be given when boarding the RER A, since all positions are equally good for the transfer to the RER B.

But maybe that's too complicated. Then again, it's something a GTFS consumer could easily infer from an "ALL" value, or from listing all cars out of n.

vikiawv:
For shared plattforms there could be a system that looks up which carriage or door of the current train is equal to the best carriage/door in the second train.

Problem beeing not all such changes have trains with equal carriages and/or lenght. At Station Wien Längenfeldgasse for example the change is from a 6-car train to a 12-car train.
At other stations in comparison the trains have equal lenghts and carriages such as Berlin Mehringdamm, there you can just take the reccomended carriage for the second train also for the first one, if you change on the same plattform.

jfabi:
> Moreover, direction_id is not enough to pinpoint a "stable" direction within the station. A verifiable case is Pontoise and Saint-Ouen-l'Aumône (line H) : in these two stations, trips sharing the same direction_id can cross the station in opposite directions. Because of this, the (stop_id, route_id, direction_id) tuple from platform_directions.txt is not sufficient to model the direction in these stations.

Re @ORHAZAL's initial comment re Transilien and the RER, I don't this type of case is common, one prominent example is the [New Jersey's PATH](https://www.panynj.gov/path/en/index.html) Journal Square–33 Street line. During the day, a one-way trip goes straight from one end to the other. At night, the trip makes an extra stop at Hoboken, a stub end station, at which the train reverses its direction. Adding an optional "previous station" field could help in that case, though it could alternatively go to the trip level.

jfabi:
_General comment:_ I'm a bit concerned that this proposal, as currently worded, may exclude the usage of boarding areas for a platform. Any feeds including both boarding areas and pathways would necessarily have their pathways going to the boarding areas, rather than to the platform stop. While the use of boarding areas is not too widespread, I am hearing increased interest among larger agencies. How do we ensure they can coexist?

orhazal:
@jfabi thanks for your input !

> I don't think this type of case is common

It is the case for a lot of stations on the Transilien / RER network : [68 stations here](https://github.com/user-attachments/files/31964266/transilien_problematic_stations.txt)
@gcamp mentioned this dataset in the proposal, and i think it's important to point out that the proposal would not cover a significant part of that network.

> Adding an optional "previous station" field could help in that case, though it could alternatively go to the trip level.

That is what we do today, and it is quite limited, for similar reasons as your example :

> We currently use a combination of (previous station, next station, terminus), but this approach turns out to be quite unreliable when operating patterns change (skipped stations , etc.).

EDIT : What about a `trip_id` field? Producers could compute it easily, without having to guess from the previous and next stations.

miklcct:
> Is it really useful? In the end, the purpose of this information is to give the closest boarding position relative to the next connection or exit. Let's say subway A stops at a station and the doors open on both sides: the platform on the right has Exit 1 at the rear, and the platform on the left has a connecting tunnel to subway B in the middle. If the traveller wants to transfer to subway B, a single position (car 4 out of 8, or MIDDLE) is enough. If they want to leave through Exit 1, it's the same: car 8 out of 8, or BACK. We had data on left/right door opening: it was hard to maintain and ultimately not that useful.

Leaving a train from the wrong side when both sides open is a massive accessibility fail, and I expect that a good navigation app should show which side I should leave the train for my destination.

cswilson252:
IMO this addition will be very useful for agencies to signal where users must board/alight when there are temporary train boarding area restrictions (whether due to construction/renovation/weather conditions/etc). 

I take a train line weekly where several stops have boarding/alighting restrictions (i.e not all doors open), and without fail, there are always passengers rushing to the side of the train that opens its doors (often not in time to exit the train before it departs :pensive:)

As it stands, the agency (Metrolinx) [signals this with diagrams on their website and at the station](https://www.gotransit.com/en/the-future-and-go/construction-notices/guelph-central-go-station-improvement), but this clearly isn't enough (as stated above)


(Actually, one of the aforementioned stops [doesn't even show that restriction info on the "Station Info" page](https://www.gotransit.com/en/find-a-station-or-stop/ac/station-details))! :frowning:

gcamp:
Happy to see more discussion on that proposal! Thanks for the feedback 

@miklcct On serving the case where not all trains have the same carriage count (or even vehicle type difference) : right now it's not supported but the plan was to take vehicle types into account when the vehicle proposal is merged. From the initial post : 
> https://github.com/google/transit/pull/636 proposes vehicles.txt along with a vehicle_class on routes.txt and trips.txt. This proposal does not depend on it and stays usable on its own through carriage_count, but if https://github.com/google/transit/pull/636 is adopted, an optional vehicle_class in carriage_positions.txt would be a natural follow-up to select the right configuration for a trip.

@miklcct On the side or opening, you're right that we could add `door_side` on the `carriage_positions.txt` as a new primarily key. We're unsure if we will be supporting it at this point, but if an other consumer/producer having this needs wants to test it happy to add it.  

@orhazal on all platform equality good, I was thinking that yes you would list all the possible cars, but we can add a value `-1` to represent "any"? I don't think I would support transitive positioning at least for now 

@orhazal on direction/stop not enough to differentiate all the cases, trip_id make sense but it's very heavy. I don't see this case in the open IDFM data, is that modelled anywhere else? `previous_stop` would also work I think, since this is static data only and skipped stops in real time don't apply. 

@vikiawv @orhazal On door granularity, this is also something I feel we should do when vehicle definition include door-level info, but I hope we can ship this change without waiting this requirement. (we should still plan for it however).

orhazal:
> I was thinking that yes you would list all the possible cars, but we can add a value -1 to represent "any"? I don't think I would support transitive positioning at least for now

Both (list all vs -1) work for me. On transitivity, it can be derived on the consumer side, so i agree it shouldn't be carried in the data.

> I don't see this case in the open IDFM data, is that modelled anywhere else?

It isn't modelled in the IDFM dataset for the same reason: their format (from_stop_id, from_line_id, to_stop_id) can't express it. We did manage to model those cases by adding all (previous_stations, next_stations and terminuses) for every station and it turned out heavy, but it works 100% of the time so we kept it.
_EDIT_: they do have some data for this in the same dataset, but only for RER A and B, and with stop ids that don't exist in the GTFS. From what I remember checking two years ago, those look like the NeTEx platform ids, so it's useless.

> previous_stop would also work I think, since this is static data only and skipped stops in real time don't apply.

The problem shows up in static data as well. I ran a query on today's dataset, here are [the stations that have more than one previous_stop in a given direction](https://github.com/user-attachments/files/32029905/stations_with_more_than_one_previous_stop.json). I have also seen this case in the past, with a line partly rerouted onto another one for engineering works as part of the static data (line P running over a section of line E if i remember correctly), and it clearly showed the limits of previous_stop-like solutions.

> trip_id make sense but it's very heavy.

I think trip_id is good enough as long as it's only used for the problematic stations (the 68 stations from my previous message for example) : producers have a reliable way to determine the direction in which a station is traversed, so when generating the GTFS they can attach only the trip ids concerned. What do you think @gcamp ?

PR #651, opened 2026-07-29
Mark images in Service Alerts as final

gcamp:
## Summary

`image`, `image_alternative_text` and the `TranslatedImage` message have been in the spec since [September 2021](https://github.com/google/transit/pull/283). This PR removes their experimental tag.

## Type of change

GTFS Realtime
- [X] Specification Change

## Proposed Discussion Period

7 calendar day (minimum)

## Testing Details

- Consumer(s): Transit
- Producer(s): Metro Transit (Minneapolis), TransLink (Vancouver), MBTA (Boston), VTA (Bay Area), [Mecatran](mecatran.com).

One caveat on the consumer side: Transit parses both fields and can display the image alongside the alert, but it isn't enabled by default today — it has to be manually enabled for the image to be shown to riders. There's plans to show it by default in the future. In the list of producers, we currently only show Metro Transit and TransLink. 

| Minneapolis | Vancouver |
| ------  | ----  |
| Capture d’écran, le 2026-07-29 à 13 39 26 | Capture d’écran, le 2026-07-29 à 13 37 53 |

## Proposal Update Tracker

| 2026-07-29 | Initial PR submission |

### DIFF
--- a/gtfs-realtime/proto/gtfs-realtime.proto
+++ b/gtfs-realtime/proto/gtfs-realtime.proto
@@ -708,12 +708,10 @@ message Alert {
 
   // TranslatedImage to be displayed along the alert text. Used to explain visually the alert effect of a detour, station closure, etc. The image must enhance the understanding of the alert. Any essential information communicated within the image must also be contained in the alert text.
   // The following types of images are discouraged : image containing mainly text, marketing or branded images that add no additional information. 
-  // NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional TranslatedImage image = 15;
 
   // Text describing the appearance of the linked image in the `image` field (e.g., in case the image can't be displayed
   // or the user can't see the image for accessibility reasons). See the HTML spec for alt image text - https://html.spec.whatwg.org/#alt.
-  // NOTE: This field is still experimental, and subject to change. It may be formally adopted in the future.
   optional TranslatedString image_alternative_text = 16;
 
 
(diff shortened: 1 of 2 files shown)


ckraatz:
+1 from SimplifyTransit when you're ready for a vote. Our Alerts module has been publishing the Image field since 2022 because map images are very helpful for riders. Our Subscriptions module also consumes images and embeds images when sending Alerts via email/SMS, so we support this field from a consumer side too.

I would encourage Producers to ensure there's a requirement to include good Alt Text for accessibility. It's important to always describe what's in the image, even if the image is a map that can't be fully described in words and its contents are accessibly represented in text in the Header/Description fields. We don't allow Images to be uploaded without entering Image Alt Text, so that field is "conditionally required."

gcamp:
I'm assuming few comments is a good thing! I'm **starting a vote** to adopt `image`, `image_alternative_text` and `TranslatedImage` formally into the spec.

The voting period ends 2026-08-13 at 23:59:59 UTC.

leonardehrenfried:
+1 OpenTripPlanner

skinkie:
-0 OpenGeo, I think we must state more explicitly that the image must not be "just text", and rich content. Hence the report that "Transit parses both fields and can display the image alongside the alert," shows to me that it does not seem the intention. While I think that should be.

ckraatz:
+1 SimplifyTransit, as a Producer

phil-swiftly:
+1 Swiftly

jfabi:
+1 from the MBTA

We add images to certain planned rail alerts, and display them on selected first-party touchpoints, including our own website. We would encourage consumers to consider in which contexts it might be more or less appropriate to show alert images (eg, if a consumer's dynamic map already marks parts of lines and stations as closed, it might be less necessary, in certain places, to show a static image likely to contain the same info).

Screenshot 2026-08-07 at 18 44 04

dan-engler-arcadis:
+1 Arcadis as a producer

gcamp:
Vote concluded last night, it passed! 

5 votes in support, one abstention.

skalexch:
Congratulations @gcamp! We will merge this PR soon.

Issue #595, opened 2025-12-22
transfers.txt: Add optional field service_id


manfredbrandl:
### Describe the problem

I suggest adding the optional field service_id to transfers.txt so that secured connections can be restricted to the days on which they are actually secured.

### Use cases

Our network has guaranteed connections (transfer_type = 1 with from_trip_id and to_trip_id) between journeys, but not on all days on which these journeys are operated. On days without guaranteed connections, there may be a through connection (block transfer), a guaranteed connection at another stop or – for operational reasons – no guaranteed connection at all.

### Proposed solution

Add the optional field service_id to transfers.txt

luca-d-axon:
I support this suggestion, we have similar cases in Switzerland too.
HRDF format correctly maps these cases with a validity (BITFELD) in DURCHBI file.

felixguendling:
I don't really have an opinion pro/against on this, just want to add that:
- this does not enable a new use case because at the moment you can just split the trip into two trips (one with `transfer_type=1` and one with `transfer_type=4`) and have the same effect
- it would not be downward compatible which is a no-go since GTFS does not have versioning
- when transforming from HRDF you have to build different trips for different track sequences anyway (unrolling track service days), so why not do it for transfers as well
- it would definitely be an advantage to be able make conversion HRDF to GTFS easier and less error prone

ue71603:
SBB supports the proposal. We agree with @felixguendling  that we could do this by splitting trips. We have then trips that we would need to split multiple times for different technical reasons. To remove one of those reasons we believe is helpful for understanding what is happening.
We will also do an implementation for the Swiss data.

manfredbrandl:
I also agree with @felixguendling that this could be achieved by splitting the journeys.
However, we do not wish to split journeys, as we will be using a global ID for journeys that is used in all formats, including GTFS and GTFS-R, in order to minimise the effort required for quality assurance. We would have to split the journeys not only in GTFS but also in GTFS-R, which is too labour-intensive for us and our software supplier. 
@felixguendling Could you please explain in more detail why the extension is not backwards compatible? The proposed additional field is intended to be optional, so existing GTFS data will remain valid with its current meaning.

felixguendling:
> @felixguendling Could you please explain in more detail why the extension is not backwards compatible?

[...]

PR #659, opened 2026-09-09
Added service_id column to transfers.txt

orhazal:
## Summary
This proposal adds an optional `service_id` field to `transfers.txt` that restricts the service days on which a transfer rule is in effect. The rule applies only when the service day of the arriving trip is included in the days referenced by service_id.

**The field is a restriction only, never a variant**: it does not join the primary key, does not add specificity, and two records must not differ only by service_id. This keeps the change backward compatible : legacy consumers ignore the column and apply each rule on all days, which is exactly what they do today.

## Describe the Problem
Discussion started in [#595](https://github.com/google/transit/issues/595). Thanks to all the participants for their comments and suggestions!

`transfers.txt` rules currently apply on every day the referenced trips operate, but several producers have rules that only hold on some of those days: guaranteed connections that are only secured on certain days, in-seat transfers that only happen when the same vehicle operates both trips, or forbidden/minimum-time rules with an application period.

The existing workaround is to split trips by transfer validity (one pair with transfer_type=4, one with transfer_type=1, each on its own service_id). Several producers have explicitly ruled this out because they use stable, global trip IDs shared between GTFS and GTFS-Realtime among other reasons.

## Use Cases

- France (SNCF) : SNCF Voyageurs publishes [minimum connection times and forbidden transfers](https://ressources.data.sncf.com/explore/dataset/temps-correspondance-minimaux/information/) (EU Passenger Rights Regulation) that convert cleanly to transfers.txt ; except train-number rules carrying an application period. Without service_id, a transfer forbidden for 3 days ends up forbidden for the trip's entire service period. This field is the last missing piece for an exhaustive transfers.txt on that feed ; today every consumer of the SNCF GTFS displays connections that are not actually feasible or bookable **_with one ticket_**.

- Other use cases exist in Switzerland and Austria, and more broadly for any operator using HRDF internally.

## Proposed Solution
Add one optional `service_id` field to `transfers.txt` that identifies a set of service days on which the transfer rule is in effect.

- **Operating days, not calendar dates.** `service_id` is matched against the service day of the arriving trip. A calendar-date semantic (as in `timeframes.txt`) was considered and dropped.
- **Restriction, never a variant.** The primary key (`from_stop_id`, `to_stop_id`, `from_route_id`, `to_route_id`, `from_trip_id`, `to_trip_id`) is unchanged and stays unique. Per-day *variants* of a rule (e.g. `min_transfer_time` of 5 min Monday–Friday and 4 min Saturday–Sunday as two rows) are explicitly **out of scope** : expressing them would duplicate the primary key, which is the backward-compatibility blocker identified in the issue. **The restriction-only semantic was chosen deliberately to keep existing consumers working**.
- **Fallback via existing precedence.** On days a rule is not in effect, consumers fall back to the next less specific rule using the existing ranking of specificity. No new selection logic is introduced.
- **Not limited to trip-to-trip rows.** A stop-to-stop or route-to-route rule may also carry a `service_id`; it then applies to every matching arriving trip operating on a day in the set.

**Backward compatibility**: legacy consumers ignore the column and apply each rule on all days, identical to today's behavior for the rules they already see. Since restricted rules are minimum times, forbidden transfers, or guarantees, a legacy consumer is at worst too strict, or over-promising a guarantee on excluded days; it never encounters duplicate primary keys or conflicting rows.

## Type of change
GTFS Schedule
- [x] Functional Change

## Proposed Discussion Period
I recommend reserving at least **2 weeks** for discussion to ensure everyone has sufficient time to discuss the proposal.

## Testing Details
- Consumer(s):
- Producer(s):
- Estimated Testing Period:

## Proposal Update Tracker 
| 2026-09-09 | Initial PR submission |
| 2026-09-23 | End of discussion period |


skalexch:
Thank you @orhazal for launching this proposal! Just wanted to mention that your link to the issue ([#595](https://github.com/google/transit/issues/595)) is returning a 404. It's because the link points to your fork. Could you please fix that?

gtfs/spec/en/reference.md
```diff
@@ -683,7 +683,7 @@ Fields `from_trip_id`, `to_trip_id`, `from_route_id` and `to_route_id` allow hig
 5. One `route_id` defined: `from_route_id` or `to_route_id`.
 6. Only `from_stop_id` and `to_stop_id` defined: no route or trip related fields set.
 
-For a given ordered pair of arriving trip and departing trip, the transfer with the greatest specificity that applies between these two trips is chosen. For any pair of trips, there should not be two transfers with equally maximal specificity that could apply.
+For a given ordered pair of arriving trip and departing trip, the transfer with the greatest specificity that applies between these two trips is chosen. A transfer that is not in effect on the service day of the arriving trip (see `service_id`) is ignored, and the transfer with the next greatest specificity that applies is chosen instead. For any pair of trips, there should not be two transfers with equally maximal specificity that could apply.
```
skalexch: I would change the wording to clarify "the next ~~greatest~~ specificity". Something to clarify that it's the first next valid level of specificity going down.

orhazal:
@skalexch Thank you for checking, it's done !

Here's an example of a generated transfers.txt with `service_id` for the SNCF GTFS feed: https://github.com/orhazal/sncf-gtfs-toolkit/blob/master/output/transfers.txt

More details in the repo: https://github.com/orhazal/sncf-gtfs-toolkit/

manfredbrandl:
I repeat from https://github.com/google/transit/issues/595 
I’ve thought of an important use case that requires an extension to the primary key.
We have two journeys that are linked as a through service on school days (transfer_type 1) and as a guaranteed connection on non-school days (transfer_type 4).
To map this correctly, we need either transfer_type or service_id to be part of the primary key.

orhazal:
Thanks @manfredbrandl,

Including service_id or transfer_type in the PK would change the scope of this PR, which currently treats service_id as a restriction only and leaves the PK untouched.

As mentioned in #595, i don't have a strong preference on this, and if the community agrees, i'm fine with adding it to the PK here. It would still be a breaking change for consumers : any consumer that assumes the current key is unique would start seeing duplicate rows.

Let's see what others think before I update the proposal.
I'm extending the discussion period by two more weeks.

Issue #643, opened 2026-06-18
File shapes.txt should be Conditionally Required


fabianabarca:
### Describe the problem

The field `trips. shape_id` (a foreign ID referencing `shapes.shape_id`) is **Conditionally Required**, with the following condition:

> Required if the trip has a continuous pickup or drop-off behavior defined either in `routes.txt` or in `stop_times.txt`

When that condition is true, `shapes.txt` becomes required as well, because `trips.shape_id` cannot be valid without a corresponding `shapes.shape_id` value.

### Use cases

It is relevant for consumers when planning the database schemas, and for evaluating correctness of the feed.

### Proposed solution

I see two possible paths:

1. Preferred: make `shapes.txt` conditionally required

Mark `shapes.txt` as **Conditionally Required** in the _Presence_ column of the _Dataset Files_ table (GTFS Schedule), with this condition:

> Required if the trip has a continuous pickup or drop-off behavior defined either in `routes.txt` or in `stop_times.txt`.

This matches the existing condition on `trips.shape_id` and makes the transitive requirement explicit.

2. Alternative: remove the reference from `trips.shape_id`

If `shapes.txt` is not intended to become conditionally required under that condition, then `trips.shape_id` should not be a Foreign ID in `trips.txt`. In that case, its Type should be changed to **Text** (similar to `direction_id`), because there would be no guaranteed referenced table.

### Additional information

This issue is related to #459, but it has a different purpose, as that one aims to include `shapes.txt` in the recommended tables.

As mentioned in the discussion of that issue, 85% of the feeds in the Mobility Database (as of May 2024) already have `shapes.txt`. Therefore, it shouldn’t be a significant change for producers to mark it as Conditionally Required.

fabianabarca:
I noticed that there is the option "Recommended" in the "Presence" column for some fields, for example for `feed_info. feed_start_date`. As a third proposed solution, `shapes.txt` could be marked as "Recommended" in the Dataset Files table, as long as that's a valid option for the presence of files. Still, my preferred option is "Conditionally Required", due to the transitive requirement discussed above.

etienne0101:
Hello @fabianabarca and thanks for opening this issue! Feel free to reach out if you need any guidance on contributing to GTFS.

Once you feel ready, and ideally after this issue has had some discussion (though that's not required), feel free to open a proposal for this change.

Since this looks like a fix for a gap in the spec, you could make the case for it being a documentation change or a non-functional change. In that case, it would need either no vote or just one vote.

[...]

PR #470, opened 2024-06-03
Add that shapes should be included

isabelle-dr:
## Context
This PR follow-up on issue https://github.com/google/transit/issues/45, which received good engagement. 
This originated form an issue by @e-lo for the [GTFS Schedule Best Practices](https://gtfs.org/schedule/best-practices/). 
We'd like to eventually have the Best Practices and the specifications consolidated into one document, so we are incorporating new Best Practices directly into the spec (see efforts to merge the specifications and Best Practices [here](https://github.com/google/transit/issues/396) and [here](https://github.com/google/transit/issues/451)). 

## Proposal
This PR adds the following mention in [shapes.txt](https://gtfs.org/schedule/reference/#shapestxt) description: 

> The shapes.txt file should be included for all route-based services (not for zone-based demand-responsive services).

Given that zone-based services can now be modeled in GTFS and don't need `shapes.txt` defined, I didn't modify the [Presence type](https://gtfs.org/schedule/reference/#presence) of shapes.txt to _Recommended_, and rather chose a "should" statement in its description. 

Note that merging this PR means we'd add a WARNING in the [Canonical GTFS Schedule Validator](https://gtfs-validator.mobilitydata.org/), which is the severity level for all GTFS Best Practices (spec _should_ & _recommended_ + everything in the Best Practices).

PR #660, opened 2026-09-09
Update shapes.txt to be conditionally required and recommended otherwise

fabianabarca:
Update `shapes.txt` to be conditionally required based on `trips.shape_id` and the continuous pickup or drop-off behavior defined either in `routes.txt` or in `stop_times.txt`.

## Summary

Make `shapes.txt` conditionally required in the table in Dataset Files of the GTFS Schedule reference.

## Describe the Problem

According to issue https://github.com/google/transit/issues/643, the field `trips. shape_id` (a foreign ID referencing `shapes.shape_id`) is **Conditionally Required**, with the following condition:

> Required if the trip has a continuous pickup or drop-off behavior defined either in `routes.txt` or in `stop_times.txt`

When that condition is true, `shapes.txt` **becomes required as well**, because `trips.shape_id` cannot be valid without a corresponding `shapes.shape_id` value.

## Use Cases

It is relevant for consumers when planning the database schemas, and for evaluating correctness of the feed.

## Proposed Solution

Originally:

|  File Name | Presence | Description |
|  ------ | ------ | ------ |
|  [shapes.txt](#shapestxt)  | Optional | Rules for mapping vehicle travel paths, sometimes referred to as route alignments. |

Becomes:

|  File Name | Presence | Description |
|  ------ | ------ | ------ |
|  [shapes.txt](#shapestxt)  | **Conditionally Required** | Rules for mapping vehicle travel paths, sometimes referred to as route alignments.<br><br>Conditionally Required: <br>- **Required** if the trip has a continuous pickup or drop-off behavior defined either in [routes.txt](#routestxt) or in [stop_times.txt](#stop_timestxt). <br>- Recommended otherwise. |

This matches the existing condition on `trips.shape_id` and makes the transitive requirement explicit. Also, and following PR https://github.com/google/transit/pull/470, `shapes.txt` should be **recommended** otherwise.

## Type of change

 

GTFS Schedule
- [ ] Functional Change
- [ ] Non-Functional Change
- [x] Documentation Maintenance

GTFS Realtime
- [ ] Specification Change

## Checklist


ghost:
The presence conditions of files are also mentioned at the beginning of each file, it would need to be updated there as well :).

fabianabarca:
That's right! Thanks. I updated this:

---

### shapes.txt

👉 **Conditionally Required** 👈

Primary key (`shape_id`, `shape_pt_sequence`)

...

---

I noticed that the description of other files that are conditionally required don’t include the condition that makes them necessary because it is in the Dataset Files table, so I'm not adding anything else.

etienne0101:
Thank you @fabianabarca for this pull request.
We will merge this as soon as it reaches the 7-day mark.

